Issue #30, opened 2015-01-05
Add type for variable-length, uniform tuples


gvanrossum:
See also https://github.com/JukkaL/typing/issues/6.

Some ideas:
- Tuple[int, ...]
- Sequence[int]

gvanrossum:
It was pointed out on python-ideas that the language about this in the PEP is somewhat ambiguous, e.g. `Tuple[t1, ..., t2]` or `Tuple[..., t1]` should be clearly ruled out.

gvanrossum:
Also, does the empty tuple match `Tuple[int, ...]`?

JukkaL:
I think an empty tuple should match `Tuple[int, ...]`.

gvanrossum:
This is decided and written up in the PEP: the syntax is `Tuple[t1, ...]` (the `...` are a literal ellipsis).

PR #43, opened 2015-01-16
Clarify expression for arbitrary-length tuples

encukou:
This is hopefully clearer, and it rules out things like `Tuple[t1, ..., t2]`.

### DIFF
--- a/pep-0484.txt
+++ b/pep-0484.txt
@@ -314,10 +314,10 @@ holds a set of classes representing builtin types with generics, namely:
 
 * FrozenSet, used as ``FrozenSet[element_type]``
 
-* Tuple, used as ``Tuple[index0_type, index1_type, ...]``.
-  Arbitrary-length tuples might be expressed using ellipsis, in which
-  case the following arguments are considered the same type as the last
-  defined type on the tuple.
+* Tuple, used by listing the element types (for example
+  ``Tuple[str, int, str]``).
+  Arbitrary-length homogeneous are expressed using one type and ellipsis,
+  as in ``Tuple[int, ...]``.
 
 It also introduces factories and helper members needed to express
 generics and union types:


encukou:
This is for #30

encukou:
Hm, Now I see that, plain `Tuple` isn't really defined in the PEP. I assume it would be equivalent to `Tuple[Any, ...]`, and similarly for other generics?
I can add some words for it if that's the expected behavior.

encukou:
Also, should `Tuple[t1, t2, ...]` be ruled out as well?

pep-0484.txt
```diff
@@ -313,9 +313,11 @@ holds a set of classes representing builtin types with generics, namely:
...
 
 * Tuple, used as ``Tuple[index0_type, index1_type, ...]``.
-  Arbitrary-length tuples might be expressed using ellipsis, in which
-  case the following arguments are considered the same type as the last
-  defined type on the tuple.
+  Arbitrary-length tuples are expressed using a final ellipsis, in which
+  case the type before the ellipsis may be repeated zero or more times.
+  For example, ``Tuple[int, ...]`` represents any tuple of ints, including
+  the empty tuple.
+  The form ``Tuple[...]`` is equivalent to plain ``Tuple``.
```
gvanrossum: I don't like this.

pep-0484.txt
```diff
@@ -313,9 +313,11 @@ holds a set of classes representing builtin types with generics, namely:
 * FrozenSet, used as ``FrozenSet[element_type]``
 
 * Tuple, used as ``Tuple[index0_type, index1_type, ...]``.
-  Arbitrary-length tuples might be expressed using ellipsis, in which
-  case the following arguments are considered the same type as the last
-  defined type on the tuple.
+  Arbitrary-length tuples are expressed using a final ellipsis, in which
+  case the type before the ellipsis may be repeated zero or more times.
```
gvanrossum: This should be stricter still -- Tuple[t1, t2, ...] should be disallowed. Only one type followed by an ellipsis should be acceptable.

encukou:
Updated.
I removed the original specification because using `...` as both syntax and metasyntax got confusing.

gvanrossum:
Thanks, I've committed something based on your words.

Issue #18, opened 2014-10-16
An Intersection type?


gvanrossum:
Sometimes you want to say that an argument must follow two or more different protocols. For example, you might want to state that an argument must be an Iterable and a Container (for example, it could be a Set, a Mapping or a Sequence). It would be nice if this could be spelled like this:
```
def assertIn(item: T, thing: Intersection[Iterable[T], Container[T]]) -> None:
    if item not in thing:
        # Debug output
        for it in thing:
            print(it)
```

gvanrossum:
(In README.text, Łukasz suggested to use Any[] instead of Union[] so that All[] can be used as Intersection[]. Feels a bit cute to me.)

ambv:
Having intersections is important since we don't have structural sub-classing and protocols will likely not be composable with ABCs.

Any/All is not only cute, it's also shorter to type and more obvious to the reader :)

gvanrossum:
Łukasz, can you show a few examples of how Any/All would be used? Your README note was rather cryptic. I agree Intersection is verbose, but it sounds like a minor use case, and I don't want to lose unparameterized Any as the top (or is that bottom? :-) of the type tree/graph.

JukkaL:
I think this would be occasionally useful, but clearly this would be a minor feature. I vote for leaving this out for now and waiting until we have seen some real-world use cases that would benefit from intersection types.

gvanrossum:
Agreed. Let's skip it for now then.

ambv:
What I proposed is alternative names: Union becomes Any, Intersection becomes All. typevar(values=) could become OneOf, with the bracket syntax, which would be more consistent.

Examples:
```
AnyStr = OneOf[bytes, str]
def memcache_get(key:AnyStr) -> Optional[AnyStr]:
  ...

OneOrMany = OneOf[AnyStr, List[AnyStr]]
def memcache_set(key:OneOrMany, value:OneOrMany) -> int:
  ...

# but:
def retry(callback:Callable[AnyArgs, Any[int, None]], timeout:Any[int, None]=None, retries=Any[int, None]=None) -> None:
  ...

# equiv:
OptionalInt = Any[int, None]
def retry(callback:Callable[AnyArgs, OptionalInt], timeout:OptionalInt, retries:OptionalInt) -> None:
  ...
```

Note that `All` provides a form of structural typing:
```
# The following absurd function uses retryable.__len__ and retryable.__call__ 
def retry(retryable:All[Sized, Callable]):
[...]

Issue #16, opened 2014-10-16
Do we need a standard way of silencing checkers?


JukkaL:
It may be useful to have a way of saying "don't complain about this type error; trust me, I know what I'm doing". For example, consider this code:

```python
def foo(f: Function[...]) -> None:
    f._attr_ = True   # Error? But what if this is intentional?
```

A type checker would probably not allow assigning to an attribute of a function. However, sometimes programmers do this, and it works for user-defined functions. Maybe we could have a standard way of asking the type checker to shut up and ignore this particular warning only.

If we don't specify this in the PEP, there could end up being multiple tool-specific ways of doing this.

Apparently pylint supports something like this:

```python
# pylint: disable=fixme
```

Hack also has something like this (example copied from Hack docs):

```php
<?hh
function unsafe_foo(int $x, int $y): int {
  if ($x > $y) {
    // UNSAFE
    return "I am not checked by the type checker"; // Covered by UNSAFE
  }
  return 34; // NOT covered by UNSAFE
}
```

Link to relevant Hack docs: http://docs.hhvm.com/manual/en/hack.modes.unsafe.php

ambv:
The Hack approach would basically be:
```
with typing.off:
  ...
```

The pylint approach has much more granularity but requires unique IDs to all checks.

gvanrossum:
I like the simple with-statement. I think it is possible to write something that is both a decorator and a context manager. IIRC the unittest.mock package does this, so perhaps we could use the same name. When used as typing.off the cute name is fine; but after "from typing import off" it's too short IMO. (It turns off *what*?)

JukkaL:
The with statement is somewhat slow in CPython (the same applies to things like `List[int]` and `Undefined(...)`) and introduces an extra block nesting level, which makes code more difficult to read.

We could also have a comment that disables/enables all errors:

```
# typing: off
1 + 'x'  # don't report an error
# typing: on
1 + 'x'  # report an error
```

The comment should probably only extend until the end of the block that introduced it:

```
if foo:
    # typing: off
    1 + 'x'   # don't report an error
1 + 'x'  # report an error
```

gvanrossum:
Hm. I'm not so keen on introducing more magic comments. See https://github.com/ambv/typehinting/issues/35

gvanrossum:
So we're introducing magic comments, but we might want to have something for stub modules to declare that a module exports absolutely anything: https://mail.python.org/pipermail/python-ideas/2015-January/031381.html

[...]

PR #44, opened 2015-01-17
Support interoperability with other annotations and type systems

scoder:
Support interoperability with other annotations and type systems by requiring separate namespaces.

### DIFF
--- a/pep-0484.txt
+++ b/pep-0484.txt
@@ -248,15 +248,56 @@ Compatibility with other uses of function annotations
 -----------------------------------------------------
 
 A number of existing or potential use cases for function annotations
-exist, which are incompatible with type hinting.  These may confuse a
-static type checker.  However, since type hinting annotations have no
-run time behavior (other than evaluation of the annotation expression
-and storing annotations in the ``__annotations__`` attribute of the
-function object), this does not make the program incorrect -- it just
-makes it issue warnings when a static analyzer is used.
+exist, which are incompatible with type hinting.  These must not confuse
+a static type checker.  Instead, these tools must ignore any annotations
+that they do not understand.
+
+Given the precedence that type hints take when it comes to annotations,
+this means that any other uses of annotations must be easily distinguishable
+from type hints.  This PEP proposes a type system that lives partly in the
+builtin namespace and partly in the special "typing" module.  Annotations
+that want to coexist with type hints must therefore use different namespaces
+(i.e. objects that live in other modules), and tools that deal with type hints
+must ignore any annotations that live in namespaces that they do not know. If
+multiple annotations are used on the same argument or return value, they can
+be chained in tuples, as long as each annotation comes from a clearly defined
+namespace.
+
+This namespace separation makes it possible to use annotations for multiple
+purposes, such as documentation and type hinting, for example::
+
+    from documentation import doc
+    from typing import integer
+
+    def func(x: (integer, doc("a number")) -> integer:
+        ...
+
+In this case, "doc()" would create some kind of documentation object that wraps
+a docstring for the function argument.  Type checkers would recognise it as
+coming from a foreign namespace and only consider the "integer" annotation as
+relevant for their own purpose.
+
+Similarly, different type systems can be mixed with the proposed type system,
+such as the type system of the Cython compiler, which supports types from the
+C programming language and compiles Python code to C.  It could allow List
+items to be defined as C types, which would then be checked at runtime by
+compiled code.  This could look as follows::
+
+    from typing import List
+    import cython
+
+    def func(x: List[cython.int]) -> cython.long:
+        ...
+
+Static Python type checkers would understand the "List" type coming from
+the typing module and could check that the type of the input argument is
+always a Python list, but would ignore the unknown "cython.int" item type
+and thus not validate the items type of this list.  The Cython compiler,
+on the other hand, would understand both namespaces and could use it to
+generate corresponding C code.
 
 To mark portions of the program that should not be covered by type
-hinting, use the following:
+hinting at all, use the following:
 
 * a ``@no_type_checks`` decorator on classes and functions
 


gvanrossum:
I don't like the tuple notation. I see several options that you might pursue:

- decorators; the type checker can understand these (in most cases)
- use custom annotations and don't run the type checker
- use custom annotations and disable the type checker per module or per function using a magic comment (see #16)

Also, it would be more useful to open an issue for discussion before proposing a textual change (unless the change in non-controversial, which you know this isn't). And, "typing.integer" isn't a thing.

scoder:
I thought it was ok to propose a text change as a basis for discussion. Pull requests are quite cheap and they're essentially just issues with changes.

I actually think that tuples are ok as long as their content is unambiguous, which is what I'm trying to assure with the namespace requirement. Lists (i.e. square brackets) might look a little nicer syntactically, but wouldn't catch the idea of different kinds of annotations as much as tuples do. The alternative would be dicts, but they are very verbose and that verbosity is not needed if each item in the tuple is semantically well defined by its namespaced type. I think tuples are as simple and straight forward as it can get.

Are you suggesting that interoperability isn't even worth the overhead of a tuple? What's your opinion on an extensible type system? Don't you think it'll be necessary?

gvanrossum:
BTW, Steven D'Aprano wrote something pretty compelling on python-ideas. Maybe we can use some of his language in the PEP:

"""
In the past I have strongly defended the idea that type-hinting
annotations need to co-exist with other uses for annotations. By this I
mean that there should be a simple way for the author of a module who
uses annotations for something else to flag that module, or parts of the
module, so that the type-checker skips it.

There is no good way to have multiple uses of annotations be used in the
same function. You have suggested using a dict, but what of the module
that wants to give a completely different meaning to dicts as
annotations? Your interpetation of annotations is not compatible with
that module. Like multiple inheritence, multiple use of annotations is
only possible if all parties cooperate and agree on semantics.

Besides, the more information you try to squeeze into the function
parameter list, the more unwieldy, unreadable and ugly it gets.

Type-hinting was Guido's original motivation for introducing
annotations, and there is no other use of annotations which has become
popular or widespread enough to justify calling it a "standard use". So
in the absense of any other standard use of annotations, I am
comfortable with giving type-hints special status:

- whether type-checking is enabled by default or not, we understand that
  annotations are primarily for type-hinting, and any annotations are to
  be understood as type-hints by default;

- other uses for annotations are permitted, but it is up to the user to
  flag the module/class/function so the type-checker skips it;

- no provision is made for mixing type-hints and arbitrary other uses in
  the same annotation (in other words, if you want to use annotations
  for type-hinting and Foo *at the same time*, you're on your own);

- but of course, nothing stops you from creating your own custom
  annotation scheme that includes Foo + type-hints.

If some other use of annotations becomes wildly successful, then it
may be worth rethinking the special status of type-hinting in the
future.

If you want to interoperate with type-hinting, decorators may be a
second-class solution, but they are a solution.
"""

gvanrossum:
@scoder:
> I thought it was ok to propose a text change as a basis for discussion.
> Pull requests are quite cheap and they're essentially just issues with changes.

The problem with a PR is that it obscures the essence of the proposal with a lot of editing details. That may make sense for code (where trying to explain it in English is often more verbose than showing the code) but not for text (where the diff is significantly more verbose than just the proposal, clearly worded). Also, a PR doesn't show the text in the same "stream" window as the rest of the discussion.

> I actually think that tuples are ok as long as their content is unambiguous, [...]

Well, I really don't like them. And I *want* type hints to become the primary use of annotations.

> Are you suggesting that interoperability isn't even worth the overhead of a tuple?
> What's your opinion on an extensible type system? Don't you think it'll be necessary?

I'm not sure what you mean by an extensible type system. Clearly you're not referring to the ability to define new types. But then what? New constructs like Intersection (#18) or Protocols (#11)? Those can be discussed and added in a future version; it's hard to see how you could add those without changes to the type checker, and an "escape" notation like you propose isn't going to help. And sticking documentation in an annotation (like your `doc()` example) isn't extending the type system, it's an attempt to add something completely different to annotations. That just sounds like a bad idea to me -- you can only cram so much in the annotation for an argument before it just becomes a jumble.

The-Compiler:
I'm one of the users of custom annotations (to export functions to the user as "commands" they can invoke in my application), and I've tried to follow the discussions so far.

> but of course, nothing stops you from creating your own custom
annotation scheme that includes Foo + type-hints.

How would that work? I'd imagine something like a decorator `@typing.custom_annotation(callable)`, where *callable* is a function which gets the annotation of a parameter, and returns the type for the checker to use. That way, having a dict is easy (`@typing.custom_annotation(operator.itemgetter('pytype'))` for example, if I'm not mistaken), and more complex things are possible.

I don't like the idea of being unable to use the type checker for functions with custom annotations, i.e. only being able to skip things via a decorator. I'd prefer to make them work using a decorator.

gvanrossum:
The problem with your proposal is that it's not enough to specify how you would get the type annotation for a parameter -- you must specify it *in a way that the static type checker can use*. The static type checker reads your code but doesn't ever import it. IMO it's unreasonable to expect the type checker to understand ``operator.itemgetter('pytype')`` and expect then the annotation to be a dict from which is picks the `''pytype"` key as the type. IOW you're still thinking of the type as something to be conjured up somehow at runtime -- while actually it is intentional that the type checker looks at your source code without any help from the runtime.

I'm not sure what Steven D'Aprano meant with "Foo + type-hints" but I don't see a reasonable interpretation except that it refers to you writing your own type checker, which you can make behave however you want to.

The-Compiler:
That is a good point. Then I'd still be in favour of a dict-syntax being supported optionally. That'd work for me (I currently use dicts), but indeed, I'm not sure how useful that'd be universally.

An example of how I use the annotations:

```python
    @cmdutils.register(...)
    def scroll_perc(self, perc: {'type': float}=None,
                    horizontal: {'flag': 'x'}=False,
                    count: {'special': 'count'}=None):
```

This is indeed rather heavy to read, but very convenient to me. This would expose a command with a `perc`-argument which is verified to be a float, a `horizontal`-flag with `x` as shorthand, and a `count` argument which is automatically filled with the count the user entered (think vim commands).

gvanrossum:
But are you really interested in running mypy over your code and over the application code that is calling your commands? How does an application even call those commands? How doe the non-type-hint annotations come into play? If an application does a dynamic lookup of commands in your library and then calls them, the type checker isn't useful anyway -- the type checker only works for (shall we say) "static" calls where the name of the called function (or method) is in the caller's code. Example:
```
from somewhere.something import whatever
def main():
    whatever('Hello', 'world')
```
can easily be type-checked (assuming there are annotations on ``whatever()``). However, in the following example, the static checker can do nothing, even though the exact same function is called (and the same annotations appear at runtime):
```
import importlib
def main():
    something = importlib.import_module('somewhere.something')
    whatever = getattr(something, 'whatever')
    whatever('Hello', 'world')
```
Now, if your use case looks more like the former, type annotations on whatever *are* useful. You can do two things: either add type annotations directly in the source code for somewhere/something.py, or you can create a stub file that is seen by the static checker but not at run time. If you use a stub file, you can still use annotations for your own nefarious (:-) purposes, but you must disable type checking own code (i.e. something.py) using a per-method or per-class decorator or a "# type: OFF" comment. Only the application will be type-checked (and only if the app developer runs mypy). If you want your own code to be checked by mypy too (or primarily), you will have to give up on using annotations for your own flags -- you're going to have to invent some decorator. (Because I still don't believe that the ugliness of supporting your dict proposal is worth it.)

The-Compiler:
Indeed, I just realized the same while catching up with the thread on python-ideas. Those functions are indeed called dynamically, so it'd probably suffice to have a "don't check this" decorator.

Am I right when I assume the decorator has to be an actual *decorator* above the function so it can be inferred statically - i.e. I couldn't apply the decarator in my `@cmdutils.register` decorator?

gvanrossum:
Yeah, see #51.

gvanrossum:
Sorry, should have closed this back in January.

Issue #35, opened 2015-01-08
The use of comments to communicate with the type checker


gvanrossum:
mypy supports comments of the form `# type: <some_type>` and there are proposals for other uses of comments (e.g. `# typing: off`). Comments are a convenient solution for things that are hard or inefficient to express in existing Python syntax (i.e. code that works at runtime). But they have the downside that they are inaccessible to runtime machinery that might want to use type annotations, and the Python AST module doesn't preserve comments, so code that needs access to the comments must implement its own parser (like mypy does).

I ask: should we define such comments in the PEP or not?

JukkaL:
One of the most common cases in mypy code where annotations are needed is when creating an empty collection. I like it better with '# type:' comments (and mypy contributors seem to agree):

```
x = []  # type: List[int]

# vs

x = Undefined(List[int])
x = []
```

Using `cast()` is another option, but it's semantically different, since it can hide type errors (see https://github.com/ambv/typehinting/issues/15). 

If we absolutely decide that comments are a no-no, another option would be to define an additional helper that is similar to `cast()` but that can't be used to hide type errors. It could be called `coerce`, for example:

```
x = coerce(List[int], [])
```

All of the non-comment alternatives have the problem that especially complex types can introduce non-trivial runtime overhead, which would go against the goals of mypy.

gvanrossum:
I have to agree, and I don't see much value in forcing the type expression
to be evaluated at run time. So at least # type: comments should probably
be defined by the PEP.

On Wednesday, January 7, 2015, Jukka Lehtosalo <notifications@github.com>
wrote:

> One of the most common cases in mypy code where annotations are needed is
> when creating an empty collection. I like it better with '# type:' comments
> (and mypy contributors seem to agree):
>
> x = []  # type: List[int]
>
> # vs
>
> x = Undefined(List[int])
> x = []
>
> Using cast() is another option, but it's semantically different, since it
> can hide type errors (see #15
> <https://github.com/ambv/typehinting/issues/15>).
>
> If we absolutely decide that comments are a no-no, another option would be
> to define an additional helper that is similar to cast() but that can't
> be used to hide type errors. It could be called coerce, for example:
>
> x = coerce(List[int], [])
>
> All of the non-comment alternatives have the problem that especially
> complex types can introduce non-trivial runtime overhead, which would go
> against the goals of mypy.
>
> —
> Reply to this email directly or view it on GitHub
[...]

PR #46, opened 2015-01-17
Replace comments with string literals

flying-sheep:
As outlined in #35, both `assert isinstance(…)` and syntactically relevant comments have severe drawbacks.

Parallel to docstrings, using string literals for inline type hinting provides hints that eliminate those disadvantages:

1. they appear in the standard  Python AST and are therefore easier to retrieve (no 3rd party AST necessary)
2. no significant runtime cost (like `assert isinstance(…)` or `Undefined(…)` have)
3. equally lightweight syntax as comments (triple quouted or single quoted string literals instead of comment)

Also they are already used in e.g. PyCharm’s type hinting.

Open question: Do we prefer

```python
'''type: List[Employee]'''
x = []
```

or

```python
'type: List[Employee]'
x = []
```

### DIFF
--- a/pep-0484.txt
+++ b/pep-0484.txt
@@ -260,9 +260,9 @@ hinting, use the following:
 
 * a ``@no_type_checks`` decorator on classes and functions
 
-* a ``# type: ignore`` comment on arbitrary lines
+* a ``'''type: ignore'''`` string literal on arbitrary lines
 
-.. FIXME: should we have a module-wide comment as well?
+.. FIXME: should we have a module-wide string literals as well?
 
 
 Type Hints on Local and Global Variables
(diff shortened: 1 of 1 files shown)


gvanrossum:
It's a clever hack. However it takes up an extra line, whereas the `# type: ...` comment goes on the same line. I think PyCharm would be quite happy to switch to PEP 484 type comments (it's written in Java so I presume they have their own "forgiving" parser anyways). @JukkaL or @ambv, any comments on this proposal?

gvanrossum:
Please don't submit PRs for ideas still under active discussion.

flying-sheep:
> Please don't submit PRs for ideas still under active discussion.

oh, sorry, i didn’t know about this policy

JukkaL:
I don't really like this proposal, because a type annotation will take an extra line and it's less obviously related to the variable, so it doesn't read as well in my opinion. Using `Undefined(...)` is almost as concise, at least for short variable names, though it's less efficient:

```
'type: List[Employee]'
# vs
x = Undefined(List[Employee])
x = []
```

Mypy allows `Undefined` with a string literal argument. This way it's a little more efficient when using complex types and it would still be there in the AST. For example:

```
x = Undefined('List[Employee]')
x = []
```

Issue #12, opened 2014-10-16
Should we encourage abstract types?


gvanrossum:
In principle yes, in practice there are issues (long type names, strings and bytes being iterables and sequences).

ambv:
Yes, variables and attributes have concrete types and it's all good. Function return types might want to be concrete or abstract, depending on who defines the interface (see: Protocols). __add__ in lists is a good example to show how this might be an issue. Can mypy discover this? If not, we might want to warn about this subtlety.

gvanrossum:
Łukasz, I think you can close this issue once you add a sentence to the PEP text.

flying-sheep:
hmm, i don’t like non-abstract types here.

what advantages does the existence of List, DIct, Set, and FrozenSet have?

why not just use consistent abstract classes: in shell-expand: `{Mutable,}{Sequence,Mapping,Set}`

gvanrossum:
Remember that Python is pretty pragmatic. List has a sort() method, which MutableSequence does not (and should not) have. List+List works, while Sequence+Sequence doesn't. Etc.

flying-sheep:
maybe we should add new mixin methods that address this?

`Sequence` could define `__add__`, and `MutableSequence` could define `sort` (should be no problem as it already does define `reverse`)

alternatively, we could add new generic abstract classes that define the functionality not covered by the current ABCs, but not bound to concrete builtin types.

abarnert:
`[1] + (2, 3)` is a `TypeError`. I don't think you're suggesting to change that. (Otherwise, `['abc'] + 'def'` would succeed, presumably returning `['abc', 'd', 'e', 'f']`, and I doubt you want that?)

Of course you could define `Sequence.__add__` requires the `other` parameter to be covariant with `self` instead of contravariant (although I don't think there's any way to write that in the current syntax, but it's actually a general problem with `__add__` and other operator methods that are effectively multiple dispatch…). But that still wouldn't help. If a user writes `f() + [1, 2]`, and `f` is only defined to return `MutableSequence` instead of `list`, it will fail the check.

Also, a mixin method for `sort` would be a pretty ugly thing, because anyone calling `x.sort()` is presumably expecting an in-place sort, but the only way `MutableSequence.sort` could be implemented i something like `self[:] = sorted(self)`.

So, I think it's reasonable that methods should be able to specify concrete return types like `list`, or `List[int]`.

abarnert:
I have a different worry here.

Everyone seems to agree with this:

[...]

PR #47, opened 2015-01-17
Specify that ABCs should be used in signatures

flying-sheep:
As discussed in #12, concrete types are useful as return values or inline variables, but bad in function signatures, as most functions will e.g. loop over their sequence arguments, or index into it, instead of mutating them by invoking `sort` on them (`sorted` works on `Sequence`s anyway!) or concatenating them via `+`.

It’s not part of the Zen Of Python, but [the robustness principle](https://en.wikipedia.org/wiki/Robustness_principle) is good to adhere to:

> Be conservative in what you send, be liberal in what you accept

### DIFF
--- a/pep-0484.txt
+++ b/pep-0484.txt
@@ -317,6 +317,12 @@ holds a set of classes representing builtin types with generics, namely:
   case the following arguments are considered the same type as the last
   defined type on the tuple.
 
+.. attention::
+   None of these is intended for use in function or method signatures.
+   Unless the function needs methods that ``collections.abc`` does not
+   provide, function signatures should always use the abstract classes
+   ``Sequence``, ``Mapping``, ``AbstractSet`` and so on.
+
 It also introduces factories and helper members needed to express
 generics and union types:
 


gvanrossum:
The currently proposed wording is too strong. There are many, many situations where requiring a programmer to think about which abstract type to use is just not worth their time. Practicality beats purity.

flying-sheep:
OK. any specific phasing you’d like? How about:

> These types are mainly for use as return values or inline. In case you
> intend to use a collection in function signatures, please consider using
> one of the abstract types like ``Sequence``, ``Mapping``, or ``AbstractSet``.

and what rST directive is the best in your eyes? there are (in the order i think they are most fitting):

*    attention
*    hint
*    note
*    important
*    tip
*    caution
*    warning
*    danger
*    error

gvanrossum:
Thanks, I think we'll borrow most of your words. I don't want to start using any of those rST callouts (this is not end user documentation, and this particular issue isn't normative). No need to submit another PR for this.

flying-sheep:
OK, great!

gvanrossum:
I committed something inspired by your words. :-)

flying-sheep:
Sounds great, thanks!

PR #66, opened 2015-03-26
Clarifications for cast and overload

ceridwen:
Two small suggestions:

1) I'm assuming cast is not a new built-in, in which case it should be imported in the example.

2) I'd suggest the utility functions should also be included in the namespace section.


gvanrossum:
Thanks, I solved this and a number of similar items in revision df02fc3b40b600e4bf7779fad5b99b4c0c096744.

Issue #97, opened 2015-04-22
Additional use case for stub files


JimJJewett:
Some of the nightmare signatures are correct, and even desirable in certain circumstances.  They're just too noisy for easy reading.

As long as separate files for the interface are available (and can even be supplied by 3rd parties, such as IDE developers), that seems like a good place to put the extra-verbose annotations.

gvanrossum:
I wonder if in some cases it would make sense to use Any for the arguments (or something simple like `Iterable[Any]`) but specify the return type more carefully. That way the imprecise stubs won't catch type errors on the call site but at least they won't prevent further analysis using the result.

JimJJewett:
There are absolutely cases where it makes sense for the primary source file to annotate only the return type.  Sometimes the type of the parameters seems obvious by the name (or local convention), but the return type would otherwise require reading the function. 

In some of those cases, an IDE might appreciate an interface file that is explicit about the parameter types anyhow, even if that would be noise for the human reader.

gvanrossum:
What's the action item here?

PR #104, opened 2015-04-28
Jim J Jewett stubs for extra precision

JimJJewett:


JimJJewett:
This covers the original point of Issue #97 
I'm not sure what the action item should be about Guido's additional questions, so I'll just let those questions about partial signatures settle for a while.

gvanrossum:
I don't think the list of use cases needs to be exhausted. I also believe the readability issue is overblown. So rejecting this edit.

PR #117, opened 2015-05-20
Clarify how constants are set at runtime.

The-Compiler:
I think "these always reflect the current platform" is too ambiguous, as there are many subtleties in how platforms can be compared. This makes this more clear in the PEP.


gvanrossum:
Thanks! I committed a slight variation.

PR #118, opened 2015-05-20
Always call stub files stub (not interface) files.

The-Compiler:
There's one place where stub files are called interface files (which was a proposed alternative which was ignored/rejected - though I prefer stub anyways).

Either way, they should be called the same everywhere to avoid confusion.

### DIFF
--- a/pep-0484.txt
+++ b/pep-0484.txt
@@ -158,8 +158,8 @@ third-party modules).
 
 While annotations are normally the best format for type hints,
 there are times when it is more appropriate to represent them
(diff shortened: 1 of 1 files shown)


The-Compiler:
There's a typo in my name (Florina -> Florian) :wink:

But I have to admit it's slightly amusing!

gvanrossum:
Fixed, pushed using "git push -f" which may cause you some trouble when you pull. :-)

Thanks again for your careful reading of the spec!

PR #141, opened 2015-06-07
Added "Roadmap".

guettli:


gvanrossum:
Hi Thomas,

I'm sorry, I think it is not necessary to mention this in the PEP. Maybe it can be added to the README file of the typeshed repo. I've added a note about this to that repo's tracker: https://github.com/JukkaL/typeshed/issues/10

matthiaskramm:
Added a note in typeshed's README.

Issue #134, opened 2015-05-22
Should all parameters in Generic be enforced distinct?


ilevkivskyi:
Currently PEP says: Each type variable argument to Generic must be distinct. This is
thus invalid  `class Pair(Generic[T, T]):  ...`

However, this is not enforced, one could create such class without any warning. Is this a bug?
If yes, one could add following to `__getitem__` in `GenericMeta` at the end of block considering `self.__parameters__ is None`:
```python
if len(set(params)) != len(params):
    raise TypeError("All type variables in Generic[...] must be distinct.")
```

There is a related question: Should be two `TypeVar`s with the same name, constraints, etc. considered equal? For example, if I have:
```python
T = TypeVar('T')
T1 = TypeVar('T')
```
then is it allowed to write `class Pair(Generic[T1, T]):  ...`? If no, then one could add something like
```python
def __hash__(self):
    return hash((self.__name__, self.__covariant__,
                         self.__contravariant__, self.__bound__,
                         self.__constraints__))

def __eq__(self, other):
    if not isinstance(other, TypeVar):
        return NotImplemented
    return (self.__name__ == other.__name__ and
                self.__covariant__ == other.__covariant__ and
                self.__contravariant__ == other.__contravariant__ and
                self.__bound__ == other.__bound__ and
                self.__constraints__ == other.__constraints__)
```
or similar to `TypeVar`.

ilevkivskyi:
I have found the following in GENERICS.rst:
"""
Note that the argument to TypeVar() must be a string, and it must be assigned to a variable with exactly that name. Type variables cannot be redefined in the same module. These constraints are enforced by the type checker (but not by the runtime implementation of TypeVar())
"""
I think it is important, since it means that the example above with `T1 = TypeVar('T')` is not allowed. Maybe add it to the PEP?

gvanrossum:
The first thing (Generic[T,T] is invalid) is a check that I plan to fix eventually. The PEP already mentions this.

The second thing (T1 = TypeVar('T') is invalid) is unrealistic to check for. It should be added to the PEP.

gvanrossum:
I've added words to the PEP describing restrictions on type variables (rev 71bf400). The check against Generic[T,T] is still to do.

ilevkivskyi:
I don't think it requires a separate issue, but currently `typing.py` contains this
```python
class Mapping(Sized, Iterable[KT], Container[KT], Generic[KT, VT_co],
              extra=collections_abc.Mapping):
[...]

PR #144, opened 2015-06-17
Prohibit Generic[T,T]

ilevkivskyi:
A simple-minded fix that prohibits Generic[T, T] and makes a minor change discussed in the same issue #134 .

### DIFF
--- a/prototyping/test_typing.py
+++ b/prototyping/test_typing.py
@@ -557,8 +557,11 @@ class GenericTests(TestCase):
 
     def test_init(self):
         T = TypeVar('T')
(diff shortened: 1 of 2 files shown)


gvanrossum:
I apologize, is this still up to date?

ilevkivskyi:
As you said in #134 , "I've added words to the PEP describing restrictions on type variables (rev 71bf400). The check against Generic[T,T] is still to do." I just tried `Generic[T, T]` with the latest version and it still works. Also the commits after Jun 17 should not interfere with this PR.

Therefore, yes it looks like it is still up to date.

PR #153, opened 2015-08-04
Use `__slots__` in `typing` module.

o11c:
This is primarily useful for subclassing `Generic[T]`, but it's easiest
to do it all at once.

### DIFF
--- a/prototyping/typing.py
+++ b/prototyping/typing.py
@@ -128,6 +128,8 @@ class TypingMeta(type):
 class Final:
     """Mix-in class to prevent instantiation."""
 
+    __slots__ = ()
+
     def __new__(self, *args, **kwds):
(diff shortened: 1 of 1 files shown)


gvanrossum:
I'm sorry, what problem is this solving? Too often, `__slots__` is a symptom of premature optimization.

o11c:
@gvanrossum 

I agree that `__slots__` is *often* a premature optimization. But sometimes it is not.

The current state of the `typing` module is that if you want to write your own generic classes, you *cannot* use the optimization.

gvanrossum:
OK, agreed. This matches collections.abc. I'll merge.

o11c:
@JukkaL It's merged now. Note that there were already other changes (mostly deletions) when I applied it.

Issue #160, opened 2015-10-15
Make 'typing' package Python 2 compatible


vlasovskikh:
The current `typing` module doesn't support Python 2 as well as the third-party `backports.typing` module. It's a good idea to have the source code of the `typing` module installed on your system when you type check Python 2 code, e.g. against Python stub files with types. It will simplify things for a static type checker.

I guess we could create a backported version of `typing`. I mean something similar to 3.5.0b1 with all the recent API changes. The 2/3 single source approach won't work here because the module heavily relies on metaclasses and uses type hints in many places.

I will start working on a backport in a separate branch.

JukkaL:
Mypy has a somewhat functional Python 2 port of `typing` as well (https://github.com/JukkaL/mypy/blob/master/lib-typing/2.7/typing.py), though others could be more feature-complete (haven't checked).

PR #162, opened 2015-10-23
Fix Python 3.2 compatibility

### LINKED ISSUE #161: Cannot install typing==3.5.0 for Python 3.2
The problem is with the Unicode literals in setup.py:

```
    Complete output from command python setup.py egg_info:
      File "<string>", line 20, in <module>
      File "/tmp/pip-mzzueb-build/setup.py", line 43
        author=u'Guido van Rossum, Jukka Lehtosalo, Łukasz Langa',
                                                                 ^
    SyntaxError: invalid syntax
```

I'll fix it and add a tox configuration for running tests for Python 3.2-3.4.

vlasovskikh:
This fixes #161.


vlasovskikh:
Please review it.

tox.ini
```diff
@@ -1,2 +1,12 @@
+[tox]
+envlist = py32, py33, py34
+
+[testenv]
+changedir = src
+commands = python -m unittest discover
+
+[testenv:py32]
+deps = mock
```
o11c: `mock` is not actually used, it would be better to just remove the import from `test_typing`.

tox.ini
```diff
@@ -1,2 +1,12 @@
+[tox]
+envlist = py32, py33, py34
```
o11c: I know the pypi package is *currently* unneeded for python 3.5, but I think that tox should test with it.

Python 3.6 is certain to ship with a newer version of `typing`, and people running 3.5 will want to install it then in order to use the new classes.

Also, possibly comment about the 2.7 possibility (see #160).

Also enable `pypy3` (and once #160 is fixed, `pypy` and maybe `jython` - I don't have an environment installed for jython or ironpython but *ideally* they would both be supported, though they might need completely custom versions of `typing`)

.gitignore
```diff
@@ -1,7 +1,9 @@
 MANIFEST
 build/
 dist/
+.tox/
 .idea/
 __pycache__/
+tmp/
```
o11c: What is `tmp/` needed for?

gvanrossum:
Andrey, you can commit this yourself, right? Just do it.

tox.ini
vlasovskikh: @o11c Thanks for spotting this.

tox.ini
vlasovskikh: @o11c We can add other Python implementations eventually. Here I just wanted to fix the problem with 3.2 and establish a way to test typing against various versions. Supporting 2.7 is the next thing I am going to do.

.gitignore
vlasovskikh: @o11c It's the directory for my temporary scripts for running tox with my local parameters and options. Not sure they worth sharing.

o11c: That really belongs in `.git/info/exclude` then.

PR #166, opened 2015-10-28
Add Travis CI config file

vlasovskikh:
I suggest running tox tests automatically via the Travis CI that integrates well with GitHub.

@ambv Could you set up a build configuration at Travis CI for this repository? You're the owner of the repository and it seems only you can do it.

Then we could add the Travis badge to the README file to browse the results of automatic builds.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,5 +1,6 @@
 language: python
 env:
+  - TOXENV=py27
   - TOXENV=py32
   - TOXENV=py33
   - TOXENV=py34
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,10 +1,16 @@
+sudo: false
 language: python
-env:
-  - TOXENV=py27
-  - TOXENV=py32
-  - TOXENV=py33
-  - TOXENV=py34
 install:
   - pip install tox
 script:
   - tox
+matrix:
+  include:
+    - python: "2.7"
+      env: TOXENV=py27
+    - python: "3.2"
+      env: TOXENV=py32
+    - python: "3.3"
+      env: TOXENV=py33
+    - python: "3.4"
+      env: TOXENV=py34
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,0 +1,9 @@
+language: python
+env:
+  - TOXENV=py32
+  - TOXENV=py33
+  - TOXENV=py34
+install:
+  - pip install tox
+script:
+  - tox


kirbyfan64:
Instead of using Tox, you can just use different Python versions like [Hy does](https://github.com/hylang/hy/blob/master/.travis.yml). That way, at a glance Travis will always say which versions passed and which ones failed. You could do something like:

```yaml
sudo: false # Use the container infrastructure for faster builds
language: python
python: # Different Python versions
  - pypy
  - "2.7"
  - "3.2"
  - "3.3"
  - "3.4"
script:
  - "[ $TRAVIS_PYTHON_VERSION = 2.7 ] && cd python2 || cd src"
  - python -m unittest discover
```

Now, if a build fails, the overview screen will show the failing Python versions.

I set up Travis for Mypy and downloaded custom PyPy versions for Hy, so I'm pretty experienced in this part. :D

vlasovskikh:
It looks like a good idea. Is there any way we could reuse our tox config for particular Travis environments like "2.7"? I'm aware of a way to run tox for a specific environment:

```bash
tox -e py27
```

Maybe we could do something like this:

```
...
script:
  - tox -e ${TRAVIS_PYTHON_VERSION}  # We need to drop dots or use a map from Travis Python names to tox names
```

kirbyfan64:
I think you can do:

```
- tox -e py`echo ${TRAVIS_PYTHON_VERSION} | sed 's/\.//'`
```

o11c:
bash can do substitutions without sed: `py${TRAVIS_PYTHON_VERSION/./}`

[but I specify the `TOX_ENV`s manually via `matrix: include` rules, since it's more obvious that way, and also supports pypy.](https://github.com/o11c/mypy/blob/master/.travis.yml)

vlasovskikh:
@o11c `matrix` looks like the best option for our case.

vlasovskikh:
I've tested this .travis.yml file for my own clone, it works OK. Builds are failing for Python 2.7 because this branch doesn't include the latest commits from master.

vlasovskikh:
We still need some help from @ambv with setting up the build configuration at the Travis website.

vlasovskikh:
@ambv Could you please set up a build configuration at the Travis CI server? Since it's your repository you are the only person who can do it. We've prepared the config file in this PR.

gvanrossum:
I did something different and simper, since I'm unfamiliar with tox. Closing this one.

PR #168, opened 2015-12-01
Add Async classes to typehinting.

rowillia:


vlasovskikh:
The corresponding PR with the stubs is https://github.com/python/typeshed/pull/30.

o11c:
What's the upgrade plan, since Python 3.5 has been released *without* this, and the standard library takes precedence over third-party installs?

gvanrossum:
PEP 484 and typing.py are provisional, so we can add it to 3.5.2 (it's too
late for 3.5.1 though, rc1 already went out).

gvanrossum:
So I merged this into Python 3.5 and 3.6, but after I was done I realized
that the tests are now broken for Python 3.2-3.4 (because the added tests
use the async/await keywords). Would you be willing to make a new PR that
moves those tests (and only those) to a separate file? We're using the same
pattern for asyncio, where there's a separate file test_pep492.py
containing tests that are syntactically invalid in earlier Python versions.
(Because it's syntax, you can't use @skip decorators.)

rowillia:
@gvanrossum sure thing.

vlasovskikh:
We really need some help from @ambv to set up a build configuration at the Travis CI server so we can catch errors like these earlier. We've prepared the config file in #166.

gvanrossum:
I think @ambv has run out of time for this project. Maybe we can convince him to transfer ownership to https://github.com/python/?

PR #190, opened 2016-03-22
Annotation of arbitrary argument lists

### LINKED ISSUE #185: The PEP doesn't say how types for *args, **kwds work
The PEP doesn't explicitly say that you can write
```
def foo(*args: str, **kwds: int): ...
```
and that this means that e.g. the following calls are valid:
```
foo('a', 'b')
foo(x=1, y=2)
foo('', z=0)
```

ilevkivskyi:
Add short description to the PEP about type hinting of arbitrary argument lists, fixes #185


gvanrossum:
Thanks -- great!

PR #194, opened 2016-03-24
Use collections.abc.Reversible if it exists

### LINKED ISSUE #172: Reversible should reference collections.abc.Reversible, if it exists
See http://bugs.python.org/issue25987

ilevkivskyi:
Fixes #172 . I assumed that if Reversible is present in collections.abc then it means that http://bugs.python.org/issue25987 is resolved as Andrew Barnert proposes (as I understand @gvanrossum agrees with this), i.e., Sequence is subclass of Reversible, and not directly of Iterable.

### DIFF
--- a/python2/test_typing.py
+++ b/python2/test_typing.py
@@ -527,7 +527,7 @@ class ProtocolTests(TestCase):
 
     def test_protocol_instance_type_error(self):
         with self.assertRaises(TypeError):
-            isinstance([], typing.Reversible)
+            isinstance(0, typing.SupportsAbs)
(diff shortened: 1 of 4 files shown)


gvanrossum:
I don't think that CPython issue has been fixed, so this is just a fantasy, right. I just put a new message in there asking Andrew if he's still interested.

ilevkivskyi:
@gvanrossum I decided not to wait for Andrew and submitted a patch for http://bugs.python.org/issue25987 following his proposal.

Issue #129, opened 2015-05-21
Confusing __class__ for instance of non-generic class that subclasses generic class


JukkaL:
The following program prints `__main__.A[int]` even though `A` is not generic. I'd expect it to print something like `<class '__main__.A'>`.

```
from typing import Iterator

class A(Iterator[int]):
    def __next__(self): pass

print(A().__class__)
```

Also, this does not generate an exception, even though `A` is not generic:

```
def f(x: A[int]) -> None: pass
```

gvanrossum:
I'm on the fence about calling this a bug. What I implemented in rev 25af77debdff7f5d0b7cf2a167ead6023789165f is that no matter how you instantiate A, you always get the original class A as the instance's `__class__`. And that is what's happening here. It just so happens that the actual class A remembers its parameters. Regarding using `A[int]` later, that probably shouldn't be allowed -- I need to rework the logic here. This is also related to issue #115. I will have to come back to this later.

gvanrossum:
OK, fixed by #195.

PR #195, opened 2016-03-25
Revamped generic class behavior to conform to updated PEP 484

### LINKED ISSUE #115: Decide meaning of alias with one generic and one concrete parameter
See https://github.com/JukkaL/mypy/issues/606. We allow the following, and they are well-defined:
```
from typing import TypeVar, Dict
T = TypeVar('T')
U = TypeVar('U')
X = Dict[int, str]
Y = Dict[T, U]
```
But what does this mean?
```
Z = Dict[T, int]
```
In typing.py this currently defines Z as something with two parameters, the second constrained to int or subclasses thereof. But another reasonable interpretation would be that Z has *one* parameter, corresponding to T only.

The second interpretation is actually more reasonable. (As long as we require that all parameters are either concrete or a simple type variable, not another type expression or alias containing a type variable.)

If we agree, we should define this in the PEP and implement it in typing.py.

gvanrossum:
Fixes #115

### DIFF
--- a/python2/test_typing.py
+++ b/python2/test_typing.py
@@ -684,6 +684,36 @@ class GenericTests(TestCase):
 
         assert C.__parameters__ == (VT, T, KT)
 
+    def test_nested(self):
+
+        G = Generic
+
+        class Visitor(G[T]):
+
+            a = None
+
+            def set(self, a):
+                self.a = a
+
+            def get(self):
+                return self.a
+
+            def visit(self):
+                return self.a
+
+        V = Visitor[typing.List[int]]
+
+        class IntListVisitor(V):
+
+            def append(self, x):
+                self.a.append(x)
+
+        a = IntListVisitor()
+        a.set([])
+        a.append(1)
+        a.append(42)
+        assert a.get() == [1, 42]
+
     def test_type_erasure(self):
         T = TypeVar('T')
 
--- a/python2/test_typing.py
+++ b/python2/test_typing.py
@@ -1032,6 +1032,15 @@ class NamedTupleTests(TestCase):
         assert Emp._fields == ('name', 'id')
         assert Emp._field_types == dict(name=str, id=int)
 
+    def test_pickle(self):
+        global Emp  # pickle wants to reference the class by name
+        Emp = NamedTuple('Emp', [('name', str), ('id', int)])
+        jane = Emp('jane', 37)
+        for proto in range(pickle.HIGHEST_PROTOCOL + 1):
+            z = pickle.dumps(jane, proto)
+            jane2 = pickle.loads(z)
+            self.assertEqual(jane2, jane)
+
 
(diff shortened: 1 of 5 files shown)


ilevkivskyi:
It looks great! Still I have found something that I think could be confusing. 

First, I would expect that `Iterable[T][int]` and `Iterable[int]` is the same type, but `Iterable[T][int] == Iterable[int]` gives me `False`.

Second (maybe related) issue is that now `__repr__` shows the whole "history" of a type. For example, `Iterable[Tuple[T,T]][int]` is printed as `__main__.Iterable<+T_co>[__main__.Tuple[~T, ~T]]<~T>[int]`. This reads as first there was an `Iterable` parameterized by `T_co`, then it was indexed by `Tuple[T, T]` and we had something parameterized by `T`, and then it was indexed by `int`. The question is whether we need to expose all this information (that is rather internal) to the user? I think it would be more clear if `__repr__` in this case will show something like `Iterable[Tuple[int,int]]`.

gvanrossum:
Ah, yes. I thought of making `Iterable[T][int] == Iterable[int]`, but I ran out of steam on how to implement it. Then again, thinking about it I'm not sure that it's important to do this -- the runtime behavior of these things is of marginal interest and mypy doesn't let you write that anyway (the PEP is silent on whether this is allowed).

Ditto for repr(). The one important thing about repr() is that when you have
```
class C(Generic[T]): ...
```
then repr(C) shows `C<~T>` rather than `C[~T]` as it used to. I think the new form is better. And then `C[Tuple[int, int]]` will actually be printed as `C<~T>[Tuple[int, int]]`. The longer form (with alternating `<...>` and `[...]`) is only of theoretical importance.

Anyway, thanks for the encouragement. I hope others look over this important change too! I intend to land this in Python 3.5.2 (whenever it is released).

ilevkivskyi:
I am now thinking about one additional thing that might be not completely clear: covariance. If I initially have `Iterable[T_co]` covariant in its type parameter and write
```python
class C(Iterable[Tuple[T,S]]):
    ...
```
then one could be not sure whether `C` is covariant in its variables or invariant. I think the natural answer is that covariance is determined only by the final list of type parameters, so that the class above is invariant in both variables. However, I didn't find anywhere in the PEP any explicit explanation of this point. Maybe it is obvious and it is only my problem of understanding, but if not then maybe one should add a short explanation to the PEP?

gvanrossum:
Good point. IMO nothing having to do with variance is obvious, so it would
be good to spell it out. Right now I'm too tired to reason this through
with a few examples, but my intuition says that your "natural" answer is
probably correct. When I feel more rested I'll figure this out and add it
to the PEP.

ilevkivskyi:
I just noticed that this PR also fixes #129 . In general, I think this new "revamped" behavior is much better than the old one.

PR #199, opened 2016-04-08
Port ContextManager from cpython repository

brettcannon:
https://hg.python.org/cpython/rev/841a263c0c56

### DIFF
--- a/src/test_typing.py
+++ b/src/test_typing.py
@@ -1,3 +1,4 @@
+import contextlib
 import pickle
 import re
 import sys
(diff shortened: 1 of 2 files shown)


gvanrossum:
Wonder if this should also be copied into the 3.5 branch -- even though it has no effect there, I kind of like all copies of typing.py and test_typing.py to be identical. (At least as long as it is provisional -- and not counting the PY2 backport which isn't affected by this at all.)

brettcannon:
OK, I'll apply it to the 3.5 branch as well when I have a chance.

brettcannon:
https://hg.python.org/cpython/rev/257dd57dd732 has the backport to sync 3.5 and default in python/cpython.

PR #201, opened 2016-04-15
Consider ellipsis in TupleMeta.__eq__

bintoro:
Before:

```python
>>> Tuple[int] == Tuple[int, ...]
True
```

After:

```python
>>> Tuple[int] == Tuple[int, ...]
False
```


gvanrossum:
Good catch -- can you also update the Python 2 backport?

bintoro:
Done!

gvanrossum:
@bintoro I have one request: Can you create a username at bugs.python.org and submit a contributor form? It's online: https://www.python.org/psf/contrib/contrib-form/

bintoro:
It's now marked as received (http://bugs.python.org/user23780)

Issue #136, opened 2015-05-22
Kill __subclasscheck__


gvanrossum:
Mark Shannon wants me to drop `__subclasscheck__` from all type objects (Any, Union etc.). This is somewhat major surgery since it is used by Union simplification and for various type checks. See also #133 and #135.

ilevkivskyi:
This is a bit sad. People would want to do some interactive experimentation to see what is `issubclass(int, Union[int, str])`, but probably there are some good reasons behind this decision.

1st1:
Maybe it is a good thing, at least for 3.5.  This will effectively force type objects to be used only in annotations, which gives us more room to tweak their behaviour before 3.6.

ilevkivskyi:
Maybe one could introduce a helper function `acceptable(obj, hint)` in typing.py that tells whether `obj` is OK for type `hint`. Like `acceptable(42, Union[int, str]) == True`.

gvanrossum:
Here's what Mark wrote about the subject:
```
 My remaining concern is the isinstance and issubclass support.
There is a real different difference between whether a value is a member of a type and whether it is an instance of a class.

Support for isinstance and issubclass applied to types just breeds confusion.

As you said, for most users, the distinction between types and classes is subtle. But I think it is important to make people aware that there is a distinction, even if they fully understand it.

For example,
List[int] and List[str] and mutually incompatible types, yet
isinstance([], List[int]) and isinstance([], List[str))
both return true.

Just delete all the __instancecheck__ and __subclasscheck__ methods and I'll be happy.
```

1st1:
> Maybe one could introduce a helper function acceptable(obj, hint) in typing.py that tells whether obj is OK for type hint. Like acceptable(42, Union[int, str]) == True.

I think that we should leave the task of validating types to external tools (and compilers), not user code. So I'd be -1 on this.

gvanrossum:
The equivalent of `issubclass()` should probably be called `is_consistent_with()` (or some variation of that) to match the terminology used by Jeremy Siek in his Gradual Typing blog post (http://wphomes.soic.indiana.edu/jsiek/what-is-gradual-typing/). I'm not sure what to call the `isinstance()` equivalent, perhaps `is_acceptable_for()`. But I plan to ask others to contribute such functions.

ilevkivskyi:
The intended use would be not to validate types in code but just do some quick interactive tests in REPL. But OK, maybe you are right people anyway will start using it for validation.

[...]

PR #205, opened 2016-04-18
Specific asserts

zware:
This PR converts the tests from simple `assert` statements to `TestCase.assert*` calls.  This ensures that the tests actually test something even if Python is run with -O, and makes the tests fit the style of the rest of the standard library tests.

Most of the changes were made by simple `s/assert .../self.assert.../` replacements, with a few manual touchups.

### DIFF
--- a/src/test_typing.py
+++ b/src/test_typing.py
@@ -131,7 +131,7 @@ class TypeVarTests(TestCase):
         # T is a subclass of itself.
         self.assertIsSubclass(T, T)
         # T is an instance of TypeVar
-        assert isinstance(T, TypeVar)
+        self.assertIsInstance(T, TypeVar)
 
     def test_typevar_instance_type_error(self):
         T = TypeVar('T')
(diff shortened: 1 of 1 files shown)


gvanrossum:
I would actually prefer going the other way -- change the remaining `self.assertEqual(x, y)` calls to `assert x == y`. I mostly run the tests using py.test and it much prefers the latter idiom.

Does anyone routinely run the stdlib tests with -O?

zware:
> Does anyone routinely run the stdlib tests with -O?

Not that I know of currently, but I would like for there to be buildbots
running with both -O and -OO, which should be easily doable if I rewrite
the build master as I recently mentioned on python-dev.

src/test_typing.py
```diff
@@ -20,6 +20,25 @@
 import typing
 
 
+class TestCase(_TestCase):
+
+    # vanilla TestCase doesn't have assert*Subclass, see bugs.python.org/14819
```
gvanrossum: Looks like it's not going to happen any time soon. I propose not to count on it. The rename dance TestCase -> _TestCase feels kind of awkward -- why not just pick a decent new name for TestCase + assert[Not]IsSubclass, and use that where it's needed.

gvanrossum:
OK, I concede your point (even though I don't like noisy commits -- you will end up owning all those lines in hg/git blame forever). Once this is merged here, can you take care of merging it into 3.5+3.6?

src/test_typing.py
zware: Right, the comment was not meant to imply that `assertIsSubclass` would eventually be a standard feature.  The rename dance was an effort to avoid churn, but as you pointed out, the whole PR is pretty much churn so I'll just change it to BaseTestCase.

zware:
My ownership of the lines is unfortunate, agreed.

Yes, I can sync `cpython` after merge here.

bintoro:
Note that, once #136 is resolved, most `issubclass` assertions will have to be changed again.

gvanrossum:
OK, go ahead and merge into CPython 3.5 and 3.6.

zware:
Thanks, Guido.  I've merged into cpython 3.5 and default.

The only remaining question is whether you want `python2/test_typing.py` similarly treated.  I have no opinion about it, but am willing to put forth the effort to do it if you want to keep the diff between `src` and `python2` smaller.

gvanrossum:
Yea, please do the python2/test_typing.py -- I do try to keep them in sync. And there are plenty of PRs coming up.

PR #206, opened 2016-04-20
Python2 specific asserts

zware:
As discussed in #205, here's the specific asserts backported to `python2/test_typing.py`.


gvanrossum:
Thanks again!

PR #207, opened 2016-04-20
Fix generics inheritance and drop __subclasscheck__

### LINKED ISSUE #136: Kill __subclasscheck__

### LINKED ISSUE #202: Inheritance issues with generic collections
I have a problem with how `__extra__` is currently inherited:

```python
>>> class CustomMapping(typing.Mapping): ...
... 
>>> issubclass(dict, CustomMapping)
True
```

```python
>>> class CustomDict(typing.Dict): ...
... 
>>> issubclass(dict, CustomDict)
True
```

This can't be right.

My initial impression is that `__extra__` should only be propagated on parameterization, not when subclassing, and that the concrete types `List`, `Dict`, etc. should have their own extras (`list`, `dict`, and so on). Thoughts? A tentative patch is [here](https://github.com/python/typing/compare/master...bintoro:extra-inheritance).

Another thing: The module distributed on PyPI still has the issue where a generic doesn't include itself in the subtype's MRO during parameterization. This was fixed for Python 3.5.1 in aab2c59. It would be great if we could get a more recent revision on PyPI as well.

3.5.1:

```python
>>> issubclass(Iterable[int], Iterable)
True
```

3.5.0:

```python
>>> issubclass(Iterable[int], Iterable)
False
```

And finally, a question: The `collections` ABC replacements do not provide the abstract (or any other) methods of the originals. In particular, `Iterable` is technically not abstract at all. Is this intentional or just a temporary state of affairs?

### LINKED ISSUE #203: typing.MutableMapping not a drop-in replacement for collections.abc.MutableMapping

```
import collections.abc
import typing

    class MyUserDict(base_class):
        def __getitem__(self, k):
            return None
        def __setitem__(self, k, v):
            pass
        def __delitem__(self, k):
            pass
        def __iter__(self):
            return iter(())
        def __len__(self):
            return 0
    return MyUserDict

```

```
    UserDict2().update()
```

bintoro:
I don't expect this to be reviewed anytime soon, but the effort is now fairly complete, so might as well put it out there for comments.

Resolves #136
Resolves #202
Resolves #203

At the heart of #136 is a conflation between type compatibility and subtype relationships. Even though `Sequence[Manager]` is compatible with `Sequence[Employee]`, it's not a subclass of it. Rather, both derive from `Sequence[+T_co]`.

To address this, the PR introduces a new helper `is_compatible()`, which implements the "is consistent with or a subtype of" relation. In short, it does what `issubclass()` has been doing until now.

The practical end result is that

* `issubclass()` is no longer applicable to the special types `Any`, `Union`, etc.

* `is_compatible()` always works — if neither argument is a special type, `is_compatible(T1, T2)` reduces to `issubclass(T1, T2)`

* `issubclass()` acquires its normal semantics w.r.t. generics:

    ```python
    >>> issubclass(Sequence[Manager], Sequence[Employee])
    False
    >>> issubclass(Sequence[Manager], Sequence)
    True
    ```

   whereas:

    ```python
    >>> is_compatible(Sequence[Manager], Sequence[Employee])
    True
    ```

#### Commit summary

There are three commits:

* The first simply stops the `__extra__` linkage from leaking down to generics' subclasses. This fixes the bug where `issubclass(dict, CustomMapping) == True`. [[Diff](a22b68e...2ef2a50)]

* The second adds the `__extra__` ABCs as actual base classes of their `typing` twins. It also refactors `GenericMeta.__subclasscheck__` into a compatibility check and a `__subclasshook__` dedicated to handling the `__extra__` connection. [[Diff](2ef2a50...0ec4dd2)] This achieves a few things:

  * User-defined generics become recognized as subclasses of the associated `collections` ABCs.
  * Abstract and mixin methods from `collections.abc` become available.
  * `issubclass(list, Iterable[X])` no longer returns true for arbitrary `X` (#136).

* The final commit is largely a mechanical change: the special types are made to raise a `TypeError` in response to `issubclass()`. Every type metaclass gets a method named `__is_compatible__` to host the prior `__subclasscheck__` logic. [[Diff]](0ec4dd2...fd4a1b0)

#### Notes

* Full parity is preserved in using unparameterized types in place of the corresponding `collections.abc` classes in `isinstance()` and `issubclass()` calls:

    ```python
    issubclass(type, typing.Callable) == issubclass(type, collections.Callable) == True
    issubclass(list, typing.Iterable) == issubclass(list, collections.Iterable) == True
    ...
    ```

* The intent has been not to mess with the types' behavior. That said, the introduction of `is_compatible()` yields a couple of automatic improvements regarding `Any`:

    * The relation evaluates to true when either argument is `Any`; previously `issubclass(Any, C)` would return false for some arbitrary class `C` but true for a `typing` type.
    * `is_compatible(Iterable[Any], Iterable[int])` correctly returns true.

* The merging of #205 caused a huge conflict in the tests. I will look into it eventually if there's interest in merging this.

* The proposal is compatible with the plan to convert some of the collection types to protocols, if that's still on the table.

### DIFF
--- a/src/test_typing.py
+++ b/src/test_typing.py
@@ -1,3 +1,4 @@
+import abc
 import contextlib
 import collections
 import pickle
@@ -14,6 +15,7 @@ from typing import Callable
 from typing import Generic
 from typing import cast
 from typing import get_type_hints
+from typing import is_compatible
 from typing import no_type_check, no_type_check_decorator
 from typing import NamedTuple
 from typing import IO, TextIO, BinaryIO
@@ -81,15 +83,15 @@ class AnyTests(TestCase):
     def test_any_is_subclass(self):
         # Any should be considered a subclass of everything.
         assert issubclass(Any, Any)
-        assert issubclass(Any, typing.List)
-        assert issubclass(Any, typing.List[int])
-        assert issubclass(Any, typing.List[T])
-        assert issubclass(Any, typing.Mapping)
-        assert issubclass(Any, typing.Mapping[str, int])
-        assert issubclass(Any, typing.Mapping[KT, VT])
-        assert issubclass(Any, Generic)
-        assert issubclass(Any, Generic[T])
-        assert issubclass(Any, Generic[KT, VT])
+        assert is_compatible(Any, typing.List)
+        assert is_compatible(Any, typing.List[int])
+        assert is_compatible(Any, typing.List[T])
+        assert is_compatible(Any, typing.Mapping)
+        assert is_compatible(Any, typing.Mapping[str, int])
+        assert is_compatible(Any, typing.Mapping[KT, VT])
+        assert is_compatible(Any, Generic)
+        assert is_compatible(Any, Generic[T])
+        assert is_compatible(Any, Generic[KT, VT])
         assert issubclass(Any, AnyStr)
         assert issubclass(Any, Union)
         assert issubclass(Any, Union[int, str])
@@ -794,13 +796,13 @@ class VarianceTests(TestCase):
     def test_invariance(self):
         # Because of invariance, List[subclass of X] is not a subclass
         # of List[X], and ditto for MutableSequence.
-        assert not issubclass(typing.List[Manager], typing.List[Employee])
-        assert not issubclass(typing.MutableSequence[Manager],
-                              typing.MutableSequence[Employee])
+        assert not is_compatible(typing.List[Manager], typing.List[Employee])
+        assert not is_compatible(typing.MutableSequence[Manager],
+                                 typing.MutableSequence[Employee])
         # It's still reflexive.
-        assert issubclass(typing.List[Employee], typing.List[Employee])
-        assert issubclass(typing.MutableSequence[Employee],
-                          typing.MutableSequence[Employee])
+        assert is_compatible(typing.List[Employee], typing.List[Employee])
+        assert is_compatible(typing.MutableSequence[Employee],
+                             typing.MutableSequence[Employee])
 
     def test_covariance_tuple(self):
         # Check covariace for Tuple (which are really special cases).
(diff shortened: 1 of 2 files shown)


ilevkivskyi:
I like this PR, it looks like it could fix many issues. But I have some questions:

1. You have two different functions `is_consistent` and `is_compatible`. As I understand we need only one function that should answer whether assigning a value with type `t1` to a variable with declared/inferred type `t2` considered an error or not. Do I understand correctly that `is_compatible` supposed to do exactly this, and why do you need a second function?

2. Maybe you should update your tests taking into account more of this: https://github.com/python/typing/pull/205 (maybe at least use `self.assertIsInstance` on lines 1247-1250, etc.)

3. This class definition `class List(list, MutableSequence[T], extra=list)` and the same for `set, frozenset, dict, defaultdict` violate DRY (you need to repeat the base class in extra every time), and, to be honest, look a bit ugly. Is it possible to somehow simplify this?

bintoro:
@ilevkivskyi The short answer to 1 and 3 is that, when it comes to a few details, the PR represents a WIP due to the exact requirements not being known.

1.  `is_consistent` is just a helper performing an `Any`-aware identity comparison. It's quite possible that it will simply be folded into `is_compatible`. It should be marked as private, but for all I know, neither function might end up as part of the module's public API. (I'm not aware of a decision laying out what, if any, type checking features `typing.py` should provide.)

2.  I wrote 99% of the PR before #205 was submitted, so the tests are based on an earlier revision. I can certainly update them to follow the `unittest` style, but that alone won't fix the conflict.

3.  The duplicates in the definitions of `List` & co. are only necessary to make the first commit a self-contained patch that can work without the rest. They are there to satisfy the test that checks for `issubclass(list, List) == True`.

    In the context of the full PR, the equality becomes very questionable. And even if we do want it to hold, the duplication won't be necessary. (The conversion of extras into bases would no longer skip builtins [here](https://github.com/python/typing/pull/207/files#diff-15eca194f829b962c9d102b4ac8ab378R1010).)

gvanrossum:
Superseded by #283. Many of the ideas here went into that PR, and it closes most of the same issues. Thanks!

PR #214, opened 2016-05-06
Be more precise about types vs. classes

ilevkivskyi:
Also remove the mention of issubclass in the description of Union, and add a note about this for other special constructs.

### DIFF
--- a/pep-0484.txt
+++ b/pep-0484.txt
@@ -819,11 +819,10 @@ Example::
           e = [e]
       ...
 
-A type factored by ``Union[T1, T2, ...]`` is compatible with
-``T1`` and any of its subtypes, ``T2`` and
-any of its subtypes, and so on. In other words, values of all types
-``T1``, ``T2``, etc., and their subtypes are acceptable for
-an argument annotated by ``Union[T1, T2, ...]``.
+A type factored by ``Union[T1, T2, ...]`` is a supertype
+of all types ``T1``, ``T2``, etc., so that a value that
+is a member of one of these types is acceptable for an argument
+annotated by ``Union[T1, T2, ...]``.
 
 One common case of union types are *optional* types.  By default,
 ``None`` is an invalid value for any type, unless a default value of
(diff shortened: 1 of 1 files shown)


pep-0484.txt
```diff
@@ -819,9 +819,11 @@ Example::
...
       ...
 
-A type factored by ``Union[T1, T2, ...]`` responds ``True`` to
-``issubclass`` checks for ``T1`` and any of its subtypes, ``T2`` and
-any of its subtypes, and so on.
+A type factored by ``Union[T1, T2, ...]`` is compatible with
+``T1`` and any of its subtypes, ``T2`` and
+any of its subtypes, and so on. In other words, values of all types
+``T1``, ``T2``, etc., and their subtypes are acceptable for
+an argument annotated by ``Union[T1, T2, ...]``.
```
gvanrossum: We haven't defined "compatible" previously. Maybe you can avoid it and just use "supertype"? Or briefly define it here? (The IOW doesn't really do that IMO. :-)

pep-0484.txt
```diff
@@ -1294,6 +1296,11 @@ collections (e.g. ``List``), types representing generic
 collection ABCs (e.g. ``Sequence``), and a small collection of
 convenience definitions.
 
+Note that special type constructs, such as ``Any``, ``TypeVar``,
+``Union``, and ``Generic`` are only supported in the type annotation
+context and will raise ``TypeError`` if appear in ``isinstance``
+or ``issubclass``.
```
gvanrossum: Actually TypeVar is a very different construct, though maybe type variables defined *using* TypeVar are indeed such special constructs. Also, Generic is not supported in a type annotation context, it may only be used as a base class.

gvanrossum:
Could you also review PEP 483? It's referenced from PEP 484 and while it isn't part of the standard it intends to explain the theory behind things without being so distracted by the mechanics of introducing typing into Python.

ilevkivskyi:
I included your proposed changes. I tried to make the definition of `Union` as simple as possible, so that I used supertype.

It is good that you mentioned PEP 483. Actually, I already started working on extending PEP 483 along the following lines:
- explain type systems in simple but precise words
- clarify classes vs types
- explain shortly why gradual typing is a good solution for Python
- extend the discussion of generic types
- add an extensive explanation of variance of generic types

I think I could make a PR soon (probably next week).

gvanrossum:
Thank! I've merged this downstream (in the peps hg repo) too.

If you're going to edit PEP 483, the best approach may be to copy that here first and then to build on that. I'll do that now.

gvanrossum:
(Done, and for good measure I've added PEP 482 too.)

ilevkivskyi:
Thanks, indeed, it is a very good idea to have all the related PEPs here.

PR #216, opened 2016-05-07
Corrections according to comments

ilevkivskyi:
I decided to make this PR so that it is easy to check all the corrections (also it is easier for me).

### DIFF
--- a/pep-0483.txt
+++ b/pep-0483.txt
@@ -333,7 +333,7 @@ are still controversial or not fully specified.)
 
     def get(key: KT, default: VT = None) -> VT: ...
 
-- Type variables could be declared in unconstrained, constrained,
+- Type variables can be declared in unconstrained, constrained,
   or bounded form. The variance of a generic type could also
   be indicaded using a type variable declared with special keyword
   arguments, thus avoiding any special syntax, e.g.::
--- a/pep-0483.txt
+++ b/pep-0483.txt
@@ -333,6 +333,20 @@ are still controversial or not fully specified.)
 
     def get(key: KT, default: VT = None) -> VT: ...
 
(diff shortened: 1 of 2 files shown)


pep-0483.txt
```diff
@@ -333,6 +333,20 @@ are still controversial or not fully specified.)
 
     def get(key: KT, default: VT = None) -> VT: ...
 
+- Type variables could be declared in unconstrained, constrained,
+  or bounded form. The variance of a generic type could also
+  be indicaded using a type variable declared with special keyword
+  arguments, thus avoiding any special syntax, e.g.::
```
gvanrossum: Could -> can

gvanrossum:
What's up with the commit history?

ilevkivskyi:
> What's up with the commit history?

That is a good question. I am not sure. But I have found a Stackowerflov question on this, so that I hope next time I will make everything clean.

ilevkivskyi:
Replaced could with can. This time two commits because I simply have overlooked the second "could", sorry for that

gvanrossum:
Maybe use rebase?

ilevkivskyi:
I just did `reset --hard upstream/master` and it looks like that fixed everything, I see all my commits in history and github reports that `python:master and ilevkivskyi:master are identical`

bintoro:
@ilevkivskyi I see you're using pull requests to bring your own repo up to date. If you do that, then the commit histories will always diverge. Generally you want to pull from upstream into your local repo, then push to origin to update.

ilevkivskyi:
@bintoro Exactly, this is what I have found on Stackoverflow, will do this starting from now.

Issue #107, opened 2015-05-08
Can we have a generic Type[C]?


NYKevin:
The `type` object occupies a rather strange place in the type hierarchy:

    >>> type(type) is type
    True

(I'm pretty sure that's a flat lie, since you can't instantiate something from itself, but regardless...)

    >>> isinstance(type, object)
    True
    >>> isinstance(object, type)
    True
    >>> isinstance(type, type)
    True

In Java, the (very rough) equivalent is a class, specifically [`Class<T>`](http://docs.oracle.com/javase/7/docs/api/java/lang/Class.html).  It's also generic; the type variable refers to the instance.  Java has it easy because they don't support metaclasses.  Classes are not first class in Java, so their type hierarchy doesn't have to deal with the strange loops shown above.

I realize metaclasses in their full generality are out of scope (at least for now), but a generic `Type[T]` like Java's would be nice to have.  So far as I can tell from the Mypy documentation, it doesn't currently exist.

Here's some example code which might like to have this feature:

    def make_foo(class_: Type[T]) -> T:
        # Instantiate and return a T

gvanrossum:
Unless this is trivial to add to mypy I propose to punt this to a future version of type hints.

FWIW type(X) is a shorthand for X.__class__ and type.__class__ is indeed type. Not a lie.

NYKevin:
Yeah, but I'm pretty sure you have to reassign `type.__class__` at some point to get there...

Anyway, I recognize that this is probably a lot more complex than it looks, and that the first 3.5 beta is rather alarmingly close, but I do feel this is essential for anyone using first-class classes in any significant way.  Otherwise, you just have a `type` object, which is basically `Callable[[???], object]` (any idea what to put for the args?).  You'd need a lot of casting to make that useful.  I mean, I suppose you can treat it like `Type[Any]`/`Callable[Any, Any]`, but then you're just being too permissive instead of too strict.

gvanrossum:
`Any` is your friend. :-)

agronholm:
Here's an example use case, straight from the code of my new framework:
```python
@typechecked
def register_extension_type(ext_type: str, extension_class: type, replace: bool=False):
    """
    Adds a new extension type that can be used with a dictionary based configuration.

    :param ext_type: the extension type identifier
    :param extension_class: a class that implements IExtension
    :param replace: ``True`` to replace an existing type
    """

    assert_subclass('extension_class', extension_class, IExtension)
    if ext_type in extension_types and not replace:
[...]

PR #218, opened 2016-05-13
Add text about Type[C].

gvanrossum:
This addresses #107 (but it doen't close it because the issue also
calls for an implementation in typing.py).

### DIFF
--- a/pep-0484.txt
+++ b/pep-0484.txt
@@ -910,17 +910,21 @@ example calling e.g. ``new_user(Union[BasicUser, ProUser])`` is
 rejected by the type checker (in addition to failing at runtime
 because you can't instantiate a union).
 
-Note that it *is* legal to use a union as the parameter for
+Note that it is legal to use a union of classes as the parameter for
 ``Type[]``, as in::
 
   def new_non_team_user(user_class: Type[Union[BasicUser, ProUser]]):
       ...
 
 However the actual argument passed in at runtime must still be a
-concrete class object, e.g.::
+concrete class object, e.g. in the above example::
 
   new_non_team_user(ProUser)  # OK
-  new_non_team_user(TeamUser)  # Disallowed
+  new_non_team_user(TeamUser)  # Disallowed by type checker
+
+``Type[Any]`` is also supported (see below for its meaning).  However,
+other special constructs like ``Tuple`` or ``Callable`` are not
+allowed.
 
 There are some concerns with this feature: for example when
 ``new_user()`` calls ``user_class()`` this implies that all subclasses
--- a/pep-0484.txt
+++ b/pep-0484.txt
@@ -919,6 +919,7 @@ indicated base class (``User`` in the example above) should be
 allowed.  A program containing a complex or extensible class hierarchy
 might also handle this by using a factory class method.
 
+When ``Type`` is parameterized it requires exactly one parameter.
 Plain ``Type`` without brackets is equivalent to using ``type`` (the
 root of Python's metaclass hierarchy).  This equivalence also
 motivates the name, ``Type``, as opposed to alternatives like
--- a/pep-0484.txt
+++ b/pep-0484.txt
@@ -861,11 +861,11 @@ allow all operations on it, and a value of type ``Any`` can be assigned
 to a variable (or used as a return value) of a more constrained type.
 
 
-Meta-types
-----------
+The type of class objects
+-------------------------
 
-A meta-type can be used to indicate a value that is itself a class
-object that is a subclass of a given class.  It is spelled as
+Sometimes you want to talk about class objects, in particular class
+objects that inherit from a given class.  This can be spelled as
 ``Type[C]`` where ``C`` is a class.  To clarify: while ``C`` (when
 used as an annotation) refers to instances of class ``C``, ``Type[C]``
 refers to *subclasses* of ``C``.  (This is a similar distinction as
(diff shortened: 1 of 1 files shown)


pep-0484.txt
```diff
@@ -861,6 +861,73 @@ allow all operations on it, and a value of type ``Any`` can be assigned
 to a variable (or used as a return value) of a more constrained type.
 
 
+Meta-types
+----------
```
ilevkivskyi: I think "Meta-types" is not a good name for this section for four reasons:

1. This could add to the classes vs. types confusion: Unlike metaclasses that are quite different from classes, types do not have such hierarchy, since one should not type annotate types, one can only type annotate classes.

2. ``Type[C]`` has normal semantics of type: This is just a set of objects that respond ``True`` to ``issubclass(objc, C)``.

3. I think someone who is looking for a documentation of this feature will not look for "Meta-types", therefore I would propose to change this to a simpler title, like "Type annotating class objects".

4. Finally, using such terminology could lead to proliferation of "meta": If we call annotations of values that are class objects by ``Type[C]`` or by ``ABCMeta`` meta-types, then how are we supposed to call annotations of values that are metaclasses (by ``Type[ABCMeta]`` for example)? Meta-meta-types? :-)

pep-0484.txt
```diff
@@ -861,6 +861,73 @@ allow all operations on it, and a value of type ``Any`` can be assigned
...
+
+A meta-type can be used to indicate a value that is itself a class
```
ilevkivskyi: Following the comment above, I would replace this by "An additional fundamental building block can be used to indicate a value that..." Putting ``Type`` in a row with other building blocks like ``Tuple`` etc.,will  emphasize, that although ``C`` must be a class, ``Type[C]`` is a normal type that can be used elsewhere:
```python
registry = [] # type: List[Tuple[int, Type[FrameworkBase]]]
```

pep-0484.txt
```diff
@@ -861,6 +861,73 @@ allow all operations on it, and a value of type ``Any`` can be assigned
...
+  U = TypeVar('U', bound=User)
+  def new_user(user_class: Type[U]) -> U:
+      ...
+
+Now when we call ``new_user()`` with a specific subclass of ``User`` a
+type checker will infer the correct type of the result::
+
+  joe = new_user(BasicUser)  # Inferred type is BasicUser
+
+At runtime the value corresponding to ``Type[C]`` must be an actual
```
ilevkivskyi: I would remove "At runtime". Just "The value corresponding to..."

pep-0484.txt
```diff
@@ -861,6 +861,73 @@ allow all operations on it, and a value of type ``Any`` can be assigned
...
+class object that's a subtype of ``C``, not a special form.  IOW, in
+the above example calling e.g. ``new_user(Union[BasicUser, ProUser])``
+is not allowed.
```
ilevkivskyi: Maybe replace here "is not allowed" by "is rejected by the type checker"? Calling ``new_user(Union[BasicUser, ProUser])`` will trigger ``TypeError`` because it tries to instantiate ``Union``, but I think it is not what you wanted to say.

pep-0484.txt
```diff
@@ -861,6 +861,73 @@ allow all operations on it, and a value of type ``Any`` can be assigned
...
+``new_user()`` calls ``user_class()`` this implies that all subclasses
+of ``User`` must support this in their constructor signature.  However
+this is not unique to ``Type[]``: class methods have similar concerns.
+A type checker ought to flag violations of such assumptions, but by
+default constructor calls that match the constructor signature in the
+indicated base class (``User`` in the example above) should be
+allowed.  A program containing a complex or extensible class hierarchy
+might also handle this by using a factory class method.
+
+When ``Type`` is parameterized it requires exactly one parameter.
```
JukkaL: I would also discuss `Type[Any]`. `type` should likely be equivalent to `Type[Any]`, similar to `list` / `List[Any]`. It's not entirely obvious how `Type[Any]` should behave. How about this:

* It can be called with arbitrary arguments, and the return type is `Any`.
* It only provides attributes and methods defined in `type`. Arbitrary class methods and attributes aren't supported.

ilevkivskyi: @JukkaL I agree, but I would add a third option of having both behaviours that you proposed: ``Type[Any]`` behaves like described in the first option, while ``Type[object]`` behaves like described in the second option. Note also that if we want to discuss ``Type[Any]`` we probably need to change wording in the first paragraph to "where ``C`` is a class or ``Any``".

JukkaL:
Miscellaneous things:

* `Type[T]` should likely be covariant, as discussed before.
* Treatment of `__init__` signatures is going to be nontrivial, as discussed in the issue. Perhaps mention that may require a future PEP update? In particular, we discussed the variant `Type[[A, ...], C]` which would be like `Type[C]` but with `__init__` constrained to supporting positional arguments `A, ...`. This would be an alternative for `Callable` that is also a type object. Is this potentially going to be included in phase 2?

gvanrossum:
Sent another omnibus update; please review!

pep-0484.txt
```diff
@@ -861,6 +861,97 @@ allow all operations on it, and a value of type ``Any`` can be assigned
...
+
+  joe = new_user(BasicUser)  # Inferred type is BasicUser
+
+The value corresponding to ``Type[C]`` must be an actual class object
+that's a subtype of ``C``, not a special form.  IOW, in the above
+example calling e.g. ``new_user(Union[BasicUser, ProUser])`` is
+rejected by the type checker (in addition to failing at runtime
+because you can't instantiate a union).
+
+Note that it *is* legal to use a union as the parameter for
```
ilevkivskyi: As I understand, you mean here "a union _of classes_", since in general ``Union`` can accept ``Tuple``, ``Callable`` etc. If yes, then maybe it is better to mention this?

pep-0484.txt
```diff
@@ -861,6 +861,97 @@ allow all operations on it, and a value of type ``Any`` can be assigned
...
+``Type[]``, as in::
+
+  def new_non_team_user(user_class: Type[Union[BasicUser, ProUser]]):
+      ...
+
+However the actual argument passed in at runtime must still be a
+concrete class object, e.g.::
+
+  new_non_team_user(ProUser)  # OK
+  new_non_team_user(TeamUser)  # Disallowed
```
ilevkivskyi: This is probably a really minor comment, by I would replace "Disallowed" by "Fails type check" or similar, since at first I was confused that it is disallowed because ``TeamUser`` is not a concrete class object. But maybe it is just me.

gvanrossum:
Clarified both in 2192c76.

ilevkivskyi:
It looks good to me. I have no further comments.

JukkaL:
Looks good to me as well.

PR #219, opened 2016-05-14
Generics and variance sections of PEP 483

ilevkivskyi:
I expanded significantly these sections (mainly because of adding many examples), but I only described the points on which there are consensus and tried to omit unnecessary details.

### DIFF
--- a/pep-0483.txt
+++ b/pep-0483.txt
@@ -404,7 +404,7 @@ Type annotation of generic functions is allowed by *type valiables*.
 Their semantics with respect to generic types is somewhat similar
 to semantics of parameters in functions. But one does not assign
 concrete types to type variables, it is the task of a statis type checker
-to find thier possible values and warn the user if it cannot.
+to find their possible values and warn the user if it cannot.
 Example::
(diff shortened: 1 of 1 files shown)


gvanrossum:
Merged. That doesn't mean I've reviewed it -- I doubt I'll get to it any time soon, there just are too many other things I also have to do. I propose to add your name as co-author (assuming you don't mind) and call it a day. I will then push this to the PEPs repo and once it's live on python.org I'll Tweet it and hopefully some folks will read it. Sounds good?

What email can I use for you?

ilevkivskyi:
> I propose to add your name as co-author

This sounds great!

> I doubt I'll get to it any time soon, there just are too many other things I also have to do.

It would be great if you could review this at some point. You have managed to improve "Background" section significantly.

> What email can I use for you?

levkivskyi@gmail.com

gvanrossum:
Pushed. I'll watch it and once it's live I'll Tweet.

I will try to eventually review it -- at least for factual mistakes. ;-)

PR #224, opened 2016-05-19
Add Type[C] "implementation"

gvanrossum:
It turns out this is a fairly vanilla generic class inheriting from type.

Should the CT variable be declared covariant?

This is one of the items we need to do before closing #107

### DIFF
--- a/python2/test_typing.py
+++ b/python2/test_typing.py
@@ -1120,7 +1120,7 @@ class TypeTests(BaseTestCase):
         class ProUser(User): pass
 
         def new_user(user_class):
-            # type: (Type[User]) -> None
+            # type: (Type[User]) -> User
             return user_class()
 
         joe = new_user(BasicUser)
--- a/python2/test_typing.py
+++ b/python2/test_typing.py
@@ -14,6 +14,7 @@ from typing import Tuple
 from typing import Callable
 from typing import Generic
 from typing import cast
+from typing import Type
 from typing import NamedTuple
 from typing import IO, TextIO, BinaryIO
 from typing import Pattern, Match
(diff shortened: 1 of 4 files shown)


python2/test_typing.py
```diff
@@ -1110,6 +1111,36 @@ def __len__(self):
...
+class TypeTests(BaseTestCase):
+
+    def test_type_basic(self):
+
+        class User(object): pass
+        class BasicUser(User): pass
+        class ProUser(User): pass
+
+        def new_user(user_class):
+            # type: (Type[User]) -> None
```
jstasiak: @gvanrossum This may be a stupid question but shouldn't this (and https://github.com/python/typing/pull/224/files#diff-f9eeb4530dadc70935c11fff001ec7acR1385) be marked as returning ``User``?

gvanrossum: It doesn't matter (mypy can't analyze typing.py) but you're right. Fixing...

jstasiak: Thanks, I just thought for a minute I was missing some crucial detail here.

JukkaL:
Similar to https://github.com/python/typeshed/pull/216#issuecomment-220676318, it might be better to create a special type for `Type[C]`, similar to `Callable`. I'm not sure if using `Generic` will result in any immediate problems, but we may have to reimplement it in the future if we decide to add support for `__init__` argument types, for example, and this could break stuff if people start assuming that `Type[]` is a subclass of `Generic`.

gvanrossum:
Yeah, here I'm just being lazy. The proper way requires an extra metaclass.
Maybe we can put a check in mypy that prevents misusing this? Or we could
just document "only use this in annotations".

Issue #119, opened 2015-05-20
Support coroutines


1st1:
Can we add support for coroutine functions?  Mainly, I'm curious how we will annotate the return type.  We basically have two options:

    async def foo() -> typing.Coroutine[str]:
         return 'spam'

or

    async def foo() -> str:
         return 'spam'

I'd really prefer the latter style.  While the result of 'foo()' is a Coroutine, this is something that type checkers will know without a hint, what really matters is the expected result of `await foo()` expression.

vlasovskikh:
The first style is better for `asyncio` coroutines of Python 3.4. There is `typing.Generator` type in the PEP 484 parameterized with in- and out- types as well as with the return type of the generator, so the coroutine type could be an alias `Coroutine = typing.Generator[Any?, Any?, V_co]`.

The second style is more convenient for `async def` functions. We could standardize the type annotation in the context of a `async def` to have a special meaning: it could be translated by a static checker to `Coroutine[...]` automatically.

1st1:
@vlasovskikh I agree.

We also need `typing.Awaitable[..]`, for cases where it is not (or cannot be) specified if we return a coroutine or a future:

    def op() -> typing.Awaitable[int]:
         if cond:
              return coro_func()
         else:
              return asyncio.Future(...)

    await op() # <- the result is of type int

gvanrossum:
I like the latter style. I'm labeling this "enhancement" because I don't want this to derail the PEP approval process. FWIW we should distinguish between the return annotation (which can be `-> str`) and the type of `foo()` (which should still be `Coroutine[str]`).

1st1:
Bumping this issue... 

Definitely not a high priority one, but it would be nice if we can have Awaitable and Coroutine types in the typing module.

gvanrossum:
We will at some time, and we can add it in 3.5.1, but I'm still on vacation
and out of round tuits for another couple of weeks...

On Mon, Aug 3, 2015 at 8:19 PM, Yury Selivanov <notifications@github.com>
wrote:

> Bumping this issue...
>
> Definitely not a high priority one, but it would be nice if we can have
> Awaitable and Coroutine types in the typing module.
>
> —
> Reply to this email directly or view it on GitHub
> <https://github.com/ambv/typehinting/issues/119#issuecomment-127360245>.
>



-- 
--Guido van Rossum (python.org/~guido)

vlasovskikh:
[...]

PR #225, opened 2016-05-24
Short subsection on annotating coroutines

ilevkivskyi:
Finishes #119 , mypy implementation is tracked in https://github.com/python/mypy/issues/1453

### DIFF
--- a/pep-0484.txt
+++ b/pep-0484.txt
@@ -1015,6 +1015,39 @@ In such cases the default value may be specified as a literal
 ellipsis, i.e. the above example is literally what you would write.
 
 
+Annotating generator functions and coroutines
+---------------------------------------------
+
+The return type of generator functions can be annotated by
+the generic type ``Generator`` provided by ``typing.py`` module::
+
+  def echo_round() -> Generator[int, float, str]:
+      res = yield
+      while res:
+          res = yield round(res)
+      return 'OK'
+
+Coroutines introduced in PEP 492 are annotated with the same syntax as
+ordinary functions. However, the return type annotation corresponds to the
+type of ``await`` expression, not to the coroutine type::
+
+  async def spam(ignored: int) -> str:
+      return 'spam'
+
+  async def foo():
+      bar = await spam(42) # type: str
+
+The ``typing.py`` module also provides generic ABCs ``Awaitable``,
+``AsyncIterable``, and ``AsyncIterator`` for situations where more precise
+types cannot be specified::
+
+  def op() -> typing.Awaitable[str]:
+      if cond:
+          return spam(42)
+      else:
+          return asyncio.Future(...)
+
+
 Compatibility with other uses of function annotations
 =====================================================
 
--- a/pep-0484.txt
+++ b/pep-0484.txt
@@ -1019,7 +1019,8 @@ Annotating generator functions and coroutines
 ---------------------------------------------
 
 The return type of generator functions can be annotated by
-the generic type ``Generator`` provided by ``typing.py`` module::
+the generic type ``Generator[yield_type, send_type,
+return_type]`` provided by ``typing.py`` module::
 
   def echo_round() -> Generator[int, float, str]:
       res = yield


pep-0484.txt
```diff
@@ -1015,6 +1015,39 @@ In such cases the default value may be specified as a literal
 ellipsis, i.e. the above example is literally what you would write.
 
 
+Annotating generator functions and coroutines
+---------------------------------------------
+
+The return type of generator functions can be annotated by
+the generic type ``Generator`` provided by ``typing.py`` module::
+
+  def echo_round() -> Generator[int, float, str]:
```
gvanrossum: From this example alone it's hard to see how [int, float, str] map to the various aspects of generators. Personally I use the fact that they were introduced historically at different times: yielded value (int), value returned by yield expression (float), value returned by return statement (str). But for others that may not be a great help...

ilevkivskyi: I agree, I added the type parameters in a new commit.

gvanrossum:
Thanks! I suspect there are still some organizational loose ends, e.g. there's a bullet list somewhere that claims to give everything exported by `typing.py` but I think we forgot to add some recent additions there.

ilevkivskyi:
Actually, ``Generator``, ``Awaitable``, ``AsyncIterable``, and ``AsyncIterator`` are already in this list. However, ``ContextManager`` is not in this list.

pep-0484.txt
```diff
@@ -1015,6 +1015,40 @@ In such cases the default value may be specified as a literal
...
+      return 'OK'
+
+Coroutines introduced in PEP 492 are annotated with the same syntax as
+ordinary functions. However, the return type annotation corresponds to the
+type of ``await`` expression, not to the coroutine type::
+
+  async def spam(ignored: int) -> str:
+      return 'spam'
+
+  async def foo():
```
gvanrossum: Whoa, I think this needs `-> None` otherwise the body won't be type-checked (despite the presence of type comments).

ilevkivskyi: In the initial version I had this `` -> None``, but then removed to simplify the example. If I understand correctly this sentence in the PEP "Any function without annotations should be treated as having the most general type possible, or ignored, by any type checker", then it is up to type-checker whether to type check the body of ``foo`` here. I do not have any preference here, if you think it is better to add ``-> None``, then should I make another PR (this one is already merged), of you will add it?

gvanrossum: I just updated the PEP. I think it's better to add, because the only known
example of a type checker does indeed ignore the body of unannotated
functions, so it's better to show examples that are fully annotated (except
when talking about annotated vs. unannotated).

On Thu, May 26, 2016 at 12:37 AM, Ivan Levkivskyi <notifications@github.com>
wrote:

> In pep-0484.txt
> <https://github.com/python/typing/pull/225#discussion_r64702492>:
>
> > +return_type]`` provided by ``typing.py`` module::
> > +
> > +  def echo_round() -> Generator[int, float, str]:
> > +      res = yield
> > +      while res:
> > +          res = yield round(res)
> > +      return 'OK'
> > +
> > +Coroutines introduced in PEP 492 are annotated with the same syntax as
> > +ordinary functions. However, the return type annotation corresponds to the
> > +type of ``await`` expression, not to the coroutine type::
> > +
> > +  async def spam(ignored: int) -> str:
> > +      return 'spam'
> > +
> > +  async def foo():
>
> In the initial version I had this -> None, but then removed to simplify
> the example. If I understand correctly this sentence in the PEP "Any
> function without annotations should be treated as having the most general
> type possible, or ignored, by any type checker", then it is up to
> type-checker whether to type check the body of foo here. I do not have
> any preference here, if you think it is better to add -> None, then
> should I make another PR (this one is already merged), of you will add it?
>
> —
> You are receiving this because you modified the open/close state.
> Reply to this email directly or view it on GitHub
> <https://github.com/python/typing/pull/225/files/30ec2751a96be037a02879b6c551b7a00d05a596#r64702492>
>



-- 
--Guido van Rossum (python.org/~guido)

Issue #189, opened 2016-03-21
Add NewType() to create simple unique types with zero runtime overhead


gvanrossum:
The description is best gleaned from the following mypy issue: https://github.com/python/mypy/issues/1284#issuecomment-199021176 and following. We're going with option (A).

Really brief example:
```python
from typing import NewType
UserId = NewType('UserId', int)
```
Now to the type checker UserId is a new type that's compatible with int, but converting an int to a UserId requires a special cast form, UserId(x). At runtime UserId instances are just ints (not a subclass!) and UserId() is a dummy function that just returns its argument.

ilevkivskyi:
I think this is a useful feature.

I have added PR #226  with a simple minded implementation of ``NewType``, some tests, and a short discussion in the PEP.

Also I added ``NewType``, ``Type`` and ``ContextManager`` to the list of things provided by ``typing.py``.

I do not discuss whether ``NewType('EURtoUSD', Callable[[float], float])`` etc. are allowed. If necessary, we could clarify this later.

gvanrossum:
I would like to limit the arg to a class. Also note that that class must have a constructor that is compatible with that assumed by the type checker -- because whenever you write `UserId(42)` the runtime will actually see `int(42)` -- which is fine, and ditto if you use e.g. strings, since `str('x')` is fine, but if you have your own custom class that you want to diversify this way, you have to make sure that your constructor accepts instances of itself -- even though the customary way of constructing instances might be something totally different. This won't work:
```
class PacketId:
    def __init__(self, major: int, minor: int) -> None:
        self._major = major
        self._minor = minor
    # Etc.

TcpPacketId = NewType('TcpPacketId', PacketId)

a = TcpPacketId(127, 0)
x = PacketId(100, 100)
b = TcpPacketId(x)  # ???
```

ilevkivskyi:
> I would like to limit the arg to a class.

OK

> This won't work: ...

Actually, at runtime, ``UserId(42)`` will be just ``42`` not ``int(42)``. So that actually ``b = TcpPacketId(x)``
will work, but ``a = TcpPacketId(127, 0)`` will not work. This is intentional (as discussed in mypy issue),
``TcpPackedId`` works as a cast function, that should be fed by ``PacketId``. Do I need to clarify this in the PEP?

gvanrossum:
[...]

PR #226, opened 2016-05-27
Add NewType

ilevkivskyi:
Simple minded implementation of ``NewType`` (see #189 and https://github.com/python/mypy/issues/1284#issuecomment-199021176) plus some tests and a short discussion in the PEP.
Also I added ``Type`` and ``ContextManager`` to the list of types provided by ``typing.py``.


bintoro:
@ilevkivskyi I know you're intent on maintaining the distinction between types and classes, so I think here "class" is the correct choice, no?

What about multiple inheritance? If there's any chance that it will be supported, perhaps rename the introspection attribute `__supertype__` in a way that accommodates that. I wonder if setting a faux `__bases__` would be too egregious.

Finally, from an introspection POV, an instance of a class with `def __call__(self, x): return x` would be easier to detect than a plain function while still avoiding the instantiation overhead.

gvanrossum:
@jukkal, what do you think of this design?

ilevkivskyi:
@bintoro 
> I know you're intent on maintaining the distinction between types and classes, so I think here "class" is the correct choice, no?

I am intentionally a bit "general" here, see my comment from the issue:
"I do not discuss whether ``NewType('EURtoUSD', Callable[[float], float])`` etc. are allowed. If necessary, we could clarify this later."

Concerning your two other points, the idea here is to be really minimal, to allow minimum overhead. In particular your idea with many base classes is almost equivalent to ``Intersection``, that might be implemented later (there is an issue on this).

gvanrossum:
Note, see my comment/concern I posted to the corresponding issue. I don't
think this is ripe yet...

pep-0484.txt
```diff
@@ -1173,6 +1173,46 @@ checker should blindly believe the programmer.  Also, casts can be used
 in expressions, while type comments only apply to assignments.
 
 
+NewType helper function
+-----------------------
+
+There are also situations, where a programmer might want to avoid logical
```
gvanrossum: No comma.

pep-0484.txt
```diff
@@ -1454,6 +1512,8 @@ Fundamental building blocks:
 
 * Generic, used to create user-defined generic classes
 
+* Type, used to annotate class objects
```
gvanrossum: This and the next one I'll apply right now. They're unrelated to this issue.

ilevkivskyi: OK, sure, I just posted a link to this PR to python-dev asking for comments.

pep-0484.txt
```diff
@@ -1212,6 +1212,24 @@ from ``UserId`` where ``int`` is expected. Examples::
 
         num = UserId(5) + 1     # type: int
 
+``NewType`` accepts only one argument that shoud be a proper class,
+i.e., not a type construct like ``Union``, etc. Function returned
```
gvanrossum: The Function

pep-0484.txt
```diff
@@ -1212,6 +1212,24 @@ from ``UserId`` where ``int`` is expected. Examples::
...
+by ``NewType`` accepts only one argument, this is equivalent to supporting
```
gvanrossum: Semicolon instead of comma.

pep-0484.txt
```diff
@@ -1212,6 +1212,24 @@ from ``UserId`` where ``int`` is expected. Examples::
...
+only one constructor accepting an instace of the base class (see above).
```
gvanrossum: instance

pep-0484.txt
```diff
@@ -1212,6 +1212,24 @@ from ``UserId`` where ``int`` is expected. Examples::
...
+      def __init__(self, major: int, minor: int) -> None:
+          self._major = major
+          self._minor = minor
+
+  TcpPacketId = NewType('TcpPacketId', PacketId)
+
+  packet = PacketId(100, 100)
+  tcp_packet = TcpPacketId(packet)  # OK
+
+  tcp_packet = TcpPacketId(127, 0)  # This fails
```
gvanrossum: Both in the type checker and at runtime, right?

JukkaL:
A few things:

* I'd add explicit language about `isinstance` not being supported for these types (and why not).
* What should we call types defined using `NewType`? Are they "`NewType` types"?

pep-0484.txt
ilevkivskyi: Yes, you are right.

I have made all corrections in a new commit.

ilevkivskyi:
@JukkaL 

> I'd add explicit language about isinstance not being supported for these types (and why not).

Good point. I added this in new commit

> I'd add explicit language about isinstance not being supported for these types (and why not).

Here I am not sure. Maybe ad-hoc types?

gvanrossum:
Let's think some more about the name, it feels important. Even if the
function is called NewType.

--Guido (mobile)

bintoro:
*Appointed types* — because they are only nominally distinct.

lgautier:
- `AdhocType`- because (presumably) just created for one situation 

- `NamedType` - to play on a naming similarity with `namedtuple`

yedpodtrzitko:
`LambdaType`

SpotlightKid:
TypeDef

sietse:
I like @lgautier's `NamedType` a lot: to me, it expresses "this is otherwise the same type, but it has a different name (and therefore its own meaning)". Also, like he says, it meshes nicely with `namedtuple` (and `typing.NamedTuple`).

How I would read it:
> `UserID = NamedType('UserID', int)` -> This is a type of `int` named `UserID`

----

If `NamedType` does not make it, what about `ValueType`?

* `ValueType`: Scala has the same concept, and calls them [Value Classes](http://docs.scala-lang.org/overviews/core/value-classes.html). It expresses nicely that this type does not say something about the data structure, but about the value it contains.

How I would read it:
> `UserID = ValueType('UserID', int)` -> This is an int with a user ID value type.

NamedType works better, though.

gvanrossum:
I could not follow anything from the Scala ValueType description (no news
there :-).

The explanation "it is the same type with a different name" feels wrong --
if it really was the same name it would be a type alias. In fact it is not
the same type, it has a specific relationship to the original type, which
is exactly that it is a subtype (but not every subtype is a NewType).

I will check with Jukka but unless he has strong objections I am going for
Distinct Type (two words) when we're discussing it in English text, but I
will keep NewType (one function) in code.

On Wed, Jun 1, 2016 at 1:42 AM, Sietse Brouwer <notifications@github.com>
wrote:

> I like @lgautier <https://github.com/lgautier>'s NamedType a lot: to me,
> it expresses "this is otherwise the same type, but it has a different name
> (and therefore its own meaning)". Also, like he says, it meshes nicely with
> namedtuple (and typing.NamedTuple).
>
> How I would read it:
>
> UserID = NamedType('UserID', int) -> This is a type of int named UserID
>
> ------------------------------
>
> If NamedType does not make it, what about ValueType?
>
>    - ValueType: Scala has the same concept, and calls them Value Classes
>    <http://docs.scala-lang.org/overviews/core/value-classes.html>. It
>    expresses nicely that this type does not say something about the data
>    structure, but about the value it contains.
>
> How I would read it:
>
> UserID = ValueType('UserID', int) -> This is an int with a user ID value
> type.
>
> NamedType works better, though.
>
> —
> You are receiving this because you commented.
> Reply to this email directly, view it on GitHub
> <https://github.com/python/typing/pull/226#issuecomment-222929750>, or mute
> the thread
> <https://github.com/notifications/unsubscribe/ACwrMiamuJVBxyBPxt4bA0yULC0M3MzVks5qHUXcgaJpZM4IoPSY>
> .
>



-- 
--Guido van Rossum (python.org/~guido)

gvanrossum:
I'm going to abandon the "English description" discussion, and merge this as-is. The new function will be called NewType() and in the discussion we'll use "unique type" (like in this PR).

gvanrossum:
Mypy issue is here: https://github.com/python/mypy/issues/1284

PR #232, opened 2016-06-06
Fix an indentation error in the rST for NewType

kirbyfan64:
In the version of [PEP 484 on python.org](https://www.python.org/dev/peps/pep-0484/#newtype-helper-function), this causes:

> System Message: WARNING/2 ( `pep-0484.txt` , line 1194)
> 
> Literal block expected; none found.


gvanrossum:
Thanks! I usually check for these, somehow must have missed one.

PR #233, opened 2016-06-06
Allow functions as Union arguments.

### LINKED ISSUE #229: Union runtime check fails for xml.etree.cElementTree.Element
```py
from typing import *
from xml.etree.cElementTree import Element
T = Union[Element, Optional[str]]
```

Under python 2, this fails at runtime, but it works for static type checking. Currently the best workaround seems to be to define T in a `if False:` block, which allows the static check to pass without causing runtime errors.

gvanrossum:
Fixes #229.

### DIFF
--- a/python2/test_typing.py
+++ b/python2/test_typing.py
@@ -4,7 +4,7 @@ import collections
 import pickle
 import re
 import sys
-from unittest import TestCase, main
+from unittest import TestCase, main, SkipTest
 
 from typing import Any
 from typing import TypeVar, AnyStr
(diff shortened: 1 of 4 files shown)


gvanrossum:
NOTE: this fix is skating on thin ice -- it's allowing some stuff through that's not allowed, but mypy will catch that. It may also just defer the problem (maybe some other type constructor that balks on this -- note that fixing _type_check() wasn't quite enough) but I've spot-checked a few and not found any further errors.

@JukkaL, @matthiaskramm -- any thoughts?

matthiaskramm:
This seems wrong. A function is not a type.

I'd prefer that API implementation peculiarities like xml.etree.cElementTree.Element are masqueraded by a handcrafted `.pyi`. (Which declares `Element` as a class, and moves the method signature into `__init__`)

gvanrossum:
Ah, but this API peculiarity *is* masqueraded by the `.pyi`. The problem is that when you then *run* the code from the example (#229) it crashes with a TypeError, and that's what I'm fixing here.

gvanrossum:
(If it was up to @markshannon Union[1, 2, 3] would also be valid at runtime. :-)

matthiaskramm:
Oh, I see. I misunderstood your fix.

Then I'm +1. Type annotations are just annotations, and should never actually *do* anything, at runtime.
(Fun fact: In our Python 2 backport of PEP 3107, we don't even evaluate them. Not that that would help if you declare a T like in #229.)

markshannon:
I don't mind if `Union[1,2,3]` is a runtime error, as long as it happens quickly using very little memory :)

JukkaL:
This looks okay to me.

gvanrossum:
Thanks for the reviews; merged!

PR #240, opened 2016-07-04
Add explanation of singletons in unions to PEP484

ilevkivskyi:
I tried to keep discussion to a minimum, only including simple examples. I think it is better to add this section right after discussion of Optional, that is actually also a Union with a singleton type (NoneType).

### DIFF
--- a/pep-0484.txt
+++ b/pep-0484.txt
@@ -846,6 +846,51 @@ This is equivalent to::
 
   def handle_employee(e: Optional[Employee] = None) -> None: ...
 
+
+Support for singleton types in unions
+-------------------------------------
+
+A singleton instance is frequently used to mark some special condition,
+in particular in situations when ``None`` is a valid value for a variable.
+Example::
+
+  _empty = object()
+
+  def func(x=_empty):
+      if x is _empty:  # default argument value
+          return 0
+      elif x is None:  # argument was provided and it's None
+          return 1
+      else:
+          return x * 2
+
+To allow precise typing in such situations, the user should use
+the ``Union`` type in conjuction with the ``enum.Enum`` class provided
+by the standard library, so that type errors could be caught statically::
+
+  from typing import Union
+  from enum import Enum
+
+  class Empty(Enum):
+      token = 0
+  _empty = Empty.token
+
+  def func(x: Union[int, None, Empty]=_empty) -> int:
+
+      boom = x * 42  # This fails type check
+
+      if x is _empty:
+          return 0
+      elif x is None:
+          return 1
+      else:  # At this point typechecker knows that x can only have type int
+          return x * 2
+
+Since the subclasses of ``Enum`` could not be further subclassed,
+the type of variable ``x`` can be statically inferred in all branches
+of the above example.
+
+
 The ``Any`` type
 ----------------
 
--- a/pep-0484.txt
+++ b/pep-0484.txt
@@ -875,7 +875,7 @@ by the standard library, so that type errors can be caught statically::
       token = 0
   _empty = Empty.token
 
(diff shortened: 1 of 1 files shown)


pep-0484.txt
```diff
@@ -846,6 +846,51 @@ This is equivalent to::
...
+      boom = x * 42  # This fails type check
+
+      if x is _empty:
+          return 0
+      elif x is None:
+          return 1
+      else:  # At this point typechecker knows that x can only have type int
+          return x * 2
+
+Since the subclasses of ``Enum`` could not be further subclassed,
```
gvanrossum: Cannot

pep-0484.txt
```diff
@@ -846,6 +846,51 @@ This is equivalent to::
...
+      if x is _empty:  # default argument value
+          return 0
+      elif x is None:  # argument was provided and it's None
+          return 1
+      else:
+          return x * 2
+
+To allow precise typing in such situations, the user should use
+the ``Union`` type in conjuction with the ``enum.Enum`` class provided
+by the standard library, so that type errors could be caught statically::
```
gvanrossum: Can

pep-0484.txt
```diff
@@ -846,6 +846,51 @@ This is equivalent to::
 
   def handle_employee(e: Optional[Employee] = None) -> None: ...
 
+
+Support for singleton types in unions
+-------------------------------------
+
+A singleton instance is frequently used to mark some special condition,
+in particular in situations when ``None`` is a valid value for a variable.
```
gvanrossum: Also

gvanrossum:
Thanks! Some language nits.

I wonder if we should add another example showing an enum with multiple values?

ilevkivskyi:
I have made corrections and added another example. I think it clarifies that it is not necessary to make many ``Enum``'s if one wants more than one special value.

pep-0484.txt
```diff
@@ -846,6 +846,66 @@ This is equivalent to::
...
+by the standard library, so that type errors can be caught statically::
+
+  from typing import Union
+  from enum import Enum
+
+  class Empty(Enum):
+      token = 0
+  _empty = Empty.token
+
+  def func(x: Union[int, None, Empty]=_empty) -> int:
```
JukkaL: Style nit: add spaces around `=`.

pep-0484.txt
```diff
@@ -846,6 +846,66 @@ This is equivalent to::
...
+the type of variable ``x`` can be statically inferred in all branches
+of the above example. The same approach is applicable if more than one
+singleton object is needed: one can use an enumeration that has more than
+one value::
+
+  class Reason(Enum):
+      timeout = 1
+      error = 2
+
+  def process(response: Union[str, Reason]='') -> str:
```
JukkaL: Similar to above, spaces around `=`.

ilevkivskyi:
@JukkaL Done.

gvanrossum:
Pushed to the peps repo and filed https://github.com/python/mypy/issues/1803.

Maybe now that the peps repo has moved to GitHub we should switch to doing PRs there?

ilevkivskyi:
> Pushed to the peps repo and filed python/mypy#1803.

Thanks!

> Maybe now that the peps repo has moved to GitHub we should switch to doing PRs there?

Yes, I think we could switch to new peps repo. By the way, what are the plans for PEP484, will it be provisional in 3.6? Or will it "graduate" soon?

gvanrossum:
I would like PEP 484 to drop the provisional status for 3.6, but we have to
discuss this on python-dev.

For asyncio we did two provisional release (3.4 and 3.5) but IMO that was
too long (and yet not long enough -- Yury recently asked me to extend that
to 3.6 as well).

PR #261, opened 2016-08-13
Add 'FrozenSet' to __all__

Vgr255:
Mypy would complain if importing 'FrozenSet', since it wasn't in `__all__`. It happily works now.

I can also submit a patch to b.p.o if needed.

Edit: Yep, I fail at testing, and mypy still complains. Is there something else to do for mypy to recognize `FrozenSet` is in typing?

### DIFF
--- a/python2/typing.py
+++ b/python2/typing.py
@@ -54,6 +54,7 @@ __all__ = [
     'DefaultDict',
     'List',
     'Set',
+    'FrozenSet',
     'NamedTuple',  # Not really a type.
     'Generator',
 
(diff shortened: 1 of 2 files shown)


gvanrossum:
No need to patch the stdlib version, I will eventually copy this over.

Regarding mypy: I think this needs updates to typing.pyi in typeshed (both versions). For that use the https://github.com/python/typeshed repo to submit a PR.

Vgr255:
Ah, typing.pyi; I didn't think of that. Will do that and submit a PR right away, thanks!

PR #265, opened 2016-08-19
Add Collection to typing

ilevkivskyi:
This PR accompanies my patch to http://bugs.python.org/issue27598


ilevkivskyi:
@gvanrossum I added ``Collection`` in such way that it is compatible with 2.7 through 3.6 (i.e. base classes are chosen to agree with ``collections.abc``)

PR #281, opened 2016-09-10
Make no_type_check apply on classes too

ilevkivskyi:
@gvanrossum 
Since we now have variable annotations, no_type_check should attach ``__no_type_check__`` also to classes (not only to functions/methods as before). I think it also should be recursive on classes so that here:
```python
@no_type_check
class C:
    class Inn:
        def __init__(self, x: 'not a type'): ...
```
we will have ``C.Inn.__init__.__no_type_check__ == True``

### DIFF
--- a/python2/test_typing.py
+++ b/python2/test_typing.py
@@ -1353,6 +1353,15 @@ class AllTests(BaseTestCase):
         # Check that Text is defined.
         self.assertIn('Text', a)
 
+    def test_respect_no_type_check(self):
+        @no_type_check
+        class NoTpCheck(object):
+            class Inn(object):
+                def __init__(self, x): pass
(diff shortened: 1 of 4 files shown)


ilevkivskyi:
There are some corner cases here. Generics have attributes that point to a base class, and ``@no_type_check`` should not affect base classes. I will fix this soon.

ilevkivskyi:
@gvanrossum 
OK, now it is ready to be merged. Please take a look.

gvanrossum:
Right. But this needs to be back-ported to the Python 2 version too.

ilevkivskyi:
@gvanrossum 
I agree. Back-ported it to Python 2.

PR #283, opened 2016-09-18
Kill __subclasscheck__

### LINKED ISSUE #133: Bare Union, Callable and Tuple should not be classes
There is no reason for "bare" Union, Callable or Tuple (i.e. without [...]) to be a class. You shouldn't ever write `isinstance(x, Union)` for example, and in their bare form these are not acceptable as annotations either.

### LINKED ISSUE #136: Kill __subclasscheck__

### LINKED ISSUE #203: typing.MutableMapping not a drop-in replacement for collections.abc.MutableMapping

ilevkivskyi:
This PR:
* Fixes #136 
* Fixes #133 
* Partially fixes #203 (fixes the ``isinstance`` part, and multiple inheritance, still ``typing.Something`` is not a drop-in replacement for ``collections.abc.Something`` in terms of *implementation*).
* Also fixes http://bugs.python.org/issue26075, http://bugs.python.org/issue25830, and http://bugs.python.org/issue26477
* Makes almost everything up to 10x faster.
* Is aimed to be a minimalistic change. I only removed ``issubclass`` tests from ``test_typing`` and main changes to ``typing`` are ``__new__`` and ``__getitem__``.

The idea is to make most things not classes. Now ``_ForwardRef()``, ``TypeVar()``, ``Union[]``, ``Tuple[]``, ``Callable[]`` are not classes (i.e. new class objects are almost never created by ``typing``).

Using ``isinstance()`` or ``issubclass()`` rises ``TypeError`` for almost everything. There are exceptions:
* Unsubscripted generics are still OK, e.g. ``issubclass({}, typing.Mapping)``. This is done to (a) not break existing code by addition of type information; (b) to allow using ``typing`` classes as a replacement for ``collections.abc`` in class and instance checks. Finally, there is an agreement that a generic without parameters assumes ``Any``, and ``Any`` means fallback to *dynamic* typing.
* ``isinstance(lambda x: x, typing.Callable)`` is also OK. Although ``Callable`` is not a generic class, when unsubscribed, it could be also used as a replacement for ``collections.abc.Callable``.
* The first rule for generics makes ``isinstance([], typing.List)`` possible, for consistency I also allowed ``isinstance((), typing.Tuple)``.

Finally, generics should be classes, to allow subclassing, but now the outcome of ``__getitem__`` on classes is cached. I use an extended version of ``functools.lru_cache`` that allows fallback to non-cached version for unhashable arguments.

This is still WIP, since I did this only for Python 3.

@gvanrossum @JukkaL Please take a look.

### DIFF
--- a/src/typing.py
+++ b/src/typing.py
@@ -545,7 +545,6 @@ class _Union(Final, _root=True):
         # It's not a union if there's only one type left.
         if len(all_params) == 1:
             return all_params.pop()
-        # Create a new class with these params.
         self.__union_params__ = tuple(t for t in params if t in all_params)
         self.__union_set_params__ = frozenset(self.__union_params__)
         return self
--- a/src/typing.py
+++ b/src/typing.py
@@ -195,12 +195,10 @@ class _ForwardRef(TypingBase):
 class _TypeAlias(TypingBase, metaclass=TypingMeta, _root=True):
     """Internal helper class for defining generic variants of concrete types.
 
-    Note that this is not a type; let's call it a pseudo-type.  It can
-    be used in instance and subclass checks, e.g. isinstance(m, Match)
-    or issubclass(type(m), Match).  However, it cannot be itself the
-    target of an issubclass() call; e.g. issubclass(Match, C) (for
-    some arbitrary class C) raises TypeError rather than returning
-    False.
+    Note that this is not a type; let's call it a pseudo-type.  It cannot
+    be used in instance and subclass checks in parameterized form, i.e.
+    ``isinstance(42, Match[str])`` raises ``TypeError`` instead of returning
+    ``False``.
     """
 
     __slots__ = ('name', 'type_var', 'impl_type', 'type_checker')
@@ -247,18 +245,21 @@ class _TypeAlias(TypingBase, metaclass=TypingMeta, _root=True):
             if not issubclass(parameter, self.type_var.__constraints__):
                 raise TypeError("%s is not a valid substitution for %s." %
                                 (parameter, self.type_var))
-        if isinstance(parameter, TypeVar) and parameter != self.type_var:
-            raise TypeError("%s is not a valid substitution for %s." %
-                                (parameter, self.type_var))
+        if isinstance(parameter, TypeVar):
+            raise TypeError("%s cannot be re-parameterized." % self.type_var)
         return self.__class__(self.name, parameter,
                               self.impl_type, self.type_checker)
 
     def __instancecheck__(self, obj):
-        raise TypeError("Type aliases cannot be used with isinstance().")
+        if not isinstance(self.type_var, TypeVar):
+            raise TypeError("Parameterized type aliases cannot be used "
+                            "with isinstance().")
+        return isinstance(obj, self.impl_type)
 
     def __subclasscheck__(self, cls):
         if not isinstance(self.type_var, TypeVar):
-            raise TypeError("Type aliases cannot be used with issubclass().")
+            raise TypeError("Parameterized type aliases cannot be used "
+                            "with issubclass().")
         return issubclass(cls, self.impl_type)
 
 
--- a/src/typing.py
+++ b/src/typing.py
@@ -617,7 +617,7 @@ class _Tuple(Final, _root=True):
     to type variables T1 and T2.  Tuple[int, float, str] is a tuple
     of an int, a float and a string.
 
-    To specify a variable-length tuple of homogeneous type, use Sequence[T].
+    To specify a variable-length tuple of homogeneous type, use Tuple[T, ...].
     """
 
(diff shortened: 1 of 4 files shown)


ilevkivskyi:
@gvanrossum @JukkaL 
It looks like this PR also fixes http://bugs.python.org/issue26075, http://bugs.python.org/issue25830, and http://bugs.python.org/issue26477

I added this to the description.

JukkaL:
I did a quick pass on the diff and tried it with mypy, and I like it. The changes seem to speed up a no-op mypy invocation (`mypy`) by roughly 20%, which is nice (though I didn't do very careful benchmarking). There still seems to  be maybe a 5-10% slowdown in startup time due to type annotations, but this seems acceptable already.

The generic type object cache hit rate when running `mypy -x` was about 78% -- increasing the cache size to 50 improved that to 80%. A further increase only had a marginal effect. The improvement with a larger cache might be higher to larger codebases, but it's hard to tell.

ilevkivskyi:
@gvanrossum 
Thank you for review! I agree with all comments and implemented them (also in ``/python2``). Except for one: I added ``_`` to ``_TypingBase`` and ``_FinalTypingBase`` as you requested, but not to ``TypingMeta``, there are few reasons for this:
* It was around for a long time (probably since beginning).
* Adding underscore there would probably also require adding a lot of them to ``TupleMeta`` etc. in ``/python2``.
* I think we could keep ``TypingMeta`` as a part of "semi-public" API (just in case someone wants to make extensions to ``typing`` module or ``mypy``).

ilevkivskyi:
@JukkaL 
I agree that the cache size should be increased. The idea is that projects that will use ``typing`` and ``mypy`` are probably very large (this is also Guido's opinion). Now the cache size is 128.

src/typing.py
```diff
@@ -562,110 +653,45 @@ def __instancecheck__(self, obj):
...
-    - To cut a tie: Union[object, Any] == Union[Any, object] == Any.
-
-    - You cannot subclass or instantiate a union.
-
-    - You cannot write Union[X][Y] (what would it mean?).
+class _Optional(_FinalTypingBase, _root=True):
+    """Optional type.
 
-    - You can use Optional[X] as a shorthand for Union[X, None].
+    Optional[X] is equivalent to Union[X, type(None)].
```
gvanrossum: While type(None) is technically correct I'm not sure how much it adds to mention it here (especially since Union[x, None] and Union[x, type(None)] are always equivalent :-).

gvanrossum:
Feel free to iterate more on this! Especially the Union[Any, int] issue. But I figured we might as well take a checkpoint, so this is now merged (also into CPython 3.5, 3.6 and 3.7==default). I closed the three CPython bugs you mentioned.

ilevkivskyi:
@gvanrossum 
Thank you for merging this!
I will submit a new PR for ``Union[int, Any]`` in ``typing.py`` and probably also PR for PEPs 484 and 483.

This is a good catch, I didn't notice this while editing PEP 483. The point is that type is characterized by _two_ sets: set of values and set of functions (methods), as mentioned in PEP 483. With such definition neither ``int`` is not a subtype of ``Any`` (since ``int`` has less methods) nor vice versa (since ``Any`` has more values). The fact that ``Any`` is at the same time at the top of type hierarchy (it has all values) and at its bottom (it has all methods) requires us to introduce two different concepts: subtype and compatibility.

Actually, now I think that ``Any`` should be considered a subtype of ``object``. Indeed, both have all values, but ``Any`` has more methods, therefore it is a subtype of ``object``. This, in turn, means that ``Union[object, Any]`` should be simplified to just ``object``. For example, if ``x`` has type ``Union[object, Any]`` then the only way to call ``x.lower()`` is when ``x`` is not ``object``, but this is not possible:
```python
x: Union[object, Any]
x.lower()  # this should be rejected
if not isinstance(x, object):
    x.lower()  # this should be allowed, but it is not reachable
```

What do you think?

(If it will help, the relation between ``object`` and ``Any`` is similar to relation between, e.g., ``Mapping`` and ``MutableMapping``, they both have the same set of values, but ``MutableMapping`` has more methods, therefore it is a subtype of ``Mapping``.)

gvanrossum:
Good reasoning! I like talking about "has more methods" and "has more
values".

PR #287, opened 2016-09-29
Include __extra__ base in generics MRO

### LINKED ISSUE #203: typing.MutableMapping not a drop-in replacement for collections.abc.MutableMapping

ilevkivskyi:
Fixes #203 
@gvanrossum 
This is a simple-minded fix. Basically, I just copied some parts from #207 + small tweaks + backport to Python 2. Note that this makes classes inherited from ``typing.Mapping`` etc. real ABCs, i.e. they could not be instantiated without implementing all the abstract methods (in this sense ``typing.X`` and ``collections.abc.X`` are identical).

Please take a look.

### DIFF
--- a/python2/test_typing.py
+++ b/python2/test_typing.py
@@ -573,36 +573,32 @@ class GenericTests(BaseTestCase):
         self.assertNotIsInstance({}, MyMapping)
         self.assertNotIsSubclass(dict, MyMapping)
 
-    def test_multiple_abc_bases(self):
+    def test_abc_bases(self):
+        class MM(MutableMapping[str, str]):
+            def __getitem__(self, k):
+                return None
+            def __setitem__(self, k, v):
+                pass
+            def __delitem__(self, k):
+                pass
+            def __iter__(self):
+                return iter(())
+            def __len__(self):
+                return 0
+        # this should just work
+        MM().update()
+        self.assertIsInstance(MM(), collections_abc.MutableMapping)
+        self.assertIsInstance(MM(), MutableMapping)
+        self.assertNotIsInstance(MM(), List)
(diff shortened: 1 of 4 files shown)


gvanrossum:
Thanks! I will not pretend I understand everything that's going on, but we have enough tests that I sort of believe it's better, not worse.

Have you thought more about Union[Any, x]?

bintoro:
@ilevkivskyi #207 added a few more tests that I believe are important. Did you omit them intentionally?

ilevkivskyi:
@gvanrossum I will make PRs about ``Union[Any, x]`` later today, we could discuss details in those PRs.

@bintoro Could you please point more precisely to the tests (line numbers/names)? Most probably I just overlooked them (I added some tests in my previous PR).

bintoro:
@ilevkivskyi Have a look from [this line](https://github.com/python/typing/pull/207/files#diff-f9eeb4530dadc70935c11fff001ec7acR1383) on. Not all of them are applicable anymore, but most are.

The code interacting with the ABC machinery is somewhat intricate (and could actually use a thorough review), so it's important to test it works with things like `register()` and custom subclasshooks.

ilevkivskyi:
@bintoro Thanks! I will make a PR with these tests later, right after I finish with ``Union[X, Any]``.

PR #288, opened 2016-09-29
Follow up for #283 (response to comments) including change in Union[X, Any]

### LINKED ISSUE #285: Union[X, Any] should not be simplified to Any
See https://github.com/python/mypy/issues/1642

ilevkivskyi:
Fixes #285 

Later I will make a PR to ``python/peps`` and will submit a documentation patch to b.p.o.

### DIFF
--- a/python2/test_typing.py
+++ b/python2/test_typing.py
@@ -201,6 +201,13 @@ class UnionTests(BaseTestCase):
     def test_union_any(self):
         u = Union[Any]
         self.assertEqual(u, Any)
+        u1 = Union[int, Any]
+        u2 = Union[Any, int]
+        u3 = Union[Any, object]
+        self.assertEqual(u1, u2)
+        self.assertNotEqual(u1, Any)
+        self.assertNotEqual(u2, Any)
+        self.assertNotEqual(u3, Any)
 
     def test_union_object(self):
         u = Union[object]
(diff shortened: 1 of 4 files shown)


python2/test_typing.py
```diff
@@ -201,10 +201,6 @@ def test_subclass_error(self):
     def test_union_any(self):
         u = Union[Any]
         self.assertEqual(u, Any)
-        u = Union[int, Any]
-        self.assertEqual(u, Any)
-        u = Union[Any, int]
-        self.assertEqual(u, Any)
```
gvanrossum: Why not keep the test but change the asserts to test for the new behavior?

python2/typing.py
```diff
@@ -598,16 +606,6 @@ class Manager(Employee): pass
         Union[Manager, int, Employee] == Union[int, Employee]
         Union[Employee, Manager] == Employee
 
-    - Corollary: if Any is present it is the sole survivor, e.g.::
-
-        Union[int, Any] == Any
-
-    - Similar for object::
-
-        Union[int, object] == object
```
gvanrossum: I'd keep this even though it's not a special case -- it's a nice edge case that confirms the rule.

ilevkivskyi:
@gvanrossum I agree with both points. Implemented this in a new commit.

PR #291, opened 2016-10-04
Add a (somewhat simple-minded) cache to the Python 2 version too.

### LINKED ISSUE #290: Python 2 typing.py needs to cache generics too
In #283 we added a cache to things like `List[str]` that gives two orders of magnitude speedup. But the cache is only Python 3. We should add this to Python 2 as well. (Sadly the Python 2 stdlib doesn't have functools.lru_cache.)

CC: @ilevkivskyi @gnprice

gvanrossum:
Fixes #290. Timing example:

Without cache:
```
$ python2 -m timeit -s 'import typing' 'typing.List[typing.List[str]]'
10000 loops, best of 3: 542 usec per loop
```
With cache:
```
$ python2 -m timeit -s 'import typing' 'typing.List[typing.List[str]]'
1000000 loops, best of 3: 1.65 usec per loop
```
That's about a 300x speedup!

### DIFF
--- a/python2/typing.py
+++ b/python2/typing.py
@@ -563,6 +563,28 @@ T_contra = TypeVar('T_contra', contravariant=True)  # Ditto contravariant.
 AnyStr = TypeVar('AnyStr', bytes, unicode)
 
 
+def _tp_cache(func, maxsize=128):
+    cache = {}
+
+    @functools.wraps(func)
+    def inner(*args):
+        key = args
+        try:
+            return cache[key]
+        except TypeError:
+            # Assume it's an unhashable argument.
+            return func(*args)
+        except KeyError:
+            value = func(*args)
+            if len(cache) >= maxsize:
+                # If the cache grows too much, just start over.
+                cache.clear()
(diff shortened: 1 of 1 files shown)


gvanrossum:
@ilevkivskyi Do you see anything wrong with this?

python2/typing.py
```diff
@@ -563,6 +563,28 @@ def __subclasscheck__(self, cls):
 AnyStr = TypeVar('AnyStr', bytes, unicode)
 
 
+def _tp_cache(func, maxsize=128):
```
ilevkivskyi: A small comment: With this definition we could not later write:
```python
@_tp_cache(maxsize=200)
def __getitem__(self, parameters):
    ...
```
This causes an error (``function takes at least 1 argument...``). I would propose to either make maxsize a constant ``MAXSIZE`` or use the decorator with args pattern here.

Except for this small comment, everything LGTM.

gvanrossum:
It's not tunable in the Python 3 version either, so I just made it a local.

PR #295, opened 2016-10-09
A real fix for issue #250 (failure with mock)

### LINKED ISSUE #250: Inheriting from instantiated generic class causes mocking to fail
My reproduction (python 2)

```
import mock

T = TypeVar('T')

class Bar(Generic[T]):
    def __init__(self):
        pass

    def do_bar(self):
        print("bar")

class Foo(Bar[int]):
    def do_foo(self):
        self.do_bar()

@mock.patch.object(Bar, 'do_bar')
def mocked(mock_do_bar):
    mock_do_bar.return_value = None
    foo = Foo()
    foo.do_foo()

mocked()
```

will unexpectedly print `bar`.

The inheritance here is crucial.

ilevkivskyi:
@gvanrossum 
This PR actually fixes #250

The main idea here is optimizing generics for cases where a type information is added to existing code.
Now:
* ``Node[int]`` and ``Node`` have identical ``__bases__`` and identical ``__mro__[1:]`` (except for the first item, since it is the class itself).
* After addition of typing information (i.e. making some classes generic), ``__mro__`` is changed very little, at most one bare ``Generic`` appears in ``__mro__``.
* Consequently, only non-parameterized generics appear in ``__bases__`` and ``__mro__[1:]``.

Interestingly, this could be achieved in few lines of code and no existing test break.

On the positive side of this approach, there is very little chance that existing code (even with sophisticated "magic") will break after addition of typing information.

On the negative side, it will be more difficult for _runtime_ type-checkers to perform decorator-based type checks (e.g. enforce method overriding only by consistent methods). Essentially, now type erasure happens partially at the class creation time (all bases are reduced to origin).

### DIFF
--- a/python2/test_typing.py
+++ b/python2/test_typing.py
@@ -598,6 +598,19 @@ class GenericTests(BaseTestCase):
             class MM2(collections_abc.MutableMapping, MutableMapping[str, str]):
                 pass
 
+    def test_orig_bases(self):
+        T = TypeVar('T')
+        class C(typing.Dict[str, T]): pass
+        self.assertEqual(C.__orig_bases__, (typing.Dict[str, T],))
+
+    def test_multi_subscr_base(self):
+        T = TypeVar('T')
+        U = TypeVar('U')
+        V = TypeVar('V')
+        # these should just work
+        class C(List[T][U][V]): pass
+        class D(C, List[T][U][V]): pass
+
     def test_pickle(self):
         global C  # pickle wants to reference the class by name
         T = TypeVar('T')
--- a/python2/test_typing.py
+++ b/python2/test_typing.py
@@ -607,25 +607,27 @@ class GenericTests(BaseTestCase):
         def naive_dict_check(obj, tp):
             # Check if a dictionary conforms to Dict type
             if len(tp.__parameters__) > 0:
-                return NotImplemented
+                raise NotImplementedError
             if tp.__args__:
                 KT, VT = tp.__args__
                 return all(isinstance(k, KT) and isinstance(v, VT)
                    for k, v in obj.items())
         self.assertTrue(naive_dict_check({'x': 1}, typing.Dict[typing.Text, int]))
         self.assertFalse(naive_dict_check({1: 'x'}, typing.Dict[typing.Text, int]))
-        self.assertIs(naive_dict_check({1: 'x'}, typing.Dict[typing.Text, T]), NotImplemented)
+        with self.assertRaises(NotImplementedError):
+            naive_dict_check({1: 'x'}, typing.Dict[typing.Text, T])
 
         def naive_generic_check(obj, tp):
             # Check if an instance conforms to the generic class
             if not hasattr(obj, '__orig_class__'):
-                return NotImplemented
+                raise NotImplementedError
             return obj.__orig_class__ == tp
         class Node(Generic[T]): pass
         self.assertTrue(naive_generic_check(Node[int](), Node[int]))
         self.assertFalse(naive_generic_check(Node[str](), Node[int]))
         self.assertFalse(naive_generic_check(Node[str](), List))
-        self.assertIs(naive_generic_check([1,2,3], Node[int]), NotImplemented)
+        with self.assertRaises(NotImplementedError):
+            naive_generic_check([1,2,3], Node[int])
 
         def naive_list_base_check(obj, tp):
             # Check if list conforms to a List subclass
--- a/python2/test_typing.py
+++ b/python2/test_typing.py
@@ -603,13 +603,54 @@ class GenericTests(BaseTestCase):
         class C(typing.Dict[str, T]): pass
         self.assertEqual(C.__orig_bases__, (typing.Dict[str, T],))
 
+    def test_naive_runtime_checks(self):
+        def naive_dict_check(obj, tp):
+            # Check if a dictionary conforms to Dict type
+            if len(tp.__parameters__) > 0:
+                return NotImplemented
+            if tp.__args__:
+                KT, VT = tp.__args__
+                return all(isinstance(k, KT) and isinstance(v, VT)
+                   for k, v in obj.items())
+        self.assertTrue(naive_dict_check({'x': 1}, typing.Dict[typing.Text, int]))
+        self.assertFalse(naive_dict_check({1: 'x'}, typing.Dict[typing.Text, int]))
+        self.assertIs(naive_dict_check({1: 'x'}, typing.Dict[typing.Text, T]), NotImplemented)
+
+        def naive_generic_check(obj, tp):
+            # Check if an instance conforms to the generic class
+            if not hasattr(obj, '__orig_class__'):
+                return NotImplemented
+            return obj.__orig_class__ == tp
+        class Node(Generic[T]): pass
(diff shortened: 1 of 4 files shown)


python2/typing.py
```diff
@@ -1092,6 +1086,15 @@ def __new__(cls, name, bases, namespace,
                          ", ".join(str(g) for g in gvars)))
                 tvars = gvars
 
+        if extra is None:
+            extra = namespace.get('__extra__')
+        if extra is not None and type(extra) is abc.ABCMeta and extra not in bases:
+            bases = (extra,) + bases
+        bases = tuple(_gorg(b) if isinstance(b, GenericMeta) else b for b in bases)
+        if any(isinstance(b, GenericMeta) and b is not Generic for b in bases):
```
gvanrossum: This could use a comment explaining what's happening -- it looks like the idea is to remove bare Generic from the bases if there are other generic-derived bases, but to leave generic if if not.

markshannon:
I do indeed want a much smaller typing.py
After all, if whoever is writing this mythical runtime type-checker can implement a useful type-checker, then they can manage some variation on `ast.parse(inspect.getsource(...))`

ilevkivskyi:
@gvanrossum 
I added the comment you requested and I propose a compromise solution: in latest commit I added a field ``__orig_bases__`` to non-parameterized generics that preserves original bases before we start all manipulations with ``__extra__`` and ``__bases__``. This way:
* The fix still works (and should prevent other bugs in existing code after addition of typing info)
* Only some classes in ``typing.py`` get slightly bigger.
* One can extract all the necessary original typing info at runtime using ``__orig_bases__`` and little effort.

gvanrossum:
Can you explain how this fixes #250 better than the previous fix for that
issue?

ilevkivskyi:
@gvanrossum 
> Can you explain how this fixes #250 better than the previous fix for that issue?

Oh, sorry, I should have started from this. There was no fix. I just mentioned it in #289 
> A side note, while playing with this I have found an easy way to fix #250. I will submit a PR later when we will decide on the permanent fix for this one.

And there by "a PR later" I meant exactly this PR, but it looks like you closed #250 by #289.
Sorry for misunderstanding.

gvanrossum:
Actually 250 was auto-closed by GitHub because of the phrase "fix #250" in
your commit message.

--Guido (mobile)

src/test_typing.py
```diff
@@ -625,6 +625,19 @@ class MM1(MutableMapping[str, str], collections_abc.MutableMapping):
...
+    def test_orig_bases(self):
+        T = TypeVar('T')
+        class C(typing.Dict[str, T]): ...
+        self.assertEqual(C.__orig_bases__, (typing.Dict[str, T],))
+
+    def test_multi_subscr_base(self):
+        T = TypeVar('T')
+        U = TypeVar('U')
+        V = TypeVar('V')
+        # these should just work
```
gvanrossum: That's slang for "these shouldn't crash", but it would be pedagogical to show what the various new dunder attributes of C and D actually are.

src/test_typing.py
```diff
@@ -625,6 +625,19 @@ class MM1(MutableMapping[str, str], collections_abc.MutableMapping):
             class MM2(collections_abc.MutableMapping, MutableMapping[str, str]):
                 pass
 
+    def test_orig_bases(self):
+        T = TypeVar('T')
+        class C(typing.Dict[str, T]): ...
+        self.assertEqual(C.__orig_bases__, (typing.Dict[str, T],))
```
gvanrossum: I know there are no docs for `__parameters__` etc. either, but I would still like to see some kind of example code that takes a class like this and recovers what it means given the various new dunder methods (`__origin__`, `__orig_bases__`, `__parameters__`). E.g. if I had a dict `{x: y}` how would I check that it's a subclass of `Dict[str, T]`? (The answer should reveal that the type of x must be str and T must be the type of y.)

src/typing.py
```diff
@@ -983,12 +979,24 @@ def __new__(cls, name, bases, namespace,
...
+            bases = tuple(b for b in bases if b is not Generic)
+        self = super().__new__(cls, name, bases, namespace, _root=True)
+
         self.__parameters__ = tvars
         self.__args__ = args
         self.__origin__ = origin
         self.__extra__ = extra
         # Speed hack (https://github.com/python/typing/issues/196).
         self.__next_in_mro__ = _next_in_mro(self)
+        if origin is None:
```
ilevkivskyi: I changed this a bit, to make it more clear and convenient to use. Now ``__orig_bases__`` is always created on subclassing, and then copied on ``__getitem__``. So that one don't need to use ``_gorg`` to find original bases, they are accessible immediately. I added a comment in code as well.

src/test_typing.py
ilevkivskyi: OK, fixed this.

ilevkivskyi:
@gvanrossum 
I agree with your comments, please take a look at the new version (details of changes are in replies to your comments in code).

python2/test_typing.py
```diff
@@ -603,13 +603,54 @@ def test_orig_bases(self):
         class C(typing.Dict[str, T]): pass
         self.assertEqual(C.__orig_bases__, (typing.Dict[str, T],))
 
+    def test_naive_runtime_checks(self):
+        def naive_dict_check(obj, tp):
+            # Check if a dictionary conforms to Dict type
+            if len(tp.__parameters__) > 0:
+                return NotImplemented
```
gvanrossum: NotImplemented is not a great value to return here -- it's really only supposed to be used by binary operators (e.g. `__add__`) to indicate that the reverse variant should be tried (i.e. `__radd__`). Maybe raising NotImplementedError would be better?

While this is just a test, it will be used as an example and copied, and raising makes it clearer that that's an unimplemented feature of the naive check.

src/typing.py
```diff
@@ -1128,6 +1130,8 @@ def __new__(cls, *args, **kwds):
         else:
             origin = _gorg(cls)
             obj = cls.__next_in_mro__.__new__(origin)
+            if '__dict__' in cls.__dict__:
+                obj.__orig_class__ = cls
```
gvanrossum: I think I'd just catch AttributeError here -- the "dict in dict" idiom feels obscure.

ilevkivskyi:
@gvanrossum Thank you for review! Fixed both points.

gvanrossum:
OK godspeed!

gvanrossum:
Python 3.6 beta 3 goes out on Oct 31. We have some time to think about this still. Did you have other things you really wanted to get in?

ilevkivskyi:
@gvanrossum 
> Python 3.6 beta 3 goes out on Oct 31. We have some time to think about this still. Did you have other things you really wanted to get in?

I have an idea of how to fix all three issues #298 #299 #301 in a single PR. It should be not big. However, it will change how the types store parameters. Now, e.g. ``Union`` uses ``__union_params__``, ``Callable`` uses ``__args__`` and ``__result__``, etc. I am going to change this so that all use the same scheme as ``Generic``: ``__parameters__``, ``__args__``, and ``__origin__``. This will allow to "unify" several pieces of code and fix the three mentioned issues. I think I will finish this by Monday or Tuesday.

gvanrossum:
SGTM. I originally gave them all different names to ensure I would not
confuse myself, but I think in retrospect it's better to unify.

PR #296, opened 2016-10-10
Alternative (simpler) repr for generics

ilevkivskyi:
@gvanrossum 
Currently, ``repr`` for generics acts in naive way, so that sometimes this leads to cumbersome representations. For example,
```python
class C(Generic[T]):
    ...

C[Tuple[S, T]][T, int][str]
```
prints the "whole history" ``C<~T>[typing.Tuple[~S, ~T]]<~S, ~T>[~T, int]<~T>[str]``. While the new ``repr`` will print simply the "final result" ``C[typing.Tuple[str, int]]``. Also for simpler cases, ``repr(List[T])`` would be simply ``typing.List[~T]``, not ``typing.List<~T>[~T]<~T>``.

The new ``repr`` does not use angular brackets in addition to square brackets (free parameters are anyway marked with ``~``, ``+`` or ``-``), see added tests.

The implementation might look a bit hacky, but this is done in order to be sure that the ``repr`` will not interfere with those of ``super()`` and will treat correctly complex cases with repeated occurrences of type variables.

### DIFF
--- a/src/typing.py
+++ b/src/typing.py
@@ -878,10 +878,6 @@ def _geqv(a, b):
     return _gorg(a) is _gorg(b)
 
 
-def _arg_repr(cls):
-    return '[%s]' % (', '.join(map(_type_repr, cls.__parameters__)))
-
-
 def _next_in_mro(cls):
     """Helper for Generic.__new__.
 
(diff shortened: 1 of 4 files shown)


gvanrossum:
I like this. I've been struggling with the repr() quite a bit (in Python 3.5.0 it was quite different). I don't know if I'll have time to merge today but I hope soon.

gvanrossum:
I just found a bug here, with very light QA.
```python
from typing import Any, Dict

class C(Dict[Any, Any]):
    pass

print(C.__mro__)
```
Gives a traceback:
```
  File "y.py", line 6, in <module>
    print(C.__mro__)
  File "/Users/guido/src/typing/src/typing.py", line 1009, in __repr__
    return super().__repr__() + self._argrepr()
  File "/Users/guido/src/typing/src/typing.py", line 1017, in _argrepr
    for i in range(len(self.__args__)):  # replace free parameters with args
TypeError: object of type 'NoneType' has no len()
```

ilevkivskyi:
Oops! I played with this on top of #295 most of the time, so I didn't notice this. Thanks!
I just pushed a new commit where I carefully treat ``Generic[T]`` and ``_Protocol[T]`` (with more tests).

ilevkivskyi:
@gvanrossum I have a simple patch for http://bugs.python.org/issue27989 (better representation of function signatures by ``pydoc``). I didn't submit it, because it works much better on top of this PR.

Are you going to merge this or you have some additional comments?

gvanrossum:
Sorry, just temporarily ran out of time. Will get to it soon.

ilevkivskyi:
@gvanrossum 
> Sorry, just temporarily ran out of time. Will get to it soon.

Not a problem at all.
In the meantime I implemented the PEP 526 syntax for ``NamedTuple`` in mypy: https://github.com/python/mypy/pull/2245 so that if mypy PR is merged, then you could also apply the patch in http://bugs.python.org/issue28107 (earlier we decided to postpone documenting new ``NamedTuple`` syntax until it is implemented in mypy).

src/typing.py
```diff
@@ -1006,14 +1006,17 @@ def _get_type_vars(self, tvars):
...
         if self.__origin__ is None:
             if not self.__parameters__:
                 return ''
-            return '[%s]' % (', '.join(map(_type_repr, self.__parameters__)))
-        r = self.__origin__._argrepr()
+            return par_repr
+        r = self.__origin__._arg_repr()
         for i in range(len(self.__args__)):  # replace free parameters with args
             par = stdlib_re.escape(_type_repr(self.__origin__.__parameters__[i]))
             r = stdlib_re.sub(par + '(?=[,\]])', '{%r}' % i, r)
```
gvanrossum: I'm sorry for only just now looking at this code in detail. I really don't like the idea that we're taking some other class's `_arg_repr()` and then doing a regex substitution on it. Especially since this is just calling itself recursively. There's got to be a better way.

ilevkivskyi:
@gvanrossum 
Agreed, probably I am just being lazy here :-) Now there is no regexp and no recursion (except for the case one calls ``repr(List[List[List[T]]][int])``, in this case there will be three recursive calls). It was necessary to add a ``_subs_repr`` method to some types that can contain type variables but are not instances of ``GenericMeta`` (this is similar to what you did with ``_get_type_vars``). Note that no new types are created, all manipulations are done with lists of parameters/arguments.

While working on this, I have found three very minor bugs (I have fixed them right here) and three bigger things that I think are bugs, I will open issues on tracker soon.

ilevkivskyi:
@gvanrossum I forgot to add, I changed the repr for bare generics, now ``repr(List) == 'typing.List'``, not ``typing.List[~T]`` as before. There are two reasons for this:
* It is more consistent with other things in ``typing``. For example ``repr(Optional)`` is simply ``'typing.Optional'``, etc.
* By agreement, bare ``List`` means ``List[Any]``, so that if one wants a function that takes an arbitrary list and returns an arbitrary list, then it would be typed as ``def func(x: List) -> List``. But then in ``pydoc`` and other doc-tools its signature will be displayed as ``func(x: List[~T]) -> List[~T]``, that looks like a generic function. I could imagine this causing a lot of confusion.

gvanrossum:
Yeah, that's why the previous repr would return 'List<~T>' (and that was an
improvement over the original that returned 'List[~T]'). I agree that just
plain 'List' is fine too; if we want to know the numbers and names of the
parameters we can write `List.__parameters__`.

gvanrossum:
I'm giving you the benefit of the doubt here -- it looks fine but I'm out of time this week to really understand how this works. I trust it's easy enough to understand when someone finds a bug though.

PR #304, opened 2016-10-21
Skip some tests if asyncio is not awailable

ilevkivskyi:
Fixes http://bugs.python.org/issue28482


gvanrossum:
I will try to find time to merge into CPython later today.

gvanrossum:
Done.

PR #305, opened 2016-10-21
Two minor fixes

ilevkivskyi:
Here are two minor fixes for things that I have found in ``typing.py``:
* Type variables should be compared by ``id`` not by name (in one of recent PRs I compared them by ``__name__``, but after reading discussions at mypy tracker I think it was not right, for example two variables with the same name defined in different modules should be considered different).
* All names should be included in ``get_type_hints`` for modules. Initially, only names that could be accessed on modules (i.e. those annotated as ``x: int = 5`` with initial values) were included in ``get_type_hints``. This was done to sync the behaviour with behaviour of ``inspect.getannotations`` (it discards manually added items to ``__annotations__``). Now we don't use ``inspect``, so that we no longer should exclude the undefined names.

### DIFF
--- a/python2/typing.py
+++ b/python2/typing.py
@@ -1004,7 +1004,7 @@ def _replace_arg(arg, tvars, args):
         return arg._subs_repr(tvars, args)
     if isinstance(arg, TypeVar):
         for i, tvar in enumerate(tvars):
-            if arg.__name__ == tvar.__name__:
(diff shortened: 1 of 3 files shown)


gvanrossum:
Merged to CPython 3.5, 3.6, 3.7.

gvanrossum:
And to mypy.

PR #308, opened 2016-10-25
Refactor Union, Tuple, and Callable

### LINKED ISSUE #156: Type variable substitution in type instances
[In http://bugs.python.org/issue25087 Stefan Krah wrote:]

If a type scheme is instantiated, should the type variables in the
class body be substituted?  This is an example (typed by hand on
a locked down Windows machine, may contain errors):

```
alpha = TypeVar('alpha')
beta = TypeVar('beta')

class ABTuple(Generic[alpha, beta]):
    def __init__(self, a : alpha, b : beta):
        self.value = (a, b)

get_type_hints(ABTuple.__init__)
 ==> {'b': ~beta, 'a': ~alpha}


IntIntTuple = ABTuple[int, int]

IntIntTuple
 ==> __main__.ABTuple[int, int]

get_type_hints(IntIntTuple.__init__)
{'b': ~beta, 'a': ~alpha}
      ^^^^^^      ^^^^^^
```

Since the type has been specialized, these should ideally be 'int'.

### LINKED ISSUE #298: All geneneric types should be indexable.
Now one can write ``Dict[T, int][str]``, while ``Callable[[T], int][str]`` raises an error. However, most of the arguments for allowing indexable generics also apply for ``Union, Tuple, Callable``, etc. Actually, these types are also generic in the sense that they could depend on other types or type variables (for example ``Callable`` takes two types ``str`` and ``int``, and returns a new type ``Callable[[str], int]``).

@gvanrossum If you agree, then you can assign this issue to me.

### LINKED ISSUE #299: Callable and Tuple should be subclassable.
One can subclass ``collections.abc.Mapping`` as well as ``typing.Mapping[str, int]``. However, it is currently prohibited to subclass ``typing.Callable`` and ``typing.Tuple``:
```python
class C(collections.abc.Callable): ...  # works
class C(typing.Callable): ...           # fails

class C(typing.List[T]): ...            # works
class C(typing.Tuple[T, ...]): ...      # fails
```
I think that both ``Callable`` and ``Tuple`` should be subclassable (at least in the simple form shown above).

@gvanrossum If you agree, then you can assign this issue to me.

### LINKED ISSUE #301: Better __eq__ for generics
Currently, ``Iterable[T][int] == Iterable[int]`` returns ``False``. It could be convenient to implement an ``__eq__`` that returns ``True`` for generics that are considered the same from the point of view of static type checkers.

Of course, there is a very simple way to do this: just compare the new ``repr`` for two classes :-)
But probably the right way would be to factor out the algorithm from the new ``repr`` that "substitutes" type variables and use it for both ``repr`` and ``__eq__``.

ilevkivskyi:
Fixes #301 now ``List[Tuple[T, T]][int] == List[Tuple[int, int]]``
Fixes #298 now ``Table = Tuple[T, List[T]]`` can be used as generic type alias (as in PEP 484 example)
Fixes #299 now ``class MyTup(Tuple[int, ...]): ...`` is allowed
Fixes #156 (well, it does not substitute the annotations, but makes this task simple even in complex cases, see example in tests)

Also this PR fixes some minor things that I have found while working on this:
* ``List[Union]`` (with bare ``Union``, i.e. without arguments), ``List[Optional]``, ``List[Generic[T]]``, and ``List[ClassVar[int]]`` are not valid and are prohibited now.
* ``Generic`` did not evaluate forward references when asked, now it does.
* ``__qualname__`` was not copied on generic class subscription.
* Type was not erased on instantiation of subclasses of concrete containers (``List``, ``Set``, etc).
* There was an obscure bug in Python 2: sometimes ``_abc_registry`` was erased on instantiation.

The main idea of this PR is to fix the issues mentioned at the top by reusing the existing code. Namely, I pulled flattening and removing duplicates code from ``_Union`` and the tree calculation function ``_subs_tree`` from ``GenericMeta``. As well I moved ``Tuple`` and ``Callable`` _after_ ``GenericMeta`` and made them inherit the latter. So that now all types that could be generic store their info in common way using ``__origin__``, ``__parameters__``, ``__args__``.

I tried to polish this, to be sure that nothing was broken in the process of "refactoring" (also to improve speed). There is no recursion, the substitution tree is recalculated only when necessary. Also I added a lot of tests and many comments/docstrings (also for things added in my recent PRs).

@gvanrossum Could you please take a look at this? I would like this to get into 3.6b3, so that it will have a chance to be thoroughly tested.


ilevkivskyi:
@gvanrossum 
OK, I have copied this to Python 2 and updated the PR description accordingly. Please take a look.

ilevkivskyi:
@gvanrossum Is there a chance this PR gets into 3.6b3?

(If this one gets merged, I could make a PR for https://github.com/python/typing/issues/306 within an hour (if I will not sleep at that time :-))

gvanrossum:
Sorry, not  much chance. :-( I'm overloaded.

gvanrossum:
I'm not sure if I'll *ever* be able to actually review this. It's over 1000
lines. The only way this will ever land is on the basis of trust alone...

ilevkivskyi:
@gvanrossum Changes to Python 2 and 3 are practically identical, so this is rather 500 lines (even less if one does not count tests). But in no case I want you to feel pressure because of this PR :-)
> The only way this will ever land is on the basis of trust alone...

I would say this is an argument in favour of including this PR in 3.6b3 (it will be more tested by wider community before 3.6 final).

Also note that this PR will enable more features from https://github.com/python/mypy/pull/2378 , namely generic aliases like ``TInt = Tuple[T, int]``.

gvanrossum:
OK, here's what I'll do, right now. I'll merge this right now, sight unseen, so you can fix #306.

I will create a CPython issue representing typing.py upgrades, and mention that when I merge this into CPython.

Hopefully the buildbots will stay green, but if not, we'll find out on that issue.

When you have #306 fixed I will merge that into CPython too, referencing the same issue.

I will also try to decide on the parameters of Coroutine.

gvanrossum:
OK, those changes were pushed, now let's watch the buildbots: https://www.python.org/dev/buildbot/

ilevkivskyi:
Thank you for merging this! It looks like all buildbots are OK :-)

PR #311, opened 2016-10-29
Fix the problem with copy and deepcopy; also improve pickling a bit

### LINKED ISSUE #306: Can't copy.copy or deepcopy any type that inherits from _FinalTypingBase
Appears to be since #283, according to git bisect.

To repro:

```
$ python --version
Python 3.6.0b2
$ python -c 'import typing, copy; copy.copy(typing.Any)'
  File "<string>", line 1, in <module>
  File "/usr/local/Cellar/python3/3.6.0b2_3/Frameworks/Python.framework/Versions/3.6/lib/python3.6/copy.py", line 106, in copy
    return _reconstruct(x, None, *rv)
  File "/usr/local/Cellar/python3/3.6.0b2_3/Frameworks/Python.framework/Versions/3.6/lib/python3.6/copy.py", line 274, in _reconstruct
    y = func(*args)
  File "/usr/local/Cellar/python3/3.6.0b2_3/Frameworks/Python.framework/Versions/3.6/lib/python3.6/copyreg.py", line 88, in __newobj__
    return cls.__new__(cls, *args)
  File "/usr/local/Cellar/python3/3.6.0b2_3/Frameworks/Python.framework/Versions/3.6/lib/python3.6/typing.py", line 191, in __new__
    raise TypeError("Cannot instantiate %r" % cls)
TypeError: Cannot instantiate typing.Any
```

Motivation for copying a type: deepcopying a function's `__annotations__` makes it possible to do manipulations on the annotations to get only the data you want, like type annotations for just the function arguments.

```python
annotations = copy.deepcopy(func.__annotations__)
annotations.pop('return', None)
```

ilevkivskyi:
Fixes #306 
@gvanrossum Here is the PR that I promised. I also improved pickling a bit, but as you could see from tests, the list of things that could be pickled is much shorter than list of those that can be copied/deepcopied (note that ``copy`` module in Python 3 treats classes as immutable, and therefore return the class unchanged)

### DIFF
--- a/src/test_typing.py
+++ b/src/test_typing.py
@@ -4,6 +4,7 @@ import pickle
 import re
 import sys
 from unittest import TestCase, main, skipUnless, SkipTest
+from copy import copy, deepcopy
 
 from typing import Any
 from typing import TypeVar, AnyStr
@@ -845,6 +846,24 @@ class GenericTests(BaseTestCase):
             self.assertEqual(x.foo, 42)
             self.assertEqual(x.bar, 'abc')
             self.assertEqual(x.__dict__, {'foo': 42, 'bar': 'abc'})
+        simples = [Any, Union, Tuple, Callable, ClassVar, List, typing.Iterable]
+        for s in simples:
+            for proto in range(pickle.HIGHEST_PROTOCOL + 1):
+                z = pickle.dumps(s, proto)
+                x = pickle.loads(z)
+                self.assertEqual(s, x)
+
+    def test_copy_and_deepcopy(self):
+        T = TypeVar('T')
+        class Node(Generic[T]): ...
+        things = [Union[T, int], Tuple[T, int], Callable[..., T], Callable[[int], int],
+                  Tuple[Any, Any], Node[T], Node[int], Node[Any], typing.Iterable[T],
+                  typing.Iterable[Any], typing.Iterable[int], typing.Dict[int, str],
+                  typing.Dict[T, Any], ClassVar[int], ClassVar[List[T]], Tuple['T', 'T'],
+                  Union['T', int], List['T'], typing.Mapping['T', int]]
+        for t in things + [Any]:
+            self.assertEqual(t, copy(t))
+            self.assertEqual(t, deepcopy(t))
 
     def test_errors(self):
         with self.assertRaises(TypeError):
--- a/python2/test_typing.py
+++ b/python2/test_typing.py
@@ -5,6 +5,7 @@ import pickle
 import re
 import sys
 from unittest import TestCase, main, SkipTest
+from copy import copy, deepcopy
 
 from typing import Any
 from typing import TypeVar, AnyStr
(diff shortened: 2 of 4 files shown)


ilevkivskyi:
OK, it looks like Pythin 3.2 behaves the same way as Python 2, so that I added ``__copy__`` also to Python 3 (it is simply ignored be newer versions).

gvanrossum:
Also merged into CPython.

src/test_typing.py
```diff
@@ -845,6 +846,24 @@ class C(B[int]):
...
+    def test_copy_and_deepcopy(self):
+        T = TypeVar('T')
+        class Node(Generic[T]): ...
+        things = [Union[T, int], Tuple[T, int], Callable[..., T], Callable[[int], int],
+                  Tuple[Any, Any], Node[T], Node[int], Node[Any], typing.Iterable[T],
+                  typing.Iterable[Any], typing.Iterable[int], typing.Dict[int, str],
+                  typing.Dict[T, Any], ClassVar[int], ClassVar[List[T]], Tuple['T', 'T'],
+                  Union['T', int], List['T'], typing.Mapping['T', int]]
+        for t in things + [Any]:
+            self.assertEqual(t, copy(t))
```
mitar: @ilevkivskyi: Are you sure this works correctly? From [documentation](https://docs.python.org/3/library/copy.html):

>  It does “copy” functions and classes (shallow and deeply), by returning the original object unchanged; this is compatible with the way these are treated by the pickle module.

Test:

```python
class Metaclass(type):
  def __copy__(self):
    return 1

class Foo(metaclass=Metaclass):
  pass

>>> copy.copy(Foo)
<class '__main__.Foo'>
>>> Foo.__copy__()
1
>>> copy.copy(Foo) is Foo
True
```

I do not think your test is really testing this correctly. Nor copying is really working correctly.

ilevkivskyi: Yes, we exactly follow the docs for ``copy`` module (classes are treated as immutable, except in Python 3.2 and 2.7, this is why we have ``__copy__`` defined), what is the problem?

mitar: Oh, so idea is that `copy.copy` should return same reference, and `__copy__` is here just for compatibility with 3.2 and 2.7 to return something sane? But why is then doing `self.__class__(...)` and not just returning self? So why is really making a copy? Or you want in 3.2 and 2.7 to be a copy?

Should a test include that `t is copy(t)` on versions which are not 2.7 and 3.2?

ilevkivskyi: >  So why is really making a copy?

IIRC, to match the behaviour of 3.2 for other classes.

> Should a test include that ``t is copy(t)`` on versions which are not 2.7 and 3.2?

This is not something super-important, but if you want to make a PR (with a comment quoting ``copy`` docs), then we will add it.

mitar: Done: https://github.com/python/typing/pull/460

PR #314, opened 2016-11-05
Fix exception for invalid objects in get_type_hints

### LINKED ISSUE #313: get_type_hints raises AttributeError instead of TypeError
From what I can tell, the [code here](https://hg.python.org/cpython/file/tip/Lib/typing.py#l1419) is a typo

> `if isinstance(object, type):`

This causes `get_type_hints` to raise 

```
AttributeError: 'int' object has no attribute '__code__'
```
instead of 
```
TypeError: 123 is not a module, class, method, or function.
```

ilevkivskyi:
Fixes #313 
@gvanrossum I did this only in Python 3 since ``get_type_hints`` is not supported in Python 2.

### DIFF
--- a/src/test_typing.py
+++ b/src/test_typing.py
@@ -1318,11 +1318,6 @@ gth = get_type_hints
 
 class GetTypeHintTests(BaseTestCase):
     def test_get_type_hints_from_various_objects(self):
-        # Should not fail for buil-in classes and functions.
-        self.assertEqual(gth(int), {})
-        self.assertEqual(gth(type), {})
-        self.assertEqual(gth(dir), {})
-        self.assertEqual(gth(len), {})
         # For invalid objects should fail with TypeError (not AttributeError etc).
         with self.assertRaises(TypeError):
             gth(123)
--- a/src/test_typing.py
+++ b/src/test_typing.py
@@ -1317,6 +1317,20 @@ if PY36:
 gth = get_type_hints
 
 class GetTypeHintTests(BaseTestCase):
+    def test_get_type_hints_from_various_objects(self):
+        # Should not fail for buil-in classes and functions.
+        self.assertEqual(gth(int), {})
+        self.assertEqual(gth(type), {})
+        self.assertEqual(gth(dir), {})
+        self.assertEqual(gth(len), {})
+        # For invalid objects should fail with TypeError (not AttributeError etc).
+        with self.assertRaises(TypeError):
+            gth(123)
+        with self.assertRaises(TypeError):
+            gth('abc')
+        with self.assertRaises(TypeError):
+            gth(None)
+
     @skipUnless(PY36, 'Python 3.6 required')
     def test_get_type_hints_modules(self):
         self.assertEqual(gth(ann_module), {1: 2, 'f': Tuple[int, int], 'x': int, 'y': str})
--- a/src/typing.py
+++ b/src/typing.py
@@ -1476,7 +1476,7 @@ else:
         elif localns is None:
             localns = globalns
         defaults = _get_defaults(obj)
-        hints = obj.__annotations__
+        hints = dict(obj.__annotations__)
         for name, value in hints.items():
             if isinstance(value, str):
                 value = _ForwardRef(value)
--- a/src/typing.py
+++ b/src/typing.py
@@ -1332,11 +1332,7 @@ def cast(typ, val):
 
 def _get_defaults(func):
     """Internal helper to extract the default arguments, by name."""
-    try:
-        code = func.__code__
-    except AttributeError:
-        # Some built-in functions don't have __code__, __defaults__, etc.
-        return {}
+    code = func.__code__
     pos_count = code.co_argcount
     arg_names = code.co_varnames
     arg_names = arg_names[:pos_count]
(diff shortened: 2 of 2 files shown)


gvanrossum:
Remind me again why there are two versions of get_type_hints(), one fore Python 3.2 and one for 3.3 and higher? Is it because of the ChainMap? I'm not sure I have ever encountered that data structure before and I'm not sure what benefit it adds (I see clearly a downside, which is that I don't understand the code -- this happened presumably because I rushed into accepting this just before 3.6b3). I don't have any intuitive understanding of how it works and despite reading the docs I find it difficult to reason about the code in get_type_hints() that uses it.

Another thought: instead of explicitly checking for and rejecting certain object types, could it at least accept any object that has `__annotations__`?

Sadly the spec for get_type_hints() in PEP 484, PEP 526 and docs.python.org is rather minimal...

ilevkivskyi:
@gvanrossum 
> this happened presumably because I rushed into accepting this just before 3.6b3

Probably you mean 3.6b1 not 3.6b3, this was added together with ``ClassVar`` few months ago. And indeed there are two versions because of ``ChainMap``. The idea behind this is to provide more precise information, where a particular class member appears in MRO. We discussed this with you and Lisa over e-mail in August, quoting you:
> Oh, and it's definitely going to be interesting to extract basic annotations for a class, because you have to carefully map the ``__annotations__`` for each class in the MRO

``ChainMap`` is optimal for this because it mimics exactly how class namespaces are resolved, for example:
```python
>>> class B:
...    x: int
>>> class C(B):
...    y: int
>>> class D(C):
...    y: float
>>> get_type_hints(D)
ChainMap({'y': <class 'float'>}, {'y': <class 'int'>}, {'x': <class 'int'>}, {})
```
at the same time ``ChainMap`` looks like an ordinary ``dict`` from API point of view, so that people who are not interested in details could use it like a normal ``dict``, for example:
```python
>>> for k, v in get_type_hints(D).items():
...     print(k, v)
x <class 'int'>
y <class 'float'>
```
Concerning your other question:
> Another thought: instead of explicitly checking for and rejecting certain object types, could it at least accept any object that has ``__annotations__``?

I was actually also thinking about this. We could also unify code for modules and functions in 3.3+ version and make it more compact. Should I do this here or in a separate PR?

Concerning the docs, I was going to submit a patch at b.p.o. today or tomorrow with docs updates for thing added/changed during last month, I will also include more info about ``get_type_hints`` for classes.

gvanrossum:
Sorry for the memory lapse. :-(

But the more I think about the ChainMap the less I like the idea of returning one here. Once we do that we can never take it back. But I don't think it is serving an actual purpose here. If you really want to know what the difference is between the type hints for a given class and those for its its bases, you should just call get_type_hints() on the class and on the bases you're interested (whether a specific list, or from `__bases__` or from `__mro__`) and compare the dicts. I want get_type_hints() to return a plain dict, so everything you might do with it (including printing it, comparing it to something else, whatever) is completely understood by everyone involved.

I agree (and this is what I meant in that message you quoted) that it's important for get_type_hints() of a class to walk the `__mro__` asking each class whether it has an `__annotations__` or not (probably through introspection of its `__dict__` rather than just using getattr()), because otherwise you'd only get the `__annotations__` of the first one in the chain. And that's what the code using the ChainMap is doing.

But I don't believe returning a ChainMap is the right solution. I would prefer to just update a single result dict rather than collecting all information in a ChainMap. Basically all the manipulations involving cmap and new_map obscure what's going on.

Could you live with it if the block looked like this?
```python
        if isinstance(obj, type):
            hints = {}
            for base in reversed(obj.__mro__):
                ann = base.__dict__.get('__annotations__', {})
                for name, value in ann.items():
                    if value is None:
                        value = type(None)
                    if isinstance(value, str):
                        value = _ForwardRef(value)
                    value = _eval_type(value, globalns, localns)
                    hints[name] = value
            return hints
```

ilevkivskyi:
@gvanrossum 
> return a plain dict, so everything you might do with it (including printing it, comparing it to something else, whatever) is completely understood by everyone involved.

Probably this makes sense. ``ChainMap`` is not a well known structure (although I like it for "nested namespaces"). I am fine with a simpler solution (note that tests should be updated, especially ``test_grammar`` uses ``get_type_hints`` in one test to check how ``__annotations__`` special attribute is created, now I think it should just look directly in ``__annotations__``).
But one question remains unanswered :-)
> Should I do this here or in a separate PR?

gvanrossum:
> Should I do this here or in a separate PR?

I'm strongly in favor of small focused PRs, so I'd actually split this one on two (one for the typo, one for the "similar one for built-in functions"). Then you can fix the ChainMap thing at your leisure (although it would be good if it landed well before 3.6b4).

ilevkivskyi:
@gvanrossum 
So that there will be a series of four PRs if I also make one for this comment
> Another thought: instead of explicitly checking for and rejecting certain object types, could it at least accept any object that has ``__annotations__``?

* fix  original bug ``get_type_hints(123)``
* fix ``get_type_hints(len)``
* allow arbitrary objects with ``__annotations__`` defined
* abandon ``ChainMap``

Probably I will make first three today.

ilevkivskyi:
@gvanrossum OK, I modified this PR so that only original bug is fixed. Two more PRs will follow

ilevkivskyi:
@gvanrossum It looks like I could not make the third PR mentioned above without merge conflicts with two existing. Moreover, after the ``ChainMap`` is removed, there is no more need for two separate versions for 3.2 and 3.3+ and arbitrary objects with defined ``__annotations__`` could be naturally allowed. I will make a single PR tomorrow if the two current PRs are merged.

gvanrossum:
Fixes #313

PR #320, opened 2016-11-06
Prohibit Callable[[...], X]

ilevkivskyi:
@gvanrossum 
Now ``Callable[[...], X]`` is accepted and treated as equivalent to ``Callable[..., X]`` (also ``Callable[[()], X]`` is treated as equivalent to ``Callable[[], X]``). PEP 484 does not specify the former forms, so that this PR makes them prohibited.

I do not have any preference on this, so please feel free to either merge or close this.

### DIFF
--- a/src/typing.py
+++ b/src/typing.py
@@ -1214,13 +1214,13 @@ class CallableMeta(GenericMeta):
             raise TypeError("Callable must be used as "
                             "Callable[[arg, ...], result].")
         args, result = parameters
-        if args is ...:
-            parameters = (..., result)
+        if args is Ellipsis:
+            parameters = (Ellipsis, result)
         else:
             if not isinstance(args, list):
                 raise TypeError("Callable[args, result]: args must be a list."
                                 " Got %.100r." % (args,))
-            parameters = tuple(args), result
+            parameters = (tuple(args), result)
         return self.__getitem_inner__(parameters)
 
     @_tp_cache
@@ -1228,7 +1228,7 @@ class CallableMeta(GenericMeta):
         args, result = parameters
         msg = "Callable[args, result]: result must be a type."
         result = _type_check(result, msg)
-        if args == ...:
+        if args is Ellipsis:
             return super().__getitem__((_TypingEllipsis, result))
         msg = "Callable[[arg, ...], result]: each arg must be a type."
         args = tuple(_type_check(arg, msg) for arg in args)
--- a/src/typing.py
+++ b/src/typing.py
@@ -1218,23 +1218,21 @@ class CallableMeta(GenericMeta):
         args, result = parameters
         if args is ...:
             parameters = (..., result)
-        elif args == []:
-            parameters = ((), result)
         else:
-            if not isinstance(args, list) or ... in args or () in args:
+            if not isinstance(args, list):
                 raise TypeError("Callable[args, result]: args must be"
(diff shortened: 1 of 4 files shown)


gvanrossum:
Strange that was ever accepted. Was this aways there? I don't recall intending to support this.

ilevkivskyi:
@gvanrossum It look like it was introduced in #308 when I split ``__getitem__`` in two parts, so that it could be cached. So that this is my fault. This PR adds tests that prohibit this.

src/typing.py
```diff
@@ -1221,9 +1221,9 @@ def __getitem__(self, parameters):
         elif args == []:
             parameters = ((), result)
         else:
-            if not isinstance(args, list):
-                raise TypeError("Callable[args, result]: args must be a list."
-                                " Got %.100r." % (args,))
+            if not isinstance(args, list) or ... in args or () in args:
```
gvanrossum: Hm... It really shouldn't be necessary to check this here. I think the reason may be that you're trying to make the hash key for e.g. `Callable[[int, str], float]` be the tuple `(int, str, float)` rather than just nesting the original structure. The exceptions in `__getitem_inner__` seem due to this.

Or maybe you're trying to make `__parameters__` a single-level tuple? But there are two exceptions for this, one for `Callable[..., t]` and one for `Callable[[], t]` (IIUC). I don't think this is really worth it...

src/typing.py
```diff
@@ -1221,9 +1221,9 @@ def __getitem__(self, parameters):
         elif args == []:
```
gvanrossum: I don't think you need this special case, and if you remove it and the matching special case in `__getitem_inner__` the tests still passes. (It's different here than for Tuple -- while `Tuple[]` is invalid due to Python's syntax, `Callable[[], t]` is fine, and there's no need to treat an empty list special.

PS. I can't add a comment there, but your code below has `[...,]` and `[(),]` -- those trailing commas are not needed.

src/typing.py
ilevkivskyi: @gvanrossum Yes, indeed I wanted to flatten ``__parameters__`` to just call ``super().__getitem__``. But you are right, it is not necessary to do this here, but only right before I actually call ``super()``.

ilevkivskyi:
@gvanrossum Thanks for review! I pushed new commits taking into account your comments. Now this is much simpler (and probably right) way to fix this.

src/typing.py
```diff
@@ -1218,24 +1216,20 @@ def __getitem__(self, parameters):
...
         return self.__getitem_inner__(parameters)
 
     @_tp_cache
     def __getitem_inner__(self, parameters):
-        *args, result = parameters
+        args, result = parameters
         msg = "Callable[args, result]: result must be a type."
         result = _type_check(result, msg)
-        if args == [...,]:
+        if args == ...:
```
gvanrossum: I'd spell this as `Ellipsis` -- seeing `args == ...` is just too weird. :-) Plus it's a singleton, so you should use `is`.

ilevkivskyi:
@gvanrossum 

I implemented your latest comments in a new commit.

> At some point in the future we should also add comments explaining the inner workings of all these functions. (Or docstrings, but make it clear they are NOT public or even protected APIs.)

This is a good idea, I will make a PR later this week.

PR #321, opened 2016-11-06
Allow keyword syntax for NamedTuple

### LINKED ISSUE #302: Allow keyword syntax for NamedTuple
It was proposed by Guido to add the following syntax:

``` python
Point = typing.NamedTuple('Point', x=int, y=int)
p = Point(x=1, y=2)
p.x
```

I also like it. Probably, we should also support it for `TypedDict`.

ilevkivskyi:
Fixes #302 
@gvanrossum Here is a simple implementation for keyword syntax that you proposed:
```python
Employee = NamedTuple('Employee', name=str, id=int)
```
Note that this requires 3.6+ since order of kwargs should be preserved.

### DIFF
--- a/src/typing.py
+++ b/src/typing.py
@@ -1987,7 +1987,11 @@ if sys.version_info[:2] >= (3, 6):
         The resulting class has one extra attribute: _field_types,
         giving a dict mapping field names to types.  (The field names
         are in the _fields attribute, which is part of the namedtuple
-        API.) Backward-compatible usage::
+        API.) Alternative equivalent keyword syntax is also accepted::
+
+            Employee = NamedTuple('Employee', name=str, id=int)
+
+        Backward-compatible usage::
 
             Employee = NamedTuple('Employee', [('name', str), ('id', int)])
         """
--- a/src/typing.py
+++ b/src/typing.py
@@ -1962,65 +1962,55 @@ def _make_nmtuple(name, types):
     return nm_tpl
 
 
-if sys.version_info[:2] >= (3, 6):
-    class NamedTupleMeta(type):
+_PY36 = sys.version_info[:2] >= (3, 6)
 
-        def __new__(cls, typename, bases, ns, *, _root=False):
-            if _root:
-                return super().__new__(cls, typename, bases, ns)
-            types = ns.get('__annotations__', {})
-            return _make_nmtuple(typename, types.items())
 
-    class NamedTuple(metaclass=NamedTupleMeta, _root=True):
-        """Typed version of namedtuple.
+class NamedTupleMeta(type):
 
-        Usage::
+    def __new__(cls, typename, bases, ns):
+        if ns.get('_root', False):
+            return super().__new__(cls, typename, bases, ns)
+        if not _PY36:
+            raise TypeError("Class syntax for NamedTuple is only supported"
+                            " in Python 3.6+")
+        types = ns.get('__annotations__', {})
+        return _make_nmtuple(typename, types.items())
 
-            class Employee(NamedTuple):
-                name: str
-                id: int
+class NamedTuple(metaclass=NamedTupleMeta):
+    """Typed version of namedtuple.
 
-        This is equivalent to::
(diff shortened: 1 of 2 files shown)


src/typing.py
```diff
@@ -1985,20 +1987,29 @@ class Employee(NamedTuple):
         The resulting class has one extra attribute: _field_types,
         giving a dict mapping field names to types.  (The field names
         are in the _fields attribute, which is part of the namedtuple
-        API.) Backward-compatible usage::
+        API.) Alternative equivalent keyword syntax is also accepted::
```
gvanrossum: Can you change the earlier part of the docstring (line 1975) to emphasize that the both the PEP 526 syntax and the keyword args syntax are only supported in Python 3.6+? (I can't add github comments that far back.)

src/typing.py
```diff
@@ -1985,20 +1987,29 @@ class Employee(NamedTuple):
...
+
+        Backward-compatible usage::
 
             Employee = NamedTuple('Employee', [('name', str), ('id', int)])
         """
 
-        def __new__(self, typename, fields):
+        def __new__(self, typename, fields=None, **kwargs):
+            if fields is None:
+                fields = kwargs.items()
```
gvanrossum: This should be disallowed for Python <= 3.5. (Remember that this code will be in the "typing" package which should work for Python 3.2 and up, and it will be copied verbatim into the 3.5 branch as well.)

src/typing.py
```diff
@@ -1985,20 +1987,29 @@ class Employee(NamedTuple):
         The resulting class has one extra attribute: _field_types,
         giving a dict mapping field names to types.  (The field names
         are in the _fields attribute, which is part of the namedtuple
-        API.) Backward-compatible usage::
+        API.) Alternative equivalent keyword syntax is also accepted::
+
+            Employee = NamedTuple('Employee', name=str, id=int)
+
+        Backward-compatible usage::
```
gvanrossum: So "backward-compatible" sounds a bit pejorative. I think this should just be referred to as the way to do it for Python versions <= 3.5.

gvanrossum:
FWIW I'm fine with having this in 3.6 without supporting it in mypy. Once 3.6 is released we can't change typing.py easily (not until 3.6.1 anyways) but we can add support to mypy at any time.

src/typing.py
ilevkivskyi: @gvanrossum Note that this docstring will be seen _only_ by people who use Python 3.6+, this docstring is inside a big ``if sys.version_info[:2] >= (3, 6):`` above. Anyway, I updated the docstring to cover this and next comment.

src/typing.py
ilevkivskyi: @gvanrossum This code is also inside ``if sys.version_info[:2] >= (3, 6):`` I mentioned above. So that the ``else:`` branch is used for other versions, and both keyword and class syntax fail with a ``TypeError``.

I will add a new commit now, where I will "unify" the code, this will give better error messages.

src/typing.py
ilevkivskyi: @gvanrossum OK, now there is no big ``if sys.version_info[:2] >= (3, 6):``, there is the same code and docstring for all versions.

ilevkivskyi:
@gvanrossum Thank you for review! I have implemented your comments.
In addition, think it is better to unify the code. First it will be easier to maintain a single version, second this gives better error messages saying explicitly
```
Class syntax for NamedTuple is only supported in Python 3.6+
```
 and
```
Keyword syntax for NamedTuple is only supported in Python 3.6+
```
Please take a look!

ilevkivskyi:
@gvanrossum Does this look OK now?

gvanrossum:
OK. Another leap of faith. :-) Did you have other changes to typing.py? Or can I do the final merge into CPython and mypy now?

ilevkivskyi:
@gvanrossum Thanks! :-)
> Did you have other changes to typing.py?

There is #322 , but note that mypy support for ``TypedDict`` is not complete yet.

PR #325, opened 2016-11-09
Add cleanup handles for _tp_cache

ilevkivskyi:
@gvanrossum This is the second PR that we discussed on b.p.o. (typing part of Serhiy's patch)


gvanrossum:
Can you backport this too? I've got the feeling that it's important to be able to clear the caches everywhere (even though for Python 2 we don't have refleak.py).

gvanrossum:
I'll merge it now just so we can make progress on the Python 3 bug. But I'd still like a backport.

PR #331, opened 2016-11-21
Improve docstrings and error messages

ilevkivskyi:
@gvanrossum 

As I promised here is a PR that improves some docstrings and error messages.
Mainly theses are small updates for docstring following recent changes, also as we discussed before I marked many classes/functions as internal (sometimes even when this is probably obvious).

In addition, I changed ``__repr__`` for ``_ForwardRef``, so that now ``List['C']`` is represented as ``typing.List['C']`` (before it was ``List[_ForwardRef('C')]``). The main motivation is that ``_ForwardRef`` is an internal class and users should not see it. Also it is a bit shorter/nicer.

### DIFF
--- a/src/typing.py
+++ b/src/typing.py
@@ -247,7 +247,7 @@ class _ForwardRef(_TypingBase, _root=True):
         raise TypeError("Forward references cannot be used with issubclass().")
 
     def __repr__(self):
-        return repr(self.__forward_arg__)
+        return '_ForwardRef(%r)' % (self.__forward_arg__,)
 
 
 class _TypeAlias(_TypingBase, _root=True):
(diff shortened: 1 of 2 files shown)


src/typing.py
```diff
@@ -247,7 +247,7 @@ def __subclasscheck__(self, cls):
         raise TypeError("Forward references cannot be used with issubclass().")
 
     def __repr__(self):
-        return '_ForwardRef(%r)' % (self.__forward_arg__,)
+        return repr(self.__forward_arg__)
```
ilevkivskyi: @gvanrossum OK, I reverted this part in a new commit, I don't have a strong preference here.

gvanrossum:
Thanks! Merged into CPython 3.5, 3.6 and 3.7, see http://bugs.python.org/issue28556.

Hopefully this is it -- after 3.6b4 we only have 1-2 release candidates, and for those the bar is *very* high; we'd have to have evidence of a regression that would break some app.

PR #333, opened 2016-11-21
fix #332: Error when using Generic and __slots__

### LINKED ISSUE #332: Error when using Generic and __slots__
```python

class C(Generic[TypeVar('T')]):
	__slots__ = ('potato',)

# ValueError: 'potato' in __slots__ conflicts with class variable
C[int]
```

This is because bare `C` is created with a class member descriptor `potato`, and instantiating it tries to create a class with all the attributes in `C`.

valtron:


the-knights-who-say-ni:

Unfortunately we couldn't find an account corresponding to your GitHub username at [bugs.python.org](http://bugs.python.org/) (b.p.o). If you don't already have an account at b.p.o, please [create one](http://bugs.python.org/user?@template=register) and make sure to add your GitHub username. If you do already have an account at b.p.o then please go there and under "Your Details" add your GitHub username.

And in case you haven't already, please make sure to sign the [PSF contributor agreement](https://www.python.org/psf/contrib/contrib-form/) (CLA); we can't legally look at your contribution until you have signed the CLA.

Once you have done everything that's needed, please reply here and someone will verify everything is in order.

gvanrossum:
@valtron Why'd you close this? (There's no need to close it until you've signed the CLA -- unless you intend not to sign it, in which case we'll have to disregard your PR and implement our own fix.)

valtron:
I don't intend to sign it.  Sorry for the trouble.

gvanrossum:
@valtron Wait, that's unfortunate -- you'll never be able to contribute to Python. What's your concern?

valtron:
The CLA asks for identifying information. Though I understand why, and that it will be kept within the PSF, I prefer not disclosing it.

gvanrossum:
That's too bad, but I understand. Thanks for the bug report!

PR #334, opened 2016-11-24
Clear class variables created from __slots__

### LINKED ISSUE #332: Error when using Generic and __slots__

ilevkivskyi:
Fixes #332 

The class variables created from slots are now cleared before copying generic class ``__dict__`` on subscription, copying, and evaluation. This is necessary only for Python 3, but I added tests for both 3 and 2.

@gvanrossum Please, take a look.

### DIFF
--- a/python2/test_typing.py
+++ b/python2/test_typing.py
@@ -866,6 +866,8 @@ class GenericTests(BaseTestCase):
             c_int.tomato = 0
 
         self.assertEqual(typing._eval_type(C['C'], globals(), locals()), C[C])
+        self.assertEqual(typing._eval_type(C['C'], globals(), locals()).__slots__,
+                         C.__slots__)
         self.assertEqual(copy(C[int]), deepcopy(C[int]))
 
     def test_errors(self):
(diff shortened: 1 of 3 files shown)


gvanrossum:
LG, except I wonder if `_safe_copy` is too general a name for that helper function.

Also, could you add some tests for cases where `__slots__` is actually a dict? (In that case the values will be ignored, but the idea is that some other metaclass might look at them.)

ilevkivskyi:
@gvanrossum 
> _safe_copy is too general a name for that helper function

I changed the name to simply ``_no_slots_copy``. I also added the tests where ``__slots__`` is a dictionary.

gvanrossum:
LG. I'm curious why this issue is not a problem in Python 2? (You added the same tests but no change to `python2/typing.py` and those tests pass.)

ilevkivskyi:
@gvanrossum In Python 2 all descriptors are silently overwritten with anything that is found in ``__slots__`` (now I actually recall that I have seen this problem somewhere while porting from Python 2 to Python 3).

gvanrossum:
Filed http://bugs.python.org/issue28790 to track merging into CPython.

PR #337, opened 2016-11-30
Add type for WSGI Applications in typing

rowillia:
I'm working on adding stubs for Flask, and WSGI Applications are one of the types that
get frequently passed around.  This seems pretty core to Python itself so it seems
to make sense to add to `typing` itself.

### DIFF
--- a/python2/test_typing.py
+++ b/python2/test_typing.py
@@ -1588,6 +1588,13 @@ class RETests(BaseTestCase):
                          "Cannot subclass typing._TypeAlias")
 
 
+class WSGIApplicationTests(BaseTestCase):
+
+    def test_wsgi_application(self):
+        from typing import Callable, WSGIApplication
+        self.assertIsSubclass(WSGIApplication, Callable)
+
+
 class AllTests(BaseTestCase):
     """Tests for __all__."""
 
--- a/python2/typing.py
+++ b/python2/typing.py
@@ -69,6 +69,7 @@ __all__ = [
     'overload',
     'Text',
     'TYPE_CHECKING',
+    'WSGIApplication',
 ]
 
 # The pseudo-submodules 're' and 'io' are part of the public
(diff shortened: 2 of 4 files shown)


the-knights-who-say-ni:

Unfortunately we couldn't find an account corresponding to your GitHub username at [bugs.python.org](http://bugs.python.org/) (b.p.o). If you don't already have an account at b.p.o, please [create one](http://bugs.python.org/user?@template=register) and make sure to add your GitHub username. If you do already have an account at b.p.o then please go there and under "Your Details" add your GitHub username.

And in case you haven't already, please make sure to sign the [PSF contributor agreement](https://www.python.org/psf/contrib/contrib-form/) (CLA); we can't legally look at your contribution until you have signed the CLA.

Once you have done everything that's needed, please reply here and someone will verify everything is in order.

rowillia:
@the-knights-who-say-ni Associated my GH Account!

gvanrossum:
I recommend trying to find a different place for these. typing.py is in the stdlib for Python 3.5 and up, and can only be changed once per bugfix release. In particular the 3.6 beta cycle is closed so you'd have to wait for 3.6.1.

JukkaL:
Currently there doesn't seem to be a really satisfying way to add new type aliases such as these, especially if they might be useful in user code as well. It's possible to just add them to some library stub files, but then they wouldn't exist at runtime and users would need to use hacks like this:

```
from typing import TYPE_CHECKING
if TYPE_CHECKING:
    from lib import MyAlias
def f(x: 'MyAlias') -> None: ...
```

One option would to create a separate helper module/package for additional type definitions that can used both in typeshed stubs and imported at runtime, assuming that a particular pip package has been installed. This wouldn't be problem-free either: some types might require `typing` features that require a recent `typing`. We could work around this by making the definitions conditional on what's available in the installed `typing` module. Also, these types might not be very discoverable.

rowillia:
@JukkaL @gvanrossum Thanks for the feedback.  Moved to https://github.com/python/typeshed/pull/725/files for now...I agree this would be a generally useful thing to have though!

PR #338, opened 2016-12-02
namedtuple: add defaults to the Python 3.6 syntax

JelleZijlstra:
This lets you write:

    class Employee(NamedTuple):
        name: str
        id: int
        language: str = 'python'

as suggested by @gvanrossum on python-ideas.

Should I also add this to any relevant documentation? I imagine mypy will also need changes to support this syntax.

### DIFF
--- a/src/typing.py
+++ b/src/typing.py
@@ -1917,17 +1917,18 @@ class NamedTupleMeta(type):
                             " in Python 3.6+")
         types = ns.get('__annotations__', {})
         nm_tpl = _make_nmtuple(typename, types.items())
-        saw_default = False
         defaults = []
         defaults_dict = {}
         for field_name in types:
             if field_name in ns:
-                saw_default = True
                 default_value = ns[field_name]
                 defaults.append(default_value)
                 defaults_dict[field_name] = default_value
-            elif saw_default:
-                raise TypeError('Non-default namedtuple field cannot follow default field')
+            elif defaults:
+                raise TypeError("Non-default namedtuple field {field_name} cannot follow default"
+                                " field(s) {default_names}"
+                                .format(field_name=field_name,
+                                        default_names=', '.join(defaults_dict.keys())))
         nm_tpl.__new__.__defaults__ = tuple(defaults)
         nm_tpl._field_defaults = defaults_dict
         return nm_tpl
--- a/src/typing.py
+++ b/src/typing.py
@@ -1916,7 +1916,21 @@ class NamedTupleMeta(type):
             raise TypeError("Class syntax for NamedTuple is only supported"
                             " in Python 3.6+")
         types = ns.get('__annotations__', {})
-        return _make_nmtuple(typename, types.items())
+        nm_tpl = _make_nmtuple(typename, types.items())
+        saw_default = False
+        defaults = []
+        defaults_dict = {}
+        for field_name in types:
+            if field_name in ns:
+                saw_default = True
+                default_value = ns[field_name]
+                defaults.append(default_value)
+                defaults_dict[field_name] = default_value
+            elif saw_default:
+                raise TypeError('Non-default namedtuple field cannot follow default field')
+        nm_tpl.__new__.__defaults__ = tuple(defaults)
+        nm_tpl._field_defaults = defaults_dict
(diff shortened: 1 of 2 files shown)


ruxi:
How does it work? Would this be valid:

    Employee = typing.NamedTuple("Employee", [("name", str), ("id", int)] )

    def foo(bar:Employee):
        pass

JelleZijlstra:
@ruxi Yes, that syntax already works in Python 3.5. Python 3.6's variable annotation syntax makes it possible to instead create typed NamedTuples using subclassing, like:
```
class Employee(typing.NamedTuple):
    name: str
    id: int
```

This pull request extends that syntax to use the values of variables assigned in the class body as default values.

gvanrossum:
Let's wait until after Python 3.6.0 and 3.5.3 are released. I don't want to
rock the release boat any more than necessary.

gvanrossum:
(But apart from my desire to delay merging, it's beautiful! Now of course we need support for mypy too...)

src/typing.py
```diff
@@ -1916,7 +1916,21 @@ def __new__(cls, typename, bases, ns):
             raise TypeError("Class syntax for NamedTuple is only supported"
                             " in Python 3.6+")
         types = ns.get('__annotations__', {})
-        return _make_nmtuple(typename, types.items())
+        nm_tpl = _make_nmtuple(typename, types.items())
+        saw_default = False
```
ilevkivskyi: I think ``saw_default`` is redundant. You could use ``elif defaults:`` or ``elif defaults_dict:``, instead of ``elif saw_default:`` below.

src/typing.py
```diff
@@ -1916,7 +1916,21 @@ def __new__(cls, typename, bases, ns):
...
+        defaults = []
+        defaults_dict = {}
+        for field_name in types:
+            if field_name in ns:
+                saw_default = True
+                default_value = ns[field_name]
+                defaults.append(default_value)
+                defaults_dict[field_name] = default_value
+            elif saw_default:
+                raise TypeError('Non-default namedtuple field cannot follow default field')
```
ilevkivskyi: It could be sometimes helpful to mention which fields are conflicting, for example ``f'Non-default namedtuple field {field_name} cannot follow default fields {list(defaults_dict.keys())}'``.

src/typing.py
```diff
@@ -1916,7 +1916,21 @@ def __new__(cls, typename, bases, ns):
...
+        nm_tpl.__new__.__defaults__ = tuple(defaults)
```
ilevkivskyi: Also probably you could use only ``defaults_dict`` and then write here ``= tuple(defaults_dict.values())``, since ``dict`` is ordered in Python 3.6.

ilevkivskyi:
@JelleZijlstra Thank you!
I just added few optional comments.

src/typing.py
JelleZijlstra: As I understand it, dict ordering is an implementation detail of CPython 3.6 and a conforming Python 3.6 implementation could choose to not implement ordering, so this code shouldn't rely on dict ordering (except for class namespaces and **kwargs because of PEP 520 and PEP 468).

gvanrossum:
I'm okay with this idea. @ilevkivskyi Is it good from your POV? We also need a mypy implementation, and docs for Python.

ilevkivskyi:
I like the idea. The mypy implementation should not be difficult, but it is not high on my priority list (I will make PRs for several other mypy issues soon).

ilevkivskyi:
Although I would prefer this PR to be merged without waiting for mypy support, to avoid possible conflicts with #352.

JelleZijlstra:
Thanks! I'll work on implementing this in mypy as soon as I finish implementing async generators and will add a doc patch to CPython once this is merged.

gvanrossum:
OK, I'll leave it to @ilevkivskyi to make sure all the parts land.

gvanrossum:
Also merged this to CPython.

(Note: we need a Misc/NEWS item for the various typing.py changes.)

JelleZijlstra:
Thanks! Created http://bugs.python.org/issue29310 to track documenting this in CPython.

PR #340, opened 2016-12-04
Make public and document GenericMeta

### LINKED ISSUE #335: Document and expose GenericMeta
In https://github.com/python/typeshed/pull/703 a reasonable request was made to expose GenericMeta. While that PR just adds it to the *stub* for typing.py, we should think about and carefully document what behavior subclasses of GenericMeta can depend on.

In particular, I *don't* think it should be allowed to write your own version of Generic using GenericMeta as a metaclass -- type checkers aren't expected to understand that. But overriding `__new__` and invoking `super().__new__(...)` should be allowed. I'm not sure about the other methods -- e.g. can a subclass of GenericMeta override `__getitem__`, `__copy__` or `__isinstancecheck__`? (Assuming they use `super()`.)

@ilevkivskyi Thoughts?

ilevkivskyi:
Fixes #335 

I added ``GenericMeta`` to ``__all__`` and added docstrings to ``GenericMeta`` and ``GenericMeta.__new__`` following the discussion in the issue.
I decided not to describe ``__origin__ `` and friends in docstrings. Instead, I just added a comment at the place where they are first used.

### DIFF
--- a/python2/typing.py
+++ b/python2/typing.py
@@ -28,6 +28,8 @@ __all__ = [
 
     # ABCs (from collections.abc).
     'AbstractSet',  # collections.abc.Set.
+    'GenericMeta',  # subclass of abc.ABCMeta and a metaclass
+                    # for 'Generic' and ABCs below.
     'ByteString',
     'Container',
     'Hashable',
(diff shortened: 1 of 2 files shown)


gvanrossum:
Harmless though this is I think it's also not critical to get it into 3.6.0. People who want it can use `from typing import GenericMeta`, which works regardless.

PR #346, opened 2017-01-07
add AsyncGenerator

JelleZijlstra:
Looks like this was inadvertently omitted from 3.6 as released.

Should I submit a separate patch to CPython directly to add documentation?

### DIFF
--- a/src/test_typing.py
+++ b/src/test_typing.py
@@ -1344,7 +1344,6 @@ class OverloadTests(BaseTestCase):
 
 
 ASYNCIO = sys.version_info[:2] >= (3, 5)
-ASYNC_GENERATOR = sys.version_info[:2] >= (3, 6)
 
 ASYNCIO_TESTS = """
 import asyncio
--- a/src/test_typing.py
+++ b/src/test_typing.py
@@ -1820,7 +1820,12 @@ class CollectionsAbcTests(BaseTestCase):
 
     @skipUnless(PY36, 'Python 3.6 required')
     def test_subclassing_async_generator(self):
-        class G(typing.AsyncGenerator[int, int]): ...
+        class G(typing.AsyncGenerator[int, int]):
+            def asend(self, value):
+                pass
+            def athrow(self, typ, val=None, tb=None):
+                pass
+
         ns = {}
         exec('async def g(): yield 0', globals(), ns)
         g = ns['g']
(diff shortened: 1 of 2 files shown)


gvanrossum:
Thanks! I will merge this after 3.5.3 has been released.

gvanrossum:
And yes, please send a separate doc patch.

JelleZijlstra:
http://bugs.python.org/issue29198

src/typing.py
```diff
@@ -1859,6 +1860,21 @@ def __new__(cls, *args, **kwds):
...
         return _generic_new(_G_base, cls, *args, **kwds)
 
+if hasattr(collections_abc, 'AsyncGenerator'):
+    _AG_base = collections_abc.AsyncGenerator
+
+    class AsyncGenerator(AsyncIterator[T_co], Generic[T_co, T_contra],
+                         extra=_AG_base):
+        __slots__ = ()
+
+        def __new__(cls, *args, **kwds):
```
ilevkivskyi: Since you always inherit from ``collections.abc`` there is no need to override ``__new__``. It will be anyway not possible to instantiate ``typing.AsyncGenerator`` because ``collections.abc.AsyncGenerator`` has abstract methods.

JelleZijlstra: This was mostly copied from the `Generator` implementation just above. It is possible to instantiate a subclass of `AsyncGenerator`, which would use this `__new__`.

ilevkivskyi: I don't think I understand your point. The only way to instantiate a subclass of an ABC is to implement all abstract methods. Moreover, there is no way a subclass could "use" this ``__new__``, it only exists everywhere in ``typing`` to prohibit direct instantiation of generic version of concrete collections and similar things.

``__new__`` for ``Generator`` is necessary, since it could inherit from ``types.GeneratorType``.

src/test_typing.py
```diff
@@ -1320,6 +1320,7 @@ def blah():
 
 
 ASYNCIO = sys.version_info[:2] >= (3, 5)
+ASYNC_GENERATOR = sys.version_info[:2] >= (3, 6)
```
gvanrossum: Why is this line needed? The variable is not used AFAICT.

JelleZijlstra: Yes, looks like I didn't end up needing it. I'll remove it.

gvanrossum:
@ilevkivskyi Are you happy with this code now? Then I can merge it. (Honestly I want to give you commit privs here if you don't already have them.)

ilevkivskyi:
> Are you happy with this code now?

There are two very minor points: the variable ``_AG_base`` is used only once, and the last test ``self.assertNotIsInstance(type(g), G)`` is not very useful, I would add ``self.assertNotIsInstance(g, G)``. Otherwise I am happy.

> Honestly I want to give you commit privs here if you don't already have them.

I am glad that you trust me. I don't have the privileges. I think it indeed may simplify things.

gvanrossum:
I invited you. Go ahead and commit this PR when you are happy with it.

ilevkivskyi:
Probably the best way is to merge this now and I will add those minor changes myself, so that #355 will be "unblocked" very soon.

PR #350, opened 2017-01-12
Fix edge case in comparison of Union to other objects/types

ilevkivskyi:
Fixes http://bugs.python.org/issue29246

This is an extended version of the fix proposed on b.p.o. + Python 2.

@gvanrossum There are 7 PRs open (all with minor fixes), maybe it makes sense to start merging them here now and later make a single merge to master CPython repo?

### DIFF
--- a/python2/test_typing.py
+++ b/python2/test_typing.py
@@ -278,6 +278,15 @@ class UnionTests(BaseTestCase):
         self.assertFalse(Union[str, typing.Iterable[int]] == typing.Iterable[int])
         self.assertTrue(Union[str, typing.Iterable] == typing.Iterable)
 
+    def test_union_compare_other(self):
+        self.assertNotEqual(Union, object)
+        self.assertNotEqual(Union, Any)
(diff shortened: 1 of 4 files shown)


gvanrossum:
> This is an extended version of the fix proposed on b.p.o. + Python 2.

Let's make sure to add the name of the author of that fix (Spiro Sideris) to the commit message.

> @gvanrossum There are 7 PRs open (all with minor fixes), maybe it makes sense to start merging them here now and later make a single merge to master CPython repo?

I still prefer to take care of all of these once Python 3.5.3 is out ([expected Jan 16](https://www.python.org/dev/peps/pep-0478/)). Or is there a concern about merge conflicts?

ilevkivskyi:
> Let's make sure to add the name of the author of that fix (Spiro Sideris) to the commit message.

Fixed.

> I still prefer to take care of all of these once Python 3.5.3 is out (expected Jan 16). Or is there a concern about merge conflicts?

I was a bit worried but now I tried to merge branches locally and everything is fine (+ all tests pass).

PR #354, opened 2017-01-17
Add typing.Deque

### LINKED ISSUE #353: Add typing.Deque
See http://bugs.python.org/issue29011 and commit https://hg.python.org/cpython/rev/dfefbf6f6c73 -- this was committed to the stdlib for 3.5-3.7 already, so should be copied into the typing repo and backported to Python 2.

ilevkivskyi:
Fixes #353 

I have noticed that Raymond added one blanc line between classes instead of two, and it is better to use full name "collections.deque()" in the error message (as we do in other places).

@gvanrossum If you think it is better to fix this in a separate PR then this one could be merged, otherwise I will add one more commit here.

### DIFF
--- a/python2/test_typing.py
+++ b/python2/test_typing.py
@@ -1272,14 +1272,6 @@ class CollectionsAbcTests(BaseTestCase):
         with self.assertRaises(TypeError):
             typing.Set[int]()
 
-    def test_no_set_instantiation(self):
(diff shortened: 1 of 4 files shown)


gvanrossum:
Let's do the cleanup in a separate diff.

Maybe we should have a flake8 run in the tests to remind us.

PR #360, opened 2017-01-18
Fix if statements with ambiguously indented bodies

ambv:
Since this was related to the same lines in conditions, also fixed binary
operators on line beginnings.

### DIFF
--- a/src/typing.py
+++ b/src/typing.py
@@ -355,13 +355,17 @@ def _type_check(arg, msg):
         return type(None)
     if isinstance(arg, str):
         arg = _ForwardRef(arg)
-    if (isinstance(arg, _TypingBase) and type(arg).__name__ == '_ClassVar' or
-        not isinstance(arg, (type, _TypingBase)) and not callable(arg)):
+    if (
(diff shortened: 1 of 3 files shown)


src/typing.py
```diff
@@ -1443,10 +1452,12 @@ def get_type_hints(obj, globalns=None, localns=None):
     hints = getattr(obj, '__annotations__', None)
     if hints is None:
         # Return empty annotations for something that _could_ have them.
-        if (isinstance(obj, types.FunctionType) or
+        if (
+            isinstance(obj, types.FunctionType) or
             isinstance(obj, types.BuiltinFunctionType) or
             isinstance(obj, types.MethodType) or
-            isinstance(obj, types.ModuleType)):
+            isinstance(obj, types.ModuleType)
```
ilevkivskyi: Maybe this could be rewritten as ``isinstance(obj, (types.FunctionType, types.BuiltinFunctionType, types.MethodType, types.ModuleType))``?

ambv: Indeed, I tried to make minimal changes to make the review easier.

PR #361, opened 2017-01-20
Enable flake8 for test_*.py on Travis CI

ambv:
For now this requires a test-specific config while I plow through the warnings
that are specific to tests.

To test this actually does something, I fixed F821 by putting the fake names to
make flake8 happy.  Hopefully you don't find this too terrible.  The other
choice was to disable "undefined names" warnings entirely.  I didn't want to do
that because I find them very useful.

### DIFF
--- a/src/test_typing.py
+++ b/src/test_typing.py
@@ -1381,6 +1381,9 @@ if ASYNCIO:
         exec(ASYNCIO_TESTS)
     except ImportError:
         ASYNCIO = False
+else:
+    # fake names for the sake of static analysis
+    AwaitableWrapper = AsyncIteratorWrapper = object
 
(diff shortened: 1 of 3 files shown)


src/test_typing.py
```diff
@@ -1381,6 +1381,9 @@ def __anext__(self) -> T_a:
         exec(ASYNCIO_TESTS)
     except ImportError:
         ASYNCIO = False
+else:
+    # fake names for the sake of static analysis
```
ilevkivskyi: I would add ``asyncio = None`` here (as you do for PY36), just for completeness. But I don't want to delay the PR because of this minor thing.

ilevkivskyi:
Thank you!

PR #364, opened 2017-01-23
Raise AttributeError on name collision in NamedTuple

ilevkivskyi:
This is a follow-up for #362 

This explicitly prohibits overwriting ``NamedTuple`` fields and special attributes:
```python
class X(NamedTuple):
    x: int
    def _fileds(self):  # This will be an error
        ...
```

### DIFF
--- a/src/test_typing.py
+++ b/src/test_typing.py
@@ -1418,11 +1418,6 @@ class XMeth(NamedTuple):
     x: int
     def double(self):
         return 2 * self.x
-
-class XMethBad(NamedTuple):
-    x: int
-    def _fields(self):
-        return 'no chance for this'
 """
 
 if PY36:
@@ -2057,8 +2052,14 @@ class NonDefaultAfterDefault(NamedTuple):
     def test_annotation_usage_with_methods(self):
         self.assertEquals(XMeth(1).double(), 2)
         self.assertEquals(XMeth(42).x, XMeth(42)[0])
-        self.assertEquals(XMethBad(1)._fields, ('x',))
-        self.assertEquals(XMethBad(1).__annotations__, {'x': int})
+
+        with self.assertRaises(AttributeError):
+            exec("""
+class XMethBad(NamedTuple):
+    x: int
+    def _fields(self):
+        return 'no chance for this'
+""")
 
     @skipUnless(PY36, 'Python 3.6 required')
     def test_namedtuple_keyword_usage(self):
--- a/src/test_typing.py
+++ b/src/test_typing.py
@@ -1425,8 +1425,7 @@ if PY36:
 else:
     # fake names for the sake of static analysis
     ann_module = ann_module2 = ann_module3 = None
-    A = B = CSub = G = CoolEmployee = CoolEmployeeWithDefault = object
-    XMeth = XMethBad = object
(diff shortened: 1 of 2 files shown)


ambv:
This is interesting. I haven't found examples before of using `AttributeError`s to denote **too many** attributes :) But looking at available base exceptions this one is actually the most sensible, maybe along `TypeError`.

gvanrossum:
LGTM

src/test_typing.py
```diff
@@ -2057,8 +2051,14 @@ class NonDefaultAfterDefault(NamedTuple):
     def test_annotation_usage_with_methods(self):
         self.assertEquals(XMeth(1).double(), 2)
         self.assertEquals(XMeth(42).x, XMeth(42)[0])
-        self.assertEquals(XMethBad(1)._fields, ('x',))
-        self.assertEquals(XMethBad(1).__annotations__, {'x': int})
+
+        with self.assertRaises(AttributeError):
```
JelleZijlstra: Can you also add a test where a method name conflicts with a NamedTuple fields? `_fields` seems like a slightly different case.

ilevkivskyi: Thanks you for suggestion. I actually tried this and I now realise that it is not possible to raise an error in this case - defining a method with the same name as a field is interpreted as a default value. But I think this is OK, currently these are all equivalent at runtime:
```python
class A(NamedTuple):
    a: Callable = lambda s: s
class A(NamedTuple):
    a: Callable
    a = lambda s: s
class A(NamedTuple):
    a: Callable
    def a(s):
        return s
```
Some of those could be prohibited in mypy however.

elazarg:
@ilevkivskyi thanks for the pointer.

I do not understand the reasoning behind this PR. If I subclass NamedTuple, I should know how it works (otherwise I shouldn't subclass it). If I know how it works and I decide to implement e.g. `__str__` or `_fields`, then why forbid it? It is inconsistent with how subclassing works, and gives incentive to write code like this:

```Python
class _A(NamedTuple):
    name: str
    val: int

class A(_A):
    def __str__(self):
        return f'({self.name} -> {self.val})'
```

This way I am losing the nice syntax and immutability. For the code I am writing now, I gave up using NamedTuple for this reason: I want to implement special methods.

ilevkivskyi:
@elazarg As I understand, what you want is unrelated to this PR. Before this PR, all attempts to overwrite a special attributes were silently ignored. Here are some points on your comment:
1. In your example you can just write ``A.__repr__ = ...`` instead of creating a subclass, or if you prefer to create a subclass, then add ``__slots__ = ()`` to keep the "immutability".
2. Namedtuple is a very popular construct, sometimes used even by beginners. This is why it is somewhat restrictive (e.g. field names could not start with an underscore). Currently, we are already more permissive than original ``collections.namedtuple`` since we allow methods to start from underscores.
3. I am **in favor** of allowing to override ``__repr__`` and ``__str__``, maybe also some other methods. This would be a one-line fix.
4. However, I am against allowing everything. If someone really knows how named tuples work, then one could simply write  ``A._fields = ...``.

elazarg:
OK. Disallowing `_fields` makes sense; but please consider allowing operators (and I don't think there are special pitfalls for beginners there).

Thank you!

PR #366, opened 2017-01-25
Add typing.Counter and typing.ChainMap

JelleZijlstra:
Similar to deque, which was added recently.

### DIFF
--- a/python2/test_typing.py
+++ b/python2/test_typing.py
@@ -1349,6 +1349,16 @@ class CollectionsAbcTests(BaseTestCase):
         with self.assertRaises(TypeError):
             typing.Counter[int]()
 
+    def test_counter_subclass_instantiation(self):
+
+        class MyCounter(typing.Counter[int]):
+            pass
+
+        d = MyCounter()
+        self.assertIsInstance(d, MyCounter)
+        self.assertIsInstance(d, typing.Counter)
+        self.assertIsInstance(d, collections.Counter)
+
     def test_no_set_instantiation(self):
         with self.assertRaises(TypeError):
             typing.Set()
--- a/python2/test_typing.py
+++ b/python2/test_typing.py
@@ -1341,13 +1341,12 @@ class CollectionsAbcTests(BaseTestCase):
         class D(typing.Deque[T]): pass
         self.assertIs(type(D[int]()), D)
 
-    def test_no_counter_instantiation(self):
-        with self.assertRaises(TypeError):
-            typing.Counter()
-        with self.assertRaises(TypeError):
-            typing.Counter[T]()
-        with self.assertRaises(TypeError):
-            typing.Counter[int]()
+    def test_counter_instantiation(self):
+        self.assertIs(type(typing.Counter()), collections.Counter)
+        self.assertIs(type(typing.Counter[T]()), collections.Counter)
+        self.assertIs(type(typing.Counter[int]()), collections.Counter)
+        class C(typing.Counter[T]): pass
+        self.assertIs(type(C[int]()), C)
 
     def test_counter_subclass_instantiation(self):
 
--- a/python2/test_typing.py
+++ b/python2/test_typing.py
@@ -1264,6 +1264,9 @@ class CollectionsAbcTests(BaseTestCase):
     def test_deque(self):
         self.assertIsSubclass(collections.deque, typing.Deque)
 
+    def test_counter(self):
+        self.assertIsSubclass(collections.Counter, typing.Counter)
+
     def test_set(self):
         self.assertIsSubclass(set, typing.Set)
         self.assertNotIsSubclass(frozenset, typing.Set)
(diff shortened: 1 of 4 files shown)


python2/typing.py
```diff
@@ -1718,6 +1719,17 @@ def __new__(cls, *args, **kwds):
         return _generic_new(collections.deque, cls, *args, **kwds)
 
 
+class Counter(collections.Counter, MutableSequence[T]):
```
ilevkivskyi: I believe this should be ``Dict[T, int]`` instead

src/typing.py
```diff
@@ -1829,6 +1830,17 @@ def __new__(cls, *args, **kwds):
         return _generic_new(collections.deque, cls, *args, **kwds)
 
 
+class Counter(collections.Counter, MutableSequence[T], extra=collections.Counter):
```
ilevkivskyi: ``Dict[T, int]``, see above comment.

ilevkivskyi:
I am not against this. If others are also not against this, then I think we could add these two new generic collections.

src/typing.py
```diff
@@ -1899,6 +1911,22 @@ def __new__(cls, *args, **kwds):
         return _generic_new(collections.defaultdict, cls, *args, **kwds)
 
 
+if hasattr(collections, 'ChainMap'):
+    # ChainMap only exists in 3.3+
+    __all__.append('ChainMap')
+
+    class ChainMap(collections.ChainMap, MutableMapping[KT, VT],
```
ilevkivskyi: typeshed defines this as ``Dict[KT, VT]``, I am not sure this is right, but typing and typeshed should be in sync for sure.

JelleZijlstra: I think MutableMapping is right; ChainMap is not a subclass of dict. I'll submit that change to typeshed.

python2/typing.py
JelleZijlstra: Yes, sorry for missing that.

python2/typing.py
```diff
@@ -1784,6 +1785,17 @@ def __new__(cls, *args, **kwds):
...
 
 
+class Counter(collections.Counter, Dict[T, int]):
+    __slots__ = ()
+    __extra__ = collections.Counter
+
+    def __new__(cls, *args, **kwds):
+        if _geqv(cls, Counter):
+            raise TypeError("Type Counter cannot be instantiated; "
+                            "use Counter() instead")
```
ilevkivskyi: I would use full name ``collections.Counter()`` in the second line, otherwise this error message reads unclear.
(The same applies to Python 3 version).

python2/test_typing.py
```diff
@@ -1228,6 +1228,9 @@ def test_list(self):
     def test_deque(self):
         self.assertIsSubclass(collections.deque, typing.Deque)
 
+    def test_counter(self):
+        self.assertIsSubclass(collections.Counter, typing.Counter)
+
```
ilevkivskyi: Please add a test checking that a subclass of counter _could_ be instantiated, as you do for ``ChainMap`` in ``test_chainmap_subclass``. (And also for Python 3)

ilevkivskyi:
Thank you!
I have added two more minor comments. I would wait for typeshed PR, then this one could be merged.

ilevkivskyi:
@gvanrossum Could you please take a look at this PR and https://github.com/python/typeshed/pull/878?

I think the PRs are ready, but I would like to have your approval.

ilevkivskyi:
@JelleZijlstra could you please update the PR allowing instantiation of ``Counter`` and ``ChainMap`` following #381? (Don't worry about merge conflict, I will take care if necessary)

JelleZijlstra:
Nice, thanks! Now we'll no longer need both `from typing import Counter` and `from collections import Counter` to make a typed Counter. I'll push the change.

JelleZijlstra:
Rebased the branch and changed it to allow instantiation.

src/test_typing.py
```diff
@@ -893,6 +893,12 @@ class MyDef(typing.DefaultDict[str, T]): ...
         self.assertIs(MyDef[int]().__class__, MyDef)
         self.assertIs(MyDef[int]().__orig_class__, MyDef[int])
 
+    @skipUnless(sys.version_info >= (3, 3), 'ChainMap was added in 3.3')
+    def test_chainmap_type_erasure_special(self):
+        class MyChain(typing.ChainMap[str, T]): ...
+        self.assertIs(MyChain[int]().__class__, MyChain)
+        self.assertIs(MyChain[int]().__orig_class__, MyChain[int])
```
ilevkivskyi: One last thing, you use ``assertIs`` here that relies on generic type cache. This is a bit dangerous if the test is run in refleak hunting mode (-R). I see two options here:
* change this to use ``assertEqual``
* add ``self.clear_caches()`` at beginning and edit the comment in the previous ``test_type_erasure_special`` to say something like ``# this test and the next one are only two that rely on type caching``

Sorry for not noticing this before.

ilevkivskyi: Note that ``assertIs`` in other places in your PR is OK since it does not depend on generic type caching.

JelleZijlstra: I copied this from similar tests for defaultdict above; should I change those too?

ilevkivskyi: No, that one is intentional. (note that it is already "protected" by ``self.clear_caches()``)

gvanrossum:
I am happy to see this go in. (I haven't done a detailed code review, but I trust the two of you.)

Once this goes in we need additional changes in typeshed, right?

ilevkivskyi:
> Once this goes in we need additional changes in typeshed, right?

Yes, @JelleZijlstra would you prefer to make a PR for typeshed?

JelleZijlstra:
Yes, I can do that later today or tomorrow. Thanks for merging it in!

PR #368, opened 2017-01-26
Make get_type_hints not throw an error on builtin methods

wheerd:
While builtin functions are supported, calling get_type_hints on a builtin method like `object.__init__` causes an error. I change this to return an empty dictionary instead.

### DIFF
--- a/src/typing.py
+++ b/src/typing.py
@@ -1461,8 +1461,8 @@ def get_type_hints(obj, globalns=None, localns=None):
             isinstance(obj, types.BuiltinFunctionType) or
             isinstance(obj, types.MethodType) or
             isinstance(obj, types.ModuleType) or
-			isinstance(obj, _wrapper_descriptor_type) or
-			isinstance(obj, _method_wrapper_type)
+            isinstance(obj, _wrapper_descriptor_type) or
+            isinstance(obj, _method_wrapper_type)
         ):
             return {}
         else:
--- a/src/typing.py
+++ b/src/typing.py
@@ -10,6 +10,12 @@ try:
     import collections.abc as collections_abc
 except ImportError:
     import collections as collections_abc  # Fallback for PY3.2.
+try:
+    from types import SlotWrapperType, MethodWrapperType, MethodDescriptorType
+except ImportError:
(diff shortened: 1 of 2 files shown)


ilevkivskyi:
Thank you for the PR!
I see that you also have opened an issue on b.p.o. http://bugs.python.org/issue29377 to add the types you need here to types module. I would prefer that these types are first added to ``types``, and then you update this PR to use ``try: from types import ... except ImportError: WrapperDescriptorType = ...``

ilevkivskyi:
@wheerd After you take a look at my latest review on b.p.o., could you please update this PR using the ``try: from types import ...`` idiom I mentioned above and fix the lint issue?

src/typing.py
```diff
@@ -1456,7 +1460,9 @@ def get_type_hints(obj, globalns=None, localns=None):
             isinstance(obj, types.FunctionType) or
             isinstance(obj, types.BuiltinFunctionType) or
             isinstance(obj, types.MethodType) or
-            isinstance(obj, types.ModuleType)
+            isinstance(obj, types.ModuleType) or
+            isinstance(obj, _wrapper_descriptor_type) or
+            isinstance(obj, _method_wrapper_type)
```
methane: I don't like this `isinstance` chain.
How about this?

```
# global
_annotatable_types = (
    types.FunctionType,
    types.BuiltinFunctionType,
    types.MethodType,
    types.ModuleType,
    _wrapper_descriptor_type,
    _method_wrapper_type)

...

        if isinstance(obj, _annotatable_types):
```

ilevkivskyi: Thank, you! I have already noticed this some time ago, I would prefer to fix this separately after this PR is merged.

ilevkivskyi:
@wheerd Could you please update the PR using the ``try: from types import ...`` idiom I mentioned in my first comment and fix the lint issue? Otherwise we could not move forward with this PR (I can't do this myself, since I can't merge a failing PR).

wheerd:
Sorry I completely forgot about the pull request. Should be fixed now.

ilevkivskyi:
Thanks!

PR #370, opened 2017-01-26
Allow overriding some methods in NamedTuple

### LINKED ISSUE #369: Allow even more flexible NamedTuple
Recently, user defined methods in ``NamedTuple`` have been allowed. @elazarg asked for more flexibility, namely to allow overriding methods defined by ``NamedTuple`` such as ``__repr__``, ``__add__``, etc. After some thinking it looks a reasonable request. The only attributes that we should protect are ``_fields``, ``_filed_types``, ``_replace``, ``_make``, ``__new__``, etc.

PR will follow soon.

ilevkivskyi:
Fixes #369 

This allows to customize ``__repr__`` and other methods using ``NamedTuple`` class syntax.
Basically I just change the logic to "everything is allowed except if explicitly prohibited".

### DIFF
--- a/src/typing.py
+++ b/src/typing.py
@@ -1982,6 +1982,7 @@ _prohibited = ('__new__', '__init__', '__slots__', '__getnewargs__',
 
 _special = ('__module__', '__name__', '__qualname__', '__annotations__')
 
+
 class NamedTupleMeta(type):
 
     def __new__(cls, typename, bases, ns):
--- a/src/typing.py
+++ b/src/typing.py
@@ -2009,10 +2009,10 @@ class NamedTupleMeta(type):
         nm_tpl._field_defaults = defaults_dict
         # update from user namespace without overriding special namedtuple attributes
         for key in ns:
-            if key not in _prohibited + _special + nm_tpl._fields:
-                setattr(nm_tpl, key, ns[key])
-            elif key in _prohibited:
+            if key in _prohibited:
                 raise AttributeError("Cannot overwrite NamedTuple attribute " + key)
+            elif key not in _special and key not in nm_tpl._fields:
+                setattr(nm_tpl, key, ns[key])
         return nm_tpl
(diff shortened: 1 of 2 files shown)


elazarg:
(There's a small typo in the title - "NamededTuple" - which will make it harder to find later)
Thanks for working on this!

ilevkivskyi:
@elazarg Thank you! Fixed.

ilevkivskyi:
@gvanrossum Could you please also take a look at this? (The PR is very small)

src/typing.py
```diff
@@ -2002,12 +2009,9 @@ def __new__(cls, typename, bases, ns):
...
         # update from user namespace without overriding special namedtuple attributes
         for key in ns:
-            if not hasattr(nm_tpl, key):
+            if key not in _prohibited + _special + nm_tpl._fields:
                 setattr(nm_tpl, key, ns[key])
-            elif (
-                key not in ['__module__', '__qualname__', '__annotations__'] and
-                key not in nm_tpl._field_defaults
-            ):
+            elif key in _prohibited:
```
gvanrossum: If you check for this first, the check on line 2012 can be simpler.

src/typing.py
```diff
@@ -2002,12 +2009,9 @@ def __new__(cls, typename, bases, ns):
         nm_tpl._field_defaults = defaults_dict
         # update from user namespace without overriding special namedtuple attributes
         for key in ns:
-            if not hasattr(nm_tpl, key):
+            if key not in _prohibited + _special + nm_tpl._fields:
```
gvanrossum: The tuple concatenation here sounds like it could be slow -- since the 'in' operator still does a linear search, you might as well do separate checks, e.g. `key not in _special and key not in nm_tmp._fields`.

ilevkivskyi:
Thanks for review!

PR #371, opened 2017-01-26
Fix hash of GenericMeta classes

### LINKED ISSUE #351: CPython 3.6 / typing 3.5.3.0 messes up custom generics with same name from different places
File module_2.py:
```
from typing import TypeVar, Generic

T_2 = TypeVar('T_2')
class A(Generic[T_2]):
	pass
```

File module_1.py:
```
from typing import TypeVar, Generic, Tuple
import module_2

T_1 = TypeVar('T_1')
class A(Generic[T_1]):
	pass

class B(Generic[T_1]):

	class A(Generic[T_1]):
		pass
	
print (A)
print (B.A)
print (module_2.A)
print ('')
print (A == B.A)
print (A == module_2.A)
print (B.A == module_2.A)
print ('')
print (A[str] == B.A[str])
print (A[str] == module_2.A[str])
print (B.A[str] == module_2.A[str])
print ('')
print (A,   Tuple[A])
print (B.A, Tuple[B.A])
print (module_2.A, Tuple[module_2.A])
print ('')
print (A[str],   Tuple[A[str]])
print (B.A[str], Tuple[B.A[str]])
print (module_2.A[str], Tuple[module_2.A[str]])
```

Output of CPython 3.6 when running module_1.py:
```
__main__.A
__main__.B.A
module_2.A

False
False
False

True
True
True

__main__.A typing.Tuple[__main__.A]
__main__.B.A typing.Tuple[__main__.B.A]
module_2.A typing.Tuple[module_2.A]

__main__.A[str] typing.Tuple[__main__.A[str]]
__main__.B.A[str] typing.Tuple[__main__.A[str]]
module_2.A[str] typing.Tuple[__main__.A[str]]
```
Note the false-positive comparisons and the unintentionally altered `Tuple.__args__` in the cases with filled-in generic value. Same issue with CPython 2.7 if running the Python 2-variant of typing-3.5.3.0.

On the other hand this example works well with CPython 3.5 and CPython 2.7 using typing-3.5.2.2.

An ugly workaround for the Tuple-issue is to manually set `__args__` again, e.g.:
```
tpl = Tuple[B.A[str]]
tpl.__args__ = (B.A[str],)
```
I suspect, an equivalent issue might arise with Union, List, and other typing-constructs, but only tested Tuple so far.
Tested on Linux Mint 18.

ilevkivskyi:
Fixes #351 

It looks like there is a very simple fix (defer hashing of bare generics to ``super()``).
Nevertheless, I add extensive tests to avoid this difficult to catch bug in the future.

Note that I added a new module, since it is necessary to check that classes with same name defined in different modules compare unequal (we cannot use ``ann_module`` for this, since it is better to have test for all versions including 2.7).

### DIFF
--- a/python2/mod.py
+++ b/python2/mod.py
@@ -2,8 +2,10 @@ from typing import TypeVar, Generic
 
 T = TypeVar('T')
 
+
 class A(Generic[T]):
     pass
+
 
 class B(Generic[T]):
     class A(Generic[T]):
--- a/python2/mod.py
+++ b/python2/mod.py
@@ -1,0 +1,10 @@
+from typing import TypeVar, Generic
+
+T = TypeVar('T')
+
+class A(Generic[T]):
+    pass
+
+class B(Generic[T]):
+    class A(Generic[T]):
+        pass
--- a/python2/test_typing.py
+++ b/python2/test_typing.py
@@ -707,6 +707,42 @@ class GenericTests(BaseTestCase):
         self.assertEqual(Union[T, int][GenericMeta], Union[GenericMeta, int])
         self.assertEqual(Callable[..., GenericMeta].__args__, (Ellipsis, GenericMeta))
 
+    def test_generic_hashes(self):
+        import mod
+        class A(Generic[T]):
+            pass
+
+        class B(Generic[T]):
+            class A(Generic[T]):
+                pass
+
+        self.assertEqual(A, A)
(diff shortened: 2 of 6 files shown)


ilevkivskyi:
Since this is a minor fix, and OP is happy with it I am going to merge it myself now.

PR #373, opened 2017-01-27
Add some tests based on coverage report: Part 1

ilevkivskyi:
I have run ``coverage`` on ``typing`` and have noticed that some statements are missing. Most of them are trivial, but some are interesting to test.


ilevkivskyi:
This just adds tests, so that I think it is safe to merge now.

PR #375, opened 2017-01-28
Add some tests based on coverage report: Part 2

ilevkivskyi:
This is a continuation of #373 

Here, however, a (very small) change to ``typing.py`` was necessary. The two PRs boost coverage to virtually 100% (if one does not count ``pass``es in abstract methods and other trivial things).

### DIFF
--- a/python2/typing.py
+++ b/python2/typing.py
@@ -225,6 +225,15 @@ class _ForwardRef(_TypingBase):
             self.__forward_evaluated__ = True
         return self.__forward_value__
 
+    def __eq__(self, other):
+        if not isinstance(other, _ForwardRef):
+            return NotImplemented
+        return (self.__forward_arg__ == other.__forward_arg__ and
+                self.__forward_value__ == other.__forward_value__)
+
+    def __hash__(self):
+        return hash((self.__forward_arg__, self.__forward_value__))
+
     def __instancecheck__(self, obj):
         raise TypeError("Forward references cannot be used with isinstance().")
 
@@ -280,6 +289,14 @@ class _TypeAlias(_TypingBase):
             raise TypeError("%s cannot be re-parameterized." % self)
         return self.__class__(self.name, parameter,
                               self.impl_type, self.type_checker)
+
+    def __eq__(self, other):
+        if not isinstance(other, _TypeAlias):
+            return NotImplemented
+        return self.name == other.name and self.type_var == other.type_var
+
+    def __hash__(self):
+        return hash((self.name, self.type_var))
 
     def __instancecheck__(self, obj):
         if not isinstance(self.type_var, TypeVar):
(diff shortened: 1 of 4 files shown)


python2/typing.py
```diff
@@ -281,6 +290,14 @@ def __getitem__(self, parameter):
         return self.__class__(self.name, parameter,
                               self.impl_type, self.type_checker)
 
+    def __eq__(self, other):
+        if not isinstance(other, _TypeAlias):
+            return NotImplemented
+        return self.name == other.name and self.type_var == other.type_var
```
JelleZijlstra: Why doesn't this check `impl_type` and `type_checker`?

ilevkivskyi: This is an internal class, we know that if ``name`` is the same, then ``impl_type`` and ``type_checker`` must also be the same.

ilevkivskyi:
OK, It looks like the simplified version depends on changes in #383.
I will leave it like this until #383 is (hopefully) merged.

@gvanrossum Do you mind if we make ``GenericMeta`` code a bit more complex for 4x memory and 2x speed optimization? (this came up on python-dev some time ago)

gvanrossum:
I don't mind in principle, but I still haven't found the energy to review your code. I hope you can find someone else on python-dev to help you review your types changes if Inada doesn't feel comfortable.

ilevkivskyi:
@gvanrossum 
I have merged discussed PRs, but there is one thing that needs to be fixed before merging this into CPython: I added a test module ``mod.py`` some time ago (it was required to verify that cross-module type caching works right, a bug was reported). I will fix import for this module, and then it will be necessary to copy it to ``/Lib/test`` (where ``ann_module.py`` etc live). I will let you know, when everything is ready (probably later today).

gvanrossum:
OK. Since the move to git is currently happening I can't merge anything
into CPython anyway right now, so there's no particular hurry. I wonder if
git can cherry-pick across repos, that would be amazing. Otherwise I guess
it will have to be a series of PRs to the new cpython repo (see
https://paper.dropbox.com/doc/CPython-workflow-changes-mx1k8G6M0rg5JLy80F1r6
).

gvanrossum:
One more thing, "mod.py" is a rather non-descript name. Can you come up with something better?

ilevkivskyi:
> Can you come up with something better?

Yes, exactly, this is among the things I am going to do today.

ilevkivskyi:
@gvanrossum I fixed the remaining issue. As I understand the current workflow, a PR should me made to master, and after it is merged, the corresponding squash-commit should be cherry-picked to other branches (3.5 and 3.6 in this case).
I have made a PR to CPython master https://github.com/python/cpython/pull/28

By the way git can in principle cherry-pick between repos, but I tried this and it feels too complicated, because ``typing`` and ``cpython`` have quite different directory structure (it is quite easy however to cherry-pick a single commit between repos).

gvanrossum:
Thanks! Godspeed.

PR #377, opened 2017-01-28
Typing inspect

ilevkivskyi:
As discussed in http://bugs.python.org/issue29262 here is a PR to add runtime inspection tools for typing types.

**EDIT:** It is no more WIP

### DIFF
--- a/typing_inspect/test_typing_inspect.py
+++ b/typing_inspect/test_typing_inspect.py
@@ -6,7 +6,7 @@ from typing_inspect import (
 from unittest import TestCase, main
 from typing import (
     Union, ClassVar, Callable, Optional, TypeVar, Sequence, Mapping,
-    MutableMapping, Iterable, Generic, List, Any, Dict, Tuple,
+    MutableMapping, Iterable, Generic, List, Any, Dict, Tuple, NamedTuple,
 )
 
 
@@ -37,7 +37,7 @@ class IsUtilityTestCase(TestCase):
 
     def test_tuple(self):
         samples = [Tuple, Tuple[str, int], Tuple[Iterable, ...]]
-        nonsamples = [int, tuple, 42, List[int]]
+        nonsamples = [int, tuple, 42, List[int], NamedTuple('N', [('x', int)])]
         self.sample_test(is_tuple_type, samples, nonsamples)
         class MyClass(Tuple[str, int]):
             pass
@@ -94,6 +94,8 @@ class GetUtilityTestCase(TestCase):
         self.assertEqual(get_last_args(ClassVar[int]), (int,))
         self.assertEqual(get_last_args(Union[T, int]), (T, int))
         self.assertEqual(get_last_args(Iterable[Tuple[T, S]][int, T]), (int, T))
+        self.assertEqual(get_last_args(Callable[[T, S], int]), (T, S, int))
+        self.assertEqual(get_last_args(Callable[[], int]), (int,))
 
     def test_args(self):
         T = TypeVar('T')
@@ -102,8 +104,11 @@ class GetUtilityTestCase(TestCase):
         self.assertEqual(get_args(Union[int, Union[T, int], str][int]),
                          (int, str))
         self.assertEqual(get_args(int), ())
+        self.assertEqual(get_args(Union[int, Tuple[T, int]][str], evaluate=True),
+                         (int, Tuple[str, int]))
         self.assertEqual(get_args(Dict[int, Tuple[T, T]][Optional[int]], evaluate=True),
                          (int, Tuple[Optional[int], Optional[int]]))
+        self.assertEqual(get_args(Callable[[], T][int], evaluate=True), ([], int,))
 
     def test_generic_type(self):
         T = TypeVar('T')
--- a/typing_inspect/test_typing_inspect.py
+++ b/typing_inspect/test_typing_inspect.py
@@ -17,4 +17,3 @@ class IsUtilityTestCase(TestCase):
 
 if __name__ == '__main__':
     main()
-
--- a/typing_inspect/test_typing_inspect.py
+++ b/typing_inspect/test_typing_inspect.py
@@ -1,5 +1,9 @@
 from typing_inspect import is_generic_type
 from unittest import TestCase, main
+from typing import (
+    Union, ClassVar, Callable, Optional, TypeVar, Sequence, Mapping,
+    MutableMapping, Iterable, Generic, List
+)
 
 
 class IsUtilityTestCase(TestCase):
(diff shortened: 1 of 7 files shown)


typing_inspect/typing_inspect.py
```diff
@@ -0,0 +1,198 @@
...
+
+
+def get_args(tp, evaluate=False):
+    """Get type arguments with all substitutions performed. For unions,
+    basic simplifications used by Union constructor are performed.
+    If `evaluate` is False (default), report result as nested tuple, this matches
+    the internal representation of types. If `evaluate` is True, then all
+    type parameters are applied (this could be expensive). Examples::
+
+        get_args(Union[int, Tuple[T, int]][str]) == (int, (Tuple, str, int))
```
flying-sheep: Why not

```py
get_args(Union[int, Tuple[T, int]][str]) == (int, Tuple[str, int])
```

Your transformation is a bit suprising if I’m concerned.

typing_inspect/typing_inspect.py
```diff
@@ -0,0 +1,198 @@
+"""Defines experimental API for runtime inspection of types defined
+in the standard "typing" module.
+
+Example usage::
+    from typing_inspect import is_generic
+"""
+
+# NOTE: This module must support Python 2.7 in addition to Python 3.x
```
flying-sheep: Is this a requirement for all `typing` modules?

And regardless: Why?

ilevkivskyi: ``typing`` on PyPI supports 2.7 (which is supported until 2020).

typing_inspect/typing_inspect.py
ilevkivskyi: This is not "transformation", this is - quoting the docstring - "internal representation". The reason is also mentioned there: creating new type objects is space and time consuming (even with cashing). If you want everything be actually substituted, just use ``evaluate=True``.

typing_inspect/typing_inspect.py
flying-sheep: ah! didn’t know it was on PyPI

typing_inspect/typing_inspect.py
flying-sheep: OK, maybe adding an example would be useful here

ilevkivskyi: There is an example with ``evaluate=True`` 4 lines below, or do you mean _one more_ example?

flying-sheep: must have missed it, sorry! now i look like a doofus 😉

ilevkivskyi: No problem, anyway, I decided to add one more example to the docstring.

typing_inspect/typing_inspect.py
```diff
@@ -0,0 +1,248 @@
+"""Defines experimental API for runtime inspection of types defined
+in the standard "typing" module.
+
+Example usage::
+    from typing_inspect import is_generic
```
JelleZijlstra: `is_generic_type`

typing_inspect/typing_inspect.py
```diff
@@ -0,0 +1,248 @@
...
+
+
+def get_last_args(tp):
+    """Get last arguments of (multiply) subscripted type. Examples::
+
+        get_last_args(int) == ()
+        get_last_args(Union) == ()
+        get_last_args(ClassVar[int]) == (int,)
+        get_last_args(Union[T, int]) == (T, int)
+        get_last_args(Iterable[Tuple[T, S]][int, T]) == (int, T)
```
JelleZijlstra: Could you add an example here for Callable? It's not obvious to me what this will return for something like `Callable[[int], str]`.

typing_inspect/typing_inspect.py
```diff
@@ -0,0 +1,248 @@
...
+    """Get type arguments with all substitutions performed. For unions,
+    basic simplifications used by Union constructor are performed.
+    If `evaluate` is False (default), report result as nested tuple, this matches
+    the internal representation of types. If `evaluate` is True, then all
+    type parameters are applied (this could be time and memory expensive).
+    Examples::
+
+        get_args(int) == ()
+        get_args(Union[int, Union[T, int], str][int]) == (int, str)
+        get_args(Union[int, Tuple[T, int]][str]) == (int, (Tuple, str, int))
```
JelleZijlstra: Callable examples would be helpful here too.

typing_inspect/typing_inspect.py
```diff
@@ -0,0 +1,248 @@
...
+        is_callable_type(Callable) == True
+        is_callable_type(Callable[..., int]) == True
+        is_callable_type(Callable[[int, int], Iterable[str]]) == True
+        class MyClass(Callable[[int], int]):
+            ...
+        is_callable_type(MyClass) == True
+
+    For more general tests use callable(), for more precise test use::
+
+        get_origin(tp) is Callable
```
JelleZijlstra: What is the difference exactly between this construct and `is_callable_type`? I think this could be explained more.

typing_inspect/test_typing_inspect.py
```diff
@@ -0,0 +1,124 @@
...
+        nonsamples = [int, type, 42, [], List[int],
+                      Union[callable, Callable[..., int]]]
+        self.sample_test(is_callable_type, samples, nonsamples)
+        class MyClass(Callable[[int], int]):
+            pass
+        self.assertTrue(is_callable_type(MyClass))
+
+    def test_tuple(self):
+        samples = [Tuple, Tuple[str, int], Tuple[Iterable, ...]]
+        nonsamples = [int, tuple, 42, List[int]]
```
JelleZijlstra: Maybe add some NamedTuples as test cases; I assume they'll be nonsamples because typing.Tuple and typing.NamedTuple are completely distinct, but it would be good to clarify anyway.

typing_inspect/typing_inspect.py
ilevkivskyi: the latter excludes subclasses, I added this to the docstring.

ilevkivskyi:
@JelleZijlstra Thank you for review! I addressed your comments (and a bug that I found) in a new commit.

JelleZijlstra:
What is blocking this PR? Having `typing_inspect` available would be useful to me.

ilevkivskyi:
@JelleZijlstra 
Actually nothing blocks this. I was just waiting for an approval from @gvanrossum. We discussed this  in http://bugs.python.org/issue29262 beginning of this year and the plan was following:
* Make few releases of ``typing_inspect`` as a separate package on PyPI
* Listen to feedback and tune API
* Eventually include this in ``typing``

@gvanrossum 
Is this still the plan? Would you like to make a review?

gvanrossum:
I'm sorry, I did not realize you were waiting for my review. I still don't have time to do this justice, so I recommend that you make this a separate distro and iterate that way. Sorry!

ilevkivskyi:
@gvanrossum @glyph @xgid @wsanchez
OK, as Guido proposed, I am closing this PR. The ``typing_inspect`` is now published as a separate package:
https://pypi.python.org/pypi/typing-inspect
https://github.com/ilevkivskyi/typing_inspect

Everyone is welcome to try it and comment on the API. Also please tell everyone else who might be interested, so that we can iterate fast. After the API is settled, some introspection functions may be added directly to ``typing``.

Issue #367, opened 2017-01-25
Allow instantiation of generic collections


ilevkivskyi:
Currently there are many places where we prohibit instantiation of generic versions of concrete collections, like
```python
    if _geqv(cls, List):
        raise TypeError("Type List cannot be instantiated; "
                                  "use list() instead")
```
This prohibits things like ``List[int]()``, ``Deque[int]()``, etc. As I remember the reason for this was that ``List[int]`` was slow. But now types are cached. Are there any other reasons to prohibit this? If not, then I propose to remove this limitation. Anyway things like ``Node[int]()`` are already allowed both in mypy and in typing.

gvanrossum:
Would `List[int]()` it return an instance of `list` or of `List[int]`? I'm against the latter, because even an empty subclass of `list` will incur a significant overhead compared to the plain `list`.

ilevkivskyi:
I also think it should be an instance of plain ``list`` (and plain ``collections.deque`` etc. for other classes).

gvanrossum:
Then let's draft a patch to PEP 484 first

markshannon:
I would prefer that the prohibition on instantiation remains.
Firstly a type is not a class. `List[int]` is not `list`, and type-hints should be kept distinct from the runtime values.
If I want a list of ints, then
`x: List[int] = []`
describes that perfectly.
If we allow
`x = List[int]()` 
are type checkers supposed to infer that l has type `List[int]`? 
PEP 484 says no, as there is no annotation, thus `x` is implicitly typed `Any`.

gvanrossum:
I didn't intend the PEP to say that x *must* be inferred as `Any` -- where do you read that? Certainly mypy make that inference when using a non-builtin class:
```python
from typing import Generic, TypeVar
T = TypeVar('T')
class Node(Generic[T]): 
    def __init__(self):
        self.left = None  # type: Node[T]
        self.right = None  # type: Node[T]
def foo() -> None:
    root = Node[int]()
    reveal_type(root)  # Shows Node[int]
```

markshannon:
You're right, it is reasonable to infer from `x = []` that `x` is a list.

I still don't like instantiating types though, mainly from wanting to keep classes and types separate.
Also, will people expect that
```python
x: List[int] = []
x.append("Hi")
```
raises a `TypeError`? 
Saying no seems to me violate the principle of least surprise, but saying yes has significant performance implications.

ilevkivskyi:
As I understand, people were asking to allow:
[...]

PR #381, opened 2017-01-30
Allow istantiation of Deque and DefaultDict

ilevkivskyi:
As discussed in #367 ``Deque`` and ``DefaultDict`` will follow the rules of user-defined classes.
Instantiating them will return instances of plain ``collections.deque`` and ``collections.defaultdict``.


ilevkivskyi:
Assuming that https://github.com/python/peps/pull/193 will be merged soon, I am merging this now.

PR #382, opened 2017-01-31
fix running tests directly in py2

JelleZijlstra:
Without this change, running `python2.7 test_typing.py` fails because A's __module__ becomes `__main__`. The current code works under py.test because the way it runs the tests the __module__ is `test_typing`.

### DIFF
--- a/python2/test_typing.py
+++ b/python2/test_typing.py
@@ -710,7 +710,7 @@ class GenericTests(BaseTestCase):
     def test_generic_hashes(self):
         import mod
         class A(Generic[T]):
-            __module__ = 'typing'
+            __module__ = 'test_typing'
 
         class B(Generic[T]):
             class A(Generic[T]):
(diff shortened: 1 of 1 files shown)


JelleZijlstra:
This is the failure I got:

```
$ python2.7 test_typing.py 
................................................................................F.............................................................................
======================================================================
FAIL: test_generic_hashes (__main__.GenericTests)
  File "test_typing.py", line 743, in test_generic_hashes
    self.assertTrue(repr(Tuple[A[str]]).endswith('typing.A[str]]'))
AssertionError: False is not true

Ran 158 tests in 0.179s

FAILED (failures=1)
```

ilevkivskyi:
Good catch! I would change the module name to ``test_typing`` (to reflect the actual situation) and update this line https://github.com/python/typing/pull/382/files#diff-de47c009e521dab38fd335991a44d3feR743 below accordingly.

PR #386, opened 2017-02-03
Prepend bases on first subscription

### LINKED ISSUE #384: singledispatch and typing.Tuple[x] interaction
Hello,

Python 3.5, but I'm using ``typing.py`` from this repo, master branch. (Trying to port a library.)

I realize this is not really the intended use of typing.

I'm not really sure if this is an issue in ``typing`` or ``singledispatch``, but it used to work before. Here's a small test case. Note that `typing.Tuple[int]` is a subclass of `typing.Tuple`, and `typing.Tuple` is a subclass of `typing.Sequence`.

```
import typing
from functools import singledispatch

s = singledispatch(lambda: 'default')
s.register(typing.Sequence, lambda: 'sequence')
s.register(typing.Tuple, lambda: 'tuple')

print(s.dispatch(typing.Tuple[int])())
```

If you run it a number of times, the results are either 'tuple' or 'sequence', based on the ordering of an internal dict in singledispatch. This is a bug in singledispatch, right?

### LINKED ISSUE #385: issubclass(Node[int], Node) returns False.
I think it would be right to return ``True`` in such cases. I have a very simple fix, that also fixes #384, will make a PR tomorrow morning.

ilevkivskyi:
Fixes #385 
Fixes #384 

In previous versions of typing, ``__bases__`` on subscription where prepend with ``__origin__``, so that MRO of, e.g., ``Mapping[int, T][str]`` was:
```python
[Mapping[int, str], Mapping[int, T], Mapping, collections.abc.Mapping, ...]
```
This was removed for several reasons. However, now it looks like complete removal of such prepending was too radical, since now e.g. ``issublcass(Node[int], Node)`` returns ``False``. I believe such runtime check should be equivalent to ``issublcass(Node, Node)``, i.e., ``True``.

Here I propose to fix this by only prepending ``__bases__`` on *first* subscription, so that, e.g, ``Mapping[int, T][str].__mro__`` will be:
```python
[Mapping[int, str], Mapping, collections.abc.Mapping, ...]
```

This PR also fixes an ambiguity when using ``typing`` ABCs with ``singledispatch()``.

### DIFF
--- a/python2/test_typing.py
+++ b/python2/test_typing.py
@@ -521,6 +521,7 @@ class GenericTests(BaseTestCase):
         Y[unicode]
         with self.assertRaises(TypeError):
             Y[unicode, unicode]
+        self.assertIsSubclass(SimpleMapping[str, int], SimpleMapping)
 
     def test_generic_errors(self):
         T = TypeVar('T')
--- a/python2/typing.py
+++ b/python2/typing.py
@@ -1169,8 +1169,11 @@ class GenericMeta(TypingMeta, abc.ABCMeta):
             _check_generic(self, params)
             tvars = _type_vars(params)
             args = params
+
+        gorg = _gorg(self)
+        prepend = (gorg,) if gorg not in self.__bases__ else ()
         return self.__class__(self.__name__,
-                              self.__bases__,
+                              prepend + self.__bases__,
                               dict(self.__dict__),
                               tvars=tvars,
                               args=args,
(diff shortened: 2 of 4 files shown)


ilevkivskyi:
This will also allow to simplify #375 to only adding  tests. So that I would like to merge this now.
@gvanrossum I am mentioning you just in case, so that this will not pass unnoticed.

PR #393, opened 2017-02-23
Treat subscripted generics as "proxies" for original class.

### LINKED ISSUE #392: Subscripted generic classes should not have independent class variables
Currently we have
```python
class A(Generic[T]):
    x = None

A[int].x = 1
A[str].x = 'a'

print(A[int].x) # prints 1
print(A[int]().x) # prints None, since type is erased on instantiation
```

@JukkaL argues and I agree with him, that this looks wrong. Conceptually, there should be only one runtime class. I think we can implement this without loosing the current runtime generic information by tweaking descriptor ``__dict__`` of ``GenericMeta``, so that all assignments and look-ups on subscripted class objects like ``A[int]`` will be transferred to the original class object ``A``.

@gvanrossum What do you think?

ilevkivskyi:
Fixes #392 

As noted by @JukkaL, conceptually, subscripting generic classes should not create independent class objects (type should be erased at runtime). In particular, class variables should be shared between subscripted generics and original class.

Here I implement this semantics using a custom ``__setattr__``.

@gvanrossum There is a question (assuming we agree that this is needed) whether this should go in 3.6.1 that has "feature freeze" in less than a week, or should we postpone this until 3.6.2?

### DIFF
--- a/python2/test_typing.py
+++ b/python2/test_typing.py
@@ -642,6 +642,33 @@ class GenericTests(BaseTestCase):
         c.bar = 'abc'
         self.assertEqual(c.__dict__, {'bar': 'abc'})
 
+    def test_subscripted_generics_as_proxies(self):
+        T = TypeVar('T')
+        class C(Generic[T]):
+            x = 'def'
+        self.assertEqual(C[int].x, 'def')
+        self.assertEqual(C[C[int]].x, 'def')
+        C[C[int]].x = 'changed'
+        self.assertEqual(C.x, 'changed')
+        self.assertEqual(C[str].x, 'changed')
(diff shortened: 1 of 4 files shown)


gvanrossum:
I think we should get this in.

--Guido (mobile)

ilevkivskyi:
> I think we should get this in

OK, I am going to merge this soon (unless you or @JukkaL express any concerns), and will then make a PR to ``python/cpython`` repo.

ilevkivskyi:
OK, so here is the PR to ``python/cpython`` https://github.com/python/cpython/pull/265

PR #397, opened 2017-03-04
Add NoReturn type

### LINKED ISSUE #165: A `NoReturn` type is needed
In order to perform branch analysis, it is necessary to know which calls will never return normally. Examples are `sys.exit` (which always returns via exception) and `os.exit` (which never returns).

`NoReturn` should drop out of `Union` and dominate `Intersection`.

To manually mark a branch as dead, use `cast(NoReturn, expression)` in an expression context, or `raise` in statement context.

This can also be used as the return type of `__init__` or `__new__` or the metaclass's `__call__` to indicate types that can't be normally instantiated.

ilevkivskyi:
Fixes #165 

Note that here I prohibit ``List[NoReturn]``, but maybe we should allow this in order to allow ``Union[int, NoReturn]`` and/or ``Callable[..., NoReturn]`` (or make exceptions for the latter two).

As discussed in #165, this PR should wait until Python 3.6.1 is released.

### DIFF
--- a/python2/test_typing.py
+++ b/python2/test_typing.py
@@ -125,15 +125,9 @@ class NoReturnTests(BaseTestCase):
     def test_repr(self):
         self.assertEqual(repr(NoReturn), 'typing.NoReturn')
 
-    def test_errors(self):
+    def test_not_generic(self):
         with self.assertRaises(TypeError):
             NoReturn[int]
-        with self.assertRaises(TypeError):
-            List[NoReturn]
-        with self.assertRaises(TypeError):
-            Union[int, NoReturn]
-        with self.assertRaises(TypeError):
-            Callable[..., NoReturn]
 
     def test_cannot_subclass(self):
         with self.assertRaises(TypeError):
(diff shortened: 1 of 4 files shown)


JelleZijlstra:
What would the use of `Union[int, NoReturn]` be? We could type functions that either return `int` or throw an exception (e.g., the `int` constructor) this way, but I'm not sure that would enable better type checking.

`Callable[..., NoReturn]` makes more logical sense to me, but it doesn't seem like something that's often useful in practice.

gvanrossum:
I think we've learned in the past that typing.py should be kept simple, stupid. A static checker can reject List[NoReturn].

ilevkivskyi:
@gvanrossum 
> I think we've learned in the past that typing.py should be kept simple, stupid.

Yes, right :-) Since this case is not obvious, we can let type checkers decide.

gvanrossum:
Code LGTM. I wonder though if we should get Mark Shannon's view of the PEP 484 changes first.

ilevkivskyi:
There is no hurry with this (this could not go into 3.6.1), let us first sort out the PEP, and then I could merge this.

gvanrossum:
Feel free to merge as soon as 3.6.1 is released.

PR #409, opened 2017-04-03
Add missing classes to __all__

ilevkivskyi:
``SupportsComplex`` and ``SupportsBytes`` were previously missing in ``__all__`` (only one of them in Python 2 version). This PR fixes this (I suppose there is no reason to exclude them from ``__all__``).


ilevkivskyi:
It looks like there are no objections, so I am merging this now.

Issue #415, opened 2017-04-25
Alternative constructors with primitive types or NewTypes


davidroeca:
Hi typing team,

I've noticed that typing.cast and NewType essentially provide the exact same functionality. This makes NewType merely a type alias for convenience.

I'd like NewType to provide the ability to either pass in a callable that serves as the constructor, or for the library to provide an alternative function called ValidatedType (or AlternativelyConstructedType or something like that), implemented in this way (or something similar):

```python
def ValidatedType(
        name,
        parent_type,
        validate_and_initialize  # this will be a Callable[[Any], parent_type]
):
    '''Follows typing NewType and runs the validation function'''
    def validated_type(x):
        return validate_and_initialize(x)
    validated_type.__name__ = name
    validated_type.__supertype__ = parent_type
    return validated_type
```

This may add some overhead, but it also adds more clarity to codebases and is worth considering.

See https://github.com/python/mypy/issues/3200 -- created this issue here since mypy might not be the right repo to direct this feature request to

JelleZijlstra:
I don't think `cast` is similar to `NewType`; you can't use a cast in a type annotation.

More broadly, we've generally avoided putting things in `typing` that affect runtime behavior. I'm not sure what you're asking for is common enough to be added to typing.

davidroeca:
I agree that it's not common, in part because it's not possible in the
current ecosystem. I also agree that it doesn't necessarily need to be
added to typing, but I'd like this to at least be possible.

In addition, the runtime overhead will be directly caused by the person who
defined the type constructor, not the typing library. This is essentially
what happens in a class constructor.

davidroeca:
Re: cast vs NewType:

Fair enough, NewType allows for type annotions, which is very valuable, and which is also why I believe that some variant would be useful. The casting of the type is essentially what happens when the inner/hidden new_type function (https://github.com/python/typing/blob/master/src/typing.py#L2178) is called, and that to me is the value in using NewType's constructor. 

To me, it seems as though this inner new_type function is an optimized special case of something more general. I may be wrong, but customization of this function will not likely add too much overhead to someone who opts out of such a customization and just uses the supported defaults.

ilevkivskyi:
> More broadly, we've generally avoided putting things in typing that affect runtime behavior. I'm not sure what you're asking for is common enough to be added to typing.

[...]

PR #416, opened 2017-04-26
Optional NewType Validation

davidroeca:
One proposed implementation of https://github.com/python/typing/issues/415

This should work as indicated by the docstring as long as this would make sense to mypy.

### DIFF
--- a/python2/test_typing.py
+++ b/python2/test_typing.py
@@ -1705,6 +1705,31 @@ class NewTypeTests(BaseTestCase):
         self.assertIsInstance(UserName('Joe'), type('Joe'))
         self.assertEqual(UserId(5) + 1, 6)
 
+    def test_construct(self):
+        def validate_user_id(user_id):
+            if user_id <= 0:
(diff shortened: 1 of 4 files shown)


the-knights-who-say-ni:

Unfortunately we couldn't find an account corresponding to your GitHub username on [bugs.python.org](http://bugs.python.org/) (b.p.o) to verify you have signed the CLA. This is necessary for legal reasons before we can look at your contribution. Please follow [the steps outlined in the CPython devguide](https://cpython-devguide.readthedocs.io/pullrequest.html#licensing) to rectify this issue.


gvanrossum:
I just have to say that this is not as easy as you think. I don't have time to elaborate, sorry.

davidroeca:
Agreed that this likely touches a number of first and third-party libraries, and that NewType should be limited in scope. Figured this could at least be a point of discussion, and finally people have responded to my issue so I believe this pull request has done its work.

PR #417, opened 2017-04-26
Protocols

### LINKED ISSUE #11: Protocols (a.k.a. structural subtyping)
In mypy's typing,py there's a neat feature called Protocol. Perhaps we should add this to the PEP. Or perhaps it's similar to a Python ABC?

ilevkivskyi:
Fixes #11 

This is a runtime implementation of protocols as per draft [PEP 544](https://www.python.org/dev/peps/pep-0544/). (The ``mypy`` counterpart is https://github.com/python/mypy/pull/3132)

The implementation is straightforward (mostly I just made current version to behave more similar to ``__subclasshook__`` in ``collections.abc`` and added support for ``__annotations__`` in Python 3.6).
Nevertheless, I add a lot of tests. Some of them may look trivial, but previous experience with generics shows that it is better to have them.

There is probably only one corner case worth noting. Consider this situation with for variable annotations:
```python
@runtime
class Proto(Protocol):
    attr: int
class OtherProto(Protocol):
    attr: int
class BadConcrete:
    attr: int
class GoodConcrete:
    attr = 1
```
Then I propose the following subclass check results:
```python
issubclass(GoodConcrete, Proto) == True
issubclass(BadConcrete, Proto) == False # This is a bit questionable, maybe return True?
issubclass(OtherProto, Proto) == True
```
The first one is probably obvious, the second one returns ``False`` to make this safe:
```python
if issubclass(BadConcrete, Proto):
    BadConcrete().attr
```
The last one although safe since one cannot instantiate protocols, and this is probably what one would expect.

Attn: @JukkaL @ambv @gvanrossum


Daenyth:
```
issubclass(BadConcrete, Proto) == False # This is a bit questionable, maybe return True?
```

This being the case means that NamedTuple classes would be unable to implement protocols, being unable to extend any parent classes other than NamedTuple

ilevkivskyi:
Something strange happened here (or maybe the comment was deleted), but here is an interesting note to make: because of how ``NamedTuple`` behaves at runtime (just making a ``collectios.namedtuple``) it implements corresponding protocols already at class level, for example:
```python
@runtime
class P(Protocol):
    x: int
class N(NamedTuple):
    x: int

assert issubclass(N, P)  # OK
```

Daenyth:
I left a comment that I realized was completely incorrect so I deleted it

ilevkivskyi:
I will soon make another PR that will add (simplified) ``Protocol`` to ``typing_extensions``. We cannot add
all the functionality while ``Protocol`` will be in ``typing_extensions`` since several classes (like ``SupportsInt`` etc.) subclass ``_Protocol`` in ``typing``, ideally they should subclass the new ``Protocol``, this is however quite minor thing.

I will keep this PR open (and update depending on the outcome of experiments with ``typing_extensions``).

gvanrossum:
So IIUC we should first merge #464 and defer this one until the PEP has been accepted, right?

ilevkivskyi:
> So IIUC we should first merge #464 and defer this one until the PEP has been accepted, right?

Yes, also if there will be something we discover while playing with ``typing_extensions``, I will update this PR accordingly (actually there are already few minor tweaks that I need to make here following the review of #464).

ilevkivskyi:
This was superseded by https://github.com/python/typing/pull/649

PR #422, opened 2017-05-03
add typing.ContextManager on all supported Python versions

### LINKED ISSUE #274: Proposal: ContextManager interface
Because context managers are a language feature (for use in `with` blocks), typing should expose an interface class for them like it does with Dict, List, etc.

The typeshed for contextlib defines an interface for a ContextManager because it needs it in the return type for `contextlib.contextmanager`, but it seems like the kind of thing that should be in typing, not in a random stub.

JelleZijlstra:
So that there is a way to type hint for a context manager even on
language versions before 3.6.

This is just copied from contextlib.AbstractContextManager in 3.6.

Fixes #274


src/typing.py
```diff
@@ -1949,7 +1949,25 @@ class ValuesView(MappingView[VT_co], extra=collections_abc.ValuesView):
...
+        @abc.abstractmethod
+        def __exit__(self, exc_type, exc_value, traceback):
+            """Raise any exception triggered within the runtime context."""
+            return None
+
+        @classmethod
+        def __subclasshook__(cls, C):
+            if cls is ContextManager:
+                if (any("__enter__" in B.__dict__ for B in C.__mro__) and
+                        any("__exit__" in B.__dict__ for B in C.__mro__)):
```
ilevkivskyi: I understand this is just copied from ``contextlib``, but I think both should be updated to support ``__enter__ = None`` pattern, see my other [comment](https://github.com/python/cpython/pull/1412/files#r114482765).

JelleZijlstra: I don't think that makes sense here, because this class will only be used in Python < 3.6, and the `= None` pattern became generally supported only in 3.6. The collections ABCs don't support anti-registration either before 3.6.

ilevkivskyi: > I don't think that makes sense here, because this class will only be used in Python < 3.6, and the ``= None`` pattern became generally supported only in 3.6. The collections ABCs don't support anti-registration either before 3.6.

OK, if @gvanrossum agrees with this, then I think this can be merged as is, I don't have any other comments.

python2/typing.py
```diff
@@ -1857,6 +1858,26 @@ class ValuesView(MappingView[VT_co]):
...
+    @abc.abstractmethod
+    def __exit__(self, exc_type, exc_value, traceback):
+        """Raise any exception triggered within the runtime context."""
+        return None
+
+    @classmethod
+    def __subclasshook__(cls, C):
+        if cls is ContextManager:
+            if (any("__enter__" in B.__dict__ for B in C.__mro__) and
+                    any("__exit__" in B.__dict__ for B in C.__mro__)):
```
gvanrossum: This is a cringe-worthy indentation (due to a stupid flake rule), since the two `any()` calls are actually siblings. I propose to stop contorting our code and instead disable that specific flake error. @ambv thoughts?

src/typing.py
gvanrossum: OK, but this is worth a comment in the code.

src/typing.py
```diff
@@ -1949,7 +1949,25 @@ class ValuesView(MappingView[VT_co], extra=collections_abc.ValuesView):
...
+else:
+    class ContextManager(Generic[T_co]):
+        __slots__ = ()
+
+        def __enter__(self):
+            return self
+
+        @abc.abstractmethod
+        def __exit__(self, exc_type, exc_value, traceback):
+            """Raise any exception triggered within the runtime context."""
```
gvanrossum: This docstring confused me. Maybe the code is better off without it?

JelleZijlstra:
PEP 484 already lists ContextManager in its laundry list of ABCs provided by typing, so I think no change is needed there.

python2/typing.py
JelleZijlstra: I'm pushing a commit to do that.

JelleZijlstra: Changing this required more changes in the flake8 configuration than I would have liked. To keep this PR focused, maybe we should keep the cringeworthy formatting here for now, then change the flake8 configuration in a separate PR?

JelleZijlstra:
Oh I didn't know that, let me try it.

gvanrossum:
OK, much better!

(Also would be nice to get rid of the separate .flake8-tests but that seems a ways away -- if I run flake8 on the test files with the config from tox.ini I get a ton of errors.)

Issue #434, opened 2017-05-22
Common test suite for Python type checkers


vlasovskikh:
It would be a good idea to have a common test suite for different Python type checkers like Mypy, PyCharm, PyType. By introducing it, we could ensure that our checkers have a common understanding of the type system.

The existing checkers detect various kinds of errors depending on the depth of code analysis and type inference. The idea is to use the common test suite just for testing basic scenarios every checker should understand according to PEP 484.

As a starting point, we could use a subset of [Mypy tests](https://github.com/python/mypy/tree/master/test-data/unit) and Mypy test format.

For example, a typical Mypy test looks like that:

```python
[case testTupleAssignmentWithTupleTypes]
from typing import Tuple
t1 = None # type: Tuple[A]
t2 = None # type: Tuple[B]
t3 = None # type: Tuple[A, A]
t4 = None # type: Tuple[A, B]
t5 = None # type: Tuple[B, A]

t1 = t2 # E: Incompatible types in assignment (expression has type "Tuple[B]", variable has type "Tuple[A]")
t1 = t3 # E: Incompatible types in assignment (expression has type "Tuple[A, A]", variable has type "Tuple[A]")
t3 = t1 # E: Incompatible types in assignment (expression has type "Tuple[A]", variable has type "Tuple[A, A]")
t3 = t4 # E: Incompatible types in assignment (expression has type "Tuple[A, B]", variable has type "Tuple[A, A]")
t3 = t5 # E: Incompatible types in assignment (expression has type "Tuple[B, A]", variable has type "Tuple[A, A]")

# Ok
t1 = t1
t2 = t2
t3 = t3
t4 = t4
t5 = t5

class A: pass
class B: pass
[builtins fixtures/tuple.pyi]
```

There are several directives in square brackets like `[case ...]` for defining a test case, `[builtins ...]` for including minimalistic stubs for built-ins, and some others. The expected error messages are written using the `# E: ...` comments.

I would like to introduce these tests in the typing repo and maybe setup Travis CI for running them.

JukkaL:
We wouldn't need to support `[builtins ...]`, which specifies a fairly minimal test-only stub to use for a particular mypy test case. These tests could probably use the real  stubs for builtins and other modules, since these would be more like acceptance tests rather than unit tests.

vlasovskikh:
@JukkaL Do you suggest to use just regular Typeshed in the tests?

JukkaL:
[...]

PR #436, opened 2017-05-24
Prototype of common test suite format and runner for Mypy

vlasovskikh:
Here is a prototype for the common test suite suggested in #434 that uses YAML as the test format. I've created a simple test runner for Mypy and added a couple of tests.

### DIFF
--- a/test-data/check-modules.yml
+++ b/test-data/check-modules.yml
@@ -1,0 +1,17 @@
+# Type checker test cases dealing with modules and imports.
+
+- case: Access imported definitions
+  files:
+    - content: |
+        import m
+        import typing
+        m.f()           # E: Too few arguments for "f"
+        m.f(object())   # E: Argument 1 to "f" has incompatible type "object"; expected "A"
+        m.x = object()  # E: Incompatible types in assignment (expression has type "object", variable has type "A")
+        m.f(m.A())
+        m.x = m.A()
+    - path: m.py
+      content: |
+        class A: pass
+        def f(a: A) -> None: pass
+        x = A()
--- a/tests/mypy_test.py
+++ b/tests/mypy_test.py
@@ -1,0 +1,89 @@
+import glob
+import os
+import re
+from subprocess import Popen, PIPE
+from typing import Text, List, Iterable, Tuple, Any
+
+import pytest
+import yaml
+
+
+DEFAULT_PATH = 'main.py'
+ENCODING = 'utf-8'
+
+
+def load_test_cases(root: Text) -> Iterable[Tuple[Text, Text, List[dict]]]:
+    paths = glob.glob(os.path.join(root, '*.yml'))
+    assert paths, 'No test data found in %s' % os.path.abspath(root)
+
+    for path in paths:
+        with open(path, 'rb') as fd:
+            raw_data = yaml.load(fd)
+        for case in raw_data:
+            _, filename = os.path.split(path)
+            name = case.get('case')
+            assert name, 'Unnamed case: %r' % case
+            files = case.get('files', [])
+            unique = {file.get('path', DEFAULT_PATH) for file in files}
+            assert len(files) == len(unique), 'Duplicate files in %s' % name
+            yield filename, name, files
+
+
+def inline_errors(errors: List[Text], files: List[dict]) -> List[dict]:
+    contents = {
+        file.get('path', DEFAULT_PATH): strip_errors(file.get('content', ''))
+        for file in files}
+
+    for error in errors:
+        index, message, path = parse_error_line(error)
+        content = contents.get(path)
+        assert content, 'Cannot find contents of %s' % path
+        lines = content.splitlines(keepends=True)
+        assert 0 <= index < len(lines)
+        line = lines[index]
+        updated = '%s# E: %s\n' % (line.rstrip('\n'), message)
+        lines[index] = updated
+        contents[path] = ''.join(lines)
+
+    results = [file.copy() for file in files]
+    for result in results:
+        result['content'] = contents.get(result.get('path', DEFAULT_PATH), '')
+    return results
+
+
+def strip_errors(s: Text) -> Text:
+    return re.sub(r'#\s*E:.*', r'', s)
+
+
+def parse_error_line(line: Text) -> Tuple[int, Text, Text]:
+    m = re.match(r'(.+):([0-9]+): (error): (.+)', line)
+    assert m, 'Cannot parse output line: %s' % line
+    path = m.group(1)
+    index = int(m.group(2)) - 1
+    message = m.group(4)
+    return index, message, path
+
+
+@pytest.mark.parametrize('filename,case,files', load_test_cases('test-data'))
+def test_yaml_case(filename: Text, case: Text, files: List[dict],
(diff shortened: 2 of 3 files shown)


the-knights-who-say-ni:

Unfortunately we couldn't find an account corresponding to your GitHub username on [bugs.python.org](http://bugs.python.org/) (b.p.o) to verify you have signed the CLA (this might be simply due to a missing "GitHub Name" entry in your b.p.o account settings). This is necessary for legal reasons before we can look at your contribution. Please follow [the steps outlined in the CPython devguide](https://cpython-devguide.readthedocs.io/pullrequest.html#licensing) to rectify this issue.


test-data/check-modules.yml
```diff
@@ -0,0 +1,17 @@
+# Type checker test cases dealing with modules and imports.
+
+- case: Access imported definitions
+  files:
```
ilevkivskyi: Is it possible to omit ``files:`` line? It looks redundant, but I have never worked with YAML.

test-data/check-modules.yml
```diff
@@ -0,0 +1,17 @@
...
+    - content: |
+        import m
+        import typing
+        m.f()           # E: Too few arguments for "f"
+        m.f(object())   # E: Argument 1 to "f" has incompatible type "object"; expected "A"
+        m.x = object()  # E: Incompatible types in assignment (expression has type "object", variable has type "A")
+        m.f(m.A())
+        m.x = m.A()
+    - path: m.py
+      content: |
```
ilevkivskyi: Is it possible to somehow make this less verbose? Mypy tests just use single line ``[file m.py]``.

vlasovskikh:
@ilevkivskyi Yes, YAML is extensible so we can add new attributes in case we need them.

For Python versions it could look like this:

```yaml
- case: Assignment and var def
  python_versions:
    - "2.7"
    - "3.6"
  files:
    - content: |
        ...
```

or, equivalently, like this (similar to Python lists of strings):

```yaml
- case: Assignment and var def
  python_versions: ["2.7", "3.6"]
  files:
    - content: |
        ...
```

test-data/check-modules.yml
vlasovskikh: No, it's not possible to omit it here. In a sense, YAML is just a simpler way to write JSON objects, which are, from the Python's point of view, just dictionaries.

So, this YAML file:

```yaml
# Type checker test cases dealing with modules and imports.

- case: Access imported definitions
  files:
    - content: |
        ...
    - path: m.py
      content: |
        ...
```

is equivalent to the following Python dict:

```python
[{'case': 'Access imported definitions',
  'files': [{'content': '...\n'}, {'content': '...', 'path': 'm.py'}]}]
```

(You can convert a YAML string to a Python dict / list using `yaml.load(yaml_string)`.)

So we need the `files` attribute, as it becomes the `'files'` key in our test case dict.

test-data/check-modules.yml
vlasovskikh: @ilevkivskyi I'm afraid not. `path` and `content` become the keys of the resulting dict: `{'path': 'm.py', 'content': '...'}`. I've made the file path `main.py` default so you don't have to type `path` for every test. But when you need a custom path, you have to provide it as an attribute.

ilevkivskyi: OK, I see. Still, YAML looks quite flexible, let's see what others will say.

tests/mypy_test.py
```diff
@@ -0,0 +1,89 @@
...
+    tmpdir.chdir()
+
+    with Popen(['mypy', '.'], stdout=PIPE, stderr=PIPE) as proc:
+        stdout, stderr = proc.communicate()
+
+    output = stdout + b'\n' + stderr
+    output_lines = output.decode(ENCODING).strip().splitlines()
+    actual_files = inline_errors(output_lines, files)
+    for expected, actual in zip(files, actual_files):
+        assert expected.get('content', '') == actual.get('content', '')
```
gvanrossum: While pytest produces nice diff-style output when an assert fails, there's one problem: if the expected output for several files doesn't match the actual output, pytest only shows the diff for the first file with a mismatch. This seems a problem (since sometimes unexpected errors in the second file provide a clue for the unexpected errors in the first file).

tests/mypy_test.py
```diff
@@ -0,0 +1,89 @@
...
+        contents[path] = ''.join(lines)
+
+    results = [file.copy() for file in files]
+    for result in results:
+        result['content'] = contents.get(result.get('path', DEFAULT_PATH), '')
+    return results
+
+
+def strip_errors(s: Text) -> Text:
+    return re.sub(r'#\s*E:.*', r'', s)
```
gvanrossum: Can we split on ' # ' (i.e. requiring at least one space on each side of '#')? See https://github.com/python/mypy/pull/3417 -- the reason is that we want to allow error messages containing URLs of the form http://host/path#fragment to deep-link directly to a specific section of the docs (like https://github.com/python/mypy/pull/3411/files#diff-7739b07c7ae4561e51cbb36f31984c90R1016 -- still WIP).

tests/mypy_test.py
```diff
@@ -0,0 +1,89 @@
...
+def parse_error_line(line: Text) -> Tuple[int, Text, Text]:
+    m = re.match(r'(.+):([0-9]+): (error): (.+)', line)
+    assert m, 'Cannot parse output line: %s' % line
+    path = m.group(1)
+    index = int(m.group(2)) - 1
+    message = m.group(4)
+    return index, message, path
+
+
+@pytest.mark.parametrize('filename,case,files', load_test_cases('test-data'))
```
gvanrossum: This appears to be calling load_test_cases() even if this test were not selected. Upon further inspection load_test_cases() is a generator function so the call doesn't actually do any work, but it still creeped me out. (Maybe load_test_cases() deserves a comment explaining why it *should* be a generator.)

tests/mypy_test.py
kirbyfan64: IMO it would look nice if the output for each file were concatenated with headers, e.g.:

```
-----x.py-----
output
-----y.py----
output
```

Then you could see it all at once.

tests/mypy_test.py
```diff
@@ -0,0 +1,89 @@
...
+def test_yaml_case(filename: Text, case: Text, files: List[dict],
```
carljm: Pytest allows easily writing a plugin to directly parse non-Python files as pytest-native test files containing individual tests, rather than using test parametrization like this. I've done this before, and it works very well and has a number of nice properties, e.g.:

- output when running tests has a line of dots per actual test file, not just a single line of dots referring uselessly to this file
- test names in failure output can be exactly the name listed in the yaml file, with no repeated `test_yaml_case` cruft
- you can use the obvious `pytest path/to/test-basic.yaml` natively, to run the tests in `test-basic.yaml` only, just like you would with a file containing Python tests

Sample code for this approach can be found at https://docs.pytest.org/en/latest/example/nonpython.html

vlasovskikh:
Since there is not a lot of interest in the common test suite at the moment among the developers of type checkers, I'm closing this PR. Maybe someday :)

PR #437, opened 2017-06-09
Extendable NamedTuples

### LINKED ISSUE #427: NamedTuple subclassing NamedTuple
I naively thought that since I can write
```python
class A(NamedTuple):
    x: int
    y: int
```
I would also be able to subclass this:
```python
class B(A):
    z: int
```
That is accepted syntactically and at runtime but no new `__new__` method is generated so calling `B(1, 2, 3)` is flagged as an error (too many arguments) by both mypy and runtime.

ilevkivskyi:
Fixes #427 

There are three notes:

First, ``NamedTuple`` must be explicitly mentioned in the bases to get the desired behavior. This is necessary since ``typing.NamedTuple`` returns a plain ``collections.namedtuple``, so that one needs ``NamedTuple`` to invoke ``NamedTupleMeta``:
```python
class A(NamedTuple):
    x: int
    y: int
class B(A, NamedTuple):
    z: int
```

Second, in the case of multiple inheritance fields are collected in reversed order, since this is similar to how MRO works, so that this:
```python
class Point(NamedTuple):
    x: int
    y: int
class Label(NamedTuple):
    label: str
class LabeledPoint(Label, Point, NamedTuple):
    pass
```
is equivalent to
```python
class LabeledPoint(NamedTuple):
    x: int
    y: int
    label: str  # this comes last
```

Finally, user defined methods are not inherited for sub-namedtuples (since the resulting ``collections.namedtuple`` has only single base -- ``tuple``). This could be patched by manually recalculating the MRO, but then ``typing.NamedTuple`` becomes overly complicated.

In general, I propose this to be the _last_ feature of ``typing.NamedTuple``, named tuples are designed to be simple compact classes. I propose to direct all future feature requests to dataclasses.

### DIFF
--- a/src/test_typing.py
+++ b/src/test_typing.py
@@ -1616,11 +1616,11 @@ class Base2(NamedTuple):
     def method2(self):
         return self.z
 
-class Derived1(Base1):
+class Derived1(Base1, NamedTuple):
     '''Named tuples can have doctrings.'''
     label: str
 
-class Derived2(Base2, Base1):
+class Derived2(Base2, Base1, NamedTuple):
     pass
 
 class TwoDefaults(NamedTuple):
@@ -2317,12 +2317,12 @@ class NonDefaultAfterDefault(NamedTuple):
 """)
         with self.assertRaises(TypeError):
             exec("""
-class BadMerged(Base1, Base2):
+class BadMerged(Base1, Base2, NamedTuple):
     pass
 """)
         with self.assertRaises(TypeError):
             exec("""
-class BadExtended(Base2):
+class BadExtended(Base2, NamedTuple):
     label: str
 """)
 
--- a/src/test_typing.py
+++ b/src/test_typing.py
@@ -2332,9 +2332,9 @@ class BadExtended(Base2, NamedTuple):
         self.assertEqual(XMeth(42).x, XMeth(42)[0])
         self.assertEqual(str(XRepr(42)), '42 -> 1')
         self.assertEqual(XRepr(1, 2) + XRepr(3), 0)
-        self.assertEqual(Derived1(1, 2, 'test').meth1(), 3)
-        self.assertEqual(Derived2(3, 4).meth1(), 7)
-        self.assertEqual(Derived2(3, 4, 5).meth2(), 5)
+        self.assertEqual(Derived1(1, 2, 'test').method1(), 3)
+        self.assertEqual(Derived2(3, 4).method1(), 7)
+        self.assertEqual(Derived2(3, 4, 5).method2(), 5)
 
         with self.assertRaises(AttributeError):
             exec("""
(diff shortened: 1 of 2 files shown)


gvanrossum:
> Finally, user defined methods are not inherited for sub-namedtuples (since the resulting collections.namedtuple has only single base -- tuple). This could be patched by manually recalculating the MRO, but then typing.NamedTuple becomes overly complicated.

This sounds disturbing -- is this only a problem when you're mixing `namedtuple` and `NamedTuple`? Then it's fine.

ilevkivskyi:
> This sounds disturbing -- is this only a problem when you're mixing ``namedtuple`` and ``NamedTuple``? 

``NamedTupleMeta`` always creates named tuple classes just by calling ``collections.namedtuple`` and then _manually_ populating the namespace of the freshly created class (which is just a simple subclass of ``tuple``) with the definitions (methods). There is no other way we can put user-defined methods into class returned by ``collections.namedtuple`` (it is created using ``exec``).

JukkaL:
If methods can't be inherited, I'd rather reject inheritance from named tuple with user methods than silently drop them. 

What about `isinstance`? Is a derived named tuple a subclass/subtype of the base named tuple or not? If not, the use of class syntax is kind of confusing, as subclassing implies subtyping elsewhere.

ilevkivskyi:
@JukkaL 
> If methods can't be inherited, I'd rather reject inheritance from named tuple with user methods than silently drop them.

Not that it is _impossible_, it is possible by manually reconstructing MRO, but I think we should not do this. If Guido is fine with raising an exception, then I will do this.

> What about ``isinstance``?

It returns ``False``. The only base of all named tuples is ``tuple``.

> If not, the use of class syntax is kind of confusing

There is an alternative option:
```python
class C(NamedTuple, extend_fields_from=[A, B]):  # kwarg name can be discussed :-)
    ...
```
Then it will be easy to explain why we drop methods from ``A`` and ``B``.

@gvanrossum 
> This all makes me worry about the entire feature and regret I even proposed it

I have seen this request few times elsewhere (on Stackoverflow?) so that we may want implement it. But we should probably only support basic/simple extending.

ilevkivskyi:
After thinking a bit I am inclined to only support something as simple as:
```python
class C(NamedTuple, extend_fields_from=B):
    ...
```
with only one "base" named tuple. If someone will ask for something more complex, then one should use dataclasses instead.

gvanrossum:
Hm, ok, the keyword arg makes it clear that this is not inheritance per se, but it's a rarely used pattern (many experienced Python users do a double-take when they see it for the first time).

Also, while it makes sense that there is no actual inheritance (neither in the static type system nor at runtime), because the created types inherits from collections.namedtuple, people will still be confused, because they can't have a function that accepts an A and pass it a B. (They can, but mypy will reject it, and isinstance(arg, A) will be false.)

At this point I think we need to either reject the feature proposal or put it off until Python 3.7. It may also make sense to continue the design discussion in the original issue rather than in this PR.

ilevkivskyi:
OK, let us close this for now. I will make an alternative proposal in the issue soon.

PR #438, opened 2017-06-09
Add AsyncContextManager generic class

ilevkivskyi:
Asynchronous context managers are defined by PEP 492, but there is no corresponding generic abstract class in ``typing``. This PR adds it for Python 3.5+.

### DIFF
--- a/src/test_typing.py
+++ b/src/test_typing.py
@@ -1554,8 +1554,8 @@ class AsyncIteratorWrapper(typing.AsyncIterator[T_a]):
             raise StopAsyncIteration
 
 class ACM:
-    async def __aenter__(self):
-        return self
+    async def __aenter__(self) -> int:
+        return 42
     async def __aexit__(self, etype, eval, tb):
         return None
 """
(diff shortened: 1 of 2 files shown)


src/test_typing.py
```diff
@@ -2165,6 +2171,19 @@ def manager():
...
+        class NotACM:
+            pass
+        self.assertIsInstance(ACM(), typing.AsyncContextManager)
+        self.assertNotIsInstance(NotACM(), typing.AsyncContextManager)
+        @contextlib.contextmanager
+        def manager():
+            yield 42
+
+        cm = manager()
+        self.assertNotIsInstance(cm, typing.AsyncContextManager)
```
gvanrossum: I'd add a test that also serves as an example of how to use AsyncContextManager (presumably in `async with ...`?

JelleZijlstra:
I was planning to do this after AbstractAsyncContextManager is merged into CPython, but there's really no need to wait. The code looks good to me. It would be nice if this can make it into 3.6.2.

gvanrossum:
Thanks!

Issue #432, opened 2017-05-17
Mark Shannon's presentation at the 2017 Language Summit


gvanrossum:
@ilevkivskyi @markshannon.

Mark observed that the typing module uses classes to represent types. This can be expensive, since e.g. the type List[int] really ought to be the tuple (List, int) but it's actually a class object which has a fair amount of overhead (though not as much as early versions of typing.py, since we now cache these class objects).

If we changed to tuples (or at least objects simpler than class object), we'd have a problem: The simpler object couldn't be subclassed from. But who subclasses List[int]? Then again, maybe simpler objects aren't the point?

Mark also pointed out that after
```python
from typing import List
class C(List[int]): pass
print(C.__mro__)
```
We find that `C.__mro__` has 17 items!

I confirmed this. The roughly equivalent code using collections.abc
```python
from collections.abc import MutableMapping
class C(MutableMapping): pass
print(C.__mro__)
```
has only 7 items. And subclassing builtins.list
```python
class C(list): pass
print(C.__mro__)
```
has only three.

This affects performance, e.g. which is faster?
```python
class C(list, Sequence[int]): pass
C().append(1)
class D(Sequence[int], list): pass
D().append(1)
```
One append() call is 10% faster than the other.

JukkaL:
I agree with Mark that this is a problem and can be a blocker for adopting type annotations for some projects. We are mostly shielded from the problem at Dropbox because we use comment annotations, but once we migrate to Python 3 we'd potentially have problems with the current approach as well.

Inheriting from `List[int]` (or, say `Dict[str, Foo]`) is occasionally useful and probably worth supporting, but Mark's suggestion to use a different syntax for this seems like a reasonable compromise. I think his suggestion was to do something like this, since `List[int]` is not a class:

```py
from typing import implements, List

@implements(List[int])
class MyList(list): ...
```

It's unclear whether this would also affect user-defined generic classes. For example, consider this code that works currently:

```py
class Box(Generic[T]): ...

class SpecialBox(Box[int]): ...

class AnotherBox(Box[T]): ...

def do_stuff(box: Box[int], box2: AnotherBox[str]) -> None: ...
```

If I understood things right, this would be written like this based on Mark's proposal:

```py
class Box(Generic[T]): ...

@implements(Box[int])
class SpecialBox(Box): ...

@implements(Box[T])
class AnotherBox(Box): ...

def do_stuff(box: Box[int], box2: AnotherBox[str]) -> None: ... # No change
```

[...]

PR #439, opened 2017-06-11
Speed improvements for typing

ilevkivskyi:
This PR contains two performance improvements based on profiling and inspired by #432:
* Inline ``_gorg`` and ``_geqv``
* Remove (IMO unnecessary) "duplicates" from MRO.

This gives roughly 5% speed-up for ``mypy mypy``. @JukkaL I will be grateful if you could do more benchmarking with some real-life large codebases.

### DIFF
--- a/python2/typing.py
+++ b/python2/typing.py
@@ -1051,25 +1051,15 @@ class GenericMeta(TypingMeta, abc.ABCMeta):
                 tvars = gvars
 
         initial_bases = bases
-        if not origin:
-            # Erase base classes
-            new_bases = []
-            for base in bases:
-                erased = base
-                if isinstance(base, GenericMeta):
-                    erased = base._gorg
-                    if extra:  # Even stronger erasure for generic ABCs in typing
-                        bextra = getattr(base, '__extra__', None)
-                        erased = bextra if bextra else erased
-                new_bases.append(erased)
+        if extra is None:
+            extra = namespace.get('__extra__')
+        if extra is not None and type(extra) is abc.ABCMeta and extra not in bases:
+            bases = (extra,) + bases
+        bases = tuple(b._gorg if isinstance(b, GenericMeta) else b for b in bases)
 
-            bases = tuple(new_bases)
-            if type(extra) is abc.ABCMeta and extra not in bases:
-                bases = (extra,) + bases
-
-            # Remove bare Generic from bases if there are other generic bases
-            if any(isinstance(b, GenericMeta) and b is not Generic for b in bases):
-                bases = tuple(b for b in bases if b is not Generic)
+        # remove bare Generic from bases if there are other generic bases
+        if any(isinstance(b, GenericMeta) and b is not Generic for b in bases):
+            bases = tuple(b for b in bases if b is not Generic)
         namespace.update({'__origin__': origin, '__extra__': extra})
         self = super(GenericMeta, cls).__new__(cls, name, bases, namespace)
         super(GenericMeta, self).__setattr__('_gorg', self if not origin
--- a/python2/typing.py
+++ b/python2/typing.py
@@ -1626,6 +1626,7 @@ class _ProtocolMeta(GenericMeta):
                     if (not attr.startswith('_abc_') and
                             attr != '__abstractmethods__' and
                             attr != '_is_protocol' and
+                            attr != '_gorg' and
                             attr != '__dict__' and
                             attr != '__args__' and
                             attr != '__slots__' and
--- a/python2/typing.py
+++ b/python2/typing.py
@@ -1074,15 +1074,25 @@ class GenericMeta(TypingMeta, abc.ABCMeta):
                 tvars = gvars
 
         initial_bases = bases
-        if extra is None:
-            extra = namespace.get('__extra__')
-        if extra is not None and type(extra) is abc.ABCMeta and extra not in bases:
-            bases = (extra,) + bases
-        bases = tuple(_gorg(b) if isinstance(b, GenericMeta) else b for b in bases)
+        if not origin:
+            # Erase base classes
+            new_bases = []
+            for base in bases:
+                erased = base
+                if isinstance(base, GenericMeta):
+                    erased = _gorg(base)
+                    if extra:  # Even stronger erasure for generic ABCs in typing
+                        bextra = getattr(base, '__extra__', None)
+                        erased = bextra if bextra else erased
+                new_bases.append(erased)
 
-        # remove bare Generic from bases if there are other generic bases
-        if any(isinstance(b, GenericMeta) and b is not Generic for b in bases):
-            bases = tuple(b for b in bases if b is not Generic)
+            bases = tuple(new_bases)
+            if type(extra) is abc.ABCMeta and extra not in bases:
+                bases = (extra,) + bases
+
+            # Remove bare Generic from bases if there are other generic bases
+            if any(isinstance(b, GenericMeta) and b is not Generic for b in bases):
+                bases = tuple(b for b in bases if b is not Generic)
         namespace.update({'__origin__': origin, '__extra__': extra})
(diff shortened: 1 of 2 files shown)


ilevkivskyi:
This now only touches Python 3 version. If Jukka's results will be promising, then I will also backport this to Python 2.

JelleZijlstra:
I'm also seeing a few % speedup running mypy on my codebase.

src/typing.py
```diff
@@ -991,14 +968,26 @@ def __new__(cls, name, bases, namespace,
         initial_bases = bases
         if extra is not None and type(extra) is abc.ABCMeta and extra not in bases:
             bases = (extra,) + bases
-        bases = tuple(_gorg(b) if isinstance(b, GenericMeta) else b for b in bases)
+        new_bases = []
+        for base in bases:
+            if isinstance(base, GenericMeta):
+                bextra = getattr(base, '__extra__', None)
```
gvanrossum: As long as we're micro-optimizing, this call is wasted when extra is falsey or origin is truthy (if I understand the condition below correctly).

src/typing.py
```diff
@@ -991,14 +968,26 @@ def __new__(cls, name, bases, namespace,
...
+                if (extra and bextra and not origin and bextra not in bases and
+                        type(bextra) is abc.ABCMeta):
```
gvanrossum: I'm not sure how this is equivalent to the original code (I would have expected something simpler, matching `_gorg(b) if isinstance(b, GenericMeta) else b`.

gvanrossum:
PS. I'm not sure how we would test this speedup with real code. Did you ask for us to have this patched into the Python version we use to run mypy and then run mypy over our large codebase? That makes sense except the time there is pretty variable (since it runs so long that all sorts of other activities on the same machine affect the timing).

src/typing.py
```diff
@@ -849,29 +849,6 @@ def __getitem__(self, arg):
...
-    The intention is to consider generic class X and any of its
-    parameterized forms (X[T], X[int], etc.) as equivalent.
-
-    However, X is not equivalent to a subclass of X.
-
-    The relation is reflexive, symmetric and transitive.
-    """
-    assert isinstance(a, GenericMeta) and isinstance(b, GenericMeta)
-    # Reduce each to its origin.
-    return _gorg(a) is _gorg(b)
```
ilevkivskyi: Probably this is just a coincidence, maybe it was different in older versions, but now ``b`` (and its ``._gorg``) is known "statically" in all places where ``_geqv`` is used.

src/typing.py
ilevkivskyi: OK, I reshuffled code here, so that we quickly escape if ``origin`` is truthy and then if ``extra`` is falsey.

src/typing.py
ilevkivskyi: The point is this is not totally equivalent. This makes more "aggressive" erasure for classes with ``__extra__``, this shortens MROs for special typing types. Compare:
Before:
```python
>>> pprint(List[int].__mro__)
(typing.List[int],
 typing.List,
 <class 'list'>,
 typing.MutableSequence,
 <class 'collections.abc.MutableSequence'>,
 typing.Sequence,
 <class 'collections.abc.Sequence'>,
 typing.Reversible,
 <class 'collections.abc.Reversible'>,
 typing.Collection,
 <class 'collections.abc.Collection'>,
 <class 'collections.abc.Sized'>,
 typing.Iterable,
 <class 'collections.abc.Iterable'>,
 typing.Container,
 <class 'collections.abc.Container'>,
 typing.Generic,
 <class 'object'>)
```
and after:
```python
(typing.List[int],
 typing.List,
 <class 'list'>,
 <class 'collections.abc.MutableSequence'>,
 <class 'collections.abc.Sequence'>,
 <class 'collections.abc.Reversible'>,
 <class 'collections.abc.Collection'>,
 <class 'collections.abc.Sized'>,
 <class 'collections.abc.Iterable'>,
 <class 'collections.abc.Container'>,
 <class 'object'>)
```

ilevkivskyi: I am actually a bit worried maybe this is too aggressive, since sometimes we even lose ``Generic``, for example:
```python
>>> pprint(MutableMapping.__mro__)
(typing.MutableMapping,
 <class 'collections.abc.MutableMapping'>,
 <class 'collections.abc.Mapping'>,
 <class 'collections.abc.Collection'>,
 <class 'collections.abc.Sized'>,
 <class 'collections.abc.Iterable'>,
 <class 'collections.abc.Container'>,
 <class 'object'>)
```
Although all tests pass, and I tried to play with this and it works as expected (it is still an instance of ``GenericMeta``). Probably, I will add ``Generic`` at the end of bases in such cases (just for consistency).

ilevkivskyi:
I am marking this as WIP until I figure out what is going on with Python 2.

@gvanrossum 
> PS. I'm not sure how we would test this speedup with real code. Did you ask for us to have this patched into the Python version we use to run mypy and then run mypy over our large codebase? That makes sense except the time there is pretty variable (since it runs so long that all sorts of other activities on the same machine affect the timing).

Yes, even if it is not super-stable it would be great to have few more data points.

ilevkivskyi:
@gvanrossum 

I decided to keep only obvious optimizations, and leave more aggressive things (that might change semantics) out for now. I think you should be happy with this now.

I will probably make another PR soon dealing with (excessively) long MROs.

python2/typing.py
```diff
@@ -1078,13 +1055,15 @@ def __new__(cls, name, bases, namespace,
...
-        bases = tuple(_gorg(b) if isinstance(b, GenericMeta) else b for b in bases)
+        bases = tuple(b._gorg if isinstance(b, GenericMeta) else b for b in bases)
 
         # remove bare Generic from bases if there are other generic bases
         if any(isinstance(b, GenericMeta) and b is not Generic for b in bases):
             bases = tuple(b for b in bases if b is not Generic)
         namespace.update({'__origin__': origin, '__extra__': extra})
         self = super(GenericMeta, cls).__new__(cls, name, bases, namespace)
+        super(GenericMeta, self).__setattr__('_gorg', self if not origin
+                                             else origin._gorg)
```
gvanrossum: Whitespace nit: I don't love how this line is broken up; either split before the comma or indent the `else` to align with the `if`.

src/typing.py
gvanrossum: Aha!

src/typing.py
```diff
@@ -1568,7 +1546,7 @@ def no_type_check(arg):
     if isinstance(arg, type):
         arg_attrs = arg.__dict__.copy()
         for attr, val in arg.__dict__.items():
-            if val in arg.__bases__:
+            if val in arg.__bases__ + (arg,):
```
gvanrossum: Why is this added `(arg,)` needed? And why not in the PY2 version? (If there's a good reason please add a comment.)

ilevkivskyi: Yes, it is better to have them in both versions, ``@no_type_chek`` can be also used in Python 2, fixed this.

gvanrossum: Oh wait, it's needed because `arg._gorg == arg`. Fine.

gvanrossum:
I'm torn about whether to merge this. Maybe we should wait until 3.6.2 has actually been released? The RC is supposed to be cut today, I expect the final release in a few weeks (see PEP 494).

src/typing.py
ilevkivskyi: > Oh wait, it's needed because ``arg._gorg == arg``

Exactly. Sorry, I have answered only half of your question.

ilevkivskyi:
> Maybe we should wait until 3.6.2 has actually been released? 

This makes sense. We can keep the master "clean" (i.e. identical with current CPython state) for any emergency fixes (just in case). Then I can merge this right after the release of 3.6.2.final.

ilevkivskyi:
OK, Python 3.6.2 was released earlier today, so that I am merging this now.

Issue #435, opened 2017-05-23
Add typing_extensions module for new / optional typing features to PyPI


vlasovskikh:
Let's add `typing_extensions` module to PyPI in addition to Mypy-specific `mypy_extensions`.

Currently new stuff in typing appears in `mypy_extensions` even though most of the new typing features are not Mypy-specific.

In order to make new typing features available to 1) previous Python versions, 2) all type checkers I would suggest to move these features to `typing_extensions`.

The good candidates for including in `typing_extensions` could be:

* `NoReturn` (apparently already in PEP 484)
* `Protocol` (draft PEP 544)
* Argspecs?
* `TypedDict`?

gvanrossum:
Agreed. I do think there should be a requirement for things to appear in a PEP (either 484 or a new one) before they go into typing_extensions. This means we should probably write up TypedDict and maybe wait for argspecs (for those there's still a lot of possible variation in the eventual design).

Michael0x2a:
Type and maybe NewType would be other good candidates. They're present in the later versions of Python 3.5, but not the earlier ones, iirc.

gvanrossum:
Oooh, might as well!

Michael0x2a:
Other ones perhaps worth adding in:

- `typing.Collection` -- added in 3.6
- `typing.Deque` -- added in 3.6.1
- `typing.ContextManager` -- added in 3.6
- `typing.Counter` -- added in 3.6.1
- `typing.ChainMap` -- added in 3.6.1
- `typing.AsyncGenerator` -- added in 3.5.4
- `typing.ClassVar` -- added in 3.5.3

(I got this list by looking for the "New in" phrase in the typing docs, though it might not be comprehensive. For example, neither `Type` nor `NewType` have the "New in ..." phrases next to them in the index at the end (which might in part be my fault, tbh))

JelleZijlstra:
Sounds like it might be worthwhile to just shadow/duplicate all of regular `typing`, so that people aren't tied to the stdlib's version of typing. The disadvantage is that type checkers would have to add special-casing for things imported from `typing_extensions`.

I've been thinking of a related idea to create a library of commonly useful type aliases. Some things that could be included are:
- An alias for `Union[bytes, Text]` (which is needed very often in typeshed)
- A generic alias for `Union[AnyStr, os.PathLike[AnyStr]]` that falls back to `AnyStr` before 3.6 (another common typeshed need)
[...]

PR #443, opened 2017-06-27
Add typing extensions

Michael0x2a:
This pull request adds a `typing_extensions` subproject to `typing`. The `typing_extensions` module backports any new additions to `typing` for Python 3.5+ users who are using older versions of `typing` that were bundled with their standard library (and so can't update to the latest versions).

See https://github.com/python/typing/issues/435 for more motivation and additional context.


ilevkivskyi:
I don't think you need all these old versions bundled in tests. I think we should just properly configure Travis (it supports Python micro releases like 3.5.2 etc. (like in mypy's ``.travis.yml``), and specific versions of ``typing`` to be installed before tests)

Michael0x2a:
@ilevkivskyi I guess my only concern with that is that it becomes harder to do sanity checks locally -- having to constantly push to run tests seems a bit annoying to me.

ilevkivskyi:
> having to constantly push to run tests seems a bit annoying to me.

You can still keep these copies locally (without tracking them in git). But I would prefer a properly configured ``.travis.yml`` to test with actual Python versions (note that Python runtime changes itself between versions, not only the stdlib, there are few places in ``typing`` where this is actually important).

typing_extensions/src_py3/typing_extensions.py
```diff
@@ -0,0 +1,612 @@
...
+T = typing.TypeVar('T')  # Any type.
+KT = typing.TypeVar('KT')  # Key type.
+VT = typing.TypeVar('VT')  # Value type.
+T_co = typing.TypeVar('T_co', covariant=True)  # Any type covariant containers.
+V_co = typing.TypeVar('V_co', covariant=True)  # Any type covariant containers.
+VT_co = typing.TypeVar('VT_co', covariant=True)  # Value type covariant containers.
+T_contra = typing.TypeVar('T_contra', contravariant=True)  # Ditto contravariant.
+
+
+def _gorg(a):
```
ilevkivskyi: Note that ``_gorg`` and ``_geqv`` will be removed soon after Python 3.6.2 is released, see PR https://github.com/python/typing/pull/439.

ilevkivskyi:
Concerning the test failure, we decided to stop supporting Python 3.2 (it reached EoL a year ago), so that you can just skip testing on Python 3.2.

Michael0x2a:
Update: I removed `run_tests.py` and the associated machinery in favor of just using Travis, as suggested. The only remaining blocker is that we can't actually install typing on Python 3.2. I figured it'd be best to make a separate pull request for that though: https://github.com/python/typing/pull/444

Once that's accepted, I can rebase, but this pull request should otherwise be ready for review (?).

ilevkivskyi:
I can review this, but I am not sure I will be able to review before week-end.

typing_extensions/src_py3/typing_extensions.py
Michael0x2a: Sounds good -- once this pull request gets accepted, I can prepare another pull request to add those changes and leave it pending just like #439.

typing_extensions/src_py2/typing_extensions.py
```diff
@@ -0,0 +1,113 @@
...
+    'ContextManager',
+    'Counter',
+    'Deque',
+    'DefaultDict',
+
+    # One-off things.
+    'NewType',
+    'overload',
+    'Text',
+    'TYPE_CHECKING',
```
ilevkivskyi: ``GenericMeta`` was made public recently (in 3.6.1 IIRC), should we also add it here?

typing_extensions/src_py3/test_typing_extensions.py
```diff
@@ -0,0 +1,636 @@
...
+import typing
+import typing_extensions
+try:
+    import collections.abc as collections_abc
+except ImportError:
+    import collections as collections_abc  # Fallback for PY3.2.
+
+TYPING_V2 = sys.version_info >= (3, 5, 1)
+TYPING_V3 = sys.version_info >= (3, 5, 3)
+TYPING_V4 = sys.version_info >= (3, 6, 1)
```
ilevkivskyi: We will have ``TYPING_V5`` very soon (but I think it is OK to not include it in this PR, just don't forget to add it later).

typing_extensions/src_py3/test_typing_extensions.py
```diff
@@ -0,0 +1,636 @@
...
+from typing import Generic
+from typing import get_type_hints
+from typing import no_type_check
+from typing_extensions import NoReturn, ClassVar, Type, NewType
+import typing
+import typing_extensions
+try:
+    import collections.abc as collections_abc
+except ImportError:
+    import collections as collections_abc  # Fallback for PY3.2.
```
ilevkivskyi: We don't support 3.2 anymore :-) this can be removed.

typing_extensions/src_py3/test_typing_extensions.py
```diff
@@ -0,0 +1,636 @@
...
+            raise self.failureException(message)
+
+    def assertNotIsSubclass(self, cls, class_or_tuple, msg=None):
+        if issubclass(cls, class_or_tuple):
+            message = '%r is a subclass of %r' % (cls, class_or_tuple)
+            if msg is not None:
+                message += ' : %s' % msg
+            raise self.failureException(message)
+
+    def clear_caches(self):
```
ilevkivskyi: Again, I am not sure you need this.

typing_extensions/src_py3/test_typing_extensions.py
```diff
@@ -0,0 +1,636 @@
...
+
+    def test_noreturn_instance_type_error(self):
+        with self.assertRaises(TypeError):
+            isinstance(42, NoReturn)
+
+    def test_noreturn_subclass_type_error_1(self):
+        with self.assertRaises(TypeError):
+            issubclass(Employee, NoReturn)
+
+    @skipUnless(TYPING_V3, "Behavior added in typing v3")
```
ilevkivskyi: It is not necessary, but it would be super-helpful to mention PR numbers where behavior was changed (also for other changes).

typing_extensions/src_py3/test_typing_extensions.py
```diff
@@ -0,0 +1,636 @@
...
+        if TYPING_V3:
+            self.assertNotIsSubclass(collections.ChainMap, MyChainMap)
+
+    def test_deque_instantiation(self):
+        self.assertIs(type(typing_extensions.Deque()), collections.deque)
+        self.assertIs(type(typing_extensions.Deque[T]()), collections.deque)
+        self.assertIs(type(typing_extensions.Deque[int]()), collections.deque)
+        class D(typing_extensions.Deque[T]): ...
+        if TYPING_V3:
+            self.assertIs(type(D[int]()), D)
```
ilevkivskyi: Potentially you might also add a test for ``__bases__`` they are also type-erased since one of the releases of ``typing``.

(But choose a class that actually derives from something subscripted, maybe ``Counter``.)

typing_extensions/src_py3/test_typing_extensions.py
```diff
@@ -0,0 +1,636 @@
...
+        exclude = {'overload', 'Text', 'TYPE_CHECKING'}
+        for item in typing_extensions.__all__:
+            if item not in exclude and hasattr(typing, item):
+                self.assertIs(
+                    getattr(typing_extensions, item),
+                    getattr(typing, item))
+
+
+if __name__ == '__main__':
+   # main(argv=[sys.argv[0]] + sys.argv[2:])
```
ilevkivskyi: Is this comment necessary? Why do we need this pattern?

typing_extensions/src_py3/typing_extensions.py
```diff
@@ -0,0 +1,612 @@
+import abc
+import collections
+import contextlib
+import sys
+import typing
+try:
+    import collections.abc as collections_abc
+except ImportError:
+    import collections as collections_abc  # Fallback for PY3.2.
```
ilevkivskyi: Again, this is not needed.

typing_extensions/src_py3/typing_extensions.py
```diff
@@ -0,0 +1,612 @@
...
+
+if hasattr(typing, '_generic_new'):
+    _generic_new = typing._generic_new
+else:
+    def _generic_new(base_cls, cls, *args, **kwargs):
+        return base_cls.__new__(cls, *args, **kwargs)
```
ilevkivskyi: I am not 100% sure this is a safe substitution. Do you have a test for this? IIRC, the point was that ``__init__`` was not called at instantiation of subscripted generics.

typing_extensions/src_py3/typing_extensions.py
```diff
@@ -0,0 +1,612 @@
...
+    'ContextManager',
+    'Counter',
+    'Deque',
+    'DefaultDict',
+
+    # One-off things.
+    'NewType',
+    'overload',
+    'Text',
+    'TYPE_CHECKING',
```
ilevkivskyi: Again, I am not sure about ``GenericMeta`` (see comment above).

typing_extensions/src_py3/typing_extensions.py
ilevkivskyi: > or whatever other course of action you think is best

OK, let's just not forget about this.

ilevkivskyi:
Actually, I have one more suggestion. Maybe edit description to reflect the fact that experimental things that are going to appear in ``typing`` later will appear here first (like ``TypedDict``, argspecs, and ``Protocol``).

Michael0x2a:
@ilevkivskyi Thanks for the review! I'll start going through your comments and making fixes in a moment.

gvanrossum:
You will have to wait until after the release unless @JukkaL merges this
despite the code freeze.

typing_extensions/setup.py
```diff
@@ -0,0 +1,70 @@
...
+
+This module allows those users to use the latest additions to the typing
+module without worrying about naming conflicts with the standard library.
+Users of Python 2.7, 3.3, and 3.4 should install the typing module
+from pypi and use that directly, except when writing code that needs to
+be compatible across multiple versions of Python.
+'''
+
+classifiers = [
+    'Development Status :: 5 - Production/Stable',
```
Michael0x2a: Done!

typing_extensions/setup.py
```diff
@@ -0,0 +1,70 @@
...
+    install_requires.append('typing >= 3.6.1')
+
+setup(name='typing_extensions',
+      version=version,
+      description=description,
+      long_description=long_description,
+      author='Guido van Rossum, Jukka Lehtosalo, Lukasz Langa, Michael Lee',
+      author_email='jukka.lehtosalo@iki.fi',
+      # TODO: Change URL
+      url='https://github.com/michael0x2a/typing_extensions',
```
Michael0x2a: Whoops, forgot to change that.

tox.ini
```diff
@@ -22,4 +22,9 @@ exclude =
   # tests have more relaxed formatting rules
   # and its own specific config in .flake8-tests
   python2/test_typing.py,
-  src/test_typing.py
+  src/test_typing.py,
+  typing_extensions/src_py2/test_typing_extensions.py,
+  typing_extensions/src_py3/test_typing_extensions.py,
+  # test_data should contain an exact snapshot of the standard
```
Michael0x2a: Same thing -- forgot to remove it.

.travis.yml
```diff
@@ -14,5 +18,8 @@ install:
 
 script:
   - export PYTHONPATH=`python -c "import sys; print('python2' if sys.version.startswith('2') else 'src')"`; py.test $PYTHONPATH
+  - python setup.py install;
```
Michael0x2a: The problem is that the `typing_extensions` depends on `typing` being installed for Python 3.4 and below since we try and re-export as much as possible from the `typing` module. That said, I've tried adjusting travis so it should install `typing` only for Python 3.4 and below -- we'll see if I did it correctly once the tests are done.

typing_extensions/README.md
```diff
@@ -0,0 +1,66 @@
...
+
+These changes are _not_ backported to prevent subtle compatibility
+issues when mixing the differing implementations of modified classes.
+
+## Running tests
+
+To run tests, navigate into the appropriate source directory and run
+`test_typing_extensions.py`. You will also need to install the latest
+version of `typing` if you are using a version of Python that does not
+include `typing` as a part of the standard library.
```
Michael0x2a: I added it to the `test_requirements.txt` file for the `typing` repo. If you prefer, I can leave that one alone and add a separate `test_requirements.txt` and `requirements.txt` file to the `typing_extensions` folder.

Since typing_extensions is basically a separate project, maybe it makes sense for it to have its own requirement files, but it does feel a little redundant...

typing_extensions/src_py2/test_typing_extensions.py
```diff
@@ -0,0 +1,134 @@
...
+            raise self.failureException(message)
+
+    def assertNotIsSubclass(self, cls, class_or_tuple, msg=None):
+        if issubclass(cls, class_or_tuple):
+            message = '%r is a subclass of %r' % (cls, class_or_tuple)
+            if msg is not None:
+                message += ' : %s' % msg
+            raise self.failureException(message)
+
+    def clear_caches(self):
```
Michael0x2a: Removed.

typing_extensions/src_py2/test_typing_extensions.py
```diff
@@ -0,0 +1,134 @@
...
+            MyAbstractCollection()
+
+    def test_contextmanager(self):
+        @contextlib.contextmanager
+        def manager():
+            yield 42
+
+        cm = manager()
+        self.assertIsInstance(cm, typing_extensions.ContextManager)
+        self.assertNotIsInstance(42, typing_extensions.ContextManager)
```
Michael0x2a: Done!

typing_extensions/src_py2/test_typing_extensions.py
```diff
@@ -0,0 +1,134 @@
...
+        self.assertIn('ClassVar', a)
+        self.assertIn('Type', a)
+        self.assertIn('Counter', a)
+        self.assertIn('DefaultDict', a)
+        self.assertIn('Deque', a)
+        self.assertIn('NewType', a)
+        self.assertIn('overload', a)
+        self.assertIn('Text', a)
+        self.assertIn('TYPE_CHECKING', a)
+
```
Michael0x2a: Removed.

typing_extensions/src_py2/test_typing_extensions.py
```diff
@@ -0,0 +1,134 @@
...
+        exclude = {'overload', 'Text', 'TYPE_CHECKING'}
+        for item in typing_extensions.__all__:
+            if item not in exclude and hasattr(typing, item):
+                self.assertIs(
+                    getattr(typing_extensions, item),
+                    getattr(typing, item))
+
+
+if __name__ == '__main__':
+    main(argv=[sys.argv[0]] + sys.argv[2:])
```
Michael0x2a: Done.

typing_extensions/src_py2/typing_extensions.py
Michael0x2a: I'm not sure if we should -- GenericMeta seems to have gone through some heavy changes since 3.5.0, and if we make it public here, we're basically guaranteeing it'll work reasonably consistently across all Python versions, and I'm not sure if it'll be easy for us to confirm that's really the case.

If you think it's still worth including though, I can add some tests for it + add it to `__all__`.

typing_extensions/src_py3/test_typing_extensions.py
Michael0x2a: Done!

typing_extensions/src_py3/test_typing_extensions.py
Michael0x2a: Will do.

typing_extensions/src_py3/test_typing_extensions.py
Michael0x2a: Yep, removed.

typing_extensions/src_py3/test_typing_extensions.py
Michael0x2a: I'll look into digging those up.

typing_extensions/src_py3/test_typing_extensions.py
Michael0x2a: This was a holdover from the code I previously had for running tests, which we removed. I removed the comment here + in the Python 2 tests.

typing_extensions/src_py3/typing_extensions.py
Michael0x2a: Also removed.

typing_extensions/setup.py
```diff
@@ -0,0 +1,70 @@
...
+                     'to install the typing package.\n')
+    exit(1)
+
+version = '3.6.1'
+description = 'Type Hint backports for Python 3.5+'
+long_description = '''\
+Typing -- Type Hints for Python
+
+This is a backport of the standard library typing module to Python
+versions 3.5+. The typing module has seen several changes since it was
```
Michael0x2a: Done! I just went with "Python 3.7" for the sake of simplicity.

typing_extensions/src_py3/typing_extensions.py
Michael0x2a: I don't know if I necessarily have tests for specifically this -- I was mostly copying what the typing module was already doing and assumed the behavior was correct. For example, in Python 3.5.0 - 3.5.2, the `List` type was defined like this:

```python
class List(list, MutableSequence[T], extra=list):
    def __new__(cls, *args, **kwds):
        if _geqv(cls, List):
            raise TypeError("Type List cannot be instantiated; "
                            "use list() instead")
        return list.__new__(cls, *args, **kwds)
```

In Python 3.5.3+, it was defined like this:

```python
class List(list, MutableSequence[T], extra=list):
    __slots__ = ()

    def __new__(cls, *args, **kwds):
        if _geqv(cls, List):
            raise TypeError("Type List cannot be instantiated; "
                            "use list() instead")
        return _generic_new(list, cls, *args, **kwds)
```

I'm also not sure if this substitution is 100% correct, but if it's broken, it'll at least be consistently broken in the same way the rest of the `typing` module was in Python 3.5.0 - 3.5.2. Defining the function this way also helped me reduce redundancy since I can define things like `Deque` just once instead of twice.

typing_extensions/src_py3/test_typing_extensions.py
```diff
@@ -15,9 +15,29 @@
 import typing_extensions
 import collections.abc as collections_abc
 
-TYPING_V2 = sys.version_info >= (3, 5, 1)
-TYPING_V3 = sys.version_info >= (3, 5, 3)
-TYPING_V4 = sys.version_info >= (3, 6, 1)
+TYPING_LATEST = sys.version_info[:3] < (3, 5, 0)
+TYPING_V2 = TYPING_LATEST or sys.version_info[:3] >= (3, 5, 1)
```
JelleZijlstra: I think this would be clearer as `TYPING_3_5_1` or similar, because "v2" is not a terminology used elsewhere.

Michael0x2a: Ok, done.

typing_extensions/src_py3/test_typing_extensions.py
Michael0x2a: I added a check for this using Counter, as you suggested.

Michael0x2a:
@ilevkivskyi -- Ok, I should be ready for a second round of review. I tried adding links to relevant pull requests and issues for the typing version-specific if statements, but found it quite hard to do so in general, so unfortunately most of them are still undocumented :(.

Michael0x2a:
Just as an FYI, I decided to remove `typing_extensions.Collection` -- in retrospect, I think attempting to backport that type was a mistake.

In order to correctly backport that type, I basically need to inject `typing_extensions.Collection` into the middle of an existing inheritance hierarchy, which I don't think is possible to do cleanly, if at all. More concretely, the following code doesn't typecheck in Python 2 and older versions of Python 3 (the typing stubs have `typing.Sequence` subclass `typing.Collection` only some of the time):

    from mypy_extensions import Collection

    def foo(x):
        # type: Collection[int] -> int
        return sum(x)

    foo([1, 2, 3])

I think this is potentially repairable at least at typechecking-time with some very careful surgery on the typing stubs, but doing that seems questionable to me. It also wouldn't fix any runtime `isinstance` or `issubclass` checks people might try doing.

(More specifically, Python 3's typing stubs has `typing.Sequence` always subclass a phantom `_Collection` type, which is only publicly exposed as `typing.Collection` for Python 3.6+. We could add a similar phantom type to the Python 2 typing stubs, then have the `typing_extension` re-export `typing._Collection` as `typing_extensions.Collection`. Users would then need to use an up-to-date version of typeshed.)

Michael0x2a:
@ilevkivskyi -- as an update, I incorporated the changes you just pushed to master earlier today (removing `_gorg` and `_geqv` and making other speed improvements).

ilevkivskyi:
@Michael0x2a thanks for continuing to work on this! I will try to make the second review today or tomorrow.

typing_extensions/README.md
ilevkivskyi: I would actually prefer separate requirements for ``typing_extensions``.

typing_extensions/src_py2/typing_extensions.py
ilevkivskyi: OK, let us not include ``GenericMeta`` for now. People who want to use it can just import it from ``typing`` (it just does not work with ``import *``).

typing_extensions/src_py3/typing_extensions.py
ilevkivskyi: > be consistently broken in the same way the rest of the ``typing`` module was in Python 3.5.0 - 3.5.2

OK, but then please mention this in a docstring/comment.

ilevkivskyi:
@Michael0x2a 
I responded to your previous comments, continuing with review now.

ilevkivskyi:
Concerning ``Collection`` , I would still add it. Maybe not in this PR, but after you make an initial PR to typeshed with ``typing_extensions`` stubs.

typing_extensions/README.rst
```diff
@@ -0,0 +1,85 @@
...
+.. image:: https://badges.gitter.im/python/typing.svg
+ :alt: Chat at https://gitter.im/python/typing
+ :target: https://gitter.im/python/typing?utm_source=badge&utm_medium=badge&utm_campaign=pr-badge&utm_content=badge
+
+Overview
+========
+
+The ``typing_extensions`` module contains both backports of changes made to
+the ``typing`` module since Python 3.5.0 as well as experimental types that
+will be eventually added to the ``typing`` module once stabilized.
```
ilevkivskyi: I think you forgot to mention that it is provisional until Python 3.7.

typing_extensions/src_py2/test_typing_extensions.py
```diff
@@ -0,0 +1,237 @@
...
+        with self.assertRaises(TypeError):
+            isinstance(cm, ContextManager[int])
+        with self.assertRaises(TypeError):
+            issubclass(type(cm), ContextManager[int])
+
+    def test_counter(self):
+        self.assertIsSubclass(collections.Counter, Counter)
+        self.assertIs(type(Counter()), collections.Counter)
+        self.assertIs(type(Counter[T]()), collections.Counter)
+        self.assertIs(type(Counter[int]()), collections.Counter)
```
ilevkivskyi: I actually wanted to also test that ``Counter.__bases__`` are already erased (before instantiation).

typing_extensions/src_py3/test_typing_extensions.py
```diff
@@ -0,0 +1,624 @@
...
+from typing import Tuple, List
+from typing import Generic
+from typing import get_type_hints
+from typing import no_type_check
+from typing_extensions import NoReturn, ClassVar, Type, NewType
+import typing
+import typing_extensions
+import collections.abc as collections_abc
+
+TYPING_LATEST = sys.version_info[:3] < (3, 5, 0)
```
ilevkivskyi: I am not sure I follow the logic here, shouldn't this be ``TYPING_OLDEST``? Or this is the version of typing you "emulate"? Could you please add a comment here?

Michael0x2a:
@ilevkivskyi -- ok, I've addressed your new comments and am ready for the 3rd round of review!

Regarding your comment regarding `Counter.__bases__`: I think I added a test for that in the Python 3 tests in response to one of your comments from the first round of review, but forgot to add the same test for the Python 2 tests. I copied the same check over.

That said, I want to make sure I actually understand both comments correctly -- are the tests I added what you were looking for, or should I need to add in more checks there?

Michael0x2a:
@ilevkivskyi stubs for typeshed can be found here: https://github.com/python/typeshed/pull/1471

I'll work on making changes to mypy itself next. iirc it contains some special-casing and tests specifically for mypy_extensions so I'll probably start by duplicating a subset of those for typing_extensions.

PR #453, opened 2017-07-23
Bump version to 3.6.2 and use setuptools instead of distutils

### LINKED ISSUE #450: Provide wheel distribution on PyPI
Hi!

Please could the PyPI package also by made available as a wheel?
https://pypi.python.org/pypi/typing/3.6.1

Wheels have the following advantages:
* Faster installation time
* Allows the wheel cache to work even when using the new pip hash-checking mode
* Allows tools to statically inspect the package metadata without having to execute untrusted code.

Since typing uses different source directories for Python 2 vs 3, I'm presuming separate wheels will have to be generated for each, per:
 https://packaging.python.org/tutorials/distributing-packages/#pure-python-wheels

The Python packaging guide recommends using [Twine](https://github.com/pypa/twine) to upload packages:
https://packaging.python.org/tutorials/distributing-packages/#uploading-your-project-to-pypi

As such, the steps for building/uploading a new release are then:

```bash
$ pip2 install -U twine setuptools wheel
$ pip3 install -U setuptools wheel
$ rm -rf dist/ build/
$ python2 setup.py sdist bdist_wheel
# Work around https://bitbucket.org/pypa/wheel/issues/147/bdist_wheel-should-start-by-cleaning-up
$ rm -rf build/
$ python3 setup.py bdist_wheel
$ twine upload dist/*
```

The PyPI package listing will then display three entries for the new release, one with type "Source" and two of type "Python Wheel" (one for each of Python 2 and 3).

Many thanks :-)

Note:
distutils doesn't support the `bdist_wheel` command, so the distutils usage in `setup.py` will also need switching to setuptools.

gvanrossum:
Also update metadata to remove compatibility with 3.5+, and add Ivan
to the list of authors.

This is the version I intend to release as typing 3.6.2, except that I just found there's a problem with the py3 wheel -- it is identical to the py2 wheel. I'll try to debug.

In a sense this fixes #450 (I intend to release from this branch but we'll also have to merge the PR to get setup.py fixed in master).

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -2,20 +2,20 @@
 # coding: utf-8
 
 import sys
-from distutils.core import setup
+from setuptools import setup
 
 if sys.version_info < (2, 7, 0) or (3, 0, 0) <= sys.version_info < (3, 3, 0):
     sys.stderr.write('ERROR: You need Python 2.7 or 3.3+ '
                      'to install the typing package.\n')
     exit(1)
 
-version = '3.6.1'
+version = '3.6.2'
 description = 'Type Hints for Python'
 long_description = '''\
 Typing -- Type Hints for Python
 
 This is a backport of the standard library typing module to Python
-versions older than 3.6.
+versions older than 3.5.
 
 Typing defines a standard notation for Python function and variable
 type annotations. The notation can be used for documenting code in a
@@ -35,8 +35,6 @@ classifiers = [
     'Programming Language :: Python :: 2.7',
     'Programming Language :: Python :: 3.3',
     'Programming Language :: Python :: 3.4',
-    'Programming Language :: Python :: 3.5',
-    'Programming Language :: Python :: 3.6',
     'Topic :: Software Development',
 ]
 
@@ -44,7 +42,7 @@ setup(name='typing',
       version=version,
       description=description,
       long_description=long_description,
-      author='Guido van Rossum, Jukka Lehtosalo, Łukasz Langa',
+      author='Guido van Rossum, Jukka Lehtosalo, Łukasz Langa, Ivan Levkivskyi',
       author_email='jukka.lehtosalo@iki.fi',
       url='https://docs.python.org/3/library/typing.html',
       license='PSF',


gvanrossum:
I didn't find a root cause (though I suspect setuptools is being overly clever), but I found a work-around. It's unsatisfactory, but I managed to create wheels that seem to work by renaming the dist directory after each setup command, like this:
```sh
rm -rf dist
python2 setup.py bdist_wheel
mv dist dist2
python3 setup.py bdist_wheel
mv dist dist3
python3 setup.py sdist
```
After this I can move the wheels back into dist and run twine, but I'll hold off until someone has reviewed this procedure.

Also the source dist created by the last step has way more files than the 3.6.1 source dist (I suspect setuptools here too). I propose to manually prune the tarball so it contains these files only:
```
typing-3.6.2/
typing-3.6.2/LICENSE
typing-3.6.2/PKG-INFO
typing-3.6.2/python2/
typing-3.6.2/README.rst
typing-3.6.2/setup.cfg
typing-3.6.2/setup.py
typing-3.6.2/src/
typing-3.6.2/src/mod_generics_cache.py
typing-3.6.2/src/test_typing.py
typing-3.6.2/src/typing.py
typing-3.6.2/python2/mod_generics_cache.py
typing-3.6.2/python2/test_typing.py
typing-3.6.2/python2/typing.py
```

graingert:
@gvanrossum hmm, this is odd. It's probably weird manifest.in stuff.

I'd probably recommend creating a universal wheel that picks module at runtime.

ilevkivskyi:
> I'd probably recommend creating a universal wheel that picks module at runtime.

I had the same idea. At least IIRC I have seen other libs on PyPI do this.

graingert:
> This is the version I intend to release as typing 3.6.2, except that I just found there's a problem with the py3 wheel -- it is identical to the py2 wheel. I'll try to debug.


I get two different wheels and an sdist when I run:

```sh
workon some3venv
rm -rf dist
python3 setup.py bdist_wheel
workon some2virtualenv
python2 setup.py bdist_wheel
python2 setup.py sdist
```

[dist.tar.gz](https://github.com/python/typing/files/1169704/dist.tar.gz)

graingert:
@gvanrossum can you repeat your problem in docker/vagrent/on travis/ ?

gvanrossum:
I'm on OS X. All I know is that the sequence you use creates two identical
wheels, even in a clean checkout, both based on the first run. There much
be some caching going on. Maybe I'll just release the dist you just
attached, since it looks fine to me...

graingert:
Try using Travis PyPI deploy

gvanrossum:
> Try using Travis PyPI deploy

I'm sorry I have no idea what that means (I'm not much of a Linux admin or devops person).

Anyways I'm sure the problem is due to some setup on my system. I have to focus on other stuff first today.

graingert:
@gvanrossum create a new user on pypi.org eg "typing-bot" and give it maintainer access to `typing` on PyPI.

then run through the steps here:

https://docs.travis-ci.com/user/deployment/pypi/

gvanrossum:
I have no intention to let Travis install anything before I've built and tested the wheels.

graingert:
@gvanrossum you can configure it to point at the testing PyPI service, then promote the wheels

gvanrossum:
Would you want to be our release manager?

edmorley:
> All I know is that the sequence you use creates two identical wheels, even in a clean checkout, both based on the first run. There much be some caching going on.

Ah this appears to be due to:
https://bitbucket.org/pypa/wheel/issues/147/bdist_wheel-should-start-by-cleaning-up

Adding an `rm -rf build/` between the two `pythonXX setup.py bdist_wheel` calls will resolve that.

gvanrossum:
I could swear I had tried that, but apparently not, because it works. I will do the release now.

gvanrossum:
OK, done. Someone please verify: https://pypi.python.org/pypi/typing/3.6.2

gvanrossum:
@ilevkivskyi When you approve of the release, can you merge this PR?

ilevkivskyi:
Thanks! Everything works perfectly.

edmorley:
Ah I guess the README section for releases needs updating too?
https://github.com/python/typing#workflow-for-pypi-releases

ilevkivskyi:
@edmorley Yes, would you like to make a PR?

PR #460, opened 2017-08-14
Make sure copy and deepcopy are returning same class

mitar:
This should hold on Python 3.3 and newer (it does not hold on 3.2 and 2.7).

From [documentation](https://docs.python.org/3/library/copy.html):

>  It does “copy” functions and classes (shallow and deeply), by returning the original object unchanged; this is compatible with the way these are treated by the pickle module.

### DIFF
--- a/src/test_typing.py
+++ b/src/test_typing.py
@@ -1069,6 +1069,9 @@ class GenericTests(BaseTestCase):
         for t in things + [Any]:
             self.assertEqual(t, copy(t))
             self.assertEqual(t, deepcopy(t))
+            if sys.version_info >= (3, 3):
+                self.assertTrue(t is copy(t))
+                self.assertTrue(t is deepcopy(t))
 
     def test_weakref_all(self):
         T = TypeVar('T')
--- a/src/test_typing.py
+++ b/src/test_typing.py
@@ -1071,7 +1071,7 @@ class GenericTests(BaseTestCase):
             self.assertEqual(t, deepcopy(t))
             if sys.version_info >= (3, 3):
                 # From copy module documentation:
-                # It does “copy” functions and classes (shallow and deeply), by returning
+                # It does "copy" functions and classes (shallow and deeply), by returning
                 # the original object unchanged; this is compatible with the way these
                 # are treated by the pickle module.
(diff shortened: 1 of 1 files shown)


the-knights-who-say-ni:

Unfortunately we couldn't find an account corresponding to your GitHub username on [bugs.python.org](http://bugs.python.org/) (b.p.o) to verify you have signed the CLA (this might be simply due to a missing "GitHub Name" entry in your b.p.o account settings). This is necessary for legal reasons before we can look at your contribution. Please follow [the steps outlined in the CPython devguide](https://devguide.python.org/pullrequest/#licensing) to rectify this issue.


ilevkivskyi:
@ethanhs could you please take a look at the test failure? May this be fixed by something like https://github.com/python/mypy/pull/3777?

ethanhs:
@ilevkivskyi yes, you are right. Opened #462.

src/test_typing.py
```diff
@@ -1069,6 +1069,9 @@ class Node(Generic[T]): ...
         for t in things + [Any]:
             self.assertEqual(t, copy(t))
             self.assertEqual(t, deepcopy(t))
+            if sys.version_info >= (3, 3):
```
ilevkivskyi: Could you please add a comment here explaining why we need this?

mitar:
Done.

src/test_typing.py
```diff
@@ -1070,6 +1070,10 @@ class Node(Generic[T]): ...
             self.assertEqual(t, copy(t))
             self.assertEqual(t, deepcopy(t))
             if sys.version_info >= (3, 3):
+                # From copy module documentation:
+                # It does “copy” functions and classes (shallow and deeply), by returning
```
ilevkivskyi: Thanks! Just one more thing: we generally prefer not to use non-ASCII characters in stdlib (and this repo is upstream for stdlib). Here you copied quotes from the HTML docs that are non-ASCII, could you please switch to ASCII quotes?

mitar: Good catch! I will fix this now.

mitar:
Done.

PR #461, opened 2017-08-14
Document the workflow for publishing wheels

edmorley:
This is based roughly on the steps used in #453.

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -156,7 +156,16 @@ Workflow for PyPI releases
 
 * Update the version number in ``setup.py``.
 
-* Build a source distribution. Install it locally and test (if you
-  were using ``tox``, you already tested source distributions).
+* Build the source and wheel distributions:
 
-* Run ``twine upload dist/typing-3.x.y.tar.gz``.
+  * ``pip3 install -U setuptools wheel``
+  * ``pip2 install -U setuptools wheel``
+  * ``rm -rf dist/ build/``
+  * ``python3 setup.py sdist bdist_wheel``
+  * ``rm -rf build/`` (Works around `a Wheel bug <https://bitbucket.org/pypa/wheel/issues/147/bdist_wheel-should-start-by-cleaning-up>`_)
+  * ``python2 setup.py bdist_wheel``
+
+* Install the built distributions locally and test (if you
+  were using ``tox``, you already tested the source distribution).
+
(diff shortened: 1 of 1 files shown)


ilevkivskyi:
@edmorley Could you please sign the CLA (and don't forget to mention your GitHub account on b.p.o.)

edmorley:
Hi! Signing the CLA means providing personal details that I would rather not do if possible. The docs state that the CLA is only required for non-trivial changes, and since this only tweaks the README I think it could count as trivial?

brettcannon:
Since this repository isn't under my purview I don't feel comfortable making a call, but since there isn't really any original thought going into the change (that's not meant to sound insulting, @edmorley , but it's what defines whether a signed CLA is necessary), it should be okay.

And what part of the CLA bothers you, @edmorley ? Is it providing your address to the PSF?

gvanrossum:
I think we should draw a hard line on the CLA. Open source is not conducted anonymously, that's not how you build trust in software that is widely used. Ed, the PSF will treat your information confidentially. (You may use the paper version of the CLA and mail or fax it in if you prefer.) If you still don't feel comfortable, that's fine too, we'll close the PR. Maybe someone else will come up with better instructions.

edmorley:
> And what part of the CLA bothers you, @edmorley ? Is it providing your address to the PSF?

It was a combination of having to provide the address, allow third party cookies, whitelist a handful of adobe domains in NoScript etc. None of which were hard blockers, it just seemed easier to ask for the (documented) exception for trivial changes than to proceed at that point.

However I'm fine with signing if you'd prefer - and have just done so. 

> I think we should draw a hard line on the CLA. Open source is not conducted anonymously, that's not how you build trust in software that is widely used.

I agree that's a valid position (if only since it removes the overhead of trying to decide what constitutes a trival change) - but it would be good for the docs to reflect that. If they had, I would have just signed the CLA from the outset. Once my CLA is processed, I'm happy to open a PR to fix the docs too (though feel free to do so yourself in the meantime if that's easier) :-)

brettcannon:
@edmorley Fixed the devguide: https://github.com/python/devguide/commit/538b386d876be3cf6f939468c616b7e65a275ac7

edmorley:
> @edmorley Fixed the devguide: python/devguide@538b386

Many thanks!

PR #462, opened 2017-08-15
Update pytest and pytest-xdist versions

ethanhs:
This solves issues with pytest-xdist requiring pytest>=3.0. This is the same change as in https://github.com/python/mypy/pull/3777


ilevkivskyi:
Thanks!

PR #463, opened 2017-08-15
Allow tuples to be used with isinstance().

bartfeenstra:
While trying to use PyContracts in my code, I stumbled upon the problem that it uses `isinstance()` internally, but tuples' `__instancecheck()` implementation explicitly states that is not supported. I could not find information about this, so decided to submit a PR after manual testing worked locally.

### DIFF
--- a/src/test_typing.py
+++ b/src/test_typing.py
@@ -395,10 +395,37 @@ class TupleTests(BaseTestCase):
             pass
         self.assertTrue(issubclass(MyTuple, Tuple))
 
+    @skipUnless(sys.version_info < (3, 7), 'Tuples could not be used with '
+                                           'isinstance() until 3.7.')
     def test_tuple_instance_type_error(self):
         with self.assertRaises(TypeError):
             isinstance((0, 0), Tuple[int, int])
         self.assertIsInstance((0, 0), Tuple)
+
+    @skipUnless(sys.version_info >= (3, 7), 'Tuples can be used with '
+                                            'isinstance() since 3.7.')
+    def test_tuple_instance_type(self):
+        # Unparameterized tuples.
+        self.assertIsInstance((), Tuple)
+        self.assertIsInstance((1, 3, 3, 7), Tuple)
+
+        # Instances of parameterized tuples.
+        self.assertIsInstance((3), Tuple[int])
+        self.assertIsInstance(('foo', 'bar'), Tuple[str, Any])
+
+        # Not tuples at all.
+        self.assertNotIsInstance('foo', Tuple)
+        self.assertNotIsInstance(123, Tuple)
+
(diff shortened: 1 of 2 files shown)


the-knights-who-say-ni:

Unfortunately we couldn't find an account corresponding to your GitHub username on [bugs.python.org](http://bugs.python.org/) (b.p.o) to verify you have signed the CLA (this might be simply due to a missing "GitHub Name" entry in your b.p.o account settings). This is necessary for legal reasons before we can look at your contribution. Please follow [the steps outlined in the CPython devguide](https://devguide.python.org/pullrequest/#licensing) to rectify this issue.


ilevkivskyi:
Thanks!
A better idea (and a preferred strategy here) would be to first open an issue. The problem is that this is _intentional_ behaviour, so that it is unfortunate to spend your time on something that most likely will be not merged. Sorry for bad news!

ilevkivskyi:
By the way, maybe this is something that could be done in PyContracts by overriding the default ``isinstance()`` and ``issubclass()`` behaviour?

bartfeenstra:
Thanks for the quick response! I already had the code locally, and was wondering if it would break any tests, hence the PR (as well as the demo factor). I'll open an issue first next time :)

bartfeenstra:
By the way, would you be able to show me where I can find *why* this behavior is intentional?

habnabit:
>The problem is that this is intentional behaviour

I can't find any reference to this intent in PEP-484 or the `typing` module documentation. Perhaps they should be amended?

ilevkivskyi:
@bartfeenstra 
> By the way, would you be able to show me where I can find why this behavior is intentional?

There were several (quite lengthy) discussions on issue tracker and on python-dev and python-ideas mailing lists. The summary is that it is:
a) hard to implement (note that you will need to add ``__subclasscheck__`` to _all_ other types and special forms),
b) sometimes ambiguous,
c) conceptually misleading (types belong to expressions/AST nodes, not to runtime objects),
d) hard to maintain.

@habnabit
The PEP currently says
> ... All of these (except for unparameterized generics) will raise ``TypeError`` if appear in ``isinstance`` or ``issubclass``.

If you want to improve this wording, you could submit a PR to python/peps repo.

habnabit:
@ilevkivskyi as someone who didn't look too much into this until hearing about this issue, the wording `Tuples cannot be used with isinstance()` in the error seemed to imply that this is specific to `Tuple` and not the common case for `typing` constructs. So, when reading through the available documentation, I focused much more on the `Tuple` sections than any information that was applied across the whole module.

gvanrossum:
I think this issue was accidentally left open.

Also, "approvals" by random strangers are pointless.

PR #464, opened 2017-08-19
Protocols in typing extensions

ilevkivskyi:
This PR is essentially an adaptation of #417, here are some comments:
* There is only infrastructure for user-defined protocols, all existing ABCs like ``Iterable`` etc. will be kept nominal for some time.
* I tried to support all versions of ``typing``, but it looks like 3.5.0 and 3.5.1 are too hard.
* There is some code duplication and recourse to private API, I tried to minimize this, but ``Protocol`` really needs to be special in many aspects (like ``Generic``).
* The implementation is on conservative/safe side with respect to structural subtyping, i.e. attributes should be actually _defined_, not just _declared_ in a subclass, see examples in the description of #417 

@Michael0x2a @vlasovskikh I will be grateful for a code review and/or playing with this and commenting whether the behaviour matches your expectations.

cc @JukkaL @ambv @gvanrossum

### DIFF
--- a/typing_extensions/src_py2/test_typing_extensions.py
+++ b/typing_extensions/src_py2/test_typing_extensions.py
@@ -474,6 +474,14 @@ class ProtocolTests(BaseTestCase):
         self.assertIsSubclass(OKClass, C)
         self.assertNotIsSubclass(BadClass, C)
 
+    def test_issubclass_fails_correctly(self):
+        @runtime
+        class P(Protocol):
+            x = 1
+        class C: pass
+        with self.assertRaises(TypeError):
+            issubclass(C(), P)
+
     def test_defining_generic_protocols(self):
         T = typing.TypeVar('T')
         S = typing.TypeVar('S')
@@ -494,6 +502,35 @@ class ProtocolTests(BaseTestCase):
             PR[int, ClassVar]
         class C(PR[int, T]): pass
         self.assertIsInstance(C[str](), C)
+
+    def test_defining_generic_protocols_old_style(self):
+        T = typing.TypeVar('T')
+        S = typing.TypeVar('S')
+        @runtime
+        class PR(Protocol, typing.Generic[T, S]):
+            def meth(self): pass
+        class P(PR[int, str], Protocol):
+            y = 1
+        self.assertIsSubclass(PR[int, str], PR)
+        self.assertIsSubclass(P, PR)
+        with self.assertRaises(TypeError):
+            PR[int]
+        with self.assertRaises(TypeError):
+            PR[int, 1]
+        class P1(Protocol, typing.Generic[T]):
+            def bar(self, x): pass
+        class P2(typing.Generic[T], Protocol):
+            def bar(self, x): pass
+        @runtime
+        class PSub(P1[str], Protocol):
+            x = 1
+        class Test(object):
+            x = 1
+            def bar(self, x):
+                return x
+        self.assertIsInstance(Test(), PSub)
+        with self.assertRaises(TypeError):
+            PR[int, ClassVar]
 
     def test_init_called(self):
         T = typing.TypeVar('T')
@@ -572,6 +609,12 @@ class ProtocolTests(BaseTestCase):
         with self.assertRaises(TypeError):
             @runtime
             class C(object): pass
+        class Proto(Protocol):
+            x = 1
+        with self.assertRaises(TypeError):
+            @runtime
+            class Concrete(Proto):
+                pass
 
     def test_protocols_pickleable(self):
         global P, CP  # pickle wants to reference the class by name
--- a/typing_extensions/src_py2/test_typing_extensions.py
+++ b/typing_extensions/src_py2/test_typing_extensions.py
@@ -344,16 +344,16 @@ class ProtocolTests(BaseTestCase):
         T = typing.TypeVar('T')
         @runtime
         class P(Protocol):
-            x = 1
+            def x(self): pass
         @runtime
         class PG(Protocol[T]):
-            x = 1
+            def x(self): pass
         class BadP(Protocol):
-            x = 1
+            def x(self): pass
         class BadPG(Protocol[T]):
-            x = 1
+            def x(self): pass
         class C(object):
-            x = 1
+            def x(self): pass
         self.assertIsSubclass(C, P)
         self.assertIsSubclass(C, PG)
         self.assertIsSubclass(BadP, PG)
@@ -372,6 +372,29 @@ class ProtocolTests(BaseTestCase):
             issubclass(P, PG[T])
         with self.assertRaises(TypeError):
             issubclass(PG, PG[int])
+
+    def test_protocols_issubclass_non_callable(self):
+        class C(object):
+            x = 1
+        @runtime
+        class PNonCall(Protocol):
+            x = 1
+        with self.assertRaises(TypeError):
+            issubclass(C, PNonCall)
+        self.assertIsInstance(C(), PNonCall)
+        PNonCall.register(C)
+        with self.assertRaises(TypeError):
+            issubclass(C, PNonCall)
+        self.assertIsInstance(C(), PNonCall)
+        # check that non-protocol subclasses are not affected
+        class D(PNonCall): pass
+        self.assertNotIsSubclass(C, D)
+        self.assertNotIsInstance(C(), D)
+        D.register(C)
+        self.assertIsSubclass(C, D)
+        self.assertIsInstance(C(), D)
+        with self.assertRaises(TypeError):
+            issubclass(D, PNonCall)
 
     def test_protocols_isinstance(self):
         T = typing.TypeVar('T')
@@ -425,12 +448,26 @@ class ProtocolTests(BaseTestCase):
         self.assertIsInstance(C(), P)
         self.assertIsInstance(C(), D)
 
-    def test_none_blocks_implementation(self):
+    def test_none_on_non_callable_doesnt_block_implementation(self):
         @runtime
         class P(Protocol):
             x = 1
         class A(object):
             x = 1
+        class B(A):
+            x = None
+        class C(object):
+            def __init__(self):
+                self.x = None
+        self.assertIsInstance(B(), P)
+        self.assertIsInstance(C(), P)
+
+    def test_none_on_callable_blocks_implementation(self):
+        @runtime
+        class P(Protocol):
+            def x(self): pass
+        class A(object):
+            def x(self): pass
         class B(A):
             x = None
         class C(object):
@@ -622,6 +659,20 @@ class ProtocolTests(BaseTestCase):
             x = None  # type: int
         class B(object): pass
         self.assertNotIsInstance(B(), P)
+        class C(object):
+            x = 1
+        class D(object):
+            x = None
+        self.assertIsInstance(C(), P)
+        self.assertIsInstance(D(), P)
+        class CI(object):
+            def __init__(self):
+                self.x = 1
+        class DI(object):
+            def __init__(self):
+                self.x = None
+        self.assertIsInstance(C(), P)
+        self.assertIsInstance(D(), P)
 
     def test_protocols_pickleable(self):
         global P, CP  # pickle wants to reference the class by name
--- a/typing_extensions/src_py2/test_typing_extensions.py
+++ b/typing_extensions/src_py2/test_typing_extensions.py
@@ -616,6 +616,13 @@ class ProtocolTests(BaseTestCase):
             class Concrete(Proto):
                 pass
 
+    def test_none_treated_correctly(self):
+        @runtime
+        class P(Protocol):
+            x = None  # type: int
+        class B(object): pass
+        self.assertNotIsInstance(B(), P)
+
     def test_protocols_pickleable(self):
         global P, CP  # pickle wants to reference the class by name
         T = typing.TypeVar('T')
--- a/typing_extensions/src_py2/test_typing_extensions.py
+++ b/typing_extensions/src_py2/test_typing_extensions.py
@@ -3,11 +3,12 @@ import os
 import abc
 import contextlib
 import collections
+import pickle
 from unittest import TestCase, main, skipUnless
 
 from typing_extensions import NoReturn, ClassVar
 from typing_extensions import ContextManager, Counter, Deque, DefaultDict
-from typing_extensions import NewType, overload
+from typing_extensions import NewType, overload, Protocol, runtime
 import typing
 import typing_extensions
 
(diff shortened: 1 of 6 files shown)


Michael0x2a:
@ilevkivskyi -- idk how much I'll be able to contribute here, but I can definitely take a look sometime during the next day or two!

ilevkivskyi:
@Michael0x2a 
> I can definitely take a look sometime during the next day or two!

Great!

typing_extensions/src_py2/typing_extensions.py
```diff
@@ -89,3 +95,259 @@ def __subclasshook__(cls, C):
...
+            gvars = None
+            for base in bases:
+                if base is Generic:
+                    raise TypeError("Cannot inherit from plain Generic")
+                if (isinstance(base, GenericMeta) and
+                        base.__origin__ in (Generic, Protocol)):
+                    if gvars is not None:
+                        raise TypeError(
+                            "Cannot inherit from Generic[...] or"
+                            " Protocol[...] multiple types.")
```
JelleZijlstra: Do you mean "multiple times"?

typing_extensions/src_py3/typing_extensions.py
```diff
@@ -620,3 +665,269 @@ def new_type(x):
...
+                     'Iterable', 'Iterator', 'AsyncIterable', 'AsyncIterator',
+                     'Hashable', 'Sized', 'Container', 'Collection', 'Reversible',
+                     'Sequence', 'MutableSequence', 'Mapping', 'MutableMapping',
+                     'AbstractContextManager', 'ContextManager',
+                     'AbstractAsyncContextManager', 'AsyncContextManager',) and
+            cls.__module__ in ('collections.abc', 'typing', 'contextlib',
+                               '_abcoll', 'abc'))
+
+
+class ProtocolMeta(GenericMeta):
```
JelleZijlstra: Should this be private?

typing_extensions/src_py3/typing_extensions.py
```diff
@@ -620,3 +665,269 @@ def new_type(x):
...
+            gvars = None
+            for base in bases:
+                if base is Generic:
+                    raise TypeError("Cannot inherit from plain Generic")
+                if (isinstance(base, GenericMeta) and
+                        base.__origin__ in (Generic, Protocol)):
+                    if gvars is not None:
+                        raise TypeError(
+                            "Cannot inherit from Generic[...] or"
+                            " Protocol[...] multiple types.")
```
JelleZijlstra: "multiple times" (I think)

typing_extensions/src_py3/typing_extensions.py
ilevkivskyi: Probably yes, this makes sense not to expose it (I think we can't give guarantees about ``ProtocolMeta`` anyway). I will update now.

typing_extensions/src_py2/typing_extensions.py
ilevkivskyi: I just use the same wording as in ``typing`` for consistency, but I actually agree that "multiple times" is more straightforward.

gvanrossum:
Would it really be too hard to support typing.py 3.5.1?

ilevkivskyi:
@gvanrossum 
> Would it really be too hard to support typing.py 3.5.1?

Unfortunately yes, as you remember, generics were completely redesigned in https://github.com/python/typing/pull/195 so that basically it would be easier to just have a completely separate implementation of ``_ProtocolMeta``, ``Protocol``, etc., under a huge ``if sys.version_info[:3] == (3, 5, 1): ...``. This is not impossible, but the main problem is that I already forgot how generics worked at the time of 3.5.1, so it will take time. A possible compromise would be to implement a simplified version of ``Protocol``, so that this will be valid:
```python
class Proto(Protocol, Generic[T]):
    ...
```
but this shorthand will error in 3.5.1
```python
class Proto(Protocol[T]):
    ...
```
This will be very easy to implement.

gvanrossum:
As parts of Dropbox are still stuck on Python 3.5.1, I think the compromise version would be great to have.

ilevkivskyi:
@gvanrossum OK, I will do this (but most probably not today or tomorrow).

gvanrossum:
No hurry. When you do, will you also add a Travis-CI check to run the tests under 3.5.1?

ilevkivskyi:
> When you do, will you also add a Travis-CI check to run the tests under 3.5.1?

The checks for ``typing_extensions`` are already run on all versions (including micro), in the current version of the PR I just skip the checks on 3.5.0. and on 3.5.1.

ilevkivskyi:
(Actually _all_ tests, not only ``typing_extensions`` generally run on all micro versions, I just checked in ``.travis.yml``.)

gvanrossum:
Ah, great. (Though maybe we should add 3.6.2 now that it's out.)

Michael0x2a:
@ilevkivskyi:

Note: I haven't really been paying close attention to Protocol-related discussion, so I apologize in advance if anything I'm mentioning before has already been brought up somewhere else.

In any case, I found the following bugs/inconsistencies (I'm running Python 3.6.1):

1. Attempting to define the following class results in an `AttributeError: type object 'Generic' has no attribute '_gorg'` error message at runtime.

    ```python
    class P1(Protocol, Generic[T]):
        def bar(self, x: T) -> str: ...
    ```

    Doing `class Foo(List[str], Protocol): ...` also resulted in the same kind of error.

    A little more broadly, I think it would be useful to add tests that intermix `Protocol`, `Generic`, and generic types in interesting ways.

2. Doing `isinstance(None, Empty)` (where "Empty" is the empty protocol as it's defined in your tests) behaves inconsistently. At runtime, that expression always evaluates to true, but when running mypy, that expression evaluates to false when strict-optional mode is enabled.

    I tested this with the following program:

    ```python
    @runtime
    class Empty(Protocol): pass

    x: str = "foo"
    if isinstance(None, Empty):
        # Raises type error only if strict-optional is disabled 
        # (binder skips this branch when strict optional is enabled
        x = 3    

        # Always prints at runtime
        print("bar")
    ```

    That said, `@runtime` isn't supposed to be fully sound anyways so maybe this is ok?

3. The isinstance checks seem a little too flexible. For example:

    ```python
    @runtime
    class P2(Protocol):
        def foo(self, x: int) -> int: ...

    class Concrete2:
        def foo(self, x: int) -> int: ...

    print(issubclass(Concrete2, P2))  # prints 'True', as expected
    print(isinstance(Concrete2, P2))  # also prints 'True'??
    ```

    I found the behavior of the last line to be surprising: I would have expected that to report 'false'.

4. Using the same definitions as above -- doing `issubclass(Concrete(), P2)` results in the following error: `AttributeError: 'Concrete' object has no attribute '__mro__'`.

5. This is probably overkill, but I think it might be nice if you checked that adding `@runtime` to a function and to a class that inherits a protocol (but isn't a protocol itself) both fail.

6. I was able to define the following class without any runtime errors: `class Bar(List[Protocol]): ...`. I was also able to define functions that used `Protocol` for the parameter or return types. Idk if we can do anything about the latter, but it might be worth adding the same checks `ClassVar` uses in the former case for consistency (and adding related tests). 

    I guess we would need to wait until `Protocol` is formally added to the typing module first in order before we can fix this, though?

   (A related, but off-topic thought: the other approach could be to remove the checks for misuse of ClassVar: since static type checkers can easily catch those sorts of things, maybe it's ok to let people get away with some silliness at runtime?)

You also asked if the behavior matched my expectations, so here are some notes related to that. I don't think these are really in-scope for this pull request though, so I can open issues on either the mypy or typing issue trackers as appropriate if you want.

1. The `@runtime` decorator -- I think calling it `@runtime` is potentially too misleading. This is partly because I think it could fool people into thinking that the decorator enables the `typing` module to do full-fledged runtime type analysis. However, since mypy and the runtime error message they'll get makes it pretty clear this isn't the case, I guess this is only a minor UX qualm. 

    The other reason is because I think it can mislead people into thinking that `@runtime` is more powerful then it is even when working with specifically protocols. For example, consider:

    ```python
    @runtime
    class P3(Protocol):
        def foo(self, x: int) -> int: ...

    class Concrete3:
        def foo(self, x: str) -> str: return x

    print(isinstance(Concrete3(), P3))
    ```

    I know that this behavior is described in the documentation (and it made sense in retrospect), but even after reading it, I forgot and thought that the last line would print `false` instead of `true`.

    As a result, I think it might be worth renaming `@runtime` to a scarier, longer name to discourage these sorts of casual accidents? Maybe something like `@enable_protocol_runtime_structural_checks` or `@enable_unsafe_protocol_runtime`? Both names are probably too long, though.

    Or alternatively, it might be worth modifying the documentation to directly warn people that `@runtime` is unsound and that they should think carefully before using it? Basically, taking the blue box that's already there and making it scarier?

2. It wasn't fully clear to me how the `NoReturn` type is supposed to interact with isinstance checks against the Empty protocol in general. Currently, doing any kind of variation of `issubclass(NoReturn, Empty)` (e.g. swapping `issubclass` with `isinstance`, swapping the `NoReturn` type with a call to a function that returns `NoReturn`) appears to evaluate at runtime to 'true'.

3. Also maybe out-of-scope for this pull request, but I'm not really comfortable with how subclasses and protocols currently interacts since it can accidentally encourage unsound behavior. For example, the following program type-checks with mypy, even with strict-optional enabled, but crashes at runtime:

    ```python
    class P4(Protocol):
        def foo(self) -> int: ...

    class Bar(P4): pass

    b = Bar()
    x: int = b.foo()
    print(x + 3)  # Typechecks but crashes at runtime!
    ```

    The documentation says (or at least implies) that the way we created `Bar` is more or less another way to make an ABC. In that case, it would be nice if `Bar` behaved as if it were defined using the `abc` module (e.g. as if the methods were annotated with `abc.abstractmethod``) to prevent this sort of thing.

4. I think it would be handy to have more documentation regarding generic protocols in general. I wasn't aware that you could do things like `class Foo(Protocol[T])` until I read through your unit tests, for example. 

5. Along the documentation vein: how are classvars and protocols supposed to interact? When I add a classvar to a protocol definition, is that classvar a part of the protocol? Also, I tried doing:

    ```python
    class P5(Protocol):
        x: ClassVar[int] = ...
        y: int = ...
    ```

    Mypy reported an error for the definition of `x` ("Incompatible types in assignment (expression has type "ellipsis", variable has type "int")") but not for 'y' -- I'm not sure if that's a bug or intentional.

6. Just to confirm, the ability to actually implement functions within a protocol class is intentional, right? In case the user wants to mix classical inheritance + nominal subtyping along with structural subtyping?

ilevkivskyi:
@Michael0x2a Thank you very much for comments and sorry for super-long delay. Here I reply to them, a bit later a will push the commits.
> 1. Attempting to define the following class results in an ``AttributeError: type object 'Generic' has no attribute '_gorg'`` error message at runtime.

Thanks, good catch, will fix this and will add tests as you suggested.
> 2. Doing ``isinstance(None, Empty)`` (where "Empty" is the empty protocol as it's defined in your tests) behaves inconsistently.

Yes, but I think in this case this is a problem with mypy. We already have issue https://github.com/python/mypy/issues/3827 to track this. In principle I think there is a good chance to make ``@runtime`` protocols almost 100% safe.
> 3. The isinstance checks seem a little too flexible. For example:

Although it is surprising, it is technically true from the point of view of runtime check, since the runtime check does not care about signatures of methods, only their presence. On the contrary, mypy of course distinguishes classes and instances in static structural subtype checks.
> 4. Using the same definitions as above -- doing ``issubclass(Concrete(), P2)`` results in the following error: ``AttributeError: 'Concrete' object has no attribute '__mro__'``.

I think this should raise a ``TypeError`` as for example ``issubclass(1, int)``. Will fix this.
> 5. This is probably overkill, but I think it might be nice if you checked that adding ``@runtime`` to a function and to a class that inherits a protocol (but isn't a protocol itself) both fail.

There is already such check, I will see if there are enough tests for it:
```python
>>> class C(Protocol):
...     x: int
>>> @runtime
... class D(C):
...     pass
  File "<stdin>", line 2, in <module>
  File "/Users/ivan/typehinting/typing_extensions/src_py3/typing_extensions.py", line 927, in runtime
    ' got %r' % cls)
TypeError: @runtime can be only applied to protocol classes, got __main__.D
```
> 6. I was able to define the following class without any runtime errors: ``class Bar(List[Protocol]): ...`` ... I guess we would need to wait until ``Protocol`` is formally added to the typing module first in order before we can fix this, though?

Yes, it will be much easier to do this in ``typing`` itself, I think I already have all the checks for wrong use of ``Protocol`` same as for ``Generic`` in the ``typing`` PR.
> 1. The ``@runtime`` decorator -- I think calling it ``@runtime`` is potentially too misleading.

This question already appeared, but this is probably the best name proposed so far. I think this is OK, since this mirrors how ABCs like ``Iterable`` behave, for example:
```python
class Test:
    __iter__ = 'Surprise!'
isinstance(Test(), collections.abc.Iterable)  # True
isinstance(Test(), typing.Iterable)  # Also True, since it simply delegates to the above.
```
I haven't heard many complains about this.
> 2. It wasn't fully clear to me how the ``NoReturn`` type is supposed to interact with isinstance checks against the ``Empty`` protocol in general.

I think that it should be ``True``, even for static structural subtype checks it is safe to say that everything is a subtype of an empty protocol. I will open an issue at mypy tracker.
 > 3. Also maybe out-of-scope for this pull request, but it seems like how classes and protocols currently interact can accidentally encourage unsound behavior?

This is not specific to protocols. Exactly the same problem exists with normal classes since mypy does not typecheck empty function bodies. I can see that maybe it is more important for protocols. It would be easy to implement in mypy, if you think it is important, then you can open an issue at mypy tracker.
> 4. I think it would be handy to have more documentation regarding generic protocols in general.

I think mypy already has some [decent docs](http://mypy.readthedocs.io/en/latest/generics.html#generic-protocols), when this will be added to ``typing`` we could discuss this more.
> 5. Along the documentation vein: how are classvars and protocols supposed to interact?

Yes, this is not mentioned in the docs, but mypy tries to carefully verify that not only types, but also "flags" (class vs instance variable, writable vs property, static method or not, etc) for attributes are satisfied in a structural subtype. Concerning the error with ellipsis, protocol attributes without default values should be declared like this (no r.h.s.):
```python
class P5(Protocol):
    x: ClassVar[int]
    y: int
```
The ellipsis is used in tub files, to indicate that there is a default value. Therefore the fact that mypy does not show an error is probably a bug, I think assigning ellipses should be allowed _only_ in stubs.
> 6. Just to confirm, the ability to actually implement functions within a protocol class is intentional, right?

Yes, the user might want to provide a default implementation (just like with ABCs), for example:
```python
class Proto(Protocol):
    @abstractmethod
    def meth(self, arg: int) -> str:
        return str(arg + 1)
class Concrete(Proto):
    def meth(arg: int) -> str:
        basic = super().meth()
        return f'Fancy {basic}'
```

typing_extensions/src_py2/typing_extensions.py
```diff
@@ -89,3 +95,259 @@ def __subclasshook__(cls, C):
...
+                              args=args,
+                              origin=self,
+                              extra=self.__extra__,
+                              orig_bases=self.__orig_bases__)
+
+
+class Protocol(object):
+    """Base class for protocol classes. Protocol classes are defined as::
+
+      class Proto(Protocol[T]):
```
JukkaL: I think that the first example should be non-generic protocol. They are likely much more common that generic ones. Also, somebody might imply from this that a protocol must be generic and is used like `Generic`.

typing_extensions/src_py2/typing_extensions.py
```diff
@@ -89,3 +95,259 @@ def __subclasshook__(cls, C):
...
+          def meth(self):
+              # type: () -> int
```
JukkaL: Maybe use `T` in the class body, since this is a generic protocol.

typing_extensions/src_py3/typing_extensions.py
```diff
@@ -5,6 +5,49 @@
...
 
+# These are used by Protocol implementation
+# We use internal typing helpers here, but this significantly reduces
+# code duplication. (Also this is only until Protocol is in typing.)
+from typing import GenericMeta, TypingMeta, Generic, Callable, TypeVar, Tuple
+NO_PROTOCOL = False
+try:
+    from typing import _type_vars, _next_in_mro, _type_check
+except ImportError:
+    NO_PROTOCOL = True
```
JukkaL: Add comment discussing when this happens and what it means.

typing_extensions/src_py3/typing_extensions.py
```diff
@@ -620,3 +665,269 @@ def new_type(x):
 else:
     # Constant that's True when type checking, but False here.
     TYPE_CHECKING = False
+
+
+def _collection_protocol(cls):
+    # Selected set of collections ABCs that are considered protocols.
+    name = cls.__name__
+    return (name in ('ABC', 'Callable', 'Awaitable',
```
JukkaL: See my comment about the similar function in the Python 2 implementation.

typing_extensions/src_py3/typing_extensions.py
```diff
@@ -620,3 +665,269 @@ def new_type(x):
...
+                              extra=self.__extra__,
+                              orig_bases=self.__orig_bases__)
+
+if NO_PROTOCOL:
+    del _ProtocolMeta
+else:
+    class Protocol(metaclass=_ProtocolMeta):
+        """Base class for protocol classes. Protocol classes are defined as::
+
+          class Proto(Protocol[T]):
```
JukkaL: See my comments about the corresponding docstring in the Python 2 implementation. They are also relevant here.

typing_extensions/src_py2/typing_extensions.py
```diff
@@ -89,3 +95,259 @@ def __subclasshook__(cls, C):
...
+            # It cannot support runtime protocol metaclasses
+            return False
+        if issubclass(instance.__class__, self):
+            return True
+        if self._is_protocol:
+            return all(hasattr(instance, attr) and getattr(instance, attr) is not None
+                       for attr in self._get_protocol_attrs())
+        return False
+
+    def __subclasscheck__(self, cls):
```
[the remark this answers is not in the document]
JelleZijlstra: That's an old-style class in Python 2; I'm not sure we should care about support for old-style classes for Protocols in Python 2 (since they are deprecated and a lot of the class machinery works differently for them).

typing_extensions/src_py2/typing_extensions.py
```diff
@@ -89,3 +95,259 @@ def __subclasshook__(cls, C):
                     return True
             return NotImplemented
 
+
+def _collection_protocol(cls):
+    # Selected set of collections ABCs that are considered protocols.
+    name = cls.__name__
+    return (name in ('ABC', 'Callable', 'Awaitable',
```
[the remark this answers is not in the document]
ilevkivskyi: I don't think this will cause problems, but just for consistency I will remove this, thus prohibiting to create subprotocols of ``Iterable`` etc. Anyway, there is a separate ``typing`` PR #417 that is oriented for post-typeshed-PR time.

typing_extensions/src_py2/typing_extensions.py
```diff
@@ -89,3 +95,259 @@ def __subclasshook__(cls, C):
...
+                        if base.__dict__[attr] is None:
+                            return NotImplemented
+                        break
+                else:
+                    return NotImplemented
+            return True
+        if '__subclasshook__' not in cls.__dict__:
+            cls.__subclasshook__ = classmethod(_proto_hook)
+
+    def __instancecheck__(self, instance):
```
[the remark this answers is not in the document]
ilevkivskyi: This maybe a glitch in the logic of treating ``None``. The ``None`` is special, since ``__iter__ = None``, explicitly indicates that a class is _not_ iterable at runtime. I will double check.

typing_extensions/src_py2/typing_extensions.py
```diff
@@ -89,3 +95,259 @@ def __subclasshook__(cls, C):
...
+        # We need this method for situations where attributes are assigned in __init__
+        if isinstance(instance, type):
+            # This looks like a fundamental limitation of Python 2.
+            # It cannot support runtime protocol metaclasses
```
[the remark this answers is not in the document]
ilevkivskyi: I will add comment. On Python 2 classes cannot be correctly inspected as _instances_ of protocols, I will double check if it is true only for old-style classes or for both.

typing_extensions/src_py2/typing_extensions.py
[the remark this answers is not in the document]
ilevkivskyi: Yes, old-stye classes are problematic, I will check if it is possible to do something reasonable with them, if not, then I will add a more meaningful error message like "Old style classes (not inheriting from object) are not supported with protocols".

ilevkivskyi:
@JukkaL Thanks for review! I will push commits soon.

ilevkivskyi:
Thank you @gvanrossum @Michael0x2a and @JukkaL! I think I addressed your comments, here is a summary of changes:
* Protocols are now also supported on 3.5.1, although one needs to write ``class P(Protocol, Generic[T]): ...`` instead of simply ``class P(Protocol[T]): ...`` (this is reflected in the docstring).
* Version 3.6.2 is added to Travis config.
* Fixed the ``_gorg`` bug.
* Fail with a clear ``TypeError`` for ``issubclass(1, Proto)``.
* Added more tests for generic protocols, and one more for ``@runtime``.
* ``None`` is now correctly treated in superclasses.
* Make current implementation consistent with mypy behaviour without the typeshed PR.
* Drop support of old-style classes on Python 2 (with a clear error message).
* Improved docstrings and comments.

typing_extensions/src_py2/typing_extensions.py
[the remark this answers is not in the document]
JukkaL: This doesn't seem quite right. Here is another Python 2 example (Python 3 is similar):

```py
from typing import Optional
from typing_extensions import Protocol, runtime

@runtime
class P(Protocol):
    x = None  # type: Optional[str]

class B(object):
    def __init__(self):  # type: () -> None
        self.x = None  # type: Optional[str]

b = B()
print isinstance(b, P) # False
b.x = ''
print isinstance(b, P) # True
b.x = None
print isinstance(b, P) # False
```

I'd expect the output to be `True`/`True`/`True`. What about only using the special `None` logic for protocol members that are callables in the protocol? They could still be used to represent a missing method. `None` is a valid value for a non-method attribute, so special casing those cases seems wrong to me.

(Overriding with `None` breaks Liskov substitutability but supporting it is reasonable since it's a Python idiom, for better or worse.)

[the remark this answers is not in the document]
ilevkivskyi: > What about only using the special None logic for protocol members that are callables in the protocol?

What do you think about doing the same that mypy currently does -- apply the ``None`` in subclass rule only to dunder attributes? (like ``__iter__``). Maybe even better idea would be to do _exactly_ the same as ABCs in ``collections.abc`` do, i.e. only check whether e.g. ``__iter__`` is set to ``None`` on class, not on instances. What do you think?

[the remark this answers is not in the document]
JukkaL: Just checking for `None` attribute in the class sounds like a good idea, thought it wouldn't be enough by itself. For example, `B` could be defined like:

```py
class B(object):
    x = None  # type: Optional[str]
```

Restriction to dunder attributes seems too ad hoc -- the `None` check is also used for non-dunder methods such as `send` and `throw` in `collections.abc` (https://github.com/python/cpython/blob/master/Lib/_collections_abc.py#L151). Mypy probably shouldn't use it either.

What about doing a `callable()` check on the attribute value defined in the protocol and requiring that the type object of the instance has a `None` value for the corresponding attribute? I think that this would be unlikely to cause problems and would cover all important use cases. It's still possible to construct an example which doesn't work right, but that's acceptable since `@runtime` was never supposed to be 100% correct anyway.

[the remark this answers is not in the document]
ilevkivskyi: > What about doing a callable() check on the attribute value defined in the protocol and requiring that the type object of the instance has a None value for the corresponding attribute?

OK, I will go with this (applying the ``None`` rule only to callable attributes, and check them on the class object). I will also open an issue in mypy, to do the same (instead of the ad-hoc dunder restriction).

typing_extensions/src_py2/test_typing_extensions.py
```diff
@@ -616,6 +616,13 @@ class Proto(Protocol):
             class Concrete(Proto):
                 pass
 
+    def test_none_treated_correctly(self):
+        @runtime
+        class P(Protocol):
+            x = None  # type: int
+        class B(object): pass
```
JukkaL: Do you test the variants where `x` is defined for the instance of `B` (`None` / not `None`) or where `x` has value `None` / not `None` in the body of `B`?

typing_extensions/src_py3/typing_extensions.py
```diff
@@ -898,14 +899,20 @@ class C:
...
 
           func(C())  # Passes static type check
 
         See PEP 544 for details. Protocol classes decorated with @typing_extensions.runtime
         act as simple-minded runtime protocols that checks only the presence of
         given attributes, ignoring their type signatures.
+
+        Protocol classes can be generic, they are defined as::
+
+          class GenProto({bases}):
```
JukkaL: Replace `{bases}` with `Protocol[T]`?

JukkaL: Nevermind, this happens below.

JukkaL:
Did you respond to this comment?

> More generally, I'm not sure if subclass checks are well-defined if a protocol has non-method attributes, since they could well be initialized in` __init__` and not visible in the type object. This could be fixed by not supporting issubclass with protocols, or only supporting it with protocols that only define methods, since it's rare that a method is defined in `__init__` (though it's not completely unheard of).

ilevkivskyi:
@JukkaL 
> Did you respond to this comment?

No, sorry, it was mentally masked by the Python 2 failure above it. This is another point where currently runtime and mypy behavior diverge. I think it makes sense to simply disallow ``issubclass()`` checks with respect to protocols that define non-method attributes. I will fix this and will open an issue to give the corresponding error in mypy. (Either late evening or tomorrow morning, don't have time right now, sorry.)

typing_extensions/src_py2/test_typing_extensions.py
ilevkivskyi: I think I test only few variants/combinations. I will add more tests.

ilevkivskyi:
@JukkaL Thanks for more review comments! I think they are now implemented. Summary:
* Prohibit ``issubclass()`` for protocols that declare non-method members (``isinstance()`` is still OK).
* Apply the ``None`` rule only to callable protocol members (methods).
* Add more tests.

I also opened two issues https://github.com/python/mypy/issues/3938 and https://github.com/python/mypy/issues/3939 to synchronize mypy behaviour. Please check if this PR can be now merged.

gvanrossum:
Everybody, I am going to try and review this tomorrow. I just read through all the discussion (whew!) to recover as much context as I can, hopefully it's still in my head tomorrow morning.

gvanrossum:
(Whoops, hit return too soon. There's more to come.)

typing_extensions/src_py3/typing_extensions.py
```diff
@@ -620,3 +664,300 @@ def new_type(x):
...
+        if orig_bases is None:
+            self.__orig_bases__ = initial_bases
+        elif origin is not None:
+            self._abc_registry = origin._abc_registry
+            self._abc_cache = origin._abc_cache
+        if hasattr(self, '_subs_tree'):
+            self.__tree_hash__ = (hash(self._subs_tree()) if origin else
+                                  super(GenericMeta, self).__hash__())
+        return self
+    if OLD_GENERICS:
```
gvanrossum: Why not just put the `def` inside `if not OLD_GENERICS:`? Just so you can copy/paste without reindenting? That feels a little weird to me.

typing_extensions/src_py3/typing_extensions.py
```diff
@@ -620,3 +664,300 @@ def new_type(x):
...
+                              origin=self,
+                              extra=self.__extra__,
+                              orig_bases=self.__orig_bases__)
+
+    if OLD_GENERICS:
+        del __getitem__
+
+
+if NO_PROTOCOL:
+    del _ProtocolMeta
```
gvanrossum: Perhaps `_ProtocolMeta` should also be inside the `if not NO_PROTOCOL:` block?

typing_extensions/src_py3/typing_extensions.py
```diff
@@ -620,3 +664,300 @@ def new_type(x):
...
+                                "it can be used only as a base class")
+            if OLD_GENERICS:
+                return _generic_new(_next_in_mro(cls), cls, *args, **kwds)
+            return _generic_new(cls.__next_in_mro__, cls, *args, **kwds)
+
+    Protocol.__doc__ = Protocol.__doc__.format(bases="Protocol, Generic[T]" if
+                                               OLD_GENERICS else "Protocol[T]")
+
+
+def runtime(cls):
```
gvanrossum: I still don't like the `@runtime` name. Maybe `@runtime_checkable`? The longer this is the better, to discourage people from using it without understanding the consequences.

typing_extensions/src_py3/typing_extensions.py
```diff
@@ -620,3 +664,300 @@ def new_type(x):
...
+    one-offs in collections.abc such as Hashable.
+    """
+    if not isinstance(cls, _ProtocolMeta) or not cls._is_protocol:
+        raise TypeError('@runtime can be only applied to protocol classes,'
+                        ' got %r' % cls)
+    cls._is_runtime_protocol = True
+    return cls
+
+
+if NO_PROTOCOL:
```
gvanrossum: Again it seems weird to define it and then delete it. (Also maybe the flag could be reverted and be named `HAVE_PROTOCOLS`?)

typing_extensions/src_py3/typing_extensions.py
ilevkivskyi: I just wanted to save some horizontal space. But I see that it is easy to miss these two lines when scrolling through the code. I will probably just put all definitions inside (nested) ``if`` statements.

typing_extensions/src_py3/typing_extensions.py
```diff
@@ -5,6 +5,45 @@
...
+    _tp_cache = lambda x: x
+try:
+    from typing import _TypingEllipsis, _TypingEmpty
+except ImportError:
+    class _TypingEllipsis: pass
+    class _TypingEmpty: pass
+try:
+    from typing import _check_generic
+except ImportError:
+    def _check_generic(cls, parameters):
```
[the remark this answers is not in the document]
ilevkivskyi: This may be just an artefact from initial attempts to allow this on 3.5.1 with minimal changes, I will check now.

ilevkivskyi:
@gvanrossum Thanks for review, I think I have now addressed all your comments.

ilevkivskyi:
@gvanrossum Ah, sorry, I forgot one your comment, about the ``@runtime`` name. Your version ``@runtime_checkable`` actually looks OK. But what do you think about releasing ``typing_extensions`` with ``@runtime`` and see how it will go?

gvanrossum:
I strongly prefer the longer name.

ilevkivskyi:
> I strongly prefer the longer name.

OK, but then we will need to also update mypy and typeshed.

ilevkivskyi:
...and the PEP.

gvanrossum:
So be it, now's the time.

typing_extensions/src_py3/test_typing_extensions.py
```diff
@@ -1188,7 +1188,7 @@ class E:
             self.assertIsInstance(E(), D)
 
 
-if NO_PROTOCOL:
+if not HAVE_PROTOCOLS:
```
gvanrossum: Arguably the entire `class ProtocolTests` should be inside `if HAVE_PROTOCOLS:` instead of this little block.

ilevkivskyi:
TBH, I am quite exhausted after last two weeks, I would prefer to postpone everything which is not urgent. I you are OK releasing ``typing_extensions`` as is, then I will be only happy.

gvanrossum:
Understood. I am okay with releasing now, without more changes. We can debate the `@runtime` decorator name more when we have some experience.

ilevkivskyi:
OK, I just uploaded ``typing_extensions`` on PyPI.

randolf-scholz:
Is the rationale for disallowing `issubclass` still valid, 6 years later? Honestly, it would be pretty useful to have `issubclass` check `ClassVar`-attributes (and ignore non-`ClassVar` attributes).

hauntsaninja:
Yes, still valid. See https://github.com/python/cpython/issues/89138
In general, please avoid commenting on long closed pull requests.

PR #470, opened 2017-09-13
get_type_hints(): find the right globalns for classes and modules

ambv:
This makes the default behavior (without specifying `globalns` manually) more
predictable for users.

Implementation for classes assumes has a `__module__` attribute and that module
is present in `sys.modules`.  It does this recursively for all bases in the
MRO.  For modules, the implementation just uses their `__dict__` directly.

This is backwards compatible, will just raise fewer exceptions in naive user
code.

### DIFF
--- a/src/mod_generics_cache.py
+++ b/src/mod_generics_cache.py
@@ -1,14 +1,21 @@
 """Module for testing the behavior of generics across different modules."""
 
-from typing import TypeVar, Generic
+from typing import TypeVar, Generic, Optional
+
+default_a: Optional['A'] = None
+default_b: Optional['B'] = None
 
 T = TypeVar('T')
 
 
 class A(Generic[T]):
-    pass
+    some_b: 'B'
 
 
 class B(Generic[T]):
     class A(Generic[T]):
         pass
+
+    my_inner_a1: 'B.A'
+    my_inner_a2: A
+    my_outer_a: 'A'  # unless somebody calls get_type_hints with localns=B.__dict__
--- a/src/mod_generics_cache.py
+++ b/src/mod_generics_cache.py
@@ -1,21 +1,53 @@
 """Module for testing the behavior of generics across different modules."""
 
+import sys
+from textwrap import dedent
 from typing import TypeVar, Generic, Optional
 
-default_a: Optional['A'] = None
-default_b: Optional['B'] = None
 
-T = TypeVar('T')
+if sys.version_info[:2] >= (3, 6):
+    exec(dedent("""
+    default_a: Optional['A'] = None
+    default_b: Optional['B'] = None
+
+    T = TypeVar('T')
 
 
-class A(Generic[T]):
-    some_b: 'B'
-
-
-class B(Generic[T]):
     class A(Generic[T]):
-        pass
+        some_b: 'B'
 
-    my_inner_a1: 'B.A'
-    my_inner_a2: A
-    my_outer_a: 'A'  # unless somebody calls get_type_hints with localns=B.__dict__
+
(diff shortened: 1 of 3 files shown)


src/typing.py
```diff
@@ -1494,7 +1495,10 @@ def get_type_hints(obj, globalns=None, localns=None):
     if getattr(obj, '__no_type_check__', None):
         return {}
     if globalns is None:
-        globalns = getattr(obj, '__globals__', {})
+        if isinstance(obj, type):
+            globalns = vars(sys.modules[obj.__module__])
```
ilevkivskyi: While you at it maybe you can fix same way also the logic below when ``__mro__`` is traversed
(I could imagine that base classes are from different modules). Or will it be too hard? If not, I would add also a  test for such scenario.

ambv: Yeah, I have a few ideas more but it's annoying that we don't ship `ann_module.py` et al. but always refer to the one in `test` for Python 3.5+. This makes it hard to iterate on them.

Let me see what I can do.

src/test_typing.py
```diff
@@ -1659,8 +1666,18 @@ def test_get_type_hints_modules(self):
...
 
     @skipUnless(PY36, 'Python 3.6 required')
+    @expectedFailure
+    def test_get_type_hints_modules_forwardref(self):
+        # FIXME: This currently exposes a bug in typing. Cached forward references
+        # don't account for the case where there are multiple types of the same
+        # name coming from different modules in the same program.
+        mgc_hints = {'default_a': Optional[mod_generics_cache.A],
+                     'default_b': Optional[mod_generics_cache.B]}
+        self.assertEqual(gth(mod_generics_cache), mgc_hints)
```
ambv: I will fix this bug in a separate PR when this one lands.

src/typing.py
ambv: Done!

src/mod_generics_cache.py
```diff
@@ -1,14 +1,53 @@
 """Module for testing the behavior of generics across different modules."""
```
ambv: I only added a bunch of variable annotations here but since this wasn't compatible with < 3.5, I had to uglify this module. Now it's actually easier to see what's going on just by looking at the entire file with the "View" button.

src/test_typing.py
ilevkivskyi: > I will fix this bug in a separate PR when this one lands.

OK, just don't forget about it (or open an issue if you want to do this later).

gvanrossum:
Also, will this go into 3.6.3?

ambv:
Yes, that's the point. I'm working on the backport as you read this.

ambv:
Hopefully I'll get to fix the invalid cache of forward refs, too, before 2017-09-18 (PEP 494 3.6.3 RC).

ilevkivskyi:
> Hopefully I'll get to fix the invalid cache of forward refs, too, before 2017-09-18 (PEP 494 3.6.3 RC).

Yes, it would be great if this gets in too.

src/test_typing.py
natgaertner: Hi, I was looking at a different issue and noticed this test code behaving non-deterministically depending on when mod_generics_cache was loaded (specifically this test would expectedly fail when I just ran the test_typing module, but unexpectedly succeed when I ran ./python -m test.)

Fixing the underlying type hint caching is beyond my scope as a new contributor, but it seems like something like this would make this test behave more reliably, if that seems reasonable to y'all: https://github.com/natgaertner/cpython/commit/0cde1032e14befe2cec21060a4dbebb891f51fc6

PR #473, opened 2017-09-17
Update typing_extensions/setup.py

ilevkivskyi:
This fixes a typo, updates required ``typing`` version to the latest one, updates the contact e-mail, and adds a brief mention of ``Protocol`` to the description.

### DIFF
--- a/typing_extensions/setup.py
+++ b/typing_extensions/setup.py
@@ -9,7 +9,7 @@ if sys.version_info < (2, 7, 0) or (3, 0, 0) <= sys.version_info < (3, 3, 0):
                      'to install the typing package.\n')
     exit(1)
 
-version = '3.6.1'
+version = '3.6.2'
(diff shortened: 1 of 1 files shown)


typing_extensions/setup.py
```diff
@@ -53,14 +53,14 @@
 
 install_requires = []
 if sys.version_info < (3, 5):
```
gvanrossum: I wonder if this could be done the way we do it for mypy now?
```
      extras_require = {
          ':python_version < "3.5"': 'typing >= 3.5.3',
      },
```
(Except using 3.6.2 for the typing version.)

IIUC that way you can build universal wheels that do the right thing.

ilevkivskyi:
> I wonder if this could be done the way we do it for mypy now?

I am not an expert in packaging, but I will try this before the next release of ``typing_extensions``.

Issue #484, opened 2017-10-18
typing breaks singledispatch on Python 2


Wilfred:
The following does not work on Python 2:

```python
import typing

from collections import Iterable
from singledispatch import singledispatch

mysum = singledispatch(sum)
mysum.register(Iterable, sum)

print mysum([])
```

Removing the `import typing` fixes the issue. This means that a codebase cannot import typing anywhere without breaking uses of singledispatch.

```
$ python foo.py
Traceback (most recent call last):
  File "foo.py", line 9, in <module>
    print mysum([])
  File "build/bdist.linux-x86_64/egg/singledispatch.py", line 210, in wrapper
  File "build/bdist.linux-x86_64/egg/singledispatch.py", line 191, in dispatch
  File "build/bdist.linux-x86_64/egg/singledispatch.py", line 142, in _find_impl
  File "build/bdist.linux-x86_64/egg/singledispatch.py", line 119, in _compose_mro
  File "build/bdist.linux-x86_64/egg/typing.py", line 1223, in __subclasscheck__
TypeError: Parameterized generics cannot be used with class or instance checks
```

This is [Python bug 28339](https://bugs.python.org/issue28339), which has been fixed in Python 3, but not in the singledispatch library which backports singledispatch to Python 2.

gvanrossum:
I'm confused -- if it should be fixed in the singledispatch backport why
don't you file the bug there?

Wilfred:
Sorry, I was unclear. (I also couldn't find the bugtracker for the singledispatch backport.)

I'm not sure where this should be fixed. As I understand it, typing.py is mutating datastructures in abc.py, so I suspect that's the best place to fix this.

Looking at the relevant code in typing.py:

```python
    def __subclasscheck__(self, cls):
        if self.__origin__ is not None:
            if sys._getframe(1).f_globals['__name__'] not in ['abc', 'functools']:
                raise TypeError("Parameterized generics cannot be used with class "
                                "or instance checks")
            return False
        if self is Generic:
            raise TypeError("Class %r cannot be used with class "
                            "or instance checks" % self)
        return super(GenericMeta, self).__subclasscheck__(cls)
```

Is this only fixing the problem for packages whose name is 'abc' or 'functools'?

I have a kludge that fixes my example script:

```python
import typing

from collections import Iterable

import singledispatch as sd
sd.__name__ = 'functools'

mysum = sd.singledispatch(sum)
mysum.register(Iterable, sum)

print mysum([])
```

[...]

PR #486, opened 2017-10-20
Override subclass check for the singledispatch library

Wilfred:
This library is a Python 2 backport of functools.singledispatch, so it
should be treated the same way as functools.

Partial fix for #484 -- this doesn't help other packages that
experience the same issue.

### DIFF
--- a/python2/typing.py
+++ b/python2/typing.py
@@ -1234,7 +1234,11 @@ class GenericMeta(TypingMeta, abc.ABCMeta):
 
     def __subclasscheck__(self, cls):
         if self.__origin__ is not None:
-            if sys._getframe(1).f_globals['__name__'] not in ['abc', 'functools']:
+            # This should only be modules within the standard
+            # library. singledispatch is the only exception, because
+            # it's a Python 2 backport of functools.singledispatch.
+            if sys._getframe(1).f_globals['__name__'] not in ['abc', 'functools',
+                                                              'singledispatch']:
                 raise TypeError("Parameterized generics cannot be used with class "
                                 "or instance checks")
             return False


the-knights-who-say-ni:

Wilfred:
I've sorted the CLA -- does the bot update its comment automatically?

gvanrossum:
I think there's a manual step as your entry into the database is verified.
I don't see your CLA bit in the bugs.python.org userbase yet (your username
is Wilfred.Hughes right?)

gvanrossum:
Actually, @ilevkivskyi does this look good to you? (It's making allowances for a 3rd party module but it's just a backport of a PY3 stdlib module.)

ilevkivskyi:
TBH, I am not very happy with this. This does not make any harm but it is kind of a "slippery slope": once we include the workaround for one library, others will want the same.

@Wilfred why don't you want to make this fix directly in ``singledispatch``? It should be quite easy (grep for places where it uses ``__subclasses__``).

Wilfred:
@ilevkivskyi yeah, it's not pretty. However, since `singledispatch` is a backport of `functools.singledispatch`, it seems appropriate to me to include `"singledispatch"` too. I don't anticipate this applying to other libraries.

I'm happy to send a patch to singledispatch, but I'd really like to see this patch in typing too.

If I write a lib_foo.py that depends on lib_bar.py, and lib_bar.py uses singledispatch, then adding typing to lib_foo will break lib_bar. With this patch, I can use the latest typing in lib_foo without worrying that all lib_foo users have upgraded lib_bar to use the (future) fixed version of singledispatch.

ilevkivskyi:
@Wilfred OK, but could you please add a comment near this line in code that this is the only non-stdlib exception and _briefly_ explain motivation?

ilevkivskyi:
@Wilfred are you still interested in this?

Wilfred:
@ilevkivskyi thanks, I've added a comment to clarify the motivation. Are you happy with the phrasing?

Wilfred:
Excellent :)

@ilevkivskyi would it be possible to cut a new release of typing? I bumped into this issue today with pylint, which uses singledispatch via astroid.

ilevkivskyi:
I think Guido will do this when he will come back from Canada.

gvanrossum:
Done, see https://github.com/python/typing/issues/532.

Issue #498, opened 2017-11-04
Error when copying instances of Generic subclasses


izquierdo:
Trying to do the following

```python
import copy
from typing import Generic, TypeVar

T = TypeVar('T')

class Test(Generic[T]):
    def __init__(self, t: T) -> None:
        self.t = t

test: Test[int] = Test(5)
copy.copy(test)
```

produces

```
Traceback (most recent call last):
  File "test.py", line 11, in <module>
    copy.copy(test)
  File "/usr/lib/python3.6/copy.py", line 88, in copy
    return copier(x)
TypeError: __copy__() takes 1 positional argument but 2 were given
```

whereas there is no error if not inheriting from `Generic`.

The most similar issue to this one is #458 but that snippet is trying to call `__copy__` directly instead of `copy.copy`. There is also #306 but that one is trying to copy the type instead of an instance, and also switching the above snippet to `copy.deepcopy` does work. Finally, there is #468, it's possible that a fix for this issue would fall within its scope.

ilevkivskyi:
Indeed, this is fixed by #468 (a.k.a. PEP 560), but the fix will be available only in Python 3.7. Therefore, I think it makes sense to also fix it in the backport (the crash also happens on Python 3.4).

hynekcer:
A solution that works:
``` python
class MyClass(Generic[T]):
    __copy__ = None
    ...
```
This is safe:  
1) The value `__copy__ = None` has the same effect for the function copy.copy() like an undefined `__copy__` attribute. A possible problem is if it would overshadow a right `__copy__` attribute.
2) The "Generic" type is the last ancestor of MyClass. (The only possible problem would be if it is used as generic mixin for a class with a `__copy__` method, but it can be easily solved by a `super(...).__copy__()`)

A general hack can be also
```
from typing import Generic

if hasattr(Generic, '__copy__') and sys.version_info[:2] == (3, 5):
    delattr(Generic, '__copy__')
```
It works in some of the first modules of an application before the first copy.

Maybe this is also possible to do in typing_extensions.

PR #502, opened 2017-11-25
Fix copying generic instances in Python 3

ilevkivskyi:
Fixes https://github.com/python/typing/issues/498 on Python 3

Unfortunately, it looks like this can't be (reasonably) fixed on Python 2, unless https://bugs.python.org/issue11480 is first backported to Python 2

### DIFF
--- a/python2/test_typing.py
+++ b/python2/test_typing.py
@@ -1024,6 +1024,30 @@ class GenericTests(BaseTestCase):
             self.assertEqual(t, deepcopy(t))
             self.assertEqual(t, copy(t))
 
+    def test_copy_generic_instances(self):
+        T = TypeVar('T')
+        class C(Generic[T]):
+            def __init__(self, attr):
+                self.attr = attr
+
+        c = C(42)
+        self.assertEqual(copy(c).attr, 42)
+        self.assertEqual(deepcopy(c).attr, 42)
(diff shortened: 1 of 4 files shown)


ilevkivskyi:
Just to clarify, on Python 2 the problem with copying classes is not specific to generics, any class with a metaclass other than ``type`` can't be copied:
```python
>>> class M(type): pass
>>> class C(object):
...     __metaclass__ = M
>>> from copy import copy
>>> copy(C)  # obscure TypeError
```
But the ad-hoc ``__copy__`` on metaclass fix that is currently used clashes with copying _instances_ of generic classes.

gvanrossum:
For Python 2, can'y you register a different copier function with the copy
module?

ilevkivskyi:
If you mean ``copy_reg.dispatch_table``, then no, ``__copy__`` is searched before. But there is an internal API ``copy._copy_dispatch`` (see new commit) that works. If you are fine with this, then I am also fine.

ilevkivskyi:
Oh, wait, but we don't need ``__copy__``. OK, then I will make a cleaner fix now.

ilevkivskyi:
Actually, ``copy_reg`` explicitly says that ``copy_reg is not intended for use with classes``, therefore I think the ``_copy_dispatch`` fix for Python 2 is OK.

PR #504, opened 2017-11-25
Fix protocols in unions (runtime problem)

### LINKED ISSUE #488: Non-`@runtime` protocols cannot be used in a `typing.Union`
Code such as the following:

```python
In [1]: import typing

In [2]: import typing_extensions

In [3]: class P(typing_extensions.Protocol):
   ...:     def f(self, s: str) -> str: ...
   ...:     

In [4]: p_or_callable = typing.Union[P, typing.Callable]
```

Leads to the following exception:

```python
../../../../.pyenv/versions/3.6.2/lib/python3.6/typing.py:682: in inner
    return func(*args, **kwds)
../../../../.pyenv/versions/3.6.2/lib/python3.6/typing.py:803: in __getitem__
    return self.__class__(parameters, origin=self, _root=True)
../../../../.pyenv/versions/3.6.2/lib/python3.6/typing.py:743: in __new__
    parameters = _remove_dups_flatten(parameters)
../../../../.pyenv/versions/3.6.2/lib/python3.6/typing.py:647: in _remove_dups_flatten
    for t2 in all_params - {t1}
../../../../.pyenv/versions/3.6.2/lib/python3.6/typing.py:649: in <genexpr>
    t2.__origin__ is not None)):
../../../../.pyenv/virtualenvs/crucible/3.6.2/lib/python3.6/site-packages/typing_extensions.py:833: in __subclasscheck__
    raise TypeError("Instance and class checks can only be used with"
E   TypeError: Instance and class checks can only be used with @runtime protocols
```

Decorating the `Protocol` subclass with `@runtime` works around the problem.

cc @ilevkivskyi

ilevkivskyi:
Fixes #488 

The fix is as proposed in https://github.com/python/typing/issues/488#issuecomment-346639532 i.e.:
* Don't simplify unions, the only simplification that stays is ``Union[X] is X``.
* Tweak the existing ``sys._getframe`` hack to accommodate it for this special case (protocols in unions) on older versions of ``typing``.

### DIFF
--- a/python2/test_typing.py
+++ b/python2/test_typing.py
@@ -173,6 +173,7 @@ class TypeVarTests(BaseTestCase):
         self.assertNotEqual(X, Y)
         self.assertEqual(Union[X], X)
         self.assertNotEqual(Union[X], Union[X, Y])
+        self.assertEqual(Union[X, X], X)
         self.assertNotEqual(Union[X, int], Union[X])
         self.assertNotEqual(Union[X, int], Union[int])
         self.assertEqual(Union[X, int].__args__, (X, int))
@@ -249,6 +250,14 @@ class UnionTests(BaseTestCase):
         self.assertNotEqual(u2, Any)
         self.assertNotEqual(u3, Any)
 
+    def test_union_object(self):
+        u = Union[object]
+        self.assertEqual(u, object)
+        u = Union[int, object]
+        self.assertEqual(u, object)
+        u = Union[object, int]
+        self.assertEqual(u, object)
+
     def test_unordered(self):
         u1 = Union[int, float]
         u2 = Union[float, int]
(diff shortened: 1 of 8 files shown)


gvanrossum:
Hm, now that I see this I am getting worried. Anyone who is writing a runtime checker that inspects unions (via `__parameters__`) could see their code broken. That worries me given that we're not planning to do much more with the version of typing.py in Python 3.6 anyways -- in 3.7 we'll have a major rewrite. I think I'd rather keep 3.6 close to the original. But the getframe hack can be applied and released separately and will relieve the problem, right? Or is it ineffective?

gvanrossum:
(And sorry I gave different guidance a week ago in the issue.)

ilevkivskyi:
> But the getframe hack can be applied and released separately and will relieve the problem, right? Or is it ineffective?

Yes, it will be still effective. I will revert the ``Union`` part for now. It is only needed to fix interaction between an old ``typing_extensions`` and a new ``typing``. We can reconsider the union simplifications later for Python 3.7.

ilevkivskyi:
OK, I reverted the ``Union`` runtime simplification part (and tests still pass).

ilevkivskyi:
I think it makes to release it after Python 3.6.4 is out and ``typing`` 3.6.4 is out (the current version scheme for ``typing_extensions`` coincides with version scheme for ``typing``).

gvanrossum:
OK.

PR #507, opened 2017-11-30
Fix interaction between generics and __init_subclass__

### LINKED ISSUE #506: GenericMeta conflicts with __init_subclass__
https://bugs.python.org/issue32162

When superclass inherits from Generic, attributes set for a subclass are incorrectly propagated to its superclass.

Without Generic attribute access raises an exception:

```python
    class X:
        def __init_subclass__(cls, **kwargs):
            super().__init_subclass__(**kwargs)
            cls.attr = 42

    class Y(X):
        pass

    Y.attr  # 42
    X.attr  # AttributeError
```

With Generic it does not:

```python
    import typing

    T = typing.TypeVar('T')

    class X(typing.Generic[T]):
        def __init_subclass__(cls, **kwargs):
            super().__init_subclass__(**kwargs)
            cls.attr = 42

    class Y(X):
        pass

    Y.attr  # 42
    X.attr  # 42
```

ilevkivskyi:
Fixes #506 with one caveat.

Note that ``X[int]`` creates an actual new class object which is a subclass of ``X``. Therefore after this code:
```python
class X(Generic[T]):
        def __init_subclass__(cls, **kwargs):
            super().__init_subclass__(**kwargs)
            cls.attr = 42

X.attr =1
X[int]
```
``X.attr == 42``. This is because ``__init_subclass__`` being called on ``X[int]`` will assign ``X[int].attr = 42`` which is equivalent to ``X.attr = 42`` (since current rules are that subscripted generics are proxies for the original class).

This is something hard to fix, unless ``X[int]`` will not be a class object. So that this unfortunate interaction between ``__init_subclass__`` and generics may be another justification for PEP 560.

@gvanrossum I think it makes sense to include this fix (as well as #504) in 3.6.4rc1 which is planned for Monday. What do you think?

### DIFF
--- a/src/test_typing.py
+++ b/src/test_typing.py
@@ -653,7 +653,6 @@ class GenericTests(BaseTestCase):
             pass
         class W(X[int]):
             pass
-        self.assertEqual(X.attr, 1)
         self.assertEqual(Y.attr, 2)
         self.assertEqual(Z.attr, 42)
         self.assertEqual(W.attr, 42)
--- a/src/test_typing.py
+++ b/src/test_typing.py
@@ -37,6 +37,9 @@ except ImportError:
     from test import mod_generics_cache
 
 
+PY36 = sys.version_info[:2] >= (3, 6)
+
+
 class BaseTestCase(TestCase):
 
     def assertIsSubclass(self, cls, class_or_tuple, msg=None):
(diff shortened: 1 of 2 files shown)


gvanrossum:
I'm not sure how valuable this fix is if `X[int]` will destroy the effect, since that is the most common use of generic classes. (I guess you could have all uses in `# type:` comments.)

ilevkivskyi:
> I'm not sure how valuable this fix is if ``X[int]`` will destroy the effect, since that is the most common use of generic classes. (I guess you could have all uses in ``# type:`` comments.)

I think this still makes sense as it fixes the original reported problem. For ``X[int]`` there are several simple workarounds like ``# type:`` comments and "forward references". Finally, with this fix, the remaining ``X[int]`` problem is a bug in _design_, while currently there is also a bug in _implementation_.

@Kentzo could you please comment on whether this fix is useful for you?

Kentzo:
@ilevkivskyi The fix I'm currently using seems to work:

```python
import typing

T = typing.TypeVar('T')

class X(typing.Generic[T]):
    def __init_subclass__(cls, **kwargs):
        super(typing.GenericMeta, cls).__setattr__('_gorg', cls)
        super().__init_subclass__(**kwargs)
        cls.attr = 42


class Y(X[T]):
    pass

print(Y.attr)  # 42
print(X[int].attr)  # 42
print(X.attr)  # AttributeError

X.attr = 1
print(X[int].attr)  # 42
print(X.attr)  # 1
```

Seems to work:

gvanrossum:
OK then feel free to merge!

PR #525, opened 2017-12-28
Check for docstring stripping before modifying Protocol's docstring

### LINKED ISSUE #524: `typing_extensions` incompatible with -OO
If docstrings are stripped (as by the -OO flag), then under Py3, `typing_extensions` will crash at import time when it tries to alter `Protocol.__doc__`:

>```
>    Protocol.__doc__ = Protocol.__doc__.format(bases="Protocol, Generic[T]" if
>                                               OLD_GENERICS else "Protocol[T]")
>```

rkr-at-dbx:
Resolves #524.

### DIFF
--- a/typing_extensions/src_py3/typing_extensions.py
+++ b/typing_extensions/src_py3/typing_extensions.py
@@ -982,8 +982,11 @@ if HAVE_PROTOCOLS:
                 return _generic_new(_next_in_mro(cls), cls, *args, **kwds)
             return _generic_new(cls.__next_in_mro__, cls, *args, **kwds)
 
-    Protocol.__doc__ = Protocol.__doc__.format(bases="Protocol, Generic[T]" if
-                                               OLD_GENERICS else "Protocol[T]")
+    # only update the docstring if docstrings are not being stripped
+    if Protocol.__doc__ is not None:
+        Protocol.__doc__ = Protocol.__doc__.format(
+            bases="Protocol, Generic[T]" if OLD_GENERICS else "Protocol[T]"
+        )
 
     def runtime(cls):
         """Mark a protocol class as a runtime protocol, so that it


the-knights-who-say-ni:

typing_extensions/src_py3/typing_extensions.py
```diff
@@ -982,8 +982,11 @@ def __new__(cls, *args, **kwds):
                 return _generic_new(_next_in_mro(cls), cls, *args, **kwds)
             return _generic_new(cls.__next_in_mro__, cls, *args, **kwds)
 
-    Protocol.__doc__ = Protocol.__doc__.format(bases="Protocol, Generic[T]" if
-                                               OLD_GENERICS else "Protocol[T]")
+    # only update the docstring if docstrings are not being stripped
+    if Protocol.__doc__ is not None:
+        Protocol.__doc__ = Protocol.__doc__.format(
+            bases="Protocol, Generic[T]" if OLD_GENERICS else "Protocol[T]"
+        )
```
ilevkivskyi: Can you also add a test for this? (I am not an expert in Travis, but I think it should be possible to run tests with flags).

Also don't forget to sign the CLA.

rkr-at-dbx:
... ouch. This repo is on a different license and license agreement than mypy and typeshed; I didn't realize that. I think I'd need approval from our legal department first. (I'm pretty sure they'd rubberstamp it, but I'm also pretty sure they're out for the holidays.)

Abandoning. 😞

typing_extensions/src_py3/typing_extensions.py
ethanhs: I don't think you can change this in Travis, and pytest seems to not like this (since `-OO` removes `asserts`), but you could change the `exec`s to `exec(compile(..., optimize=2))` which should test things.

PR #527, opened 2018-01-15
Fix for issue #524

Nickatak:
### Issue #524 summary:

Running any script that imports typing_extensions with the -OO flag will raise an AttributeError.

#### Expected behavior:
Files should be able to import typing_extensions even if they're being ran with the -OO flag.

#### Actual behavior:
Raises the attribute error seen below because `Protocol.__doc__` is a NoneType when the -OO flag is used, and therefore has no attribute .format.
```
  File "testing.py", line 1, in <module>
    import typing_extensions
  File "<frozen importlib._bootstrap>", line 971, in _find_and_load
  File "<frozen importlib._bootstrap>", line 955, in _find_and_load_unlocked
  File "<frozen importlib._bootstrap>", line 656, in _load_unlocked
  File "<frozen importlib._bootstrap>", line 626, in _load_backward_compatible
  File "C:\Users\Nickatak\desktop\github\venv_typing\lib\site-packages\typing_extensions-3.6.2.1-py3.6.egg\typing_extens
ions.py", line 986, in <module>
AttributeError: 'NoneType' object has no attribute 'format'
```

#### Steps to recreate the error:
- [x] Initial setup (Create a venv/clone the repo)
- [x] Install typing (`python setup.py install` for /typing/setup.py)
- [x] Install typing_extensions (`python setup.py install` for /typing/typing_extensions/setup.py)
- [ ] Create a Python file and try to import typing_extensions in it OR
- [ ] Start the live interpreter with -OO (`python -OO`) and then try to import typing_extensions.

#### Alternative steps to recreate the error:
- [x] Run /typing_extensions/src_py3/test_typing_extensions.py with the OO flag (`python -OO test_typing_extensions.py`)

### The fix:
 
During running the tests, we can see that the problem originates from line 985 in /typing/typing_extensions/typing_extensions.py:
```
Protocol.__doc__ = Protocol.__doc__.format(bases="Protocol, Generic[T]" if
                                               OLD_GENERICS else "Protocol[T]")
```

Placing a simple if-check to see whether Protocol.__doc__ is not None around this piece of code avoids trying to call .format on a Nonetype.

### Testing instructions:
In order to confirm whether the issue has been resolved, we can take the following steps.
- [x] Initial setup (Create a venv/clone the repo).
- [x] Install typing (`python setup.py install` for /typing/setup.py).
- [x] Install typing_extensions (`python setup.py install` for /typing/typing_extensions/setup.py).
- [ ] Create a Python file and try to import typing_extensions in it OR
- [ ] Start the live interpreter with -OO (`python -OO`) and then try to import typing_extensions.

AND

- [x] Run /typing_extensions/src_py3/test_typing_extensions.py with the OO flag (`python -OO test_typing_extensions.py`)
- [x] Run test_typing_extensions.py normally (`python test_typing_extensions.py`)
- [x] Run tox tests.

### Output from testing (just the test_typing_extensions.py):
```
..........................................................................
Ran 74 tests in 0.033s

OK
```

### Request:
- Be brutal, I'm new to github.  Anything you want me to fix/edit/add/etc., I will be happy to do.  I'm learning :D 
- If I need to find a better way to test it, let me know (I will try my best to figure it out).
- Please be a little patient when it comes to raw github commands (branching/squashing/etc.).


the-knights-who-say-ni:

Unfortunately our records indicate you have not signed the CLA. For legal reasons we need you to sign this before we can look at your contribution. Please follow [the steps outlined in the CPython devguide](https://devguide.python.org/pullrequest/#licensing) to rectify this issue.


Nickatak:
Just a heads up: I signed that (right now).  If there's an issue with it, I am more than willing to sign it again.

ethanhs:
Hm, it seems you staged and committed all of the files you modified. Did you perhaps run `git add .`? I think you probably wanted to `git add typing_extensions/src_py3/typing_extensions.py`. To fix that you can run `git reset HEAD~` (undo the last commit). Then you can `git add` like I suggest above. Then `git commit -c ORIG_HEAD` (reuse the replaced commit message). Then `git push -f origin master` (this will re-write history to be just the changes to typing_extensions.py).

Also we probably want a test for this. This is somewhat open ended. One idea is to use `compile` with the correct optimization level, and check that the typing source compiles under -OO. Or perhaps you could pass a flag to pytest (which is the underlying test framework typing uses) to run it under -OO.

Nickatak:
Hmm, thank you for your feedback :D Unfortunately, I have to go to sleep now :C 
I will resume things when I have free time tomorrow and attempt to follow your instructions (thank you for t he suggestion on the testing by the way, that was really disturbing me).

typing_extensions/src_py3/typing_extensions.py
```diff
@@ -981,8 +981,8 @@ def __new__(cls, *args, **kwds):
             if OLD_GENERICS:
                 return _generic_new(_next_in_mro(cls), cls, *args, **kwds)
             return _generic_new(cls.__next_in_mro__, cls, *args, **kwds)
-
-    Protocol.__doc__ = Protocol.__doc__.format(bases="Protocol, Generic[T]" if
+    if Protocol.__doc__ is not None:
+        Protocol.__doc__ = Protocol.__doc__.format(bases="Protocol, Generic[T]" if
                                                OLD_GENERICS else "Protocol[T]")
```
ilevkivskyi: You should also add four spaces of indent here.

Nickatak: Thank you! C: should I commit a second time (or just re-write the initial commit)?

Nickatak:
I did an overhaul of the PR.  I realized that while I understood the issue, I couldn't even reproduce it consistently [didn't understand it other than syntactically], let alone test it.  After trying many many different things (and spending a lot of time trying to test the *wrong* module), I finally was able to reproduce the error consistently AND provide a solid way of confirming the issue is resolved.  As for the review: please let me know if you'd like me to make another commit to fix that or if you would rather I reset the head->applied the review-fix -> re-committed.  Thank you very much for your patience! C:

elliott-beach:
> As for the review: please let me know if you'd like me to make another commit to fix that or if you would rather I reset the head->applied the review-fix -> re-committed. Thank you very much for your patience! C:

According to https://github.com/python/mypy/issues/4161, I guess the policy is to not rewrite commits.

ilevkivskyi:
@Nickatak 
>  As for the review: please let me know if you'd like me to make another commit to fix that or if you would rather I reset the head->applied the review-fix -> re-committed. Thank you very much for your patience! C:

You can just commit on your branch and push. As I can see from the current state of PR you did everything right.

The only thing missing now is a test. See @ethans post above
> Also we probably want a test for this. This is somewhat open ended. One idea is to use ``compile`` with the correct optimization level, and check that the typing source compiles under -OO. Or perhaps you could pass a flag to pytest (which is the underlying test framework typing uses) to run it under -OO.

Nickatak:
Indeed.  Initially, I misinterpreted that statement as "We need a way to make sure this works manually."  While the former is important,  I realize now that what he meant was we need a written test-case (and I think using `compile` is probably the easiest way to go here).  I have a few questions and will try to list them out as clearly/separately as possible:

- Wouldn't the test go inside test_typing_extensions.py (not test_typing.py), since the issue itself originated from within typing_extensions.py? Or do both need tests? (I also think it would be strange to test typing, when typing_extensions isn't imported anywhere in typing, or maybe I'm misunderstanding something?)
- In any case of the above question ^: Is there a particular section that you would recommend placing the test in (I was thinking under ProtocolTests for either)?
- Looking at all the previous tests written for both files (test_typing_extensions.py and test_typing.py), don't both use the unittest module (instead of pytest) for py3?
- Clearly, I have attempted to fix something way outside the range of doing so swiftly/independently with my current ability (and learned/am learning a lot along the way).  Is it preferable in the future that I post these types of questions inside the Issue itself instead of opening a PR?

ethanhs:
>    Wouldn't the test go inside test_typing_extensions.py (not test_typing.py), since the issue itself originated from within typing_extensions.py? Or do both need tests? (I think it would be strange to add a test for typing, when typing_extensions isn't imported anywhere in typing, or maybe I'm misunderstanding something?)

There should be a test for both typing.py and typing_extensions.py (one in each test_* file). The reasoning is that while the issue here was in typing_extensions.py, a) many things move from typing_extension.py into typing.py and b) we don't want regressions on this behavior in either file.

>    In any case of the above question ^: Is there a particular section that you would recommend placing the test in (I was thinking under ProtocolTests for either, named something like test_compiles_with_optimization)?

This seems fine to me.

>    Looking at all the previous tests written for both files (test_typing_extensions.py and test_typing.py), don't both use the unittest module (instead of pytest) for py3?

pytest is used as the test runner, but automatically picks up unittest test cases.

>    Clearly, I have attempted to fix something way outside the range of doing so swiftly/independently with my current ability (and learned/am learning a lot along the way). Is it preferable in the future that I post these types of questions inside the Issue itself instead of opening a PR?

Ideally you should research these topics independently, the internet is a great resource! However, you can also ask [on the typing Gitter chat channel](https://gitter.im/python/typing?utm_source=share-link&utm_medium=link&utm_campaign=share-link) for issues more closely pertaining to project specific things. I should usually be around to answer questions, and if I am not, someone else usually is.

Nickatak:
So, I'm having a little bit of trouble isolating the issue with compile.  Since compile itself needs some kind of string for a source, I only know of one option: read in the entire file (I tried using inspect, but since the source for that is tabbed, it just throws out SyntaxErrors).  But, doing so doesn't really specify that the issue is with the Protocol class (as when I exec() to test, the error could be anywhere within the file).  While the below does produce the failed test when using the old file, and does not produce the error when using the newly committed changes, I feel like this is a) hacky as heck, and b) very brittle.  Interestingly, even with the old file, it is able to compile it with optimize=2, the issue displays itself when you exec() it.

Just a heads up, this looks terrible, sorry about that:

```
        def test_compiles_with_opt(self):
            with open('typing_extensions.py', 'r') as fo:
                test_code = fo.read()

            test_module = compile(test_code, 'typing_extensions', 'exec', optimize=2)
            try:
                exec(test_module, locals())
            except AttributeError:
                self.fail('Module did not compile correctly.')```

File: typing_extensions/src_py3/test_typing_extensions.py [hunk 2]
```diff
@@ -1117,7 +1117,7 @@ class P(Protocol[T]): pass
 
         def test_generic_protocols_special_from_protocol(self):
             @runtime
-            class PR(Protocol):
+            class PR(Protocol):
```
ethanhs: I don't think you meant to add whitespace here.

ilevkivskyi:
Just checking that file compiles with optimizations seems OK to me. Also I would add the same tests on Python 2 (although it doesn't fail now, it would be useful to prevent this in future). There are two folders: ``python2`` and ``typing_extensions/src_py2``.

Nickatak:
I'm having some issues porting the tests over to py2 (the compile() for py2 doesn't support an optimize flag like it does in py3).  I think I may be able to come up with a (really hackish) way to check the py2 sources using os.system/subprocess.call.  I'm doing some testing with them right now, to see if I can get something to work (I am able to reproduce the error [tried to do something to a nonetype docstring due to OO] if something like the original issue happens in the other sources, and it does indeed break correctly when this is used: `python -OO -m pytest typing_extensions/src_py2`).

Nickatak:
#### I found the issue with my old tests and wrote new tests for it! :D
They're kind of (extremely) brittle, but they do test the issue successfully.  I'll try to explain some of the decisions I made regarding the test.

#### My reason for choosing to use subprocess over compile:
`compile()` itself doesn't throw errors (unless it's syntax/import/etc.).  The only way to detect whether there are runtime errors using this method is to attempt to `exec()` it after compiling.  The problem (or atleast what I understand) with this is that when you `exec()`, it forces the module to be 'reloaded' in memory and `isinstance()` will break, causing an infinite loop when other tests are loaded with `unittest.discover()`.  Instead, I've chosen to spawn an entirely different (separate) process passing the `-OO` flag as a command-line argument.  While this makes the test EXTREMELY brittle, it does work.
    
    
tox doesn't run any tests found in typing_extensions (but just incase, I tox'd them with the new tests anyway).  I decided to use a very similar testing method that I tried to use for typing_extensions_py2 because it doesn't reload the module (very important).  I have made a handy checklist for testing my tests (seriously, sorry) and will post it below.

#### Tests of tests:
- typing_extensions_py3:
    - With the old error present (without the `if Protocol.__doc__ is not None:` statement).
        - [x] `py.test typing_extensions/src_py3` fails.
        - [x] `pytest typing_extensions/src_py3` fails.
        - [x] `tox` passes (tox tests do not load typing_extensions).
    - Without the old error present (with the `if Protocol.__doc__ is not None:` statement).
        - [x] `py.test typing_extensions/src_py3` passes.
        - [x] `pytest typing_extensions/src_py3` passes.
        - [x] `tox` passes (tox tests do not load typing_extensions).
- typing_extensions_py2:
    - With a duplicated error present (Insert `__doc__ += 's'` on line 345 to force error).
        - [x] `py.test typing_extensions/src_py2` fails.
        - [x] `pytest typing_extensions/src_py2` fails.
        - [x] `tox` passes (tox tests do not load typing_extensions).
    - With the original file.
        - [x] `py.test typing_extensions/src_py2` passes.
        - [x] `pytest typing_extensions/src_py2` passes.
        - [x] `tox` passes (tox tests do not load typing_extensions).
- typing_py3 (/src/):
    - With a duplicated error present (Insert `__doc__ += 's'` on line 1697 to force error).
        - [x] `py.test src` fails.
        - [x] `pytest src` fails.
        - [x] `tox` fails.
    - With the original file.
        - [x] `py.test src` passes.
        - [x] `pytest src` passes.
        - [x] `tox` passes.
-typing_py2 (/python2/):
    - With a duplicated error present (Insert `__doc__ += 's'` on line 1602 to force error).
        - [x] `py.test python2` fails.
        - [x] `pytest python2` fails.
        - [x] `tox` fails.
    - With the original file.
        - [x] `py.test python2` passes.
        - [x] `pytest python2` passes.
        - [x] `tox` passes.

src/test_typing.py
```diff
@@ -1,7 +1,10 @@
 import contextlib
 import collections
+import inspect
```
ilevkivskyi: `inspect` is not used anymore.

python2/test_typing.py
```diff
@@ -1896,6 +1898,16 @@ def foo(x):
...
         self.assertIsNone(typing.get_type_hints(foo))
 
+    def test_typing_compiles_with_opt(self):
+        file_path = os.path.join(os.path.dirname(os.path.realpath(__file__)),
+                                 'typing.py')
+        try:
+            subprocess.check_output('python -OO {}'.format(file_path),
+                                    stderr=subprocess.STDOUT,
+                                    shell=True)
+        except subprocess.CalledProcessError as e:
```
ilevkivskyi: You are not using `e`, below, so `... as e` is not needed.

ilevkivskyi:
I think we can keep the current version of the tests (unless @ethanhs has another ideas). I can merge this after you fix the remaining comments.

ethanhs:
This testing method is acceptable to me.

src/test_typing.py
Nickatak: Removed inspect (latest commit).

python2/test_typing.py
Nickatak: Hm, do you think the specific catch is still necessary? I ended up removing it on my latest commit as a whole and can put it back if you think it is. :D

python2/test_typing.py
```diff
@@ -1896,6 +1898,16 @@ def foo(x):
...
         self.assertIsNone(typing.get_type_hints(foo))
 
+    def test_typing_compiles_with_opt(self):
+        file_path = os.path.join(os.path.dirname(os.path.realpath(__file__)),
+                                 'typing.py')
+        try:
+            subprocess.check_output('python -OO {}'.format(file_path),
+                                    stderr=subprocess.STDOUT,
+                                    shell=True)
+        except:
```
ilevkivskyi: Bare except is bad style. It cat catch something unrelated, I was talking about only removing `as e` part, since it was not used.

Nickatak: Indeed.  I will add it back immediately.

PR #538, opened 2018-02-26
Add utility __pos__ and __neg__ operators to TypeVar

mariusvniekerk:
We already include this in the `__repr__`.
This makes it easier to explicitly just use +/-T to make generics co/contra-variant.


gvanrossum:
Is this specified by PEP 484? Is it supported by mypy? IIRC we had a discussion about it and decided not to do this. Certainly there should be a discussion about this outside this PR.

mariusvniekerk:
Ah right i see in https://github.com/python/typing/issues/211.  I can close this.

PR #549, opened 2018-04-02
Fix typing_extensions to support PEP 560

ilevkivskyi:
The main fix is rewriting `Protocol`. Although the fix is straightforward, there are some observations/details I wanted to discuss before moving `Protocol` to `typing`:
* It looks like we still need a metaclass for `Protocol` even with PEP 560. The problem is a certain asymmetry in ABCs API, there is `__subclasshook__`, but there is no such thing as `__instancehook__`. Because of this we need a metaclass with a single method. `__instancehook__` (or currently the metaclass) is needed to support structural `isinstance()` checks for concrete classes that set attributes on `self` so that fallback to `issubclass()` is not possible. Although protocols with non-method members are relatively rare, in view of planned support for static type checking in Bazel (and other modular build systems) protocols describing certain data will be needed. So we have two options here:
  - Continue to use metaclass, since explicitly subclassing protocols (and thus having metaclass conflicts) is rare. **(-0)**
  - Just add `__instancehook__` exactly the same way there is `__subclasshook__`, then we can keep `typing` metaclass-free even with protocols. **(+1)**
* Currently behaviour of `.register()` with protocols is not specified. As a result it is determined by the current implementation and has some (weird) corner cases. Maybe `.register()` should be actually prohibited for protocols (normal ABCs can still use it)? Protocols are supposed to be purely structural, allowing `.register()` breaks this logic and may introduce some type un-safety. So the options here are simple:
  - Specify the rules for how `.register()` should work for protocols and try to implement these rules. **(-1)**
  - Prohibit using `.register()` on protocols (still OK for normal ABCs). **(+0)**
* Third question is about what to do with extending "built-in" protocols? Existing implementation actually prohibits `class FancyIter(Iterable[str], Protocol): ...` (because protocols can only inherit form other protocols, but technically `collections.abc.Iterable` is a nominal class). However allowing this is not easy because machinery in `abc` (and `functools`) indiscriminately calls `issubclass()` on all subclasses of `Iterable`, but it is prohibited for non-runtime protocols. Here we have three options:
  - Don't allow extending built-in protocols. I don't like this because it reminds me `UserDict` and `UserList`. **(-1)**
  - Allow this and just continue to use the "good old" `sys._getframe()` hack (which I wanted to remove). **(-0)**
  - Allow this and patch `__bases__` in `__init_subclass__` to remove built-in protocols from there. We will also need to copy the methods to allow using them by explicit subclasses. **(+0)**

For clarity I added my votes for all options. @gvanrossum what is your opinion on these points. It is not urgent to discuss them, (I didn't act on any of these points in the PR, it just mirrors the pre-PEP 560 implementation so can be merged before we agree on these points) but we should discuss these before moving protocols to `typing`.

### DIFF
--- a/typing_extensions/src_py3/test_typing_extensions.py
+++ b/typing_extensions/src_py3/test_typing_extensions.py
@@ -1148,19 +1148,9 @@ if HAVE_PROTOCOLS:
                 self.assertFalse(P._is_runtime_protocol)
             self.assertTrue(PR._is_runtime_protocol)
             self.assertTrue(PG[int]._is_protocol)
-            if not PEP_560:
-                self.assertEqual(P._get_protocol_attrs(), {'meth'})
-                self.assertEqual(PR._get_protocol_attrs(), {'x'})
-                self.assertEqual(frozenset(PG._get_protocol_attrs()),
-                                 frozenset({'x', 'meth'}))
-                self.assertEqual(frozenset(PG[int]._get_protocol_attrs()),
-                                 frozenset({'x', 'meth'}))
-            else:
-                self.assertEqual(P._protocol_attrs, {'meth'})
-                self.assertEqual(PR._protocol_attrs, {'x'})
-                self.assertEqual(frozenset(PG._protocol_attrs),
-                                 frozenset({'x', 'meth'}))
-                self.assertEqual(frozenset(PG[int]._protocol_attrs),
+            self.assertEqual(typing_extensions._get_protocol_attrs(P), {'meth'})
+            self.assertEqual(typing_extensions._get_protocol_attrs(PR), {'x'})
+            self.assertEqual(frozenset(typing_extensions._get_protocol_attrs(PG)),
                                  frozenset({'x', 'meth'}))
 
         def test_no_runtime_deco_on_nominal(self):
--- a/typing_extensions/src_py3/test_typing_extensions.py
+++ b/typing_extensions/src_py3/test_typing_extensions.py
@@ -1102,7 +1102,7 @@ if HAVE_PROTOCOLS:
             T = TypeVar('T')
             S = TypeVar('S')
             class P(Protocol[T, S]): pass
-            # After PEP 560 unsubscripted generics have standard repr.
+            # After PEP 560 unsubscripted generics have a standard repr.
             if not PEP_560:
                 self.assertTrue(repr(P).endswith('P'))
             self.assertTrue(repr(P[T, S]).endswith('P[~T, ~S]'))
@@ -1151,6 +1151,9 @@ if HAVE_PROTOCOLS:
             self.assertEqual(typing_extensions._get_protocol_attrs(P), {'meth'})
             self.assertEqual(typing_extensions._get_protocol_attrs(PR), {'x'})
             self.assertEqual(frozenset(typing_extensions._get_protocol_attrs(PG)),
+                             frozenset({'x', 'meth'}))
+            if not PEP_560:
+                self.assertEqual(frozenset(typing_extensions._get_protocol_attrs(PG[int])),
                                  frozenset({'x', 'meth'}))
 
         def test_no_runtime_deco_on_nominal(self):
--- a/typing_extensions/src_py3/test_typing_extensions.py
+++ b/typing_extensions/src_py3/test_typing_extensions.py
@@ -20,6 +20,9 @@ except ImportError:
 import typing
 import typing_extensions
 import collections.abc as collections_abc
+
+PEP_560 = sys.version_info[:3] >= (3, 7, 0)
+
 OLD_GENERICS = False
 try:
     from typing import _type_vars, _next_in_mro, _type_check
@@ -471,7 +474,10 @@ class CollectionsAbcTests(BaseTestCase):
         class C(typing_extensions.Counter[T]): ...
         if TYPING_3_5_3:
             self.assertIs(type(C[int]()), C)
-            self.assertEqual(C.__bases__, (typing_extensions.Counter,))
+            if not PEP_560:
+                self.assertEqual(C.__bases__, (typing_extensions.Counter,))
+            else:
+                self.assertEqual(C.__bases__, (collections.Counter, typing.Generic))
 
     def test_counter_subclass_instantiation(self):
 
@@ -818,9 +824,10 @@ if HAVE_PROTOCOLS:
             self.assertIsSubclass(C, P)
             self.assertIsSubclass(C, PG)
             self.assertIsSubclass(BadP, PG)
-            self.assertIsSubclass(PG[int], PG)
-            self.assertIsSubclass(BadPG[int], P)
-            self.assertIsSubclass(BadPG[T], PG)
+            if not PEP_560:
+                self.assertIsSubclass(PG[int], PG)
+                self.assertIsSubclass(BadPG[int], P)
+                self.assertIsSubclass(BadPG[T], PG)
             with self.assertRaises(TypeError):
                 issubclass(C, PG[T])
             with self.assertRaises(TypeError):
@@ -1043,7 +1050,11 @@ if HAVE_PROTOCOLS:
                 def meth(self): pass
             class P(PR[int, str], Protocol):
                 y = 1
-            self.assertIsSubclass(PR[int, str], PR)
+            if not PEP_560:
+                self.assertIsSubclass(PR[int, str], PR)
+            else:
+                with self.assertRaises(TypeError):
+                    self.assertIsSubclass(PR[int, str], PR)
             self.assertIsSubclass(P, PR)
             with self.assertRaises(TypeError):
                 PR[int]
@@ -1091,7 +1102,9 @@ if HAVE_PROTOCOLS:
             T = TypeVar('T')
             S = TypeVar('S')
             class P(Protocol[T, S]): pass
-            self.assertTrue(repr(P).endswith('P'))
+            # After PEP 560 unsubscripted generics have standard repr.
+            if not PEP_560:
+                self.assertTrue(repr(P).endswith('P'))
             self.assertTrue(repr(P[T, S]).endswith('P[~T, ~S]'))
             self.assertTrue(repr(P[int, str]).endswith('P[int, str]'))
 
@@ -1135,12 +1148,20 @@ if HAVE_PROTOCOLS:
                 self.assertFalse(P._is_runtime_protocol)
             self.assertTrue(PR._is_runtime_protocol)
             self.assertTrue(PG[int]._is_protocol)
-            self.assertEqual(P._get_protocol_attrs(), {'meth'})
-            self.assertEqual(PR._get_protocol_attrs(), {'x'})
-            self.assertEqual(frozenset(PG._get_protocol_attrs()),
-                             frozenset({'x', 'meth'}))
-            self.assertEqual(frozenset(PG[int]._get_protocol_attrs()),
-                             frozenset({'x', 'meth'}))
+            if not PEP_560:
+                self.assertEqual(P._get_protocol_attrs(), {'meth'})
+                self.assertEqual(PR._get_protocol_attrs(), {'x'})
+                self.assertEqual(frozenset(PG._get_protocol_attrs()),
+                                 frozenset({'x', 'meth'}))
+                self.assertEqual(frozenset(PG[int]._get_protocol_attrs()),
+                                 frozenset({'x', 'meth'}))
+            else:
+                self.assertEqual(P._protocol_attrs, {'meth'})
+                self.assertEqual(PR._protocol_attrs, {'x'})
+                self.assertEqual(frozenset(PG._protocol_attrs),
+                                 frozenset({'x', 'meth'}))
+                self.assertEqual(frozenset(PG[int]._protocol_attrs),
+                                 frozenset({'x', 'meth'}))
 
         def test_no_runtime_deco_on_nominal(self):
             with self.assertRaises(TypeError):
--- a/typing_extensions/src_py3/typing_extensions.py
+++ b/typing_extensions/src_py3/typing_extensions.py
@@ -997,24 +997,10 @@ if HAVE_PROTOCOLS:
         Protocol.__doc__ = Protocol.__doc__.format(bases="Protocol, Generic[T]" if
                                                    OLD_GENERICS else "Protocol[T]")
 
-    def runtime(cls):
-        """Mark a protocol class as a runtime protocol, so that it
-        can be used with isinstance() and issubclass(). Raise TypeError
-        if applied to a non-protocol class.
-
-        This allows a simple-minded structural check very similar to the
-        one-offs in collections.abc such as Hashable.
-        """
-        if not isinstance(cls, _ProtocolMeta) or not cls._is_protocol:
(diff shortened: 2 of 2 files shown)


gvanrossum:
This is a huge thing. Maybe you can find a new co-conspirator amongst Python core devs or mypy contributors? I also am not sure at all that I fully appreciate all the subtleties in the issues you bring up. I wish we could whiteboard in person -- maybe at or after PyCon?

> in view of planned support for static type checking in Bazel (and other modular build systems) protocols describing certain data will be needed.

Can you clarify? Why is Bazel used as a forcing function? What specific examples do you have in mind? Runtime checking for Protocol conformance is pretty iffy even if just methods are involved (there's no signature checking, hence the `@runtime*` decorator) and I would be okay if Protocols just didn't support this if data attributes are present. Then again maybe data attributes are a major use case somehow?

> Just add __instancehook__ exactly the same way there is __subclasshook__, then we can keep typing metaclass-free even with protocols. (+1)

I wonder what Mark Shannon would think of this (he's on record as not wanting to enable runtime checks at all).

> Prohibit using .register() on protocols (still OK for normal ABCs). (+0)

Makes some sense to me, though it depends on how good the solution is for the previous issue. I could imagine that in some cases it's easier to use .register() (and then accept instances of the registered class as members of the protocol without any attribute checks) than to make the attribute check work. Also I'm curious about the corner cases you've found.

> Third question is about what to do with extending "built-in" protocols?

I'm not following all the details of your description here -- I haven't thought about any of this code in a really long time. :-(

Sorry...

ilevkivskyi:
> This is a huge thing.

Do you mean the PR, my comment, or the protocols in general? The comment is indeed long but it was inspired by thinking about where we can go with PEP 560 and protocols. As I mentioned the PR itself _doesn't act_ yet on any of this points it just fixes the support for protocols in Python 3.7 broken by the changes in internal `typing` API.

> I wish we could whiteboard in person -- maybe at or after PyCon?

Discussion of these issues is not urgent, it can wait until then, but I would appreciate if you review the PR, so that `typing_extensions` stops being broken on Python 3.7 (now it raises on import).

> Can you clarify? Why is Bazel used as a forcing function?

We had a discussion on Slack with Jared Hance about what to do with circular imports necessary only for type checking and Bazel that doesn't like circular imports (even under `if MYPY: ...`). He reasonably proposed that all such cycles should be broken by introducing protocols for the function/method argument types where necessary. After thinking some time about this it seems a very good idea to me: modular (e.g. in the sense of having well defined Bazel targets) static type checking will require protocols. Then since many argument types for functions will have some data components, support for non-method members is needed. The question about why we need the runtime support is a bit more subtle (also see my answer below). The problem will happen if someone needs to transition from one API to another. Imagine this (simplified) scenario:
```python
class LegacyData(Protocol):
    x: int
    y: int
    label: str

class NewData(Protocol):
    coord: Tuple[int, int]
    label: str

def process(items: Iterable[Union[LegacyData, NewData]]) -> None:
    for item in items:
        if <some check here>:
             # unpack and process legacy API
        else:
             # other logic for new API
```
IMO ideally we should provide a way to destructure a union of protocols (otherwise we would induce growing amount of rigidity in any protocol specified APIs) so the check above can be just `isinstace(item, LegacyData)`. Of course one can use some `hasattr()` hacks instead, but then this function will be not statically checked.

> I wonder what Mark Shannon would think of this (he's on record as not wanting to enable runtime checks at all).

On my record he is not participating in protocol discussion at all :-) But seriously this situation is special. Namely, mypy already has reasonable support of `isinstance()` for nominal types, and this is important for example for destructuring unions. We should have the same mechanism to distinguish structural types, and it already exists -- `collections.abc.Iterable` and friends already work with `isinstance()` this way. Arguably, mypy implementation for this should be improved, and we have an issue for this https://github.com/python/mypy/issues/3827, whatever we do at runtime should be matched 1:1 statically, so that we are completely safe, for example:
```python
@runtime_checkable
class ProtoA(Protocol):
    attr: int
class ProtoB(Protocol):
    attr: str

x: Union[ProtoA, ProtoB]
if isinstance(x, ProtoA):
    x.attr + 1  # the binder type is still a union here, so mypy will complain after the issue is fixed.
```

> Maybe you can find a new co-conspirator amongst Python core devs or mypy contributors?

I hope @JukkaL is still interested in participating in this discussion. My ideal goal would be to have protocols story settled by the end of June, but I don't want to distract Jukka to much, there is still some work needed on fine-grained mode.

typing_extensions/src_py3/typing_extensions.py
```diff
@@ -729,6 +738,30 @@ def _next_in_mro(cls):
...
+                    '__args__', '__slots__',
+                    '__next_in_mro__', '__parameters__', '__origin__',
+                    '__orig_bases__', '__extra__', '__tree_hash__',
+                    '__doc__', '__subclasshook__', '__init__', '__new__',
+                    '__module__', '_MutableMapping__marker', '_gorg')):
+                attrs.add(attr)
+    return attrs
+
+
+def _callable_members_only(cls):
```
gvanrossum: It would be nice if this function followed the (very informal) naming convention for booleans to start with "has" or "is". Initially I mistook it for a filter over `_get_protocol_attrs()`.

typing_extensions/src_py3/typing_extensions.py
```diff
@@ -985,6 +997,188 @@ def __new__(cls, *args, **kwds):
         Protocol.__doc__ = Protocol.__doc__.format(bases="Protocol, Generic[T]" if
                                                    OLD_GENERICS else "Protocol[T]")
 
+
+if PEP_560:
+    # We need to rewrite these from scratch.
```
gvanrossum: Can you elaborate in the comment why? Also, it seems weird that there are now two conditional definitions of `_ProtocolMeta` and `Protocol`, one pair under `if HAVE_PROTOCOLS` above and then this pair. Why not at least execute precisely one pair of definitions at runtime?

typing_extensions/src_py3/typing_extensions.py
```diff
@@ -985,6 +997,188 @@ def __new__(cls, *args, **kwds):
...
+                if all(hasattr(instance, attr) and
+                        (not callable(getattr(cls, attr, None)) or
+                         getattr(instance, attr) is not None)
+                        for attr in _get_protocol_attrs(cls)):
+                    return True
+            return super().__instancecheck__(instance)
+
+
+    class Protocol(metaclass=_ProtocolMeta):
+        # There is quite a few overlapping code with typing.Generic.
```
gvanrossum: s/few/lot of/

gvanrossum:
Whoops, sent the review too soon. Expect more to come.

typing_extensions/src_py3/test_typing_extensions.py
```diff
@@ -818,9 +824,10 @@ def x(self): ...
             self.assertIsSubclass(C, P)
             self.assertIsSubclass(C, PG)
             self.assertIsSubclass(BadP, PG)
-            self.assertIsSubclass(PG[int], PG)
-            self.assertIsSubclass(BadPG[int], P)
-            self.assertIsSubclass(BadPG[T], PG)
+            if not PEP_560:
```
gvanrossum: Is there a similar regression for non-protocol generics? If so, fine. If not, this is a bit unfortunate (though it may be unavoidable).

typing_extensions/src_py3/test_typing_extensions.py
```diff
@@ -471,7 +474,10 @@ def test_counter_instantiation(self):
         class C(typing_extensions.Counter[T]): ...
         if TYPING_3_5_3:
```
gvanrossum: It's kind of unfortunate that these variables are named in such a way that it's not clear that they imply "or all higher versions".

typing_extensions/src_py3/test_typing_extensions.py
```diff
@@ -1091,7 +1102,9 @@ def test_generic_protocols_repr(self):
             T = TypeVar('T')
             S = TypeVar('S')
             class P(Protocol[T, S]): pass
-            self.assertTrue(repr(P).endswith('P'))
+            # After PEP 560 unsubscripted generics have standard repr.
```
gvanrossum: Nit: s/have/have a/

typing_extensions/src_py3/test_typing_extensions.py
```diff
@@ -1135,12 +1148,10 @@ def meth(self):
...
-            self.assertEqual(P._get_protocol_attrs(), {'meth'})
-            self.assertEqual(PR._get_protocol_attrs(), {'x'})
-            self.assertEqual(frozenset(PG._get_protocol_attrs()),
-                             frozenset({'x', 'meth'}))
-            self.assertEqual(frozenset(PG[int]._get_protocol_attrs()),
-                             frozenset({'x', 'meth'}))
+            self.assertEqual(typing_extensions._get_protocol_attrs(P), {'meth'})
+            self.assertEqual(typing_extensions._get_protocol_attrs(PR), {'x'})
+            self.assertEqual(frozenset(typing_extensions._get_protocol_attrs(PG)),
+                                 frozenset({'x', 'meth'}))
```
gvanrossum: Did you intentionally delete the test for `PG[int]`?

ilevkivskyi:
> I have a sense of deja vu -- did I see an earlier version of this code in a different PR?

Thanks for review! Yes, this just mirrors the logic in pre-PEP 560 version. I tried to share some code, but it is not so easy. I will check if there are some more places where common logic can be factored out. I will reply to some comments now, but will push the new version later.

typing_extensions/src_py3/typing_extensions.py
ilevkivskyi: Yes, this totally makes sense. There is no need to execute both versions in 3.7. I will re-organise the `if`s.

typing_extensions/src_py3/typing_extensions.py
```diff
@@ -985,6 +997,188 @@ def __new__(cls, *args, **kwds):
...
+        # Unfortunarely it is hard to avoid this while these live in two different modules.
+        # The duplicated code will be removed when Protocol is moved to typing.
```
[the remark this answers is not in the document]
ilevkivskyi: We can move it already in 3.7.1, however this will require `_ProtocolMeta`, since we can add `__instancehook__` only in 3.8. Although this is probably not a big deal, since I expect that explicit subclassing of protocols by concrete classes will be rare. One of the main goals of PEP 544 is to allow not to subclass them :-)

typing_extensions/src_py3/typing_extensions.py
```diff
@@ -985,6 +997,188 @@ def __new__(cls, *args, **kwds):
...
+            def func(x: Proto) -> int:
+                return x.meth()
+
+            func(C())  # Passes static type check
+
+        See PEP 544 for details. Protocol classes decorated with
+        @typing_extensions.runtime act as simple-minded runtime protocol that checks
+        only the presence of given attributes, ignoring their type signatures.
+
+        Protocol classes can be generic, they are defined as::
```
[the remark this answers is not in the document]
ilevkivskyi: Protocols are not different from normal generics in this aspect, for example:
```python
@runtime_checkable
class GenP(Protocol[T]):
    def meth(self) -> T:
        ...
class C:
    ...
issubclass(C, GenP)  # This is OK, False or True depending on C
issubclass(C, GenP[int])  # This is an error both in mypy and at runtime.
```

typing_extensions/src_py3/test_typing_extensions.py
ilevkivskyi: Yes, this is not different from normal generics. This is already mentioned in the PEP 560 as one of the few user visible changes.

typing_extensions/src_py3/test_typing_extensions.py
ilevkivskyi: I agree, but I would say it is too late to change this.

typing_extensions/src_py3/test_typing_extensions.py
ilevkivskyi: Yes, but I think it is better to keep it under `if not PEP_560:`.

typing_extensions/src_py3/typing_extensions.py
```diff
@@ -985,6 +997,188 @@ def __new__(cls, *args, **kwds):
...
 
+
+if PEP_560:
+    # We need to rewrite these from scratch.
+    del _ProtocolMeta, Protocol
+
+    from typing import _type_check, _GenericAlias, _collect_type_vars
+
+    class _ProtocolMeta(abc.ABCMeta):
+        # This metaclass is a bit unfortunate and exists only because of the lack
```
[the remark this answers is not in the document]
ilevkivskyi: Hm, actually it looks like I have found a simple solution that would avoid using metaclass and will not need the `__instancehook__` and instead using PEP 560 "on full speed". The idea is that `@runtime` doesn't need to return a class, it can return a `_GenericAlias` (this is what `typing.Iterable` etc. are instances of), what do you think about this? With this solution we can have fully functional and metaclass-free protocols in `typing` already in 3.7.1.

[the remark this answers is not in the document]
ilevkivskyi: Actually, I think it makes sense to test this solution (especially if you have some doubts/worries). So I am going to remove the metaclass from this PR tomorrow (this should be easy). So that we can test this approach for few months while `Protocol` is still in `typing_extensions`.

typing_extensions/src_py3/typing_extensions.py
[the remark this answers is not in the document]
gvanrossum: Still my suggestion stands -- `@runtime` (by whichever name) should complain when the decorated class is a generic protocol. Right? Since the only reason to add `@runtime` is to enable `isinstance()` checks and that is not allowed for generic protocols.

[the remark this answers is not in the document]
gvanrossum: Oh wait. After reading [the relevant passage in PEP 544](https://www.python.org/dev/peps/pep-0544/#runtime-decorator-and-narrowing-types-by-isinstance) again I finally realize that `@isinstance(x, GenP)` is still allowed, it's only `@isinstance(x, GenP[int])` that's disallowed. So forget everything I said here. (Except that as I said in a previous comment it's been a long time since I thought about this. :-( )

typing_extensions/src_py3/typing_extensions.py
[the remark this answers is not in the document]
ilevkivskyi: OK, makes sense. But then I would like (3) to happen relatively soon (beginning or middle of May).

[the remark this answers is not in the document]
gvanrossum: Sure. At this point I believe I'm just waiting for you to update fixes to address the code review so far and then hopefully we can converge on (1) pretty soon.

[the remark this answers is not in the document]
ilevkivskyi: > Sure. At this point I believe I'm just waiting for you to update fixes to address the code review so far and then hopefully we can converge on (1) pretty soon.

Sorry, spent lots of time today working on protocols in fine grained, so probably finish this tomorrow.

ilevkivskyi:
@gvanrossum I addressed the previous set or your comments. Could you please take a look again if this PR is ready?

PR #557, opened 2018-05-05
Type variable value instantiation

tdb-alcorn:
This PR makes it possible to call type variables as constructors within class definitions of subclasses of Generic. Concretely, this enables writing

```python
T = TypeVar('T')

class C(Generic[T]):
    def __init__(self):
        self.inner = T()
```

A realistic use case might be when writing a generic mapping class where you can add keys without specifying a value because a default one is created for you by calling `T()`. This brings type variables more in line with how they can be used in other typed languages.

### DIFF
--- a/src/test_typing.py
+++ b/src/test_typing.py
@@ -1378,6 +1378,30 @@ class GenericTests(BaseTestCase):
         self.assertEqual(c.from_b, 'b')
         self.assertEqual(c.from_c, 'c')
 
+    def test_instantiate_typevar(self):
+        T = TypeVar('T')
+
+        class C(Generic[T]):
+            def __init__(self):
+                self.inner = T()
+
+        c = C[int]
+        x = c()
+        self.assertEqual(x.inner, int())
+
+        U = TypeVar('U')
+        V = TypeVar('V')
+
+        class P(Generic[U, V]):
+            def __init__(self):
+                self.left = U()
+                self.right = V()
+
+        p = P[int, str]
+        y = p()
+        self.assertEqual(y.left, int())
+        self.assertEqual(y.right, str())
+
 
 class ClassVarTests(BaseTestCase):
 
--- a/src/typing.py
+++ b/src/typing.py
@@ -494,7 +494,7 @@ class TypeVar(_TypingBase, _root=True):
     """
 
     __slots__ = ('__name__', '__bound__', '__constraints__',
-                 '__covariant__', '__contravariant__')
+                 '__covariant__', '__contravariant__', '__value__')
 
     def __init__(self, name, *constraints, bound=None,
                  covariant=False, contravariant=False):
@@ -515,6 +515,7 @@ class TypeVar(_TypingBase, _root=True):
             self.__bound__ = _type_check(bound, "Bound must be a type.")
         else:
             self.__bound__ = None
+        self.__value__ = None
 
     def _get_type_vars(self, tvars):
         if self not in tvars:
@@ -534,6 +535,11 @@ class TypeVar(_TypingBase, _root=True):
 
     def __subclasscheck__(self, cls):
         raise TypeError("Type variables cannot be used with issubclass().")
+
+    def __call__(self, *args, **kwargs):
+        if self.__value__ is not None:
+            return self.__value__(*args, **kwargs)
+        super().__call__(*args, **kwargs)
 
 
 # Some unconstrained type variables.  These are used by the container types.
@@ -1132,6 +1138,11 @@ class GenericMeta(TypingMeta, abc.ABCMeta):
             _check_generic(self, params)
             tvars = _type_vars(params)
             args = params
+            if len(params) != len(self.__parameters__):
+                raise TypeError(
+                    "Wrong number of type parameters")
+            for c, p in zip(params, self.__parameters__):
+                p.__value__ = c
 
         prepend = (self,) if self.__origin__ is None else ()
         return self.__class__(self.__name__,


the-knights-who-say-ni:

When your account is ready, please add a comment in this pull request
and a Python core developer will remove the `CLA not signed` label
to make the bot check again.

tdb-alcorn:
I have signed the CLA

Michael0x2a:
I'm not sure this interacts cleanly with the rest of PEP 484? (e.g. type inference, forward declarations, non-concrete types...)

For example, I'm not really sure how your PR would handle the following cases?

    T = TypeVar('T')
    class Foo(Generic[T]):
        def __init__(self) -> None:
            self.value = T()

    # Inferred type is 'Foo[Any]'; what is self.value?
    f1 = Foo()

    # The inferred type is 'Foo[int]' -- can Foo's constructor tell?
    f2: Foo[int] = Foo()

    # 'Bar' is a forward reference
    f3 = Foo['Bar[int]']()

    # Which one gets instantiated?
    f4 = Foo[Union[int, str]]()

    
    S = TypeVar('S')

    # Now the typevar is forward declared
    class Bar(Generic['S']):
        def __init__(self) -> None:
            # How does 'S' know what type it is?
            self.value = S()

    b = Bar[int]()

gvanrossum:
> This brings type variables more in line with how they can be used in other typed languages.

Can you provide an example of how this works in other typed languages? Have you actually encountered a situation in real code where you needed this behavior?

Also, you are changing the runtime behavior, but a type checker like mypy currently doesn't support this -- e.g.
```py
T = TypeVar('T')
class C(Generic[T]):
    def __init__(self) -> None:  # The -> None is needed so mypy checks the method body
        self.inner = T()
```
gets an error from mypy on the last line:
```
_.py:5: error: 'T' is a type variable and only valid in type context
```
Have you thought this through from a static typing POV?

tdb-alcorn:
To your first question, here is an example of this behavior in C#:
```
using System;
					
public class Program
{
	public static void Main()
	{
		var f = new Foo<int>();
		Console.WriteLine(f.Inner);
	}
}

public class Foo<T> where T : new()
{
	public T Inner;
	
	public Foo()
	{
		Inner = new T();
	}
}
```
which you can experiment with at https://dotnetfiddle.net/aA0I4s

To your second question, I was actually motivated to implement this because I want to use this behavior in a side project related to reinforcement learning. Specifically, I have an abstract base class `Agent` that handles interactions with the task environment. I'm experimenting with different agents, so I might have `RandomAgent`, `LearningAgent` and so on. One particular agent, `LearningAgent`, in turn makes use of another abstract base class `Predictor` which codifies and interface with methods like `predict` and `update`. I'm also experimenting with different predictors, so I might have `SimplePredictor`, `FancyPredictor` and so on. As a result, `Learning Agent` is defined like
```
P = TypeVar('P', bound=Predictor)
class LearningAgent(Generic[P]):
    def __init__(self):
        self.predictor = P()
```
The logic that sets up and runs the environment needs to be able to call an agent's constructor, so I'd like to be able to supply it with something like `LearningAgent[FancyPredictor]`. Currently, I'm working around this by having a factory function that accepts the predictor type and wraps the `LearningAgent` constructor like so
```
def learning_agent_with_predictor(p):
    agent = LearningAgent[p]()
    agent.predictor = p()
    return agent
```

Regarding static typing, in my opinion the easiest thing to do would be to make a follow-up change on mypy that special cases calling a type variable e.g. `T()`. The downside is that this would open up the possibility of runtime errors due to non-instantiable type variable values like `Any` or `Union`, as discussed above. In other languages, C# supports static checking of this behaviour with the `where T : new()` type constraint as exhibited above, but Java doesn't support it at all and punts to runtime (see this [Stackoverflow thread](https://stackoverflow.com/questions/1927789/why-should-i-care-that-java-doesnt-have-reified-generics)).

gvanrossum:
The basic problem with this proposal is that it violates "type erasure". At runtime there's no difference between C[int] and C[str], they just look like C. Instead of C[int]() you should probably arrange for the call to be C(int), e.g. via `def __init__(self, t: Type[T])`.

I also don't see why T() should be special -- why not T.some_factory_method() too? Also the signature of T() would seem problematic, since `__init__` (unless other method calls) is not required to follow Liskov.

@ilevkivskyi may be able to better explain this.

ilevkivskyi:
Sorry, but I don't like this idea. The key idea of typing _in Python_ is to maintain a clear separation between static and runtime worlds (maybe there are few remaining issues, but we have been successfully moving towards this goal). Essentially, types are just annotations, i.e. "tags" attached to expressions (nodes in the syntactic tree) to be used by external tools. The (limited) presence of annotations at runtime is for introspection purposes and other third party tools.

It is very important to understand this point: types are for expressions, _not_ for objects. This proposal violates this principle, and therefore I believe we should not go this way.

gvanrossum:
The reasonable response at this point is to close the issue.

tdb-alcorn:
I understand your reasoning, keeping clear separation between static/runtime for Python makes sense. However, generics already seem to be able to be used as objects and, as a result, affect runtime semantics. Possibly this is necessary to implement checking of type constraints? For example
```
class Bar(object):
    def do(self):
        return 0

T = TypeVar('T', bound=Bar)

class Foo(Generic[T]):
    @staticmethod
    def hello(x: T) -> T:
        x.do()
        return x

y: int = Foo[Bar].hello(1)
```
passes static type checking even though we have all the information necessary to prove that it could fail at runtime.

Perhaps worse, we could write something like
```
class NotBar(object):
    def do(self):
        return -1

z: NotBar = Foo[Bar].hello(NotBar())
```
which would both type check and silently "work" at runtime, at least until something unexpected happens later. I'm not an expert but it feels to me like once you have generics in expressions at all you have to start checking the use of types as objects (i.e. expressions like `Foo[Bar].hello()`), and from there it doesn't seem like much more of a stretch to extend that to checking the use of type variables as objects too (i.e. `T()` or `T.do()`)

Do these ideas resonate? If not, I'll close this PR. I recognize that this might be pushing type checking beyond the line where it stops making sense for Python.

ilevkivskyi:
> passes static type checking even though we have all the information necessary to prove that it could fail at runtime.

It passes because of a known bug in mypy https://github.com/python/mypy/issues/3645 (although here the situation is a bit different form `@classmethod` your definition of `Foo` doesn't make sense and should be prohibited because static methods can't use `T` at all), this has nothing to do with any conceptual limitations to static type checking.

tdb-alcorn:
> static methods can't use T at all
I see, I didn't realise that this was part of PEP 484. Thanks to everyone for considering this proposal. I really appreciate the work you guys have done on introducing types to Python!

PR #571, opened 2018-07-10
include license file for typing-extensions and in wheels

dougalsutherland:
Currently `typing-extensions` doesn't include the `LICENSE` file at all, and wheels for `typing` don't either.


dougalsutherland:
I just signed the CLA.

typing_extensions/LICENSE
```diff
@@ -0,0 +1 @@
+../LICENSE
```
gvanrossum: I take it this is a symbolic link. I'm not sure we're okay with those, though I'm not sure how to do it without one. I know that in typeshed we stopped using symbolic links because they caused problems on Windows, but of course that's a much larger project.

ethanhs: Since the LICENSE will likely never change, I think it would be better just to copy it over.

gvanrossum:
Oh, that's fair.

dougalsutherland:
Changed the symlink to another copy.

PR #576, opened 2018-07-31
Fix IO.closed to be property

### LINKED ISSUE #575: bug: IO.closed should be a property rather than method
Currently the `IO` base class has the following definition for the `closed` attribute: https://github.com/python/typing/blob/ff2084eb9ab5eea778d744c706e6a6aad05ea4bd/src/typing.py#L2272-L2274

It should be `@abstractproperty` instead.

suutari:
The `closed` attribute in file objects is actually a property rather than a method.  Fix this in the `IO` base class.

Fixes #575

### DIFF
--- a/python2/typing.py
+++ b/python2/typing.py
@@ -2061,7 +2061,7 @@ class IO(Generic[AnyStr]):
     def close(self):
         pass
 
-    @abstractmethod
+    @abstractproperty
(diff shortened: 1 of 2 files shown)


suutari:
Filled in my GitHub account username in b.p.o now (and I had already signed the CLA).

PR #580, opened 2018-08-26
Fix instance/subclass checks of functions against runtime protocols.

### LINKED ISSUE #579: Runtime Protocols can't be tested against certain objects
I discovered this bug when I tried to scan a module containing PyTest fixtures for instances of a Protocol. `isinstance(fixture, protocol)` fails with `TypeError: argument of type 'getset_descriptor' is not iterable`.

I haven't dug sufficiently into what `@pytest.fixture` does to upset the `isinstance` check, but here is a relatively simple repro:

```python
In [1]: import typing_extensions

In [2]: import pytest

In [3]: @typing_extensions.runtime
   ...: class MyProtocol(typing_extensions.Protocol):
   ...:     member: str
   ...:     

In [4]: @pytest.fixture
   ...: def fix():
   ...:     return 1
   ...: 

In [5]: class C:
   ...:     def __init__(self):
   ...:         self.member = 'value'
   ...:         

In [6]: isinstance(C(), MyProtocol)
Out[6]: True

In [7]: isinstance(fix, MyProtocol)
---------------------------------------------------------------------------
TypeError                                 Traceback (most recent call last)
<ipython-input-7-f1a1c6b96adc> in <module>()
----> 1 isinstance(fix, MyProtocol)

/Users/erikwright/.pyenv/virtualenvs/crucible/3.6.5/lib/python3.6/site-packages/typing_extensions.py in __instancecheck__(self, instance)
    891                         for attr in _get_protocol_attrs(self)):
    892                     return True
--> 893             return super(GenericMeta, self).__instancecheck__(instance)
    894 
    895         def __subclasscheck__(self, cls):

/Users/erikwright/.pyenv/virtualenvs/crucible/3.6.5/lib/python3.6/abc.py in __instancecheck__(cls, instance)
    191                 return False
    192             # Fall back to the subclass check.
--> 193             return cls.__subclasscheck__(subclass)
    194         return any(cls.__subclasscheck__(c) for c in {subclass, subtype})
    195 

/Users/erikwright/.pyenv/virtualenvs/crucible/3.6.5/lib/python3.6/site-packages/typing_extensions.py in __subclasscheck__(self, cls)
    908                     not _is_callable_members_only(self)):
    909                 if sys._getframe(1).f_globals['__name__'] in ['abc', 'functools', 'typing']:
--> 910                     return super(GenericMeta, self).__subclasscheck__(cls)
    911                 raise TypeError("Protocols with non-method members"
    912                                 " don't support issubclass()")

/Users/erikwright/.pyenv/virtualenvs/crucible/3.6.5/lib/python3.6/abc.py in __subclasscheck__(cls, subclass)
    207             return False
    208         # Check the subclass hook
--> 209         ok = cls.__subclasshook__(subclass)
    210         if ok is not NotImplemented:
    211             assert isinstance(ok, bool)

/Users/erikwright/.pyenv/virtualenvs/crucible/3.6.5/lib/python3.6/site-packages/typing_extensions.py in _proto_hook(other)
    869                                 return NotImplemented
    870                             break
--> 871                         if (attr in getattr(base, '__annotations__', {}) and
    872                                 isinstance(other, _ProtocolMeta) and other._is_protocol):
    873                             break

TypeError: argument of type 'getset_descriptor' is not iterable

In [8]: 
```

erikwright:
Fixes #579 .

When performing an `issubclass` check of a type against a protocol, the `__annotations__` member of the type is accessed and assumed to be iterable. `__annotations__` is a descriptor in the case of `types.FunctionType`, so while it is iterable when accessed on a function instance it is not iterable when accessed on the type of a function. This causes the `issubclass` check to fail with an exception.

In some cases (AFAICT, non-data protocols), an `isinstance` check of an object will use, internally, a subclass check of the object's type. As a result, `isinstance` will also fail with an exception in these conditions.

The above only seemed to occur in Python 3.

This PR fixes the issue in the Python 3 implementation while adding test coverage for both Python 2 and 3 that ensures that functions (and `types.FunctionType`) can be correctly compared both against protocols that they legitimately implement as well as those that they do not implement.

### DIFF
--- a/typing_extensions/src_py2/test_typing_extensions.py
+++ b/typing_extensions/src_py2/test_typing_extensions.py
@@ -5,6 +5,7 @@ import contextlib
 import collections
 import pickle
 import subprocess
+import types
 from unittest import TestCase, main, skipUnless
 
 from typing_extensions import NoReturn, ClassVar
@@ -234,19 +235,34 @@ class ProtocolTests(BaseTestCase):
         class D(object):
             def meth(self):
                 pass
+        def f():
+            pass
         self.assertIsSubclass(D, P)
         self.assertIsInstance(D(), P)
         self.assertNotIsSubclass(C, P)
         self.assertNotIsInstance(C(), P)
+        self.assertNotIsSubclass(types.FunctionType, P)
+        self.assertNotIsInstance(f, P)
 
     def test_everything_implements_empty_protocol(self):
         @runtime
         class Empty(Protocol): pass
         class C(object): pass
-        for thing in (object, type, tuple, C):
+        def f():
+            pass
+        for thing in (object, type, tuple, C, types.FunctionType):
             self.assertIsSubclass(thing, Empty)
-        for thing in (object(), 1, (), typing):
+        for thing in (object(), 1, (), typing, f):
             self.assertIsInstance(thing, Empty)
+
+    def test_function_implements_protocol(self):
+        @runtime
+        class Function(Protocol):
+            def __call__(self, *args, **kwargs):
+                pass
+        def f():
+            pass
+        self.assertIsInstance(f, Function)
 
     def test_no_inheritance_from_nominal(self):
         class C(object): pass
--- a/typing_extensions/src_py3/test_typing_extensions.py
+++ b/typing_extensions/src_py3/test_typing_extensions.py
@@ -319,6 +319,11 @@ class XRepr(NamedTuple):
     def __add__(self, other):
         return 0
 
+@runtime
+class FunctionProtocol(Protocol):
+    __call__: typing.Callable
+
+
 async def g_with(am: AsyncContextManager[int]):
     x: int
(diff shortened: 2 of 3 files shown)


the-knights-who-say-ni:

erikwright:
I've made some inquiries about signing the CLA as a representative of Shopify. I should be able to get a response by Tuesday.

ilevkivskyi:
Hi @erikwright, what s the situation here? Have signed the CLA?

erikwright:
It's taking a bit more time than I expected. Let me send another follow-up.

erikwright:
I submitted the CLA today. I gather that I need to wait up to 24h before it is processed. I'll check back tomorrow.

erikwright:
@ilevkivskyi The CLA was received and I fixed up the tests so they run on all the releases.

PR #583, opened 2018-09-12
Add Final to typing_extensions

ilevkivskyi:
This is a runtime counterpart of an experimental feature added to mypy in https://github.com/python/mypy/pull/5522

This implementation just mimics the behaviour of `ClassVar` on all Python/`typing` versions, which is probably the most reasonable thing to do.

### DIFF
--- a/typing_extensions/src_py2/test_typing_extensions.py
+++ b/typing_extensions/src_py2/test_typing_extensions.py
@@ -128,11 +128,11 @@ class FinalTests(BaseTestCase):
             Final[int][str]
 
     def test_repr(self):
-        self.assertEqual(repr(Final), 'typing.Final')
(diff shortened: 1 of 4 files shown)


JukkaL:
If this will be moved to `typing`, I hope there is a way to share code with the implementation of `ClassVar`. This is a lot of code for a relatively minor feature, but it's less of a problem if most of it goes away in the long term.

ilevkivskyi:
In Python 3.7 it will be literally one line (not counting the docstring). But the point of `typing_extensions` is to be compatible with _all_ versions of `typing`.

ilevkivskyi:
OK, if there are no suggestions, I will merge this later today. We can release a new version of `typing_extensions` later this week.

PR #591, opened 2018-11-17
Add Literal[...] types to typing_extensions

Michael0x2a:
This diff adds a preliminary draft of literal types to `typing_extensions`. I intentionally designed `Literal[...]` so it does not attempt to perform any runtime checks on the provided parameter, to make it easier to spand the scope of what `Literal` can do in the future.

You are, however, prohibited from subclassing `Literal[...]` or performing `isinstance`/`issubclass` checks on it, as usual.

For background context, here's the associated PEP draft: https://github.com/Michael0x2a/peps/pull/1

### DIFF
--- a/typing_extensions/src_py2/typing_extensions.py
+++ b/typing_extensions/src_py2/typing_extensions.py
@@ -216,7 +216,7 @@ class _Literal(typing._FinalTypingBase):
     The type checker understands that 'var' is literally equal to the
     value 4 and no other value.
 
-    Literal[..] cannot be subclassed. There is no runtime checking
+    Literal[...] cannot be subclassed. There is no runtime checking
     verifying that the parameter is actually a value instead of a type.
     """
 
--- a/typing_extensions/src_py3/typing_extensions.py
+++ b/typing_extensions/src_py3/typing_extensions.py
@@ -525,7 +525,7 @@ elif sys.version_info[:2] >= (3, 7):
         The type checker understands that 'var' is literally equal to the
         value 4 and no other value.
 
-        Literal[..] cannot be subclassed. There is no runtime checking
+        Literal[...] cannot be subclassed. There is no runtime checking
         verifying that the parameter is actually a value instead of a type.
         """)
 elif hasattr(typing, '_FinalTypingBase'):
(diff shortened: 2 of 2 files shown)


typing_extensions/src_py2/typing_extensions.py
```diff
@@ -196,6 +197,67 @@ class Other(Leaf):  # Error reported by type checker
     return f
 
 
+class LiteralMeta(TypingMeta):
```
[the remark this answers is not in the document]
Michael0x2a: Hmm, good point, fixed. I saw that we had a few other non-private meta classes as well (FinalMeta, ClassVarMeta, NoReturnMeta...) and fixed those as well.

typing_extensions/src_py3/typing_extensions.py
```diff
@@ -503,6 +504,142 @@ class Other(Leaf):  # Error reported by type checker
...
+if hasattr(typing, 'Literal'):
+    Literal = typing.Literal 
+elif sys.version_info[:2] >= (3, 7):
+    class _LiteralForm(typing._SpecialForm, _root=True):
+
+        def __repr__(self):
+            return 'typing_extensions.' + self._name
+
+        def __getitem__(self, parameters):
+            return _GenericAlias(self, parameters)
```
ilevkivskyi: Logically this is more like `_VariadicGenericAlias`, but since we don't add any special meaning for `...` (Ellipsis), this is OK.

typing_extensions/src_py3/typing_extensions.py
```diff
@@ -503,6 +504,142 @@ class Other(Leaf):  # Error reported by type checker
...
+        """A type that can be used to indicate to type checkers that the
+        corresponding value has a value literally equivalent to the
+        provided parameter. For example:
+
+            var: Literal[4] = 4
+        
+        The type checker understands that 'var' is literally equal to the
+        value 4 and no other value.
+
+        Literal[..] cannot be subclassed. There is no runtime checking
```
ilevkivskyi: Are two dots (not three) intentional? It is a bit unusual.

Michael0x2a: Whoops, that was an oversight -- fixed!

PR #605, opened 2019-01-12
Bump typing_extensions version

ilevkivskyi:
Preparing for the release. I bumped to 3.7.2 to follow the logic (or tradition) of matching the version of the latest supported Python version.


ilevkivskyi:
If there will be no objections, then I will merge this and release `typing_extensions` later today.

PR #615, opened 2019-02-14
Run the extension tests using the current Python executable

mr-c:
Not whatever "python" might be.


mr-c:
@ilevkivskyi Thanks! This was needed for the Debian package

PR #619, opened 2019-03-07
Add Final to the README

codergator:
It looks like `Final` fell between the cracks. It's my favorite experimental type. Adding it to the list.


gvanrossum:
Thanks!

PR #628, opened 2019-04-30
Add TypedDict to typing_extensions

ilevkivskyi:
This is almost a verbatim copy of the code in `mypy_extensions` plus few small tweaks due to different imports from `typing`.

### DIFF
--- a/typing_extensions/src_py2/typing_extensions.py
+++ b/typing_extensions/src_py2/typing_extensions.py
@@ -581,22 +581,22 @@ TypedDict = _TypedDictMeta('TypedDict', (dict,), {})
 TypedDict.__module__ = __name__
 TypedDict.__doc__ = \
     """A simple typed name space. At runtime it is equivalent to a plain dict.
+
     TypedDict creates a dictionary type that expects all of its
     instances to have a certain set of keys, with each key
     associated with a value of a consistent type. This expectation
-    is not checked at runtime but is only enforced by typecheckers.
+    is not checked at runtime but is only enforced by type checkers.
     Usage::
+
         Point2D = TypedDict('Point2D', {'x': int, 'y': int, 'label': str})
+
         a: Point2D = {'x': 1, 'y': 2, 'label': 'good'}  # OK
         b: Point2D = {'z': 3, 'label': 'bad'}           # Fails type check
+
         assert Point2D(x=1, y=2, label='first') == dict(x=1, y=2, label='first')
+
     The type info could be accessed via Point2D.__annotations__. TypedDict
-    supports two additional equivalent forms::
+    supports an additional equivalent form::
+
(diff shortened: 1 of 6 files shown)


typing_extensions/src_py3/typing_extensions.py
```diff
@@ -1507,3 +1507,84 @@ def runtime(cls):
...
+
+TypedDict = _TypedDictMeta('TypedDict', (dict,), {})
+TypedDict.__module__ = __name__
+TypedDict.__doc__ = \
+    """A simple typed name space. At runtime it is equivalent to a plain dict.
+    TypedDict creates a dictionary type that expects all of its
+    instances to have a certain set of keys, with each key
+    associated with a value of a consistent type. This expectation
+    is not checked at runtime but is only enforced by typecheckers.
+    Usage::
```
JukkaL: See my comments above. They apply here as well.

typing_extensions/src_py2/typing_extensions.py
```diff
@@ -519,3 +519,84 @@ def runtime(cls):
...
+            tp_dict.__total__ = total
+        return tp_dict
+
+    __instancecheck__ = __subclasscheck__ = _check_fails
+
+
+TypedDict = _TypedDictMeta('TypedDict', (dict,), {})
+TypedDict.__module__ = __name__
+TypedDict.__doc__ = \
+    """A simple typed name space. At runtime it is equivalent to a plain dict.
```
JukkaL: Maybe add an empty line after the first sentence?

typing_extensions/src_py2/typing_extensions.py
```diff
@@ -519,3 +519,84 @@ def runtime(cls):
...
+    TypedDict creates a dictionary type that expects all of its
+    instances to have a certain set of keys, with each key
+    associated with a value of a consistent type. This expectation
+    is not checked at runtime but is only enforced by typecheckers.
+    Usage::
+        Point2D = TypedDict('Point2D', {'x': int, 'y': int, 'label': str})
+        a: Point2D = {'x': 1, 'y': 2, 'label': 'good'}  # OK
+        b: Point2D = {'z': 3, 'label': 'bad'}           # Fails type check
+        assert Point2D(x=1, y=2, label='first') == dict(x=1, y=2, label='first')
+    The type info could be accessed via Point2D.__annotations__. TypedDict
```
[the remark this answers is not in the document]
ilevkivskyi: > Do we want to make `__annotations__` part of the public API?

It is already a public API for class syntax according to PEP 526. Also it may be already used by some runtime type validators (at least `typing_inspect` uses it).

ilevkivskyi:
I am not sure what happened with pytest on Python 3.3, tests passed after I restarted it. We might want to just drop 3.3 support.

PR #630, opened 2019-05-17
Add SupportsIndex

pcd1193182:
This PR adds a runtime SupportsIndex protocol that can be used to check at runtime if something supports the \_\_index\_\_ function, which is used in python to implement `hex()`, `oct()`, and `bin()`. This would allow typeshed to enhance the code contributed in https://github.com/python/typeshed/pull/2996.

I have run the python2.7 tests locally, but have had trouble getting other versions of python installed to run the 3.3-3.6 tests.  This is a pretty simple copy and modification of SupportsInt, so I imagine it will pass the tests, but still.

### DIFF
--- a/python2/test_typing.py
+++ b/python2/test_typing.py
@@ -541,6 +541,10 @@ class ProtocolTests(BaseTestCase):
         self.assertIsSubclass(list, typing.Reversible)
         self.assertNotIsSubclass(int, typing.Reversible)
 
+    def test_supports_index(self):
+        self.assertIsSubclass(int, typing.SupportsIndex)
+        self.assertNotIsSubclass(str, typing.SupportsIndex)
+
     def test_protocol_instance_type_error(self):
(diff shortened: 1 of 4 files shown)


the-knights-who-say-ni:

ilevkivskyi:
@pcd1193182 Please note that the current status of this repo is _backport_. If you want to add this to `typing`, then you need to first open an issue on http://bugs.python.org, then make a PR to CPython repo, then after that you can backport it to older versions.

Also you need to hurry if you want this in Python 3.8, the feature freeze is in one week!

pcd1193182:
I have filed a PR in the cpython repo: https://github.com/python/cpython/pull/13448

pcd1193182:
Given that this has now landed in cpython, is it a candidate for inclusion here? Not sure exactly what the preferred workflow is.

PR #632, opened 2019-05-23
Add a python 3.7+ version of Annotated to typing_extensions (#1014)

till-varoquaux:
First commits for PEP 593..

### DIFF
--- a/typing_extensions/src_py3/test_typing_extensions.py
+++ b/typing_extensions/src_py3/test_typing_extensions.py
@@ -1536,12 +1536,11 @@ if HAVE_ANNOTATED:
                 x = 1
 
             G = Annotated[_Annotated_test_G[int], "A decoration"]
+            G.foo = 42
+            G.bar = 'abc'
 
-            c = G()
-            c.foo = 42
-            c.bar = 'abc'
             for proto in range(pickle.HIGHEST_PROTOCOL + 1):
-                z = pickle.dumps(c, proto)
+                z = pickle.dumps(G, proto)
                 x = pickle.loads(z)
                 self.assertEqual(x.foo, 42)
                 self.assertEqual(x.bar, 'abc')
--- a/typing_extensions/src_py3/test_typing_extensions.py
+++ b/typing_extensions/src_py3/test_typing_extensions.py
@@ -9,7 +9,7 @@ import types
 from unittest import TestCase, main, skipUnless
 from typing import TypeVar, Optional
 from typing import T, KT, VT  # Not in __all__.
-from typing import Tuple, List
+from typing import Tuple, List, Dict
 from typing import Generic
 from typing import get_type_hints
 from typing import no_type_check
@@ -65,6 +65,8 @@ PY36 = sys.version_info[:2] >= (3, 6)
 # Protocols are hard to backport to the original version of typing 3.5.0
 HAVE_PROTOCOLS = sys.version_info[:3] != (3, 5, 0)
 
+# Not backported to older versions yet
+HAVE_ANNOTATED = PEP_560
 
 class BaseTestCase(TestCase):
     def assertIsSubclass(self, cls, class_or_tuple, msg=None):
@@ -1458,6 +1460,87 @@ class TypedDictTests(BaseTestCase):
             self.assertEqual(Options.__total__, False)
 
 
+if HAVE_ANNOTATED:
+    from typing_extensions import Annotated
+
+    class AnnotatedTests(BaseTestCase):
+
+        def test_repr(self):
+            self.assertEqual(
+                repr(Annotated[int, 4, 5]),
+                "typing_extensions.Annotated[int, 4, 5]"
+            )
+
+        def test_flatten(self):
+            A = Annotated[Annotated[int, 4], 5]
+            self.assertEqual(A, Annotated[int, 4, 5])
+            self.assertEqual(A.__extras__, (4, 5))
+            self.assertEqual(A.__origin__, int)
+
+        def test_hash_eq(self):
+            self.assertEqual(len({Annotated[int, 4, 5], Annotated[int, 4, 5]}), 1)
+            self.assertNotEqual(Annotated[int, 4, 5], Annotated[int, 5, 4])
+            self.assertNotEqual(Annotated[int, 4, 5], Annotated[str, 4, 5])
+            self.assertNotEqual(Annotated[int, 4], Annotated[int, 4, 4])
+            self.assertEqual(
+                {Annotated[int, 4, 5], Annotated[int, 4, 5], Annotated[T, 4, 5]},
+                {Annotated[int, 4, 5], Annotated[T, 4, 5]}
+            )
+
+        def test_instantiate(self):
+            class C:
+                def __init__(self, x):
+                    self.x = x
+
+                def __eq__(self, other):
+                    if not isinstance(other, C):
+                        return NotImplemented
+                    return other.x == self.x
+
+            self.assertEqual(C(5), Annotated[C, "a decoration"](5))
+
+        def test_cannot_subclass(self):
+            with self.assertRaises(TypeError):
+                class C(Annotated):
+                    pass
+
+        def test_pickle(self):
+            x = Annotated[List[int], "a"]
+            for prot in range(pickle.HIGHEST_PROTOCOL + 1):
+                with self.subTest(protocol=prot):
+                    pickled = pickle.dumps(x, prot)
+                    restored = pickle.loads(pickled)
+                    self.assertEqual(x, restored)
+
+        def test_subst(self):
+            dec = "a decoration"
+
+            S = Annotated[T, dec]
+            self.assertEqual(S[int], Annotated[int, dec])
+
+            L = Annotated[List[T], dec]
+            self.assertEqual(L[int], Annotated[List[int], dec])
+            with self.assertRaises(TypeError):
+                L[int, int]
+
+            D = Annotated[Dict[KT, VT], dec]
+            self.assertEqual(D[str, int], Annotated[Dict[str, int], dec])
+            with self.assertRaises(TypeError):
+                D[int]
+
+            I = Annotated[int, dec]
+            with self.assertRaises(TypeError):
+                I[None]
+
+            LI = L[int]
+            with self.assertRaises(TypeError):
+                LI[None]
+
+        def test_annotated_in_other_types(self):
+            X = List[Annotated[T, 5]]
+            self.assertEqual(X[int], List[Annotated[int, 5]])
+
+
 class AllTests(BaseTestCase):
 
     def test_typing_extensions_includes_standard(self):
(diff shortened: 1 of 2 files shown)


typing_extensions/src_py3/typing_extensions.py
```diff
@@ -1595,3 +1604,93 @@ class Point2D(TypedDict):
...
+
+
+if HAVE_ANNOTATED:
+    class _Annotated(typing._GenericAlias, _root=True):
+        """Runtime representation of an annotated type.
+
+        At its core 'Annoted[t, dec1, dec2...]' is an alias for the type 't'
+        with extra annotations. The alias behaves like a normal typing alias,
+        instantiating is the same as instantiating the underlying type, binding
+        it to types is also the same...
```
ilevkivskyi:
```suggestion
        it to types is also the same.
```

typing_extensions/src_py3/typing_extensions.py
```diff
@@ -1595,3 +1604,93 @@ class Point2D(TypedDict):
     The class syntax is only supported in Python 3.6+, while two other
     syntax forms work for Python 2.7 and 3.2+
     """
+
+
+if HAVE_ANNOTATED:
+    class _Annotated(typing._GenericAlias, _root=True):
+        """Runtime representation of an annotated type.
+
+        At its core 'Annoted[t, dec1, dec2...]' is an alias for the type 't'
```
ilevkivskyi:
```suggestion
        At its core 'Annoted[t, dec1, dec2, ...]' is an alias for the type 't'
```

typing_extensions/src_py3/typing_extensions.py
```diff
@@ -1595,3 +1604,93 @@ class Point2D(TypedDict):
...
+
+        def __init__(self, origin, extras):
+            if isinstance(origin, _Annotated):
+                extras = origin.__extras__ + extras
+                origin = origin.__origin__
+            super().__init__(origin, origin)
+            self.__extras__ = extras
+
+        def copy_with(self, params):
+            new_type, = params
```
ilevkivskyi: It is easy to miss the come, I would rather write `params[0]` (you can also add something like `assert len(params) == 1`).

typing_extensions/src_py3/typing_extensions.py
```diff
@@ -1595,3 +1604,93 @@ class Point2D(TypedDict):
...
+                return NotImplemented
+            if self.__origin__ != other.__origin__:
+                return False
+            return self.__extras__ == other.__extras__
+
+        def __hash__(self):
+            return hash((self.__origin__, self.__extras__))
+
+
+    class Annotated:
```
ilevkivskyi: I would make `Annotated` a `_SpecialForm` on Python 3.7 (like `Optional`, `Any`, etc.), it doesn't need to be a class. But this is not important for the PoC implementation if it is tricky.

typing_extensions/src_py3/typing_extensions.py
```diff
@@ -1595,3 +1604,93 @@ class Point2D(TypedDict):
...
+        __slots__ = ()
+
+        def __new__(cls, *args, **kwargs):
+            raise TypeError("Type Annotated cannot be instantiated.")
+
+
+        @typing._tp_cache
+        def __class_getitem__(cls, params):
+            if not isinstance(params, tuple) or len(params) < 2:
+                raise TypeError("Annotated[...] should be instantiated "
```
ilevkivskyi:
```suggestion
                raise TypeError("Annotated[...] should be used "
```

typing_extensions/src_py3/typing_extensions.py
```diff
@@ -1595,3 +1604,93 @@ class Point2D(TypedDict):
...
+
+        - It's an error to call `Annotated` with less than two arguments.
+        - Nested Annotated are flattened::
+
+            Annotated[Annotated[int, Ann1, Ann2], Ann3] == Annotated[int, Ann1, Ann2, Ann3]
+
+        - Instantiating an annotated type is equivalent to instantiating the
+        underlying type::
+
+            Annotated[C, Ann1](5) == C(5)
```
ilevkivskyi: Should we also mentioned the generic alias use case? It expect it will be a common one. For example:
```python
Optimized = Annotated[T, runtime.Optimize]
x: Optimized[int]
OptimizedList = Annotated[List[T], runtime.Optimize]
xs: OptimizedList[int]
```

typing_extensions/src_py3/typing_extensions.py
```diff
@@ -1595,3 +1604,93 @@ class Point2D(TypedDict):
...
+if HAVE_ANNOTATED:
+    class _Annotated(typing._GenericAlias, _root=True):
+        """Runtime representation of an annotated type.
+
+        At its core 'Annoted[t, dec1, dec2...]' is an alias for the type 't'
+        with extra annotations. The alias behaves like a normal typing alias,
+        instantiating is the same as instantiating the underlying type, binding
+        it to types is also the same...
+        """
+
```
ilevkivskyi: This empty line is not needed.

typing_extensions/src_py3/test_typing_extensions.py
```diff
@@ -1458,6 +1460,87 @@ def test_total(self):
...
+            self.assertEqual(len({Annotated[int, 4, 5], Annotated[int, 4, 5]}), 1)
+            self.assertNotEqual(Annotated[int, 4, 5], Annotated[int, 5, 4])
+            self.assertNotEqual(Annotated[int, 4, 5], Annotated[str, 4, 5])
+            self.assertNotEqual(Annotated[int, 4], Annotated[int, 4, 4])
+            self.assertEqual(
+                {Annotated[int, 4, 5], Annotated[int, 4, 5], Annotated[T, 4, 5]},
+                {Annotated[int, 4, 5], Annotated[T, 4, 5]}
+            )
+
+        def test_instantiate(self):
```
ilevkivskyi: I would add more tests for instantiations (for example involving generic classes, and subscripted generic classes).

Also probably for attribute access on class objects (i.e. `A = Annotated[C, 5]; A.x == C.x`)?

typing_extensions/src_py3/test_typing_extensions.py
```diff
@@ -1458,6 +1460,87 @@ def test_total(self):
...
+                    return other.x == self.x
+
+            self.assertEqual(C(5), Annotated[C, "a decoration"](5))
+
+        def test_cannot_subclass(self):
+            with self.assertRaises(TypeError):
+                class C(Annotated):
+                    pass
+
+        def test_pickle(self):
```
ilevkivskyi: I would add more pickling tests: involving more base types (like user-defined classes, unsubscripted generics, unions nested generics), and involving subscripting of and attribute access on restored objects.

typing_extensions/src_py3/typing_extensions.py
```diff
@@ -1595,3 +1604,93 @@ class Point2D(TypedDict):
...
+        def __init__(self, origin, extras):
+            if isinstance(origin, _Annotated):
+                extras = origin.__extras__ + extras
+                origin = origin.__origin__
+            super().__init__(origin, origin)
+            self.__extras__ = extras
```
ilevkivskyi: I would not call this `__extras__` to avoid confusion with an attribute with similar name in older versions of `typing`. Maybe call it `__metadata__`?

typing_extensions/src_py3/typing_extensions.py
```diff
@@ -1595,3 +1604,93 @@ class Point2D(TypedDict):
...
+            if not isinstance(params, tuple) or len(params) < 2:
+                raise TypeError("Annotated[...] should be instantiated "
+                                "with at least two arguments (a type and an "
+                                "annotation).")
+            msg = "Annotated[t, ...]: t must be a type."
+            origin = typing._type_check(params[0], msg)
+            extras = tuple(params[1:])
+            return _Annotated(origin, extras)
+
+
```
ilevkivskyi: One empty line should be enough.

typing_extensions/src_py3/test_typing_extensions.py
```diff
@@ -1458,6 +1460,87 @@ def test_total(self):
...
+
+            LI = L[int]
+            with self.assertRaises(TypeError):
+                LI[None]
+
+        def test_annotated_in_other_types(self):
+            X = List[Annotated[T, 5]]
+            self.assertEqual(X[int], List[Annotated[int, 5]])
+
+
```
ilevkivskyi: Maybe add couple more substitution tests with special forms like `Union` and/or `Optional`.

typing_extensions/src_py3/typing_extensions.py
jstasiak:
```suggestion
        At its core 'Annotated[t, dec1, dec2,...]' is an alias for the type 't'
```

typing_extensions/src_py3/typing_extensions.py
till-varoquaux: _SpecialForm seems to be a of a no-go unless we change the standard library; it just special cases everything based on class name.

till-varoquaux:
Since I'm new to this code base I'm proceeding very incrementally. The get_type_hints diffs will come in a separate batch.

typing_extensions/src_py3/test_typing_extensions.py
```diff
@@ -1458,6 +1460,183 @@ def test_total(self):
...
+                x = 1
+
+            G = Annotated[_Annotated_test_G[int], "A decoration"]
+
+            c = G()
+            c.foo = 42
+            c.bar = 'abc'
+            for proto in range(pickle.HIGHEST_PROTOCOL + 1):
+                z = pickle.dumps(c, proto)
+                x = pickle.loads(z)
```
ilevkivskyi: Here you are testing pickling of a normal instance, instead you should (un-)pickle `G` itself.

typing_extensions/src_py3/test_typing_extensions.py
```diff
@@ -1458,6 +1460,183 @@ def test_total(self):
...
+            global _Annotated_test_G
+
+            class _Annotated_test_G(Generic[T]):
+                x = 1
+
+            G = Annotated[_Annotated_test_G[int], "A decoration"]
+
+            c = G()
+            c.foo = 42
+            c.bar = 'abc'
```
ilevkivskyi: Assign these attributes on `G` instead.

typing_extensions/src_py3/typing_extensions.py
```diff
@@ -1595,3 +1604,149 @@ class Point2D(TypedDict):
     The class syntax is only supported in Python 3.6+, while two other
     syntax forms work for Python 2.7 and 3.2+
     """
+
+
+if HAVE_ANNOTATED:
+    class _Annotated(typing._GenericAlias, _root=True):
```
ilevkivskyi: Just a bit of bikesheding, I would call this class `_AnnotatedAlias`.

till-varoquaux:
Ok, we should be ready to go! Thank you Ivan and Jakub for your all of your feedback. If you want I can squash all of those commits before we merge this request.

ilevkivskyi:
We always squash ourselves, please don't do this.

You can now post in the PEP discussion thread and propose people to try this out and play with it (one can install the library direcrly from GitHub).

PR #633, opened 2019-05-28
Drop support of Python 3.3

ilevkivskyi:
This version had EoL a year ago. Also it is currently broken on master.


ilevkivskyi:
I am going to merge this without review just to un-break master.

Another thing I noticed is that README is quite outdated. We should probably update it.

PR #635, opened 2019-05-28
fix __origin__ in annotated types (and a couple of typos)

till-varoquaux:
The __origin__ field in _AnnotatedAlias should point to `Annotated`, not the underlying type.


jstasiak:
I've tested this change and it works as expected.

ilevkivskyi:
As I explained in the mailing list, I would rather keep the internal implementation simple, and instead add a little tweak to `get_origin()` and `get_args()`.

PR #637, opened 2019-05-30
Pep 593 py3

till-varoquaux:
This adds `Annotated` for python [3.4...3.6]. I tried hard to make it behave as closely as possible to the py3.7+ version but was not able to backport the `include_extras` in get_type_hint (we relied on `<type>.copy_with` to implement it).

### DIFF
--- a/typing_extensions/src_py3/test_typing_extensions.py
+++ b/typing_extensions/src_py3/test_typing_extensions.py
@@ -11,9 +11,8 @@ from typing import TypeVar, Optional
 from typing import T, KT, VT  # Not in __all__.
 from typing import Tuple, List, Dict
 from typing import Generic
-from typing import get_type_hints
 from typing import no_type_check
-from typing_extensions import NoReturn, ClassVar, Final, IntVar, Literal, Type, NewType, TypedDict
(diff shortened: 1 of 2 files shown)


till-varoquaux:
Will fix python 3.5.0-2 and then reopen.

ilevkivskyi:
> Will fix python 3.5.0-2 and then reopen.

Please note that 3.5.0 might be _really_ hard, so don't spend too much time on it. I remember I spent tons of time trying when working on protocols, but I ultimately gave up, so currently protocols are supported on most versions of Python _except 3.5.0_, see https://github.com/python/typing/blob/master/typing_extensions/src_py3/typing_extensions.py#L139.

till-varoquaux:
Ah; that's great news. I was pulling my hair out trying to backport this.

omry:
How do you access the Annotations on Python 3.6 without get_type_hints?

gvanrossum:
Use the __annotations__ attribute.

omry:
> Use the **annotations** attribute.

Oh, of course.
I didn't start coding yet so I missed that obvious thing (and I don't think the exact object model was described in [pep 593](https://www.python.org/dev/peps/pep-0593/)).

PR #642, opened 2019-06-06
Unify the implementation of annotated in src_py2 and src_py3

till-varoquaux:


typing_extensions/src_py3/typing_extensions.py
```diff
@@ -1853,31 +1855,18 @@ def __call__(self, *args, **kwargs):
...
             else:
                 setattr(self._get_cons(), attr, value)
 
-        # We overload this because the super() error message is confusing:
-        # Parametrized generics cannot be used ...
-        def __instancecheck__(self, obj):
-            raise TypeError("Annotated cannot be used with isinstance().")
-
-        def __subclasscheck__(self, cls):
-            raise TypeError("Annotated cannot be used with issubclass().")
```
ilevkivskyi: It looks like these are actually needed on _some_ versions, see test failures.

till-varoquaux: OOps, good catch. I really need to setup a better test system on my machine.

Issue #648, opened 2019-06-16
Add Protocol, runtime_checkable, Final, final, Literal, and TypedDict to typing on Python 2


ilevkivskyi:
These should be essentially just copied from `typing_extensions`.

PR #649, opened 2019-06-17
Move Protocol, runtime_checkable, Final, final, Literal, and TypedDict to typing

ilevkivskyi:
Fixes https://github.com/python/typing/issues/648

This essentially just moves a bunch of code and tests from Python 2 version of `typing_extensions` to Python 2 version of `typing`. We don't need to copy the definitions in the former, and can just re-export them from `typing`, since there is not built-in `typing` module on Python 2, so that we can always assume the latest version of `typing` (we will just need to release them in right order, `typing` first, and then update dependency in `typing_extensions`).

Notable changes apart from the code move:
* Like on Python 3, keep `runtime` in `typing_extensions` as an alias for `runtime_checkable` for backwards compatibility (the former is not included in `__all__`)
* Update `GenericMeta.__new__` and `GenericMeta.__getitem__` instead of almost identical overrides in `_ProtocolMeta`.


gvanrossum:
Where's the dependency on a version of typing.py specified in typing_extensions?

ilevkivskyi:
> Where's the dependency on a version of typing.py specified in typing_extensions?

In `typing_extensions/setup.py`.

ilevkivskyi:
```python
install_requires = []
if sys.version_info < (3, 5):
    install_requires.append('typing >= 3.6.2')
```

PR #662, opened 2019-07-31
Fix test failure with non-default Python interpreter

Kelledin:
`test_typing_compiles_with_opt()` is prone to failure if building and testing with an interpreter that doesn't match the default python in $PATH (for example, if running `python3 setup.py ...` when the default system-wide interpreter is Python 2.x).

Using `sys.executable` (vs hardcoded unversioned "python") seems to be the most sensible solution.

### DIFF
--- a/python2/test_typing.py
+++ b/python2/test_typing.py
@@ -2676,9 +2676,8 @@ class AllTests(BaseTestCase):
         file_path = os.path.join(os.path.dirname(os.path.realpath(__file__)),
                                  'typing.py')
         try:
-            subprocess.check_output('python -OO {}'.format(file_path),
-                                    stderr=subprocess.STDOUT,
-                                    shell=True)
+            subprocess.check_output([sys.executable, '-OO', file_path],
+                                    stderr=subprocess.STDOUT)
         except subprocess.CalledProcessError:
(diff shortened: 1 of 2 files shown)


the-knights-who-say-ni:

Our records indicate we have not received your CLA. For legal reasons we need you to sign this before we can look at your contribution. Please follow [the steps outlined in the CPython devguide](https://devguide.python.org/pullrequest/#licensing) to rectify this issue.

Kelledin:
So...simple quoting appears to address spaces in the interpreter path.  Then I thought about it little more and decided we might as well account for spaces in file_path as well.

If we want to handle any more complex corner cases (like quote characters embedded in file names), we'd probably be better off breaking the command into an array and leaving shell=False.  Unless there's any particular reason we set shell=True?

gvanrossum:
I don't see any deep reason why `shell=True` was used here except convenience. So please go ahead!

ilevkivskyi:
@Kelledin Please don't force-push because GitHub doesn't send a notification in this case.

@gvanrossum It looks like your last comment is now addressed. So I think this can be merged now.

gvanrossum:
Thanks both!
-- 
--Guido (mobile)

Issue #600, opened 2018-12-13
Add support for external annotations in the typing module


till-varoquaux:
We propose adding an `Annotated` type to the typing module to decorate existing types with context-specific metadata. Specifically, a type `T` can be annotated with metadata `x` via the typehint `Annotated[T, x]`. This metadata can be used for either static analysis or at runtime. If a library (or tool) encounters a typehint `Annotated[T, x]` and has no special logic for metadata `x`, it should ignore it and simply treat the type as `T`. Unlike the `no_type_check` functionality that current exists in the `typing` module which completely disables typechecking annotations on a function or a class, the `Annotated` type allows for both static typechecking of `T` (e.g., via MyPy or Pyre, which can safely ignore `x`)  together with runtime access to `x` within a specific application. We believe that the introduction of this type would address a diverse set of use cases of interest to the broader Python community.

### READING binary data

The `struct` module provides a way to read and write C structs directly from their byte representation. It currently relies on a string representation of the C type to read in values:

```py
record = b'raymond   \x32\x12\x08\x01\x08'
name, serialnum, school, gradelevel = unpack('<10sHHb', record)
```

The [documentation](https://docs.python.org/3/library/struct.html#examples) suggests using a named tuple to unpack the values and make this a bit more tractable:

```py
from collections import namedtuple
Student = namedtuple('Student', 'name serialnum school gradelevel')
Student._make(unpack('<10sHHb', record))
# Student(name=b'raymond   ', serialnum=4658, school=264, gradelevel=8)
```

However, this recommendation is somewhat problematic; as we add more fields, it's going to get increasingly tedious to match the properties in the named tuple with the arguments in `unpack`.

Instead, annotations can provide better interoperability with a type checker or an IDE without adding any special logic outside of the `struct` module:

```py
from typing import NamedTuple
UnsignedShort = Annotated[int, struct.ctype('H')]
SignedChar = Annotated[int, struct.ctype('b')]

@struct.packed
class Student(NamedTuple):
  # MyPy typechecks 'name' field as 'str' 
  name: Annotated[str, struct.ctype("<10s")]
  serialnum: UnsignedShort
  school: SignedChar
  gradelevel: SignedChar

# 'unpack' only uses the metadata within the type annotations
Student.unpack(record))
# Student(name=b'raymond   ', serialnum=4658, school=264, gradelevel=8)
```

### dataclasses

[...]

PR #666, opened 2019-08-07
Try fixing Travis build

ilevkivskyi:


ilevkivskyi:
@gvanrossum This fixes the red master (something changed in Travis environment). The fix also discovered that `flake8` was not run for long time, so I also fix it.

Could you please take a look?

gvanrossum:
> Could you please take a look?

I would, but I am a place with internet so marginal that I can't pull up
GitHub.

Maybe Friday, otherwise the weekend.

typing_extensions/src_py3/typing_extensions.py
```diff
@@ -363,21 +368,20 @@ def __getitem__(self, parameters):
                                       '{} accepts only single type'.format(self._name))
             return _GenericAlias(self, (item,))
 
-    Final = _FinalForm('Final', doc=
-        """A special typing construct to indicate that a name
-        cannot be re-assigned or overridden in a subclass.
-        For example:
+    Final = _FinalForm('Final',
+                       doc="""A special typing construct to indicate that a name
+                       cannot be re-assigned or overridden in a subclass.
```
brugz: A style similar is what I'm recommending for `Annotated` on https://github.com/python/typing/issues/600

typing_extensions/src_py3/typing_extensions.py
```diff
@@ -1220,8 +1224,10 @@ def _proto_hook(other):
                                 return NotImplemented
                             break
                         annotations = getattr(base, '__annotations__', {})
-                        if (isinstance(annotations, typing.Mapping) and attr in annotations and
-                                isinstance(other, _ProtocolMeta) and other._is_protocol):
+                        if (isinstance(annotations, typing.Mapping) and
```
brugz: I think if all() may work here to cut down on the superfluous `and`'s

typing_extensions/src_py3/typing_extensions.py
```diff
@@ -1506,8 +1516,10 @@ def _proto_hook(other):
...
```
brugz: Same here

gvanrossum:
> Whats the #noqa tag actually doing?

That's indicating to linters (e.g. flake8, pylint) that they should not complain about that line. There are several tests that intentionally put bad code in a string type annotation or in a type comment (to verify that the typing module handles these correctly), and some linters are smart enough to look inside type comments or string annotations and complain, making the build fail. So the `# noqa` shuts up the linter and so the build passes.

typing_extensions/src_py3/typing_extensions.py
gvanrossum: (This isn't really a code review comment, it's a cross reference, so resolving -- the cross reference will still be in #666.)

typing_extensions/src_py3/typing_extensions.py
gvanrossum: I don't see how that could work since the `and` operator here is used for the benefit of its shortcut behavior (i.e. if the first condition fails, following conditions aren't evaluated). The only way that I can see `all()` used for something like `X and Y and Z` would be `all((X, Y, Z))` and that would evaluate all of `X`, `Y` and `Z` before calling `all()`.

src/test_typing.py
```diff
@@ -1564,7 +1564,7 @@ def foo(a: 'Noode[T]'):
     def test_no_type_check(self):
 
         @no_type_check
-        def foo(a: 'whatevers') -> {}:
+        def foo(a: 'int') -> {}:
```
gvanrossum: Hum, I thought it was intentional that this test references something that doesn't exist. If a linter complains it is technically in violation of PEP 484 (by not understanding the `@no_type_check` decorator), but the pragmatic is to add yet another `# noqa`.

ilevkivskyi: Indeed, I will switch these to `# noqa` now.

typing_extensions/src_py3/typing_extensions.py
brugz: That's curious; I suppose the implication here is that all() does not implement shortcut behavior. I wonder if maybe it should? I'll see if I can explore this a bit.

brugz: Well, actually, now I'm just unsure if maybe I'm not understanding your point gvanrossum@; to me, it looks like all() supports 'short-circuiting' See also: https://www.geeksforgeeks.org/short-circuiting-techniques-python/

Michael0x2a: The tuple will be fully evaluated before it's passed into `all(...)`-- doing `all((X, Y, Z))` is basically the same thing as doing `temp = (X, Y, Z); all(temp)`. So, the short-circuiting isn't helpful here.

If we want to take advantage of `all(...)`'s short-circuiting behavior, we'd need to instead pass in some sort of iterator that computes and yield values on demand. For example, we could use generator comprehensions and do something like `if all(isinstance(x, str) for x in args): ...` -- generators are a kind of iterator. But in this case, the boolean conditions are heterogeous, which makes writing this kind of nicely compact generator comprehension pretty much impossible. (Give it a shot if you're not sure why).

More broadly, a good rule of thumb is that `all(...)` is useful primarily when you have some loop that checks whether every bool in some iterable is true and you want to compress that loop into a one-liner -- it's easy to rewrite simple loops as comprehensions. (And the corollary is that if you can't rewrite the checks you want to perform into some sort of loop, `all(...)` is unlikely to be helpful.)

Issue #656, opened 2019-07-05
Tests failing on Python 2


lwfitzgerald:
Tests are failing on the `3.7.4` tag on Python 2
```
# python --version
Python 2.7.13 :: Continuum Analytics, Inc.
```
```
# python test_typing.py
.................................................................................................................................................................E.......E..E..........................................................
======================================================================
ERROR: test_protocol_instance_works (__main__.ProtocolTests)
----------------------------------------------------------------------
Traceback (most recent call last):
  File "test_typing.py", line 1057, in test_protocol_instance_works
    self.assertNotIsInstance('no', typing.SupportsAbs)
  File "/usr/lib/python2.7/unittest/case.py", line 971, in assertNotIsInstance
    if isinstance(obj, cls):
  File "/home/lwfitzgerald/typing/python2/typing.py", line 1824, in __instancecheck__
    issubclass(instance.__class__, self)):
  File "/home/lwfitzgerald/typing/python2/typing.py", line 1847, in __subclasscheck__
    return super(_ProtocolMeta, self).__subclasscheck__(cls)
  File "/home/lwfitzgerald/typing/python2/typing.py", line 1407, in __subclasscheck__
    return super(GenericMeta, self).__subclasscheck__(cls)
  File "/usr/lib/python2.7/abc.py", line 180, in __subclasscheck__
    if issubclass(subclass, scls):
  File "/home/lwfitzgerald/typing/python2/typing.py", line 1847, in __subclasscheck__
    return super(_ProtocolMeta, self).__subclasscheck__(cls)
  File "/home/lwfitzgerald/typing/python2/typing.py", line 1401, in __subclasscheck__
    raise TypeError("Parameterized generics cannot be used with class "
TypeError: Parameterized generics cannot be used with class or instance checks

======================================================================
ERROR: test_reversible (__main__.ProtocolTests)
----------------------------------------------------------------------
Traceback (most recent call last):
  File "test_typing.py", line 1049, in test_reversible
    self.assertNotIsSubclass(int, typing.Reversible)
  File "test_typing.py", line 42, in assertNotIsSubclass
    if issubclass(cls, class_or_tuple):
  File "/home/lwfitzgerald/typing/python2/typing.py", line 1847, in __subclasscheck__
    return super(_ProtocolMeta, self).__subclasscheck__(cls)
  File "/home/lwfitzgerald/typing/python2/typing.py", line 1407, in __subclasscheck__
    return super(GenericMeta, self).__subclasscheck__(cls)
[...]

PR #669, opened 2019-08-22
Fix isinstance() with generic protocol subclasses after subscripting

ilevkivskyi:
Fixes https://github.com/python/typing/issues/656

The problem is if a generic class inherits from a protocol, then `GenericMeta.__subclasscheck__()`
gets overridden with `_ProtocolMeta.__subclasscheck__()` that calls the former via `super()`, thus causing an extra frame.

### DIFF
--- a/python2/test_typing.py
+++ b/python2/test_typing.py
@@ -749,6 +749,27 @@ class ProtocolTests(BaseTestCase):
         self.assertIsInstance(C(1), P)
         self.assertIsInstance(C(1), PG)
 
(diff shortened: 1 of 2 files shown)


gvanrossum:
Be sure to also merge this upstream in the cpython repo.
-- 
--Guido (mobile)

ilevkivskyi:
> Be sure to also merge this upstream in the cpython repo.

The bug is Python 2 specific, but I wanted to add tests to `python/cpython` to prevent regressions.

PR #687, opened 2019-11-24
Track optional TypdeDict keys

Zac-HD:
Backport of https://github.com/python/cpython/pull/17214 ([BPO-38834](https://bugs.python.org/issue38834)) for @ilevkivskyi :smile:

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -2,9 +2,9 @@ language: python
 
 jobs:
   include:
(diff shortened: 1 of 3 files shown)


Zac-HD:
Looks like Travis doesn't have a `3.9-dev` yet, but should be good to go otherwise :smile:

PR #691, opened 2019-12-05
Add clarification to package description

### LINKED ISSUE #690: Misleading documentation when installing via `pip install -t . typing`
The following words are misleading because they do not hold true when installing via `pip install -t . typing`:

> NOTE: in Python 3.5 and later, the typing module lives in the stdlib,
> and installing this package has NO EFFECT.

Normally, the installation has no effect because Python standard library takes higher precedence than `site-packages`. 

However, when installing with `-t .`, `typing.py` is pulled in current directory and thus `import typing` will reach out to this library and break any Python 3.7 environment that uses `typing` with an error like:

```
AttributeError: type object 'Callable' has no attribute '_abc_registry'
  ...
  File "/var/task/sentry_sdk/_types.py", line 2, in <module>
    from typing import TYPE_CHECKING as MYPY
  File "/var/task/typing.py", line 1357, in <module>
    class Callable(extra=collections_abc.Callable, metaclass=CallableMeta):
  File "/var/task/typing.py", line 1005, in __new__
    self._abc_registry = extra._abc_registry
```

lingxiaoyang:
This PR closes #690.

- Explain why installing this package has no effect on Python >= 3.5
- Guide package maintainers to follow the preferred usage of this
  package, to avoid issues with pip install -t .

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -24,7 +24,8 @@ static and runtime type checkers, static analyzers, IDEs and other
 tools.
 
 NOTE: in Python 3.5 and later, the typing module lives in the stdlib,
-and installing this package has NO EFFECT.  To get a newer version of
+and installing this package has NO EFFECT, because stdlib takes higher
+precedence than the installation directory. To get a newer version of
 the typing module in Python 3.5 or later, you have to upgrade to a
 newer Python (bugfix) version.  For example, typing in Python 3.6.0 is
 missing the definition of 'Type' -- upgrading to 3.6.2 will fix this.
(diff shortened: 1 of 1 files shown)


the-knights-who-say-ni:

## Recognized GitHub username

@lingxiaoyang

lingxiaoyang:
Thanks @gvanrossum! I'm just reminding you my CLA is green now.

gvanrossum:
Thanks @lingxiaoyang!

PS: Next time don't force-push after your PR has been reviewed; force-pushing makes reviewing harder, and we squash all PRs anyway when we merge them.

PR #698, opened 2020-02-06
Backport get_origin() and get_args()

jstasiak:
The implementations come from CPython commit 427c84f13f77 with one small
change – the get_origin's docstring mentions Annotated as it's also
supported.

get_origin() and get_args() introduced in [1] and modified in [2] to
support Annotated.

[1] https://github.com/python/cpython/pull/13685
[2] https://github.com/python/cpython/pull/18260

### DIFF
--- a/typing_extensions/src_py3/typing_extensions.py
+++ b/typing_extensions/src_py3/typing_extensions.py
@@ -1995,7 +1995,7 @@ elif HAVE_ANNOTATED:
 
 # Python 3.8 has get_origin() and get_args() but those implementations aren't
 # Annotated-aware, so we can't use those, only Python 3.9 versions will do.
-if sys.version_info[:3] >= (3, 9, 0)
+if sys.version_info[:3] >= (3, 9, 0):
     get_origin = typing.get_origin
     get_args = typing.get_args
 elif PEP_560:
--- a/typing_extensions/src_py3/typing_extensions.py
+++ b/typing_extensions/src_py3/typing_extensions.py
@@ -151,7 +151,7 @@ __all__ = [
 HAVE_ANNOTATED = PEP_560 or SUBS_TREE
 
 if PEP_560:
-    __all__.append("get_type_hints")
+    __all__.extend(["get_args", "get_origin", "get_type_hints"])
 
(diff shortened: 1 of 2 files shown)


jstasiak:
Well, I'm not sure what to do regarding the build failure:

```
_____________ AllTests.test_typing_extensions_defers_when_possible _____________

self = <test_typing_extensions.AllTests testMethod=test_typing_extensions_defers_when_possible>

    def test_typing_extensions_defers_when_possible(self):

        exclude = {

            'overload',

            'Text',

            'TypedDict',

            'TYPE_CHECKING',

            'Final',

            'get_type_hints'

        }

        for item in typing_extensions.__all__:

            if item not in exclude and hasattr(typing, item):

>               self.assertIs(

                    getattr(typing_extensions, item),

                    getattr(typing, item))

E               AssertionError: <function get_args at 0x7fcb84063430> is not <function get_args at 0x7fcb8a0b3c10>

typing_extensions/src_py3/test_typing_extensions.py:1838: AssertionError
```

It happens because even though 3.8 has `get_args` and `get_origin` they don't support PEP 593 `Annotated` (only Python 3.9 `get_origin` and `get_args` do), so I redefine them in `typing_extensions` on 3.8 anyway, it's a bit of a mess. Is it acceptable to add them to the exclude list?

gvanrossum:
Yeah, it looks like it belongs in the exclude list (maybe only in 3.8.x?)

typing_extensions/src_py3/typing_extensions.py
```diff
@@ -1992,3 +1992,53 @@ class Annotated(metaclass=AnnotatedMeta):
             OptimizedList = Annotated[List[T], runtime.Optimize()]
             OptimizedList[int] == Annotated[List[int], runtime.Optimize()]
         """
+
+# Python 3.8 has get_origin() and get_args() but those implementations aren't
+# Annotated-aware, so we can't use those, only Python 3.9 versions will do.
+if sys.version_info[:3] >= (3, 9, 0):
```
gvanrossum: `sys.version_info[:2] >= (3, 9)`

PR #701, opened 2020-02-07
Add .vscode and .egg-info to gitignore

vemel:
### Changed

- Added `.vscode` to `.gitignore`
- Added `*.egg-info/` to `.gitignore`

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -2,9 +2,11 @@ MANIFEST
 build/
 dist/
 .tox/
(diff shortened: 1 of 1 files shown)


gvanrossum:
Sure. (Although not enough people realize that they can add such files to their *global* git ignore file.)

vemel:
Thank you! I for sure did not realize that I can use global `.gitignore`.

PR #704, opened 2020-03-02
Fix ForwardRef hash and equality checks

### LINKED ISSUE #667: Infinite recursion with get_type_hints
I encountered one condition where calling get_type_hints causes infinite recursion when dealing with forward declaration and cyclic types.

Here's an example:

```
from typing import Union, List, get_type_hints

ValueList = List['Value']
Value = Union[str, ValueList]

class A:
    a: List[Value]

get_type_hints(A, globals(), locals())
```

This reaches the recursion limit as of 3.8.0b2.

It seems that the combining `_GenericAlias` with `ForwardRef` is what triggers this condition:
- `ForwardRef._evaluate` sets `__forward_value__` on its first call on a given instance
- `_GenericAlias` tries to compare its args post evaluation
If one of the arguments is a previously evaluated forward reference containing a cycle, then it will infinitely recurse in the hash function when building a frozen set for the comparison.

The above is, of course, a very artificial example, but I can imagine this happening a lot in code with trees or similar structures.
My initial reproduction case was using `_eval_type` to resolve forward references returned by `get_args` (side note: it would be nice to have a public function to do that).

rfrowe:
This backports python/cpython#15400 to this typing module and fixes #667.

> Ideally if we stick a ForwardRef in a dictionary we would like to reliably be able to get it out again.
> 
> https://bugs.python.org/issue37953
> (cherry picked from commit e082e7c in python/cpython)
> 
> Co-authored-by: plokmijnuhby 39633434+plokmijnuhby@users.noreply.github.com

### DIFF
--- a/src/test_typing.py
+++ b/src/test_typing.py
@@ -1497,6 +1497,65 @@ class ForwardRefTests(BaseTestCase):
         self.assertEqual(fr, typing._ForwardRef('int'))
         self.assertNotEqual(List['int'], List[int])
 
+    def test_forward_equality_gth(self):
+        c1 = typing._ForwardRef('C')
+        c1_gth = typing._ForwardRef('C')
+        c2 = typing._ForwardRef('C')
+        c2_gth = typing._ForwardRef('C')
+
+        class C:
+            pass
+        def foo(a: c1_gth, b: c2_gth):
+            pass
+
+        self.assertEqual(get_type_hints(foo, globals(), locals()), {'a': C, 'b': C})
+        self.assertEqual(c1, c2)
+        self.assertEqual(c1, c1_gth)
+        self.assertEqual(c1_gth, c2_gth)
+        self.assertEqual(List[c1], List[c1_gth])
+        self.assertNotEqual(List[c1], List[C])
+        self.assertNotEqual(List[c1_gth], List[C])
+        self.assertEquals(Union[c1, c1_gth], Union[c1])
+        self.assertEquals(Union[c1, c1_gth, int], Union[c1, int])
+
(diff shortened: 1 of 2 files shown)


the-knights-who-say-ni:

## CLA Missing

Our records indicate the following people have not signed the CLA:

@rfrowe

For legal reasons we need all the people listed to sign the CLA before we can look at your contribution. Please follow [the steps outlined in the CPython devguide](https://devguide.python.org/pullrequest/#licensing) to rectify this issue.

src/test_typing.py
```diff
@@ -1516,6 +1575,62 @@ def foo(a: Tuple['T']):
...
+            ret = get_type_hints(fun, globals(), locals())
+            return a
+
+        def cmp(o1, o2):
+            return o1 == o2
+
+        r1 = namespace1()
+        r2 = namespace2()
+        self.assertIsNot(r1, r2)
+        self.assertRaises(RecursionError, cmp, r1, r2)
```
gvanrossum: So this fails in 3.4 because RecursionError doesn't exist there -- recursion raises RuntimeError. You can do something like
```
try:
    exc = RecursionError
except NameError:
    exc = RuntimeError
self.assertRaises(exc, ...)
```

rfrowe: Didn't have a chance to look at this till now. Thanks for the snippet :)

gvanrossum:
Thanks for your efforts! Is there anything else you're waiting for?

rfrowe:
Nope, that's it. Thanks for all of your help. Glad to be a part of the 3.7.7 release!

Issue #723, opened 2020-05-13
Add `SupportsIndex` to `typing_extensions`


srittau:
Cf python/typeshed#3974.

gvanrossum:
Can you clarify? I don’t understand the issue you link.

JelleZijlstra:
`typing.SupportsIndex` was added only in 3.8 (https://docs.python.org/3/library/typing.html#typing.SupportsIndex), but the concept of `__index__` is much older, so we should put the ABC in `typing_extensions` so that it can be used on older versions of Python.

gvanrossum:
Okay, fair. Just di it, I guess.

PR #724, opened 2020-05-13
Add SupportsIndex, added in Python 3.8

srittau:
Closes: #723


srittau:
flake8 seems to fail, but in tests that were untouched by this PR.

gvanrossum:
Yeah, it looks like the flake8 on Python 3.6 has some extension installed that makes it interpret forward references in type annotations like `List['CC']`. Some tests use this notation in a way that that extension apparently doesn't understand. (Probably it's relying on `typing.get_type_hints()`, which perhaps in the 3.6 stdlib is not capable enough.)

I propose to ignore this, and I'll land, since Ivan already approved.

PR #729, opened 2020-05-28
Use regex to allow `typing` or `typing_extensions` module name

frenzymadness:
This is useful for testing with Python 3.9 where:

```
>>> repr(Annotated)
"<class 'typing.Annotated'>"
```

but in  Python 3.8 and lower:

```
>>> repr(Annotated)
"<class 'typing_extensions.Annotated'>"
```

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -2,6 +2,9 @@ language: python
 
 jobs:
   include:
+    - name: "3.9"
+      dist: xenial
+      python: 3.9-dev
     - name: "3.8"
       dist: xenial
       python: 3.8.0
--- a/typing_extensions/src_py3/test_typing_extensions.py
+++ b/typing_extensions/src_py3/test_typing_extensions.py
@@ -1588,13 +1588,13 @@ class TypedDictTests(BaseTestCase):
 class AnnotatedTests(BaseTestCase):
 
     def test_repr(self):
-        self.assertEqual(
+        self.assertRegex(
             repr(Annotated[int, 4, 5]),
-            "typing_extensions.Annotated[int, 4, 5]"
+            r"typing(?:_extensions|)\.Annotated\[int, 4, 5\]"
         )
-        self.assertEqual(
+        self.assertRegex(
             repr(Annotated[List[int], 4, 5]),
-            "typing_extensions.Annotated[typing.List[int], 4, 5]"
+            r"typing(?:_extensions|)\.Annotated\[typing\.List\[int\], 4, 5\]"
         )
 
     def test_flatten(self):


gvanrossum:
Are you saying that this is the only fix necessary to make this package's tests pass with 3.9?

If so, I am okay with it. Otherwise, let's fix all the errors at once.

frenzymadness:
That's all I had to fix to make the tests of typing_extensions pass with Python 3.9b1:

```
$ python --version
Python 3.9.0b1

$ python typing_extensions/src_py3/test_typing_extensions.py
................................................................................s.....s..........s.....................s..
Ran 122 tests in 0.055s

OK (skipped=4)
```

I can add Python 3.9 to the Travis config so we can see what CI thinks about it.

gvanrossum:
> I can add Python 3.9 to the Travis config so we can see what CI thinks about it.

Please do!

frenzymadness:
The CI is green but the log contains some exceptions with Python 3.9. I need to investigate them and also find why it's not reported as a failure.

frenzymadness:
There are two problems I don't know how to solve. Problem n1 is that Callable no longer provides internal methods like `_abs_registry` or `_abc_cache` — [bpo-31333](https://bugs.python.org/issue31333) which is implemented also in Python 3.8.3.

The second one is that Python 3.9 importlib.abc imports Protocol from typing but this version of typing does not have Protocol.

```
PYTHONPATH=src python -m pytest src/test_typing.py
Could not import runpy module
  File "/usr/lib64/python3.9/runpy.py", line 15, in <module>
    import importlib.util
  File "/home/lbalhar/.virtualenvs/typing/lib64/python3.9/importlib/util.py", line 2, in <module>
    from . import abc
  File "/home/lbalhar/.virtualenvs/typing/lib64/python3.9/importlib/abc.py", line 17, in <module>
    from typing import Protocol, runtime_checkable
ImportError: cannot import name 'Protocol' from 'typing' (/home/lbalhar/Software/typing/src/typing.py)
```

This PR might be considered as a fix just for typing_extensions because its tests work well and the typing module will need more investigation.

gvanrossum:
This was fixed differently in #731. Sorry! We can close this now.

gvanrossum:
FWIW this version of typing.py is not meant to work with versions of Python that have a typing.py in the stdlib (i.e. all currently still supported versions :-). But typing_extensions.py still needs to work there. This means that if you want to run the tests with 3.9 in Travis you should only run the tests for typing_extensions/src_py3. Can you think of a way to do that?

PR #774, opened 2021-01-02
Support PEP 612 in typing_extensions (Python 3)

Fidget-Spinner:
Runtime correctness is zero due to wacky list workarounds.  Unfortunately, I'm not enough of an MRO/metaclass wizard to know how to trick ``isinstance(arg, list)`` in ``Callable`` while still returning a ``typing._GenericAlias``.

The first patch is for Python3. I'm mildly reluctant to write one for Python2 since it's EOL-ed for a year already, but we'll see ;).

### DIFF
--- a/typing_extensions/src_py3/typing_extensions.py
+++ b/typing_extensions/src_py3/typing_extensions.py
@@ -2252,6 +2252,12 @@ else:
                 prefix = '~'
             return prefix + self.__name__
 
+        def __hash__(self):
+            return hash((self.__name__,
+                         self.__covariant__,
+                         self.__contravariant__,
+                         self.__bound__))
+
         def __reduce__(self):
             return self.__name__
 
--- a/typing_extensions/src_py3/typing_extensions.py
+++ b/typing_extensions/src_py3/typing_extensions.py
@@ -2224,8 +2224,23 @@ else:
         args = object()
         kwargs = object()
 
-        def __init__(self, *args, **kwargs):
-            pass
+        def __init__(self, name, *, bound=None, covariant=False, contravariant=False):
+            self.__name__ = name
+            self.__covariant__ = bool(covariant)
+            self.__contravariant__ = bool(contravariant)
+            if bound:
+                self.__bound__ = typing._type_check(bound, "Bound must be a type.")
+            else:
+                self.__bound__ = None
+
+        def __repr__(self):
+            if self.__covariant__:
+                prefix = '+'
+            elif self.__contravariant__:
+                prefix = '-'
+            else:
+                prefix = '~'
+            return prefix + self.__name__
 
 
 # Inherits from list as a workaround for Callable checks in Python < 3.9.2.
--- a/typing_extensions/src_py3/typing_extensions.py
+++ b/typing_extensions/src_py3/typing_extensions.py
@@ -2233,7 +2233,7 @@ class _ConcatenateGenericAlias(list):
     def __init__(self, *args, **kwargs):
         pass
 
-
+@_tp_cache
 def _concatenate_getitem(self, parameters):
     if parameters == ():
         raise TypeError("Cannot take a Concatenate of no types.")
--- a/typing_extensions/src_py3/typing_extensions.py
+++ b/typing_extensions/src_py3/typing_extensions.py
@@ -2226,6 +2226,7 @@ else:
         kwargs = object()
 
         def __init__(self, name, *, bound=None, covariant=False, contravariant=False):
+            super().__init__([self])
             self.__name__ = name
             self.__covariant__ = bool(covariant)
             self.__contravariant__ = bool(contravariant)
@@ -2267,6 +2268,7 @@ else:
 # Inherits from list as a workaround for Callable checks in Python < 3.9.2.
 class _ConcatenateGenericAlias(list):
     def __init__(self, origin, args):
+        super().__init___(args)
         self.__origin__ = origin
         self.__args__ = args
 
--- a/typing_extensions/src_py3/typing_extensions.py
+++ b/typing_extensions/src_py3/typing_extensions.py
@@ -115,7 +115,9 @@ else:
 __all__ = [
     # Super-special typing primitives.
     'ClassVar',
+    'Concatenate',
     'Final',
+    'ParamSpec',
     'Type',
 
     # ABCs (from collections.abc).
@@ -146,6 +148,7 @@ __all__ = [
     'NewType',
     'overload',
     'Text',
+    'TypeAlias',
     'TYPE_CHECKING',
 ]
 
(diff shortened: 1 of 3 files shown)


Fidget-Spinner:
@gvanrossum Would you like to see Python 2 supported as well? It's possible to have runtime correctness for Python 2 since I can edit the typing PyPI module as well, though that would require more work.

gvanrossum:
Let’s skip Python 2. Sorry for the slow review, will get to it soon.

Fidget-Spinner:
> Let’s skip Python 2. Sorry for the slow review, will get to it soon.

No worries, I didn't mean to rush and I hope it didn't come off that way :).

gvanrossum:
The rush was entirely in my head. :-)

gvanrossum:
@JelleZijlstra Have you reviewed this? I have sadly dropped the ball (again). If you approve I will land it.

typing_extensions/src_py3/typing_extensions.py
```diff
@@ -2166,3 +2169,181 @@ class TypeAlias(metaclass=_TypeAliasMeta, _root=True):
...
+               return x + y
+
+        Parameter specification variables defined with covariant=True or
+        contravariant=True can be used to declare covariant or contravariant
+        generic types.  These keyword arguments are valid, but their actual semantics
+        are yet to be decided.  See PEP 612 for details.
+
+        Parameter specification variables can be introspected. e.g.:
+
+           P.__name__ == 'T'
```
JelleZijlstra, line 2216: This isn't true for your implementation right now. I'd prefer to make this part of the docstring actually true, but if that's not possible we should remove it.

typing_extensions/src_py3/typing_extensions.py
```diff
@@ -2166,3 +2169,181 @@ class TypeAlias(metaclass=_TypeAliasMeta, _root=True):
...
+           P.__covariant__ == False
+           P.__contravariant__ == False
+
+        Note that only parameter specification variables defined in global scope can
+        be pickled.
+        """
+        args = object()
+        kwargs = object()
+
+        def __init__(self, *args, **kwargs):
```
JelleZijlstra, line 2227: Why not add more specific parameters, like in CPython?

https://github.com/python/cpython/blob/master/Lib/typing.py#L782

typing_extensions/src_py3/typing_extensions.py
```diff
@@ -2166,3 +2169,181 @@ class TypeAlias(metaclass=_TypeAliasMeta, _root=True):
...
+@_tp_cache
+def _concatenate_getitem(self, parameters):
+    if parameters == ():
+        raise TypeError("Cannot take a Concatenate of no types.")
+    if not isinstance(parameters, tuple):
+        parameters = (parameters,)
+    if not isinstance(parameters[-1], ParamSpec):
+        raise TypeError("The last parameter to Concatenate should be a "
+                        "ParamSpec variable.")
+    return _ConcatenateGenericAlias(self, parameters)
```
JelleZijlstra, line 2245: CPython does some additional typechecking: https://github.com/python/cpython/blob/master/Lib/typing.py#L564

typing_extensions/src_py3/typing_extensions.py
```diff
@@ -2166,3 +2169,181 @@ class TypeAlias(metaclass=_TypeAliasMeta, _root=True):
...
+        """
+        args = object()
+        kwargs = object()
+
+        def __init__(self, *args, **kwargs):
+            pass
+
+
+# Inherits from list as a workaround for Callable checks in Python < 3.9.2.
+class _ConcatenateGenericAlias(list):
```
JelleZijlstra, line 2232: For runtime introspectability, it would be good to provide access the constructor arguments in the same way that 3.10's typing.py does it.

Here's what a 3.10 Concatenate has in its innards:
```
>>> Concatenate[int, ParamSpec("P")].__dict__
{'_inst': True, '_name': None, '__origin__': typing.Concatenate, '__slots__': None, '__args__': (<class 'int'>, ~P), '__parameters__': (~P,), '__module__': 'typing'}
```
Adding `__origin__` and `__args__` at least should be easy.

typing_extensions/src_py3/typing_extensions.py
```diff
@@ -2166,3 +2169,181 @@ class TypeAlias(metaclass=_TypeAliasMeta, _root=True):
...
+        def __repr__(self):
+            return 'typing_extensions.Concatenate'
+
+        def __getitem__(self, parameters):
+            return _concatenate_getitem(self, parameters)
+
+    Concatenate = _ConcatenateAliasBase(_root=True)
+
+else:
+    class _ConcatenateAliasMeta(typing.TypingMeta):
```
JelleZijlstra, line 2323: Looks like this is for 3.5.0 through 3.5.2, since `_FinalTypingBase` exists in 3.5.3 and 3.6.0. Might be worth a comment.

typing_extensions/src_py3/typing_extensions.py
```diff
@@ -2166,3 +2169,181 @@ class TypeAlias(metaclass=_TypeAliasMeta, _root=True):
...
+        are yet to be decided.  See PEP 612 for details.
+
+        Parameter specification variables can be introspected. e.g.:
+
+           P.__name__ == 'T'
+           P.__bound__ == None
+           P.__covariant__ == False
+           P.__contravariant__ == False
+
+        Note that only parameter specification variables defined in global scope can
```
JelleZijlstra, line 2221: But there's no test for pickling.

JelleZijlstra (changes requested):
Sorry for letting this linger! I took a look at the code and I have some suggested improvements.

JelleZijlstra:
Also, I noticed we're not running 3.9 in CI, so I opened #796 to add it.

typing_extensions/src_py3/typing_extensions.py
Fidget-Spinner: PEP 612 states that bounds, covariants, contravariant etc. just need to be accepted during runtime, but they don't actually mean anything: https://www.python.org/dev/peps/pep-0612/#declaration. So I just followed what Pyre did https://github.com/facebook/pyre-check/blob/master/pyre_extensions/__init__.py#L93 .

Though I don't think it will hurt. So I'll add them in :).

typing_extensions/src_py3/typing_extensions.py
Fidget-Spinner: Good point, thank you for catching this, I was being lazy and copied from my Cpython docstrings.

Fidget-Spinner:
@JelleZijlstra - Thanks! I really appreciate the reviews. I marked the PR as draft because you gave me some other ideas to improve the runtime information. So I'm working on them now.

This PR was originally just to implement all the basic stuff required for type checkers, which is why it's so barren in terms of runtime information.

Fidget-Spinner:
@JelleZijlstra this should be ready now. Please take a look when you have the time. Thanks!

typing_extensions/src_py3/typing_extensions.py
```diff
@@ -2166,3 +2169,226 @@ class TypeAlias(metaclass=_TypeAliasMeta, _root=True):
...
+        be pickled.
+        """
+        args = object()
+        kwargs = object()
+
+        def __init__(self, name, *, bound=None, covariant=False, contravariant=False):
```
JelleZijlstra, line 2228:
```suggestion
        def __init__(self, name, *, bound=None, covariant=False, contravariant=False):
            super().__init__([self])
```
With this change, the runtime annotations of the `add_logging` example in the PEP look like this:
```
In [10]: add_logging.__annotations__
Out[10]: 
{'f': typing.Callable[[~P], ~R],
 'return': typing.Callable[[~P], typing.Awaitable[~R]]}
```
That's not exactly correct, but it's close enough that dynamic type checkers should be able to do something with it. With your current code, it produces `Callable[[], ~R]` because it iterates over the ParamSpec and finds nothing.

This feels a little bad because we're creating a list that contains itself, but I think it might not have any ill effects except for creating a reference cycle.

typing_extensions/src_py3/typing_extensions.py
```diff
@@ -2166,3 +2169,226 @@ class TypeAlias(metaclass=_TypeAliasMeta, _root=True):
...
+
+        # Note: Can't fake ParamSpec as a TypeVar to get it to work
+        # with Generics. ParamSpec isn't an instance of TypeVar in 3.10.
+        # So encouraging code like isinstance(ParamSpec('P'), TypeVar))
+        # will lead to breakage in 3.10.
+        # This also means no accurate __parameters__ for GenericAliases.
+
+# Inherits from list as a workaround for Callable checks in Python < 3.9.2.
+class _ConcatenateGenericAlias(list):
+    def __init__(self, origin, args):
```
JelleZijlstra, line 2269:
```suggestion
    def __init__(self, origin, args):
        super().__init___(args)
```
Similar principles here, though it's a little cleaner because we don't need to make a recursive list. This produces the following for the `with_request` example in the PEP:
```
In [9]: with_request.__annotations__
Out[9]: 
{'f': typing.Callable[[__main__.Request, ~P], ~R],
 'return': typing.Callable[[~P], ~R]}
```

typing_extensions/src_py3/typing_extensions.py
```diff
@@ -2166,3 +2169,226 @@ class TypeAlias(metaclass=_TypeAliasMeta, _root=True):
...
+
+           P.__name__ == 'T'
+           P.__bound__ == None
+           P.__covariant__ == False
+           P.__contravariant__ == False
+
+        Note that only parameter specification variables defined in global scope can
+        be pickled.
+        """
+        args = object()
```
JelleZijlstra, line 2225: I'm a bit unhappy with this because it makes the runtime annotations just look like this:
```
In [11]: takes_int_str.__annotations__
Out[11]: 
{'args': <object at 0x7ff1831d5f30>,
 'kwargs': <object at 0x7ff1831d5f60>,
 'return': ~R}
```
However, it's the exact same in CPython, so it makes sense to follow the same principle here. I'll plan to submit a change to CPython to make this instead return some helper object pointing back to the ParamSpec; if that gets accepted we can apply the same change here.

For now no change is required here.

JelleZijlstra (changes requested):
Thanks! I checked out the code and played around with it some more and I have two suggestions to make the new types a bit more usable at runtime.

typing_extensions/src_py3/typing_extensions.py
Fidget-Spinner: Wow this is such a nice hack. The only gripe I have is that technically ``Callable[[~P], int]`` is invalid according to PEP 612 https://www.python.org/dev/peps/pep-0612/#id2 . And I really don't want to promote the invalid form even if its just showing it in repr, we already have enough improper usages of Callable as is. Hmm im not sure what route to take here 😔.

typing_extensions/src_py3/typing_extensions.py
Fidget-Spinner: This was definitely something I overlooked when I implemented the CPython version :(. How I'd do it in typing.py:

1. Create a `_ParamSpecGenericAlias`.
2.  `__origin__` is the ParamSpec it belongs to, `__args__` is just a tuple with 1 element - "args" or "kwargs".
3. Override the `__repr__` to print `~T.(args|kwargs)` instead of the default `~T[args]` found in genericaliases.

I look forward to seeing your PR!

JelleZijlstra: Yes, I was thinking something similar, except with different classes for args and kwargs, but either way should work.

typing_extensions/src_py3/typing_extensions.py
JelleZijlstra: That's true, but seeing `Callable[[], ~R]` in the repr (the current state) is pretty confusing too. My main motivation is to give dynamic type checkers something to work with here.

Fidget-Spinner: You're right, either way both reprs will be wrong, so let's choose the one that actually benefits a bunch of people.

typing_extensions/src_py3/typing_extensions.py
```diff
@@ -2166,3 +2169,228 @@ class TypeAlias(metaclass=_TypeAliasMeta, _root=True):
...
+        # Note: Can't fake ParamSpec as a TypeVar to get it to work
+        # with Generics. ParamSpec isn't an instance of TypeVar in 3.10.
+        # So encouraging code like isinstance(ParamSpec('P'), TypeVar))
+        # will lead to breakage in 3.10.
+        # This also means no accurate __parameters__ for GenericAliases.
+
+# Inherits from list as a workaround for Callable checks in Python < 3.9.2.
+class _ConcatenateGenericAlias(list):
+    def __init__(self, origin, args):
+        super().__init___(args)
```
JelleZijlstra, line 2271:
```suggestion
        super().__init__(args)
```
Sorry, I typoed the suggestion

JelleZijlstra:
Looks like this broke CI because ParamSpec objects aren't hashable. I'm guessing we could make them hashable by overriding `__hash__` and hoping that nobody will actually modify the underlying list.

Fidget-Spinner:
Hmm the maximum recursion depth exceeded errors due to ``_tp_cache`` in ``_concatenate_getitem`` are particularly hairy.

typing_extensions/src_py3/typing_extensions.py
```diff
@@ -2252,6 +2252,12 @@ def __repr__(self):
                 prefix = '~'
             return prefix + self.__name__
 
+        def __hash__(self):
+            return hash((self.__name__,
```
JelleZijlstra, line 2256: I think you can just do `object.__hash__(self)`. TypeVars are also hashed by identity:
```
>>> T1 = TypeVar("T")
>>> T2 = TypeVar("T")
>>> hash(T1), hash(T2)
(283358494, 283358500)
```

JelleZijlstra:
> Hmm the maximum recursion depth exceeded errors due to `_tp_cache` in `_concatenate_getitem` are particularly hairy.

It reproduces only if some other test runs before it, not if you just run the line that fails on its own. I can spend some time later today trying to figure out where it's going wrong.

JelleZijlstra:
I tracked it down to `ParamSpec("P") == ParamSpec("P2")` failing. It works if you add `def __eq__(self, other): return self is other`. Comparing by identity seems reasonable to me because that's also how TypeVars work: `TypeVar("T") == TypeVar("T")` returns False.

Fidget-Spinner:
> I tracked it down to `ParamSpec("P") == ParamSpec("P2")` failing. It works if you add `def __eq__(self, other): return self is other`. Comparing by identity seems reasonable to me because that's also how TypeVars work: `TypeVar("T") == TypeVar("T")` returns False.

Thank you for digging out the cause. I had forgotten that dictionary keys need to implement both ``__hash__`` and ``__eq__`` to work properly. This was a good refresher.

typing_extensions/src_py3/typing_extensions.py
```diff
@@ -2166,3 +2169,236 @@ class TypeAlias(metaclass=_TypeAliasMeta, _root=True):
...
+    def __init__(self, origin, args):
+        super().__init__(args)
+        self.__origin__ = origin
+        self.__args__ = args
+
+    def __repr__(self):
+        _type_repr = typing._type_repr
+        return '{origin}[{args}]' \
+               .format(origin=_type_repr(self.__origin__),
+                       args=', '.join(_type_repr(arg) for arg in self.__args__))
```
JelleZijlstra, line 2285:
```suggestion
                       args=', '.join(_type_repr(arg) for arg in self.__args__))

```

JelleZijlstra (changes requested):
Thanks! Could you also add some unit tests doing `==` and `hash()` on ParamSpec and Concatenate? We should make sure they keep working if we ever refactor these classes here. With those changes this should be good to go.

Fidget-Spinner:
> Thanks! Could you also add some unit tests doing `==` and `hash()` on ParamSpec and Concatenate? We should make sure they keep working if we ever refactor these classes here. With those changes this should be good to go.

Thanks for the reminder, that slipped my mind.

JelleZijlstra (approved):

PR #778, opened 2021-01-19
Fixed required/optional keys with old-style TypedDict

### LINKED ISSUE #761: TypedDict(...) as function does not respect "total" when setting __required_keys__ and __optional_keys__
Issue on Python bug tracker: https://bugs.python.org/issue42059

domdfcoding:
Fixes #761 

Copied from https://github.com/python/cpython/pull/22736

Co-authored-by: Alex Grönholm <alex.gronholm@nextday.fi>

### DIFF
--- a/typing_extensions/src_py3/test_typing_extensions.py
+++ b/typing_extensions/src_py3/test_typing_extensions.py
@@ -1586,11 +1586,15 @@ class TypedDictTests(BaseTestCase):
         self.assertEqual(D(), {})
         self.assertEqual(D(x=1), {'x': 1})
         self.assertEqual(D.__total__, False)
+        self.assertEqual(D.__required_keys__, frozenset())
+        self.assertEqual(D.__optional_keys__, {'x'})
 
         if PY36:
             self.assertEqual(Options(), {})
             self.assertEqual(Options(log_level=2), {'log_level': 2})
             self.assertEqual(Options.__total__, False)
+            self.assertEqual(Options.__required_keys__, frozenset())
+            self.assertEqual(Options.__optional_keys__, {'log_level', 'log_path'})
 
     @skipUnless(PY36, 'Python 3.6 required')
     def test_optional_keys(self):
--- a/typing_extensions/src_py3/test_typing_extensions.py
+++ b/typing_extensions/src_py3/test_typing_extensions.py
@@ -1530,7 +1530,7 @@ class TypedDictTests(BaseTestCase):
 
     def test_typeddict_errors(self):
         Emp = TypedDict('Emp', {'name': str, 'id': int})
-        if sys.version_info[:2] >= (3, 9):
+        if sys.version_info >= (3, 9, 2):
             self.assertEqual(TypedDict.__module__, 'typing')
         else:
             self.assertEqual(TypedDict.__module__, 'typing_extensions')
--- a/typing_extensions/src_py3/typing_extensions.py
+++ b/typing_extensions/src_py3/typing_extensions.py
@@ -1692,6 +1692,11 @@ else:
                                          ' /, *, total=True, **kwargs)')
 
     class _TypedDictMeta(type):
+        def __init__(cls, name, bases, ns, total=True):
+            # In Python 3.4 and 3.5 the __init__ method also needs to support the keyword arguments.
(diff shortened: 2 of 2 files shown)


typing_extensions/src_py3/test_typing_extensions.py
```diff
@@ -1552,11 +1552,15 @@ def test_total(self):
...
         self.assertEqual(D(x=1), {'x': 1})
         self.assertEqual(D.__total__, False)
+        self.assertEqual(D.__required_keys__, frozenset())
+        self.assertEqual(D.__optional_keys__, {'x'})
 
         if PY36:
             self.assertEqual(Options(), {})
             self.assertEqual(Options(log_level=2), {'log_level': 2})
             self.assertEqual(Options.__total__, False)
+            self.assertEqual(Options.__required_keys__, frozenset())
```
domdfcoding, line 1562: I don't know whether this test and the one below should be within the `if PY36` block or not.

typing_extensions/src_py3/typing_extensions.py
```diff
@@ -1643,14 +1643,14 @@ def _typeddict_new(*args, total=True, **kwargs):
...
-        ns = {'__annotations__': dict(fields), '__total__': total}
+        ns = {'__annotations__': dict(fields)}
         try:
             # Setting correct module is necessary to make typed dict classes pickleable.
             ns['__module__'] = sys._getframe(1).f_globals.get('__name__', '__main__')
         except (AttributeError, ValueError):
             pass
 
-        return _TypedDictMeta(typename, (), ns)
+        return _TypedDictMeta(typename, (), ns, total=total)
```
domdfcoding, line 1653: On Python 3.4 and 3.5 it seems that `_TypedDictMeta(...)` calls `__init__` rather than `__new__`, but `__init__` isn't implemented. See, for instance: https://travis-ci.com/github/python/typing/jobs/473002561#L338

domdfcoding:
There aren't any tests set up on Travis for Python 3.9, but for me they fail locally as python/cpython#22736 didn't make it into 3.9.1.
Should I change the check to only use the stdlib for `>=3.9.2`? 

https://github.com/python/typing/blob/c7a981aaab7b20b93c911f1e5dd0ab2a8e30fd6c/typing_extensions/src_py3/typing_extensions.py#L1587

typing_extensions/src_py3/test_typing_extensions.py
JelleZijlstra: `Options` only exists if PY36 is on (it's defined with variable annotations), so these tests should be within the `if PY36` block.

JelleZijlstra (commented):
Sorry for the long wait here! I just merged a PR adding 3.9.0 to CI. I think it's OK to make `typing_extensions.TypedDict` not an alias for `typing.TypedDict` in 3.9.0/1, so that we can provide the fixed version of TypedDict here.

typing_extensions/src_py3/typing_extensions.py
JelleZijlstra: That looks nasty. I wonder if it's related to this note in https://docs.python.org/3/whatsnew/3.6.html#changes-in-the-python-api : 

> As part of PEP 487, the handling of keyword arguments passed to type (other than the metaclass hint, metaclass) is now consistently delegated to object.__init_subclass__(). This means that type.__new__() and type.__init__() both now accept arbitrary keyword arguments, but object.__init_subclass__() (which is called from type.__new__()) will reject them by default. Custom metaclasses accepting additional keyword arguments will need to adjust their calls to type.__new__() (whether direct or via super) accordingly.

I just tried and failed to get pyenv to compile 3.5 for me so I can't try this out myself :( If it's too hard to get this to work I wouldn't be opposed to fixing it for 3.6+ only for now.

domdfcoding:
Right, I think I've sorted it. 

On Python 3.9.2 and above `typing.TypedDict` is used, otherwise it's `typing_extensions.TypedDict`.

For Python 3.4 and 3.5 it is necessary to implement `__init__` on `_TypedDictMeta`, as the "total" keyword gets passed to that method. 3.6+ accept arbitrary keyword arguments for `type.__init__` but earlier versions don't.

JelleZijlstra (approved):

PR #780, opened 2021-01-21
Fix protocol’s __init__ behaviour to be same like in python >= 3.8

dkraczkowski:
Hello everyone, I am author of a dependency injection container for python: https://pypi.org/project/kink/. Lately we run into some issues with older versions of python were we do use `typing_extensions` as a pollyfill for missing functionalities.

In python >= 3.8 `_no_init` function which is responsible for disabling Protocol's initialisation is defined in the global scope of a module, which allows us to do checks like here: https://github.com/kodemore/kink/blob/master/kink/inject.py#L75. Sadly in `typing_extensions` this function is defined as a local function so when doing the check memory address does not match and comparison fails. This is inconsistent with python >= 3.8 implementation and this PR fixes this inconsistency.

I have also prepared test to cover the use-case.

### DIFF
--- a/typing_extensions/src_py3/test_typing_extensions.py
+++ b/typing_extensions/src_py3/test_typing_extensions.py
@@ -1423,6 +1423,15 @@ if HAVE_PROTOCOLS:
             self.assertIsSubclass(B, Custom)
             self.assertNotIsSubclass(A, Custom)
 
+        def test_no_init_same_for_different_protocol_implementations(self):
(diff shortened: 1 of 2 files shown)


the-knights-who-say-ni:

## Recognized GitHub username

@dkraczkowski

dkraczkowski:
I have signed CLA today, wonder how long it takes to process my form.

JelleZijlstra:
Thanks for your contribution!

PR #790, opened 2021-02-25
Document availability of Annotated

### LINKED ISSUE #789: Add Annotated to typing_extensions
It feels like a great tool which I would like to use in older Python versions. Based on the documentation the does not seem to be a fundamental problem why it could not be included in that package.

septatrix:
This was previously not included in the readme although implemented. Closes #789


gvanrossum (approved):
Thanks! I wonder if other recent additions aren’t listed?

PR #798, opened 2021-04-11
backport ParamSpecArgs/Kwargs

JelleZijlstra:
From python/cpython#25298. I also added more tests for get_args/get_origin,
which previously didn't exist in in typing_extensions.

### DIFF
--- a/typing_extensions/src_py3/test_typing_extensions.py
+++ b/typing_extensions/src_py3/test_typing_extensions.py
@@ -576,13 +576,14 @@ class GetUtilitiesTestCase(TestCase):
         if sys.version_info >= (3, 9):
             self.assertEqual(get_args(list[int]), (int,))
         self.assertEqual(get_args(list), ())
-        # Support Python versions with and without the fix for
-        # https://bugs.python.org/issue42195
-        self.assertIn(get_args(collections.abc.Callable[[int], str]),
-                      (([int], str), ([[int]], str)))
-        self.assertEqual(get_args(collections.abc.Callable[..., str]), (..., str))
-        self.assertIn(get_args(collections.abc.Callable[[], str]),
-                      (([], str), ([[]], str)))
+        if sys.version_info >= (3, 9):
+            # Support Python versions with and without the fix for
+            # https://bugs.python.org/issue42195
+            self.assertIn(get_args(collections.abc.Callable[[int], str]),
+                          (([int], str), ([[int]], str)))
+            self.assertEqual(get_args(collections.abc.Callable[..., str]), (..., str))
+            self.assertIn(get_args(collections.abc.Callable[[], str]),
+                          (([], str), ([[]], str)))
         P = ParamSpec('P')
         self.assertIn(get_args(Callable[P, int]), [(P, int), ([P], int)])
         self.assertEqual(get_args(Callable[Concatenate[int, P], int]),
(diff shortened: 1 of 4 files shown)


JelleZijlstra:
(Pretty likely that this broke tests in some Python version, I'll monitor CI.)

JelleZijlstra:
Finally, all the tests pass. @Fidget-Spinner would you mind reviewing this?

Fidget-Spinner:
> Finally, all the tests pass. @Fidget-Spinner would you mind reviewing this?

Sure, give me some time, I'll try to review it tomorrow (your time).

JelleZijlstra:
Thanks, no hurry!

typing_extensions/src_py3/test_typing_extensions.py
```diff
@@ -519,6 +519,77 @@ def test_final_forward_ref(self):
...
+        self.assertEqual(get_args(Annotated[T, 'one', 2, ['three']]), (T, 'one', 2, ['three']))
+        self.assertEqual(get_args(List), ())
+        self.assertEqual(get_args(Tuple), ())
+        self.assertEqual(get_args(Callable), ())
+        if sys.version_info >= (3, 9):
+            self.assertEqual(get_args(list[int]), (int,))
+        self.assertEqual(get_args(list), ())
+        if sys.version_info >= (3, 9):
+            # Support Python versions with and without the fix for
+            # https://bugs.python.org/issue42195
```
Fidget-Spinner, line 581: Off topic: I didn't understand these tests at first, but then I saw assert**In** and the lists/tuples, and now I find them really smart :).

Also I can't wait for Python 3.14 or higher (however many years down the road that is) where we can drop 3.9 support.

I think it's worth to add a small comment saying what versions this affect and which results are expected, so in the future people don't need to scroll through that really log bpo. Maybe
```
# The first result is expected in Python 3.9.0/1, the second is valid from 3.9.2 onwards
```

typing_extensions/src_py3/typing_extensions.py
```diff
@@ -2058,11 +2058,23 @@ class Annotated(metaclass=AnnotatedMeta):
 
 # Python 3.8 has get_origin() and get_args() but those implementations aren't
 # Annotated-aware, so we can't use those, only Python 3.9 versions will do.
-if sys.version_info[:2] >= (3, 9):
+# Similarly, Python 3.9's implementation doesn't support ParamSpecArgs and
```
Fidget-Spinner, line 2061: Another off topic comment: Something I just realised: if ParamSpecArgs/Kwargs had inherited from ``_GenericAlias`` in the original implementation, we wouldn't need this. Do you think it's worth it to do that in CPython or not worth the extra complexity/feel it's too hacky and not really what ``_GenericAlias`` actually represents?

Fidget-Spinner (approved):
Great work! Thanks for spending time on this.

typing_extensions/src_py3/typing_extensions.py
JelleZijlstra: That's definitely an option. I feel like it's too different though. A `ParamSpec` isn't the type while the origin of a GenericAlias is, and ParamSpecArgs/Kwargs don't really have args that are comparable to those of a GenericAlias.

Fidget-Spinner: Hmm yeah. Since ``ParamSpec`` isn't a Generic to begin with, ``GenericAlias`` definitely feels wrong.

typing_extensions/src_py3/test_typing_extensions.py
Fidget-Spinner: Hehe. touché - I would've done the same if I couldn't remember.

JelleZijlstra: Added some comments here

Fidget-Spinner:
I'm happy with this. If you're satisfied too, I think you can merge.

PR #784, opened 2021-02-11
Update setup.py

dciborow:
We got this package by mistake because the upper pin for Python is not set. The dataclasses package is a similar backport, but contains this pin.
https://github.com/ericvsmith/dataclasses

Issue #247, opened 2016-07-14
A milestone for dropping provisional status?


ilevkivskyi:
PEP 484 will drop its provisional status at some point. I hope this will happen in time for 3.6.
I propose to create a milestone to estimate how much work should be done before this happens. 
Here are some issues that I propose to add to this milestone:

Big things:
- The notorious #136 (and #133, #203 that depend on it). This bug requires substantial changes to typing.py. However, it does not require PEP change and there is already a PR #207 so that probably it will not take too much time.
- #11 (Protocols) It looks like a very convenient tool, and it is suitable for Python's duck-typing. There is already a concrete proposal by @JukkaL https://github.com/python/typing/issues/11#issuecomment-138133867 that was positively accepted at python-ideas.

Smaller things:
- #213 (`Intersection`) This will be very easy to implement in typing.py. It will work nicely with protocols. Also I could speculate that mypy implementation will not be too difficult (it should be similar to `Union`).
- #239 and #264 (flexible `Callable`) This feature has been requested several times, and it is really easy to implement in typing.py, there are some detail to discuss.
- #241 and #246 (`Implementation`) It seems that an agreement on this is quite close.
- #208 how to handle str/unicode.
- #271 SelfType and type annotations of classmethods? I am not sure about this one.

I could volunteer to take care about all of the above except #11 (it probably will be better done by Jukka) and SelfType. By "take care" I mean make PRs to PEPs and typing.py. However, I do not exclude also working on some of these in mypy if necessary. I think it is not necessary for mypy to support all the features at the point when PEP 484 drops provisional status, they could be implemented later.

vlasovskikh:
And #208, I'm working on it.

JukkaL:
There's no issue for this yes, but I'd want us to resolve the slowness of expressions like `List[int]`, as each of these constructs a new type object (or even multiple type objects if the type is more complex). I did some quick benchmarking a while ago and it seems that type annotations can slow down program startup significantly. Also, it can affect the performance of code with nested, annotated functions and casts to complex types.

This was discussed at PyCon and somebody (@ambv?) raised this as a potential issue.

ilevkivskyi:
@JukkaL There was a proposal #130 to cache outcomes of such things. The issue is closed now.
I think that this is in fact also related to #136 . For example, ``Union[str, int]`` does not need to be a class, anyway one should not subclass it, and it will not support ``issubclass`` etc.

[...]

PR #805, opened 2021-05-01
prepare release 3.10.0.0

JelleZijlstra:
Also update the description for typing-extensions

### DIFF
--- a/typing_extensions/setup.py
+++ b/typing_extensions/setup.py
@@ -44,6 +44,7 @@ classifiers = [
     'Programming Language :: Python :: 3.7',
     'Programming Language :: Python :: 3.8',
     'Programming Language :: Python :: 3.9',
+    'Programming Language :: Python :: 3.10',
     'Topic :: Software Development',
 ]
 
--- a/typing_extensions/setup.py
+++ b/typing_extensions/setup.py
@@ -9,23 +9,24 @@ if sys.version_info < (2, 7, 0) or (3, 0, 0) <= sys.version_info < (3, 4, 0):
                      'to install the typing package.\n')
     exit(1)
 
-version = '3.7.4.3'
+version = '3.10.0.0'
 description = 'Backported and Experimental Type Hints for Python 3.5+'
 long_description = '''\
 Typing Extensions -- Backported and Experimental Type Hints for Python
 
-The ``typing`` module was added to the standard library in Python 3.5 on
-a provisional basis and will no longer be provisional in Python 3.7. However,
-this means users of Python 3.5 - 3.6 who are unable to upgrade will not be
+The ``typing`` module was added to the standard library in Python 3.5, but
+many new features have been added to the module since then.
+This means users of Python 3.5 - 3.6 who are unable to upgrade will not be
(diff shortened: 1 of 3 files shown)


typing_extensions/setup.py
```diff
@@ -9,23 +9,24 @@
...
     exit(1)
 
-version = '3.7.4.3'
+version = '3.10.0.0'
 description = 'Backported and Experimental Type Hints for Python 3.5+'
 long_description = '''\
 Typing Extensions -- Backported and Experimental Type Hints for Python
 
-The ``typing`` module was added to the standard library in Python 3.5 on
-a provisional basis and will no longer be provisional in Python 3.7. However,
```
Fidget-Spinner, line 18: Off topic: I see https://github.com/python/typing/issues/247 and https://github.com/python/cpython/pull/16204 saying typing became implicitly non-provisional in 3.7 because no new features were added, but PEP 484 still says  ``Status: | Provisional`` ? Maybe that PEP needs discussion/updating once Guido's back.

typing_extensions/setup.py
```diff
@@ -9,23 +9,24 @@
                      'to install the typing package.\n')
     exit(1)
 
-version = '3.7.4.3'
+version = '3.10.0.0'
```
Fidget-Spinner, line 12: I wonder if it's too early to also add to the ``classifiers`` list below this the following classifier: ``'Programming Language :: Python :: 3.10',`` while you're at it?

Fidget-Spinner (approved):
Thanks! Changing the tox env versions for the ``typing`` module is a good catch considering #784.

srittau (approved):

typing_extensions/setup.py
```diff
@@ -9,23 +9,24 @@
...
-``typing.Text`` or ``typing.Coroutine``.
+``typing.Protocol`` or ``typing.TypedDict``.
 
-The ``typing_extensions`` module contains both backports of these changes
-as well as experimental types that will eventually be added to the ``typing``
-module, such as ``Protocol`` or ``TypedDict``.
+The ``typing_extensions`` module contains backports of these changes.
+Experimental types that will eventually be added to the ``typing``
+module are also included in ``typing_extensions``, such as
+``typing.ParamSpec`` and ``typing.TypeGuard``.
```
JelleZijlstra, line 26: And also IntVar, which still doesn't have a PEP proposing it.

typing_extensions/setup.py
JelleZijlstra: Yeah I was thinking about that. Let me just run the test suite on current CPython master to be sure.

Fidget-Spinner: I hope I didn't do any double work here, but FWIW, the latest CPython master runs the test with no errors on my machine.

typing_extensions/setup.py
JelleZijlstra: Good catch, I opened python/peps#1942 to fix it.

typing_extensions/setup.py
JelleZijlstra: Thanks!

PR #807, opened 2021-05-03
Add PEP 655 Required and NotRequired to typing_extensions

davidfstr:
Related to https://github.com/python/mypy/pull/10370


typing_extensions/src_py3/typing_extensions.py
```diff
@@ -2803,3 +2803,115 @@ def is_str(val: Union[str, float]):
...
+
+    @_ExtensionsSpecialForm
+    def Required(self, parameters):
+        _REQUIRED_DOC
+        item = typing._type_check(parameters, f'{self} accepts only single type.')
+        return typing._GenericAlias(self, (item,))
+
+    @_ExtensionsSpecialForm
+    def NotRequired(self, parameters):
+        _NOT_REQUIRED_DOC
```
JelleZijlstra, line 2853: This doesn't actually create a docstring accessible at runtime. I'm afraid you'll just have to copy-paste the docstring.

JelleZijlstra:
Thanks for submitting this! I'll make a new typing-extensions release for this once it's ready.

I don't have time for a full review right now, but two thoughts:
- I feel like you should be able to get early 3.5 to work, since Required/NotRequired are conceptually similar at runtime to TypeGuard, which I got to work for early 3.5.
- It would be nice if we could make it easy to figure out which keys are required or not required at runtime. I guess that would involve making the `__required_keys__` attribute public; it currently seems undocumented.

davidfstr:
> I feel like you should be able to get early 3.5 to work, since Required/NotRequired are conceptually similar at runtime to TypeGuard, which I got to work for early 3.5.

Definitely possible, but is it worth the effort? I see [Python 3.5 isn't supported in general](https://en.wikipedia.org/wiki/History_of_Python#Support). But on the other hand I see typing_extensions attempts to provide support not just for 3.5 but also 3.4 and even 2.7. 🤔 

> It would be nice if we could make it easy to figure out which keys are required or not required at runtime. I guess that would involve making the __required_keys__ attribute public; it currently seems undocumented.

The `__required_keys__` and `__optional_keys__` were recently documented in [Python 3.10's typing module](https://docs.python.org/3.10/library/typing.html).

typing_extensions/src_py3/typing_extensions.py
```diff
@@ -2803,3 +2803,163 @@ def is_str(val: Union[str, float]):
...
+                r += '[{}]'.format(typing._type_repr(self.__type__))
+            return r
+
+        def __hash__(self):
+            return hash((type(self).__name__, self.__type__))
+
+        def __eq__(self, other):
+            if not isinstance(other, _Final):
+                return NotImplemented
+            if self.__type__ is not None:
```
JelleZijlstra, line 2926: This seems incorrect; it will make `Required[str] == NotRequired[str]` return True.

Also, could you add tests for `__eq__`?

davidfstr:
Thanks again for taking a first look @JelleZijlstra ! I'm prepping for PyCon this weekend, so am hoping next weekend (~5/15) to revise this PR based on feedback.

typing_extensions/src_py3/typing_extensions.py
```diff
@@ -2803,3 +2803,163 @@ def is_str(val: Union[str, float]):
...
+    def NotRequired(self, parameters):
+        """A special typing construct to mark a key of a TypedDict as
+        potentially missing. For example:
+
+            class Movie(TypedDict):
+                title: str
+                year: NotRequired[int]
+
+            m = Movie(
+                title='The Matrix',  # typechecker error if key is omitted
```
Gobot1234, line 2846: This comment needs to be removed/updated

davidfstr: The title key is required here, so I believe the comment is correct.

davidfstr:
Anyone have a tip on easily running the typing_extensions suite on very specific Python versions (especially 3.5.0-3.5.2) locally?

I tried to use Docker, via something like:
```
docker run -it -v `pwd`:/home -w /home python:3.5.2 bash
```
However it appears that this Docker image for Python lacks the Python's built-in semi-private `test` module which typing_extensions relies on.

Probably I'll try Docker again but with a specific Linux image like Ubuntu and try to install Python manually via package manager.

JelleZijlstra:
> Anyone have a tip on easily running the typing_extensions suite on very specific Python versions (especially 3.5.0-3.5.2) locally?

Honestly I've just been abusing CI for that: push a change and see what happens in 3.5. I tried and failed to get 3.5 to compile locally.

typing_extensions/src_py3/typing_extensions.py
Gobot1234: Oh yes sorry my bad.

JelleZijlstra:
@davidfstr Any updates?

davidfstr:
Sorry, going through rough patch in my life at the moment and my energy is nil. 😔

I'm projecting I could be out a month or two. Hopefully not, but I've got some rebuilding to do...

JelleZijlstra:
I'm sorry to hear that! No hurry, take your time.

huyz:
> Hopefully not, but I've got some rebuilding to do...

@davidfstr Hey man, just wanted you to wish you the best on that. Life can be quite challenging.

srittau:
@JelleZijlstra If I understand correctly, this works for Python 3.6+, right? In that case, could we just use the current version and worry about Python 3.5 compatibility later (if ever)?

JelleZijlstra:
I'd have to look at the code again to make sure it's ready for 3.6+, but I'm on board in general with not worrying about 3.5.

JelleZijlstra:
Heads up that I'm hoping to get to this PR soon to get it across the finish line. It should be easier now because we dropped support for 3.5 since the PR was created.

davidfstr:
I've regained most of my energy compared with earlier this year, so I am looking at reengaging again with PEP 655 in general starting within the next few weeks.

> Heads up that I'm hoping to get to this PR soon to get it across the finish line. It should be easier now because we dropped support for 3.5 since the PR was created.

Aye. I'll coordinate with you before picking up this PR again.

JelleZijlstra:
Welcome back to the party!

JelleZijlstra:
New version in #937.

PR #808, opened 2021-05-04
Add a missing comma to `__all__`

BvB93:
A comma is currently missing between `"DefaultDict"` and `"OrderedDict"`, 
thus concatenating both names into the non-existent `"DefaultDictOrderedDict"`.


hauntsaninja (approved):
Thanks! Maybe worth a patch release? cc @JelleZijlstra

JelleZijlstra:
Well that's embarrassing. I'm hoping we'll also get Required/NotRequired in soon, so let's give it a few more days for a release.

PR #814, opened 2021-05-24
Initial sphinx docs setup

shannonzhu:
Run Sphinx initialize to set up structure for Python typing documentation. 

See typing-sig discussion: https://mail.python.org/archives/list/typing-sig@python.org/thread/4E7V7MVO4FQVYPVELMCAYFEIYJRXDSY2/
Python docs community issue: https://github.com/python/docs-community/issues/8

### DIFF
--- a/docs/conf.py
+++ b/docs/conf.py
@@ -18,8 +18,8 @@
 # -- Project information -----------------------------------------------------
 
 project = 'typing'
-copyright = '2021, shannonzhu'
-author = 'shannonzhu'
+copyright = '2021, The Python Typing Team'
+author = 'The Python Typing Team'
 
 
 # -- General configuration ---------------------------------------------------
--- a/docs/conf.py
+++ b/docs/conf.py
@@ -1,0 +1,52 @@
+# Configuration file for the Sphinx documentation builder.
+#
+# This file only contains a selection of the most common options. For a full
+# list see the documentation:
+# https://www.sphinx-doc.org/en/master/usage/configuration.html
+
+# -- Path setup --------------------------------------------------------------
+
+# If extensions (or modules to document with autodoc) are in another directory,
+# add these directories to sys.path here. If the directory is relative to the
+# documentation root, use os.path.abspath to make it absolute, like shown here.
+#
+# import os
+# import sys
+# sys.path.insert(0, os.path.abspath('.'))
+
+
+# -- Project information -----------------------------------------------------
+
+project = 'typing'
+copyright = '2021, shannonzhu'
+author = 'shannonzhu'
+
+
+# -- General configuration ---------------------------------------------------
+
+# Add any Sphinx extension module names here, as strings. They can be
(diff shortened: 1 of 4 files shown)


docs/conf.py
```diff
@@ -0,0 +1,52 @@
...
+# import os
+# import sys
+# sys.path.insert(0, os.path.abspath('.'))
+
+
+# -- Project information -----------------------------------------------------
+
+project = 'typing'
+copyright = '2021, shannonzhu'
+author = 'shannonzhu'
```
shannonzhu, line 22: Not sure what the correct copyright and author is. Filled in placeholders but will fix based on suggestions.

docs/index.rst
```diff
@@ -0,0 +1,20 @@
...
+   You can adapt this file completely to your liking, but it should at least
+   contain the root `toctree` directive.
+
+Welcome to the Python Type System documentation!
+================================================
+
+.. toctree::
+   :maxdepth: 2
+   :caption: Contents:
+
```
shannonzhu, line 12: Open to thoughts / follow-up PRs for a good initial skeleton TOC to shape the first iteration of docs!

hauntsaninja: How much do we want to steal from mypy docs? https://mypy.readthedocs.io/en/latest/#overview-type-system-reference

docs/conf.py
JelleZijlstra: I'd suggest saying something like "The typing team"

docs/conf.py
```diff
@@ -0,0 +1,52 @@
...
+# This pattern also affects html_static_path and html_extra_path.
+exclude_patterns = ['_build', 'Thumbs.db', '.DS_Store']
+
+
+# -- Options for HTML output -------------------------------------------------
+
+# The theme to use for HTML and HTML Help pages.  See the documentation for
+# a list of builtin themes.
+#
+html_theme = 'alabaster'
```
Fidget-Spinner, line 47: I think we should consider using the python docs theme https://github.com/python/python-docs-theme to convey that this _is_ the official docs for typing and friends.

Another one to consider: recently I've seen some people in the docs WG using the furo theme https://github.com/pradyunsg/furo for new docs too since its more responsive on mobile browsers. Maybe we can get some input from them later on.

Fidget-Spinner (commented):
Wow! Thanks so much for championing this change. I just have one minor point for discussion below :).

stefanv:
How about merging this and then iterating as necessary?  I think this is a fine start.

JelleZijlstra:
Good point, I'll merge this PR now and we can write more docs in new PRs.

docs/conf.py
shannonzhu: Good idea - following up (a bit late) with https://github.com/python/typing/pull/818/ to set up python docs theming and hopefully an easier build for this. Also open to using the furo theme instead, if you'd be willing to help investigate and change it!

PR #818, opened 2021-06-17
Add docs build options and theme requirements

shannonzhu:
Add Makefile rules to pull in python docs theme requirements in a venv and shortcut sphinx builds. Not confident on the Makefile details - used python/docs Sphinx builds as a reference here.

Verified that `make venv` successfully builds a virtual environment with Sphinx and the correct theming dependencies installed, and `make html` creates a page successfully.
Output: https://gist.github.com/shannonzhu/faa9e20e34cd0a785ea9098aed69ae56

### DIFF
--- a/docs/Makefile
+++ b/docs/Makefile
@@ -3,16 +3,42 @@
 
 # You can set these variables from the command line, and also
 # from the environment for the first two.
-SPHINXOPTS    ?=
-SPHINXBUILD   ?= sphinx-build
 SOURCEDIR     = .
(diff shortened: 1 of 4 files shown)


JelleZijlstra (approved):

stefanv (approved):
This looks great, thanks @shannonzhu.

My only recommendation would be that we use [MyST](https://myst-parser.readthedocs.io/en/latest/) markdown as the markup format to make it easier for people to contribute to the docs.

PR #836, opened 2021-08-18
Replace deprecated unittest aliases

hugovk:
> *Deprecated since version 2.7:* the aliases listed in the second column

https://docs.python.org/2/library/unittest.html#deprecated-aliases


srittau (approved):
Thanks!

PR #844, opened 2021-08-23
Add type stubs document

srittau:
This was originally intended to become a stub and was written by:

Sebastian Rittau <sebastian.rittau@zfutura.de>
Rebecca Chen <rechen@google.com>
Teddy Sudol <tsudol@google.com>
Jelle Zijlstra <jelle.zijlstra@gmail.com>

Also, remove unnecessary modindex link from main index.

### DIFF
--- a/docs/stubs.rst
+++ b/docs/stubs.rst
@@ -1,0 +1,1069 @@
+.. _stubs:
+
+**********
+Type Stubs
+**********
+
+Abstract
+========
+
+Optional type hints were introduced to the Python language in PEP 484
+[#pep484]_, based on the function annotation syntax from PEP 3107
+[#pep3107]_. Static type checkers can use type hints to prevent bugs,
+documentation tools can automatically add type information,
+and IDEs can offer improved autocompletion and support safer refactorings.
+
+PEP 484 also introduced *type stubs*, also called *stub files*,
+that provide type information for untyped Python packages and modules. Type
+stubs serve multiple purposes:
+
+* They are the only way to add type information to extension modules.
+* They can provide type information for packages that do not wish to
+  add them inline.
+* They can be distributed separately from the implementation.
+  This allows stubs to be developed at a different pace or by different
+  authors, which is especially useful when adding type annotations to
+  existing packages.
+* They can act as documentation, succinctly explaining the external
+  API of a package, without including the implementation or private
+  members.
+
+This PEP aims to give guidance to both authors of type stubs and developers
+of type checkers and other tools. It describes the constructs that can be used safely in type stubs,
+suggests a style guide for them, and lists constructs that type
+checkers are expected to support.
+
+Type stubs that only use constructs described in this PEP should work with
+all type checkers that also follow this PEP.
+Type stub authors can elect to use additional constructs, but
+must be prepared that some type checkers will not parse them as expected.
+
+A type checker that conforms to this PEP will parse a type stub that only uses
+constructs described here without error and will not interpret any
+construct in a contradictory manner. However, type checkers are not
+required to implement checks for all these constructs, and
+can elect to ignore unsupported ones. Additionally type checkers
+can support constructs not described in this PEP and tool authors are
+encouraged to experiment with additional features.
+
+This PEP is intended as a living document and will be updated as new
+features are supported or best practices evolve.
+
+Syntax
+======
+
+Type stubs are syntactically valid Python 3.7 files with a ``.pyi`` suffix.
+The Python syntax used for type stubs is independent from the Python
+versions supported by the implementation, and from the Python version the type
+checker runs under (if any). Therefore, type stub authors should use the
+latest available syntax features in stubs (up to Python 3.7), even if the
+implementation supports older, pre-3.7 Python versions.
+Type checker authors are encouraged to support syntax features from
+post-3.7 Python versions, although type stub authors should not use such
+features if they wish to maintain compatibility with all type checkers.
+
+For example, Python 3.7 added the ``async`` keyword (see PEP 492 [#pep492]_).
+Stub authors should use it to mark coroutines, even if the implementation
+still uses the ``@coroutine`` decorator. On the other hand, type stubs should
+not use the positional-only syntax from PEP 570 [#pep570]_, introduced in
+Python 3.8, although type checker authors are encouraged to support it.
+
+Stubs are treated as if ``from __future__ import annotations`` is enabled.
+In particular, built-in generics and forward references can be used.
+
+Starting with Python 3.8, the ast_ module from the standard library supports
+all syntax features required by this PEP. Older Python versions can use the
+typed_ast_ package, which also supports Python 3.7 syntax and ``# type``
+comments.
+
+Distribution
+============
+
+Type stubs can be distributed with or separately from the implementation;
+see PEP 561 [#pep561]_ for more information. The typeshed_ project
+includes stubs for Python's standard library and several third-party
+packages. These are usually distributed with type checkers and do not
+require separate installation.
+
+Supported Constructs
+====================
+
+This sections lists constructs that type checkers will accept in type stubs.
+Type stub authors can safely use these constructs. If a
+construct is marked as "unspecified", type checkers may handle it
+as they best see fit or report an error. Linters should usually
+flag those constructs. Type stub authors should avoid using them to
+ensure compatibility across type checkers.
+
+Unless otherwise mentioned, type stubs support all features from the
+``typing`` module of the latest released Python version. If a stub uses
+typing features from a later Python version than what the implementation
+supports, these features can be imported from ``typing_extensions`` instead
+of ``typing``.
+
+For example, a stub could use ``Literal``, introduced in Python 3.8,
+for a library supporting Python 3.7+::
+
+    from typing_extensions import Literal
+
+    def foo(x: Literal[""]) -> int: ...
+
+Comments
+--------
+
+Standard Python comments are accepted everywhere Python syntax allows them.
+
+Two kinds of structured comments are accepted:
+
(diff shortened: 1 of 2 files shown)


srittau:
At the moment, this is a 1:1 copy of the proposed PEP, minus the header. I want to iterate on that in future PRs, but for now I wanted to have a pristine copy of our last version in the history.

srittau:
Cf. srittau/type-stub-pep#102, cc @JelleZijlstra, @rchen152, @Solumin

docs/stubs.rst
```diff
@@ -0,0 +1,1069 @@
...
+comments.
+
+Distribution
+============
+
+Type stubs can be distributed with or separately from the implementation;
+see PEP 561 [#pep561]_ for more information. The typeshed_ project
+includes stubs for Python's standard library and several third-party
+packages. These are usually distributed with type checkers and do not
+require separate installation.
```
Akuli, line 86: This appears to be a bit outdated, since now there are `types-foo` packages on pypi.

docs/stubs.rst
```diff
@@ -0,0 +1,1069 @@
...
+
+Standard Python comments are accepted everywhere Python syntax allows them.
+
+Two kinds of structured comments are accepted:
+
+* A ``# type: X`` comment at the end of a line that defines a variable,
+  declaring that the variable has type ``X``. However, PEP 526-style [#pep526]_
+  variable annotations are preferred over type comments.
+* A ``# type: ignore`` comment at the end of any line, which suppresses all type
+  errors in that line.
```
Akuli, line 122: Even though mypy is just one type checker, it might be good to emphasize that `# type: ignore[foo]` is a mypy-specific extension and should not be used in stubs. Typeshed still uses it in a few places, but we should change those.

docs/stubs.rst
```diff
@@ -0,0 +1,1069 @@
...
+* ``import X as X``
+* ``from Y import X as X``
+* ``from Y import *``
+
+Here are some examples of imports that make names available for internal use in
+a stub but do not re-export them::
+
+    import X
+    from Y import X
+    from Y import X as OtherX
```
Akuli, line 140: I thought `from Y import X as OtherX` would re-export. What is the rationale for not re-exporting in this case? If you don't want to re-export, you can always do `from Y import X as _OtherX`, which most stubs in typeshed already do in this case.

docs/stubs.rst
```diff
@@ -0,0 +1,1069 @@
...
+
+Sub-modules are always exported when they are imported in a module.
+For example, consider the following file structure::
+
+    foo/
+        __init__.pyi
+        bar.pyi
+
+Then ``foo`` will export ``bar`` when one of the following constructs is used in
+``__init__.pyi``::
```
Akuli, line 155: It might be good to emphasize that there is no good way to avoid re-exporting. For example, `tkinter/__init__.py` has some methods that accept `tkinter.font.Font` objects, even though `import tkinter` does not import `tkinter/font.py`. As far as I know, there's no way to express this in the type system, and mypy thinks that `import tkinter` also loads `tkinter.font` even though it doesn't.

docs/stubs.rst
```diff
@@ -0,0 +1,1069 @@
...
+
+By default, ``from foo import *`` imports all names in ``foo`` that
+do not begin with an underscore. When ``__all__`` is defined, only those names
+specified in ``__all__`` are imported::
+
+  __all__ = ['public_attr', '_private_looking_public_attr']
+
+    public_attr: int
+    _private_looking_public_attr: int
+    private_attr: int
```
Akuli, line 173: Is the indentation weird on purpose?

docs/stubs.rst
```diff
@@ -0,0 +1,1069 @@
...
+
+Type checkers can handle cyclic imports in stub files.
```
Akuli, line 175: I find this a bit confusing. In typeshed we expect cyclic imports to just work in typeshed, and it's not some kind of optional thing that a type checker can choose to not support.

docs/stubs.rst
```diff
@@ -0,0 +1,1069 @@
...
+        def utility(x): ...  # x has type Any
+
+But::
+
+    _T = TypeVar("_T")
+
+    class Foo:
+        def do_things(self: _T): ...  # self has type _T
+        @classmethod
+        def create_it(cls: _T): ...  # cls has type _T
```
Akuli, line 287: Even though this is just a toy example, it would be nice if it didn't misuse `TypeVar`s by mentioning them only once in the method signature. So for example `def create_it(cls: _T) -> _T: ...`

docs/stubs.rst
```diff
@@ -0,0 +1,1069 @@
...
+        rgb_value: int  # no way for type checkers to know that this is not an enum member
+
+Unsupported Features
+--------------------
+
+Currently, positional-only argument syntax (PEP 570 [#pep570]_),
+unions using the pipe operator (``|``) (PEP 604 [#pep604]_),
+``ParamSpec`` (PEP 612 [#pep612]_), and ``TypeAlias`` (PEP 613 [#pep613]_)
+are not supported by all type
+checkers and should not be used in stubs.
```
Akuli, line 476: Even though documentation shouldn't be used when it is outdated, I don't think it's great if the documentation is outdated the moment it is created :)

I think it would be better to either explain when `|` unions should be used, or wait until we can just always use them in typeshed. Not all stubs go to typeshed, but it just doesn't make much sense to me to intentionally have more or less official documentation disagreeing with typeshed's practices.

docs/stubs.rst
```diff
@@ -0,0 +1,1069 @@
...
+
+Example::
+
+    def list2cmdline(seq: Sequence[str]) -> str: ...  # undocumented
+
+Such undocumented objects are allowed because omitting objects can confuse
+users. Users who see an error like "module X has no attribute Y" will
+not know whether the error appeared because their code had a bug or
+because the stub is wrong. Although it may also be helpful for a type
+checker to point out usage of private objects, we usually prefer false
```
Akuli, line 513: What exactly does "we" here mean? This appears to be taken from typeshed's CONTRIBUTING.md where "we" clearly means the typeshed maintainers.

docs/stubs.rst
```diff
@@ -0,0 +1,1069 @@
...
+----------------
+
+Partial stubs can be useful, especially for larger packages, but they should
+follow the following guidelines:
+
+* Included functions and methods should list all arguments, but the arguments
+  can be left unannotated.
+* Do not use ``Any`` to mark unannotated arguments or return values.
+* Partial classes should include a ``__getattr__()`` method marked with an
+  ``# incomplete`` comment (see example below).
```
Akuli, line 560: In this context, specifying `# incomplete` comments looks like it is something that type checkers should parse. I believe that's not the intention here, and we should keep the `# incomplete` comment markers only in typeshed's CONTRIBUTING.md?

docs/stubs.rst
```diff
@@ -0,0 +1,1069 @@
...
+Types
+-----
+
+Generally, use ``Any`` when a type cannot be expressed appropriately
+with the current type system or using the correct type is unergonomic.
+
+Use ``float`` instead of ``Union[int, float]``.
+Use ``None`` instead of ``Literal[None]``.
+For argument types,
+use ``bytes`` instead of ``Union[bytes, memoryview, bytearray]``.
```
Akuli, line 903: I think this needs a bit more explaining, or a link to somewhere where it's documented, since it does not match runtime behaviour:

```
>>> isinstance(bytearray(), bytes)
False
```

docs/stubs.rst
```diff
@@ -0,0 +1,1069 @@
...
+    def to_string(o: Any) -> str: ...
+
+Maybe::
+
+    class MyProto(Protocol):
+        def foo(self) -> list[int]: ...
+        def bar(self) -> Mapping[str]: ...
+
+Avoid ``Union`` return types, since they require ``isinstance()`` checks.
+Use ``Any`` if necessary.
```
Akuli, line 951: Maybe this part should mention the `str | Any` trick. A return type like `str | Any` means that it is not an error to treat the return value as a string, but it could be something else too.

docs/stubs.rst
hauntsaninja: It used to re-export, this was changed / clarified in PEP 484 somewhat recently (https://github.com/python/peps/pull/1630). The reason is more relevant for .py than it is for .pyi — basically to better match the people's reexport intentions in the wild — but the discussion was here: https://mail.python.org/archives/list/typing-sig@python.org/thread/YLJPWECBNPD2K4TRIBRIPISNUZJCRREY/#FPTLFORIBPKY36UO7PLO47HJDCLG2KD5

srittau:
Thank you for your comments. I will revisit them once this gets merged. For the time being, I want to merge this "as is" to have a "pristine" version in the git history.

docs/stubs.rst
Akuli: So the rule is now simpler than ever before, `as` always re-exports?

Akuli:
>This was originally intended to become a stub

Become a PEP maybe?

What happened to this PEP? Did it never become a PEP? I'm mostly worried about having outdated information in a PEP that shows up in google searches.

JelleZijlstra:
Yes, we intended to submit it as a PEP, but never did because we couldn't find a sponsor. The typing docs are now probably a better fit. The PEP is unlikely to show up in Google searches because it just existed in @srittau's repo.

docs/stubs.rst
srittau: `as` only re-exports, if the alias is the same as the original name.

docs/stubs.rst
srittau: This should probably be more prominent in the rest of the typing documentation (it is explained in the typing module docs), but the style guide is not really the correct place to explain it, just as `float == float | int` is not explained.

PR #846, opened 2021-08-23
Various improvements to the Type Stubs document

srittau:
* Rename "Abstract" to "Introduction" and remove fluff originally intended
  for the PEP.
* "This PEP" -> "This document"
* Remove note about the living document status. This is a given, not that
  it isn't a PEP anymore.
* Third-party stubs are now distributed via PyPI.
* Mention "# type: ignore[xxx]".
* Fix indentation.
* Reformulate sentence about cyclic imports.
* Use type vars correctly in example.
* Use new union syntax.
* Remove "we" from sentence and slightly reformulate.
* Clarify that "# incomplete" is mainly intended for stub authors.
* Mention "X | Any" return type. Still needs to be explained.

### DIFF
--- a/docs/stubs.rst
+++ b/docs/stubs.rst
@@ -4,18 +4,11 @@
 Type Stubs
 **********
 
-Abstract
-========
+Introduction
+============
 
-Optional type hints were introduced to the Python language in PEP 484
-[#pep484]_, based on the function annotation syntax from PEP 3107
-[#pep3107]_. Static type checkers can use type hints to prevent bugs,
-documentation tools can automatically add type information,
-and IDEs can offer improved autocompletion and support safer refactorings.
-
-PEP 484 also introduced *type stubs*, also called *stub files*,
-that provide type information for untyped Python packages and modules. Type
-stubs serve multiple purposes:
+*type stubs*, also called *stub files*, provide type information for untyped
+Python packages and modules. Type stubs serve multiple purposes:
 
 * They are the only way to add type information to extension modules.
 * They can provide type information for packages that do not wish to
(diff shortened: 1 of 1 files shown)


docs/stubs.rst
```diff
@@ -119,7 +111,9 @@ Two kinds of structured comments are accepted:
   declaring that the variable has type ``X``. However, PEP 526-style [#pep526]_
   variable annotations are preferred over type comments.
 * A ``# type: ignore`` comment at the end of any line, which suppresses all type
-  errors in that line.
+  errors in that line. The type checker mypy supports suppressing certain
```
JelleZijlstra, line 114: If I recall correctly some other type checkers support similar constructs. A comparison table could be a useful part of the typing docs.

docs/stubs.rst
```diff
@@ -947,8 +944,8 @@ Maybe::
         def foo(self) -> list[int]: ...
         def bar(self) -> Mapping[str]: ...
 
-Avoid ``Union`` return types, since they require ``isinstance()`` checks.
-Use ``Any`` if necessary.
+Avoid union return types, since they require ``isinstance()`` checks.
+Use ``Any`` or ``X | Any`` if necessary.
```
JelleZijlstra, line 948: It would be useful to explain more about how the `X | Any` trick works: when should you use it? what does it do?

JelleZijlstra (approved):

docs/stubs.rst
srittau: Indeed, but as this is not stubs-specific, probably in a best practices document?

PR #855, opened 2021-08-24
Update supported features

srittau:
* Built-in generics are only partly supported.
* Pipe unions are not partly supported.
* ParamSpec is partly and type guards are completely supported.

Also fix a syntax problem, use pipe syntax in examples where applicable, and don't use built-in generics in examples where not applicable.

### DIFF
--- a/docs/stubs.rst
+++ b/docs/stubs.rst
@@ -59,7 +59,8 @@ not use the positional-only syntax from PEP 570 [#pep570]_, introduced in
 Python 3.8, although type checker authors are encouraged to support it.
 
 Stubs are treated as if ``from __future__ import annotations`` is enabled.
-In particular, built-in generics and forward references can be used.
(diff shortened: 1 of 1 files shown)


docs/stubs.rst
```diff
@@ -170,6 +171,35 @@ specified in ``__all__`` are imported::
 
 Type checkers support cyclic imports in stub files.
 
+Built-in Generics
+-----------------
+
+PEP 585 [#pep585]_ built-in generics are generally supported, with
+the following exceptions:
+
+* Built-in generics don't work in type aliases.
```
JelleZijlstra, line 180: To help this stay up to date, it would be useful to link to the appropriate mypy issues.

srittau: Good idea, although I would propose to link to the typeshed tracking issues as they give a bit better overview for all type checkers.

PR #866, opened 2021-08-29
Switch pipeline to Github Actions

davfsa:
This PR migrates the testing jobs from Travis CI to Github Actions

Also added python 3.10-dev to the CI

Closes #864

PR #869, opened 2021-08-30
Fixes crash on Python3.10-beta.2 and typing_extensions@3.10.0.1

sobolevn:
I got hit by it today: https://github.com/dry-python/returns/pull/1041/checks?check_run_id=3462681678


JelleZijlstra:
Thanks! We also need #866 in to make sure the CI works.

Fidget-Spinner (approved):
I arrived to the same solution too. Thank you for the fix!

sobolevn:
It was fast! 🎉 
Thanks everyone!

PR #871, opened 2021-08-30
fix _GenericAlias import

JelleZijlstra:
TypeGuard was broken in 3.9 because we weren't importing _GenericAlias in some versions after #868. I'm still working on testing other versions using pyenv locally.


srittau (approved):
LGTM if tests pass locally.

Fidget-Spinner (approved):
Thanks for digging this out!

PR #872, opened 2021-08-30
Fix tests for 3.10rc1 and up

Fidget-Spinner:
Before:
```
======================================================================
FAIL: test_get_type_hints_modules (__main__.GetTypeHintTests)
  File "D:\Ken\Documents\GitHub\typing\typing_extensions\src_py3\test_typing_extensions.py", line 473, in test_get_type_hints_modules
    self.assertEqual(gth(ann_module), ann_module_type_hints)
    ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
AssertionError: {1: 2, 'x': <class 'int'>, 'y': <class 'str'>, 'f[39 chars]loat} != {1: 2, 'f': typing.Tuple[int, int], 'x': <class '[21 chars]tr'>}
+ {1: 2, 'f': typing.Tuple[int, int], 'x': <class 'int'>, 'y': <class 'str'>}
- {1: 2,
-  'f': typing.Tuple[int, int],
-  'u': int | float,
-  'x': <class 'int'>,
-  'y': <class 'str'>}
```
Because `ann_module` added a new annotation for union tests in 3.10rc1.

### DIFF
--- a/typing_extensions/src_py3/test_typing_extensions.py
+++ b/typing_extensions/src_py3/test_typing_extensions.py
@@ -53,6 +53,7 @@ TYPING_3_5_1 = TYPING_LATEST or sys.version_info[:3] >= (3, 5, 1)
 TYPING_3_5_3 = TYPING_LATEST or sys.version_info[:3] >= (3, 5, 3)
 TYPING_3_6_1 = TYPING_LATEST or sys.version_info[:3] >= (3, 6, 1)
 TYPING_3_10_0 = TYPING_LATEST or sys.version_info[:3] >= (3, 10, 0)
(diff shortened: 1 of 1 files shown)


Fidget-Spinner:
@JelleZijlstra in case you were testing locally on 3.10rc1 and wondering why some tests were failing.

This isn't required for release, so we can commit it after 3.10.0.2 is out.

JelleZijlstra:
Thanks! I'm still running tox locally but it doesn't seem to have gotten past 3.5 yet so I may have to rethink my testing strategy. I will land this so that 3.10 tests pass on the new release.

PR #875, opened 2021-08-30
Fix linting

srittau:
* Split linting against implementation and tests into two steps.
* Add more error codes for test linting.
* Lint with Python 3.9.
* Fix a linting problem.

Ideally, we'd have just one configuration for linting both the implementation and tests.

### DIFF
--- a/.flake8-tests
+++ b/.flake8-tests
@@ -11,17 +11,25 @@ builtins = basestring, unicode
 max-line-length = 100
 ignore =
   # temporary ignores until we sort it out
+  B017,
+  E302,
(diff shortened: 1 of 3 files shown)


JelleZijlstra (approved):

JelleZijlstra:
There's another 3.10 CI issue, one that I didn't see locally.

srittau:
If I understand correctly, this is not really a 3.10 problem, but a CI problem. CI shouldn't run the typing tests for 3.10, only the typing_extension tests. Working on it.

PR #876, opened 2021-08-30
Fix and clean up CI script

srittau:
* Split Python 2.7 tests from Python 3 tests as they use different paths and require installation of the typing module.
* Split typing backport and typing_extension tests for Python 2.
* Install typing from pypi instead of using the local version for Python 2.
* Remove support for Python 3.4 and 3.5.
* Don't run typing backport tests for Python 3.
* Unpin pytest version for Python 3.
* Don't install unnecessary pytest plugins pytest-xdist and pytest-cov.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -8,6 +8,29 @@ permissions:
   contents: read
 
 jobs:
+  tests-27:
+    name: Run tests (2.7)
+    runs-on: ubuntu-latest
+
+    steps:
+      - uses: actions/checkout@v2
+
+      - name: Set up Python
+        uses: actions/setup-python@v2
+        with:
+          python-version: 2.7
+
+      - name: Install dependencies
+        run: |
+          python -m pip install --upgrade pip
+          pip install -r test-requirements.txt
+
+      - name: Run tests
+        run: |
+          pytest python2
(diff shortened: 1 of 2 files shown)


.github/workflows/ci.yml
```diff
@@ -32,16 +57,14 @@ jobs:
           python -m pip install --upgrade pip
           pip install -r test-requirements.txt
 
-      - name: Run tests
+      - name: Test typing
+        if: matrix.python-version < '3.7'
```
JelleZijlstra: typing has been in the stdlib since 3.5, so this should only be run on 2.7 and (for now) 3.4

srittau: Considering that we don't run CI for 3.4 at the moment (since it doesn't seem to exist), we can just remove this check completely (except for Python 2.7).

srittau:
Ready for review. I'm unsure why the 3.10 fails, but at least the CI script is now cleaner, and it fails in a test that it should actually run, not in the typing test, so this PR is an improvement. I will have another stab at this tomorrow, if no one beats me to it.

JelleZijlstra:
Might be an incompatibility in pytest or pytest-xdist. To get around it we could just stick to unittest.

PR #878, opened 2021-08-30
Remove the drafts directory

srittau:
This contained historic notes for PEP 484.

@gvanrossum Should we keep this around somewhere for historic purposes or is it enough if it slumbers in git history?


gvanrossum (approved):
Yeah, it's fine to get rid of this -- those notes were only important while we were discussing things (heatedly :-).

JelleZijlstra:
Pretty interesting to read these back now!

PR #880, opened 2021-08-31
Add caching to the new GHA workflow

sondrelg:
Hi!

This might be a bit of a premature optimisation, but I noticed you were transitioning to Github actions and I thought I'd propose adding caching to your current workflow 🙂 I also took the liberty of making two other minor changes which I'm happy to drop if you'd rather not include them - just let me know!

I don't think this should introduce any conflicts with https://github.com/python/typing/pull/876, but if that PR is merged first, I could rebase and add caching to that job too, if that's something you want (or the other way around hehe) 👍

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -29,7 +29,7 @@ jobs:
 
       - name: Install dependencies
         run: |
-          python -m pip install --upgrade pip
+          pip install --upgrade pip
           pip install -r test-requirements.txt
 
       - name: Run tests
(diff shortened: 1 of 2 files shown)


.github/workflows/ci.yml
```diff
@@ -58,8 +66,8 @@ jobs:
 
       - name: Install dependencies
         run: |
-          python -m pip install --upgrade pip
-          pip install -r test-requirements.txt
+          pip install --upgrade pip
+          pip install flake8
```
srittau, line 70: We are also installing flake8-bugbear (and flake8-pyi, which shouldn't matter). To avoid this kind of discrepancies, I'd prefer to continue using test-requirements.txt here, even if it installs some unneeded packages.

sondrelg: I see! I'll add caching here instead then 🙂

srittau (approved):

PR #889, opened 2021-09-09
Add "Typing Guidance for Python Libraries" document

srittau:
By Eric Traut, source: https://raw.githubusercontent.com/microsoft/pyright/main/docs/typed-libraries.md

See https://github.com/srittau/type-stub-pep/issues/94 for background.

This is a pristine copy of the original document. It needs to be
converted into RestructuredText and linked from the index as a next
step.

### DIFF
--- a/docs/libraries.md
+++ b/docs/libraries.md
@@ -1,0 +1,395 @@
+# Typing Guidance for Python Libraries
+
+Much of Python’s popularity can be attributed to the rich collection of Python libraries available to developers. Authors of these libraries play an important role in improving the experience for Python developers. This document provides some recommendations and guidance for Python library authors.
+
+These recommendations are intended to provide the following benefits:
+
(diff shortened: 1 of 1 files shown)


srittau:
Cc @erictraut

stefanv:
@srittau Another route would be to configure Sphinx to parse Markdown files, since people seem to enjoy writing in that format and we want to encourage contribution.

Option 1: Use [MyST](https://myst-parser.readthedocs.io/en/latest/sphinx/intro.html) which supports both markdown and RST.

Option 2: Use recommonmark by adding [this](https://github.com/skyportal/skyportal/blob/master/doc/conf.py#L37) and [this](https://github.com/skyportal/skyportal/blob/master/doc/conf.py#L54) line.

srittau:
While I usually prefer markdown, I'd prefer to keep using ReST exclusively here, for a few reasons:

* To stay consistent within the documentation.
* It's the standard for Python documentation.
* To allow easy reshuffling between documents.
* To be able to use advanced features like standardized crosslinks to Python documentation and index items.

stefanv:
OK, [here](https://gist.github.com/stefanv/a7476771dabc2c500abb94d3a163391f) is the Pandoc conversion to RST, in case that's helpful to start with.

srittau:
Thanks! That's quite useful indeed.

srittau:
Bumpity bump. I'd like to start integrating the document into our docs structure. We can have all discussions about the contents then. For now, I'd like this original version to be in version control.

PR #893, opened 2021-09-16
Drop Python 2 support for typing_extensions

srittau:
Update README accordingly


srittau:
As announced here: https://mail.python.org/archives/list/typing-sig@python.org/thread/65LVW6O6ET6QHWLD4SAFDFKMWOIPPXRK/

PR #916, opened 2021-11-02
Move individual pages from toplevel docs

shannonzhu:
Add a link to the live documentation page, also move the individual doc pages into a source/ directory to avoid eventually adding a lot of clutter to the toplevel docs as we add more content.


srittau (approved):

srittau:
The build failure is curious and doesn't occur when I apply the changes locally.

shannonzhu:
Yeah, I was just trying to figure out what's wrong - I also can't reproduce locally. It's also failing on another PR touching the docs in a different way: https://github.com/python/typing/pull/913

srittau:
I tried upgrading Sphinx in #917. That passes. Maybe this will fix this PR as well.

PR #929, opened 2021-11-09
Script to mail regular issue summaries to typing-sig

srittau:
Example output (with cheated "since" date, since last week I touched all issues):

```
From: Typing Bot <noreply@python.org>
To: typing-sig@python.org
Subject: Opened and changed typing issues week 2021-W44

The following is an overview of all issues and pull requests in the
typing repository on GitHub with the label 'topic: feature'
that were opened or updated last week, exluding closed issues.

---------------------------------------------------

The following issues and pull requests were opened last week: 

#928   Add a `import type` feature (like typescript)
       opened by @KotlinIsland
       https://api.github.com/repos/python/typing/issues/928

---------------------------------------------------

The following issues and pull requests were updated last week: 

#802   Wrapper/Proxy Generic Type
       opened by @hatal175
       https://api.github.com/repos/python/typing/issues/802
#926   Change PEP 647 to allow for narrowing in the negative case
       opened by @Zomatree
       https://api.github.com/repos/python/typing/issues/926

---------------------------------------------------

All issues and pull requests with the label 'topic: feature'
can be viewed under the following URL:

https://github.com/python/typing/issues?q=label%3A%22topic%3A+feature%22
```

This script would be run once a week as a cron job. Unfortunately it seems that we can't mail directly from GitHub Actions (at least not without a mail server configured for it), but I can set up a cron job on a company server.

One possible caveat: It's possible the Python mail server will reject the mail with a python.org-From from a remote web server, in which case we need to change the `From` to something else.

@gvanrossum Would this weekly report be okay for typing-sig?

Cc @shannonzhu

### DIFF
--- a/scripts/typing-summary.py
+++ b/scripts/typing-summary.py
@@ -1,0 +1,127 @@
+#!/usr/bin/env python3
+
+from __future__ import annotations
+
+import datetime
+from dataclasses import dataclass
+from typing import Any, Iterable, Sequence
+
+import requests
+
+ISSUES_API_URL = "https://api.github.com/repos/python/typing/issues"
+ISSUES_URL = "https://github.com/python/typing/issues?q=label%3A%22topic%3A+feature%22"
+
+
+@dataclass
+class Issue:
+    number: int
+    title: str
+    url: str
+    created: datetime.datetime
+    user: str
+    pull_request: bool = False
+
+
+def main() -> None:
+    since = previous_week_start()
+    issues = fetch_issues(since)
+    if len(issues) > 0:
+        new, updated = split_issues(issues, since)
+        print_summary(since, new, updated)
+
+
+def previous_week_start() -> datetime.date:
+    today = datetime.date.today()
+    return today - datetime.timedelta(days=today.weekday() + 7)
+
+
+def fetch_issues(since: datetime.date) -> list[Issue]:
+    """Return (new, updated) issues."""
+    j = requests.get(
+        ISSUES_API_URL,
+        params={
+            "labels": "topic: feature",
+            "since": f"{since:%Y-%m-%d}T00:00:00Z",
+            "per_page": "100",
+            "state": "open",
+        },
+        headers={"Accept": "application/vnd.github.v3+json"},
+    ).json()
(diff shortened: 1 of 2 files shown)


Gobot1234:
Apart from a couple of things LGTM:
- would it not be good to reform the issue urls so that they don't take you to json
- should this not include a github actions script to automate sending of this

scripts/typing-summary.py
```diff
@@ -0,0 +1,132 @@
...
+        headers={"Accept": "application/vnd.github.v3+json"},
+    ).json()
+    assert isinstance(j, list)
+    return [parse_issue(j_i) for j_i in j]
+
+
+def parse_issue(j: Any) -> Issue:
+    number = j["number"]
+    title = j["title"]
+    url = j["url"]
```
hauntsaninja, line 56:
```suggestion
    url = j["html_url"]
```

srittau:
> * should this not include a github actions script to automate sending of this

See my comment above:

> Unfortunately it seems that we can't mail directly from GitHub Actions (at least not without a mail server configured for it), but I can set up a cron job on a company server.

srittau:
I've now set up a cron job that executes this script every Monday morning. Let's see whether this get through tomorrow.

scripts/typing-summary.py
```diff
@@ -0,0 +1,144 @@
...
+            "The following is an overview of all issues and pull requests in the\n"
+            "typing repository on GitHub with the label 'topic: feature'\n"
+            "that were opened or updated last week, excluding closed issues.\n\n"
+            "---------------------------------------------------\n\n"
+        )
+    if len(new) >= 0:
+        s += "The following issues and pull requests were opened last week: \n\n"
+        s += "".join(generate_issue_text(issue) for issue in new)
+        s += "\n---------------------------------------------------\n\n"
+    if len(changed) >= 0:
```
Akuli, line 121: Maybe these should be `>`, so that they don't run every time.

srittau: Good catch!

srittau:
Ping. This has now been officially used for a few weeks. I think we should merge it.

PR #931, opened 2021-11-10
Clean-up `typing_extensions` -- meta

AA-Turner:
xref #892 
@srittau 

'Meta' PR with build changes -- note changed from the original PR as I saw setuptools intend to deprecate and remove `setup.py install` and `pyproject.toml` is the blessed solution. I was going to use setuptools as the backend for consistency but it fails on windows due to encoding problems (setuptools uses encoding=None when reading config files, and the default windows code page isn't UTF-8) with Łukasz's name -- flit seems to have no such problems, and was the first alternative when I searched for non-setuptools backends.

A


typing_extensions/pyproject.toml
```diff
@@ -0,0 +1,67 @@
...
+Typing Extensions -- Backported and Experimental Type Hints for Python
+
+The ``typing`` module was added to the standard library in Python 3.5, but
+many new features have been added to the module since then.
+This means users of older Python versions who are unable to upgrade will not be
+able to take advantage of new types added to the ``typing`` module, such as
+``typing.Protocol`` or ``typing.TypedDict``.
+
+The ``typing_extensions`` module contains backports of these changes.
+Experimental types that will eventually be added to the ``typing``
```
JelleZijlstra, line 21:
```suggestion
Experimental types that may eventually be added to the ``typing``
```
We include types from unaccepted PEPs.

JelleZijlstra (approved):

srittau (approved):

srittau:
Thanks!

PR #932, opened 2021-11-10
Clean-up `typing_extensions` -- `typing_extensions`

AA-Turner:
xref #892 
@srittau 

Actual `typing_extensions` PR. Unchanged from original.

A

### DIFF
--- a/typing_extensions/src_py3/typing_extensions.py
+++ b/typing_extensions/src_py3/typing_extensions.py
@@ -6,7 +6,7 @@ import typing
 
 # After PEP 560, internal typing API was substantially reworked.
 # This is especially important for Protocol class which uses internal APIs
-# quite extensivelly.
+# quite extensively.
 PEP_560 = sys.version_info[:3] >= (3, 7, 0)
 
 if PEP_560:
--- a/typing_extensions/src_py3/typing_extensions.py
+++ b/typing_extensions/src_py3/typing_extensions.py
@@ -22,13 +22,8 @@ if PEP_560:
     from typing import _GenericAlias
 else:
     from typing import GenericMeta, TypingMeta
-OLD_GENERICS = False
-try:
-    from typing import _type_vars, _next_in_mro, _type_check
-except ImportError:
-    OLD_GENERICS = True
 if sys.version_info[:2] == (3, 6):
-    from typing import _subs_tree  # noqa
+    from typing import _type_vars, _next_in_mro, _type_check, _subs_tree  # noqa
 
 # The two functions below are copies of typing internal helpers.
 # They are needed by _ProtocolMeta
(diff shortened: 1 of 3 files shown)


typing_extensions/src_py3/typing_extensions.py
```diff
@@ -1698,7 +1046,7 @@ def __init__(cls, name, bases, ns, total=True):
             # In Python 3.4 and 3.5 the __init__ method also needs to support the
             # keyword arguments.
             # See https://www.python.org/dev/peps/pep-0487/#implementation-details
```
srittau, line 1048: This comment seems obsolete.

typing_extensions/src_py3/typing_extensions.py
```diff
@@ -1780,7 +1128,10 @@ class Point2D(TypedDict):
     # Not exported and not a public API, but needed for get_origin() and get_args()
     # to work.
     _AnnotatedAlias = typing._AnnotatedAlias
+# 3.7-3.8
 elif PEP_560:
+    import operator
```
srittau, line 1133: Nit: While this import is only used for this branch, I suggest to import at the top for better overview.

srittau (changes requested):
Thanks!

You (sensibly) removed a few non-underscored, but also not exported via `__all__` attributes. As this is technically an API change (even though it's not exported via `__all__`), could you add these removals to `typing_extensions/CHANGELOG`:

* `TypingMeta`
* `OLD_GENERICS`
* `SUBS_TREE`
* `HAVE_ANNOTATED`
* `HAVE_PROTOCOLS`

Also could you mention in the Changelog that starting with 4.0.0 support for Python 2 and <3.6 has been dropped?

A few more remarks below.

typing_extensions/src_py3/typing_extensions.py
AA-Turner: Done

typing_extensions/src_py3/typing_extensions.py
AA-Turner: Done

typing_extensions/src_py3/typing_extensions.py
```diff
@@ -155,27 +78,16 @@ def _check_methods_in_mro(C, *methods):
...
-# Annotated relies on substitution trees of pep 560. It will not work for
-# versions of typing older than 3.5.3
-HAVE_ANNOTATED = PEP_560 or SUBS_TREE
-
 if PEP_560:
     __all__.extend(["get_args", "get_origin", "get_type_hints"])
 
-if HAVE_ANNOTATED:
-    __all__.append("Annotated")
+__all__.extend(['Annotated', 'Protocol', 'runtime', 'runtime_checkable'])
```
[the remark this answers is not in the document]
AA-Turner: Done

AA-Turner:
Sorry, forgot to add a changelog entry earlier -- have now done so.

srittau (approved):
Thanks again. For future reference: Please don't force push as it makes it harder to review changes.

AA-Turner:
Sorry - I had to rebase onto `master` to pick up the changelog -- is it better to merge in `HEAD` instead of rebase + force push?
A

srittau:
Yes, we squash merge anyway, so rebasing should not be necessary.

PR #934, opened 2021-11-11
Make library guidelines checker agnostic

Bibo-Joshi:
Hi. This PR is a spin-off of #913. Citing:

> Moreover, because the guideline was copied as-is from pyright, it strongly emphasizes pyright as typechecker, which I also wouldn't expect from a python source. For comparison: the [docs of the typing module](https://docs.python.org/3/library/typing.html#typing.Annotated) which are somewhat longer than these guidelines mention "e.g., via mypy or Pyre" exactly once.

Hence, this PR tries to make guidelines more agnostic of the used type checker. It also tries to emphasize that the guidelines are not (yet) fixed rules, but in fact guidelines - which at least is my understanding of what the document is supposed to represent.


JelleZijlstra (approved):

docs/source/libraries.rst
```diff
@@ -119,11 +119,13 @@ determine the value of ``__all__``.
 Type Completeness
 =================
 
-A “py.typed” library is said to be “type complete” if all of the symbols
+A “py.typed” library should aim to be type complete so that type
+checking and inspection can work to their full extend. Here we say that a
```
hauntsaninja, line 123:
```suggestion
checking and inspection can work to their full extent. Here we say that a
```

erictraut:
These changes look good to me. Thanks @Bibo-Joshi.

PR #935, opened 2021-11-11
Use unittest for testing instead of pytest

srittau:


hauntsaninja (commented):
Thanks, can we leave a comment about not having `pip install`s, just in case?

srittau:
Sure, where do you want the comment. In the yaml file?

.github/workflows/ci.yml
```diff
@@ -26,21 +26,10 @@ jobs:
...
-
-      - name: Install dependencies
-        run: |
-          pip install --upgrade pip
-          pip install -r test-requirements.txt
-
       - name: Test typing_extensions
-        run: pytest typing_extensions/src_py3
+        run: |
+          cd typing_extensions/src_py3
```
hauntsaninja, line 31:
```suggestion
        run: |
          # Be wary of running `pip install` here, since it becomes easy for us to
          # accidentally pick up typing_extensions as installed by a dependency 
          cd typing_extensions/src_py3
```

JelleZijlstra (approved):
Would be good to also have an assertion in the unit test file to make sure we're picking up the right copy of the file (something like `assert dirname(typing_extensions.__file__) == dirname(__file__)`.

It's going to be tricky for people testing the module locally to avoid picking up an installed copy, so it would be useful to provide this check and guidance like using an empty venv.

PR #941, opened 2021-11-14
prepare release 4.0.0

JelleZijlstra:


srittau (approved):

JelleZijlstra:
That was fast. :) I'm going to be out for a couple of hours and do the actual release after that.

PR #942, opened 2021-11-14
Add `Self` type to the README of typing-extensions

jack1142:
This type was missing from the README even though it is part of the typing-extensions package.


srittau (approved):
Thanks!

JelleZijlstra:
Thanks, good catch

PR #951, opened 2021-11-18
Update typing-extensions README

JelleZijlstra:
- Add an explicit inclusion policy: new stuff can go in here as soon as there is a PEP
- Remove obsolete discussion of the typing PyPI package
- Organize the list of contents by Python version, so it's easier to find the interesting bits. Maybe we should deprecate the stuff that was new in 3.5/3.6 at some point.

### DIFF
--- a/typing_extensions/README.rst
+++ b/typing_extensions/README.rst
@@ -9,15 +9,19 @@ Typing Extensions
 Overview
 ========
 
-The ``typing_extensions`` module contains both backports of ``typing`` features
-as well as experimental types that will eventually be added to the ``typing``
-module, such as ``Protocol`` (see PEP 544 for details about protocols and
-static duck typing) or ``TypedDict`` (see PEP 589).
+The ``typing_extensions`` module serves two related purposes:
 
-Users of other Python versions should continue to install and use
-the ``typing`` module from PyPi instead of using this one unless
-specifically writing code that must be compatible with multiple Python
-versions or requires experimental types.
+- Enable use of new type system features on older Python versions. For example,
+  ``typing.TypeGuard`` is new in Python 3.10, but ``typing_extensions`` allows
+  users on Python 3.6 through 3.9 to use it too.
+- Enable experimentation with new type system PEPs before they are accepted and
+  added to the ``typing`` module.
+  
+New features may be added to ``typing_extensions`` as soon as they are specified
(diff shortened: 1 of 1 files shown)


typing_extensions/README.rst
```diff
@@ -31,49 +35,62 @@ Included items
...
+  - ``runtime_checkable`` (see PEP 544)
+  - ``TypedDict`` (see PEP 589)
+  - ``get_origin`` (``typing_extensions`` provides this function only in Python 3.7+)
+  - ``get_args`` (``typing_extensions`` provides this function only in Python 3.7+)
+
+- In ``typing`` since Python 3.7
+
+  - ``OrderedDict``
+
+- In ``typing`` since Python 3.5 or 3.6 (see `the typing documentation
```
JelleZijlstra, line 72: Separating out 3.5 and 3.6 is hard because things were added in patch releases (for example, ChainMap was new in 3.5.4 and 3.6.1). Listing out all the details about that doesn't seem useful.

hauntsaninja (approved):

typing_extensions/README.rst
```diff
@@ -9,15 +9,19 @@ Typing Extensions
...
+  ``typing.TypeGuard`` is new in Python 3.10, but ``typing_extensions`` allows
+  users on Python 3.6 through 3.9 to use it too.
+- Enable experimentation with new type system PEPs before they are accepted and
+  added to the ``typing`` module.
+  
+New features may be added to ``typing_extensions`` as soon as they are specified
+in a PEP that has been added to the `python/peps <https://github.com/python/peps>`_
+repository. If the PEP is accepted, the feature will then be added to ``typing``
+for the next CPython release. No typing PEP has been rejected so far, so we
+haven't yet figured out how to deal with that possibility.
```
hauntsaninja, line 24: lol

srittau: Maybe we'll remove it in the next major version. But I have no strong feelings regarding this.

typing_extensions/README.rst
srittau: Especially since we don't support 3.5 anymore.

srittau (approved):

typing_extensions/README.rst
JelleZijlstra: That makes sense to me too, but we can figure that out when we get to it.

PR #963, opened 2021-11-24
PEP 646 implementation

### LINKED ISSUE #950: Implement PEP 646 in typing_extensions
PEP 646 has been accepted (subject to some confusion about an edge case that's still being discussed), and we should have an implementation of it in typing-extensions.

@mrahtz @pradeep90 is there already a runtime implementation of PEP 646 somewhere? If not, I'd be happy to work on getting the right bits into typing-extensions and typing.

JelleZijlstra:
Fixes #950.

This turned out to require even more hacks than usual, because the runtime in old versions of typing really doesn't want things that are not TypeVars to be used as TypeVars.

The tests were copied from https://github.com/python/cpython/pull/24527/files and the implementation is mostly inspired by Required/NotRequired (Unpack) and ParamSpec (TypeVarTuple).


JelleZijlstra:
The last 3.6 failures are annoying and I don't see an easy fix, I think I'll give it another try tomorrow.

JelleZijlstra:
This is now finally ready for review. This is uncommonly hacky even by typing-extensions standards, because the older versions of `typing` only know about TypeVars and now we need to convince them to somehow also accept unpacked TypeVarTuples. I solved this with monkeypatching (yuck).

For 3.6, which had a very different internal implementation of Generic, it's still not enough. Given that 3.6 is EOL and most usage of TypeVarTuple/Unpack actually works correctly, I decided to just leave this as a documented limitation.

Fidget-Spinner:
@JelleZijlstra Sorry I'll try to review this in the next week or two, swamped right now.

JelleZijlstra:
No hurry, take your time!

typing_extensions/src/test_typing_extensions.py
```diff
@@ -2271,6 +2282,138 @@ def return_tuple(self) -> TupleSelf:
...
+            Union[Unpack[Xs], int].__origin__,
+            Union
+        )
+
+    @skipUnless(PEP_560, "Unimplemented for 3.6")
+    def test_concatenation(self):
+        Xs = TypeVarTuple('Xs')
+        Tuple[int, Unpack[Xs]]
+        Tuple[Unpack[Xs], int]
+        Tuple[int, Unpack[Xs], str]
```
Fidget-Spinner, line 2346: I'm confused, what are these testing?

typing_extensions/src/typing_extensions.py
```diff
@@ -91,6 +73,88 @@ def _check_generic(cls, parameters):
...
+    if elen is _marker:
+        if not hasattr(cls, "__parameters__") or not cls.__parameters__:
+            raise TypeError(f"{cls} is not a generic class")
+        elen = len(cls.__parameters__)
+    alen = len(parameters)
+    if alen != elen:
+        if hasattr(cls, "__parameters__"):
+            parameters = [p for p in cls.__parameters__ if not _is_unpack(p)]
+            num_tv_tuples = sum(isinstance(p, TypeVarTuple) for p in parameters)
+            if num_tv_tuples > 0 and alen >= elen - num_tv_tuples:
```
Fidget-Spinner, line 106: Could you add parentheses around the operations here please? They're slightly hard to read.

typing_extensions/src/typing_extensions.py
```diff
@@ -91,6 +73,88 @@ def _check_generic(cls, parameters):
...
+    if '__slots__' in dict_copy:
+        for slot in dict_copy['__slots__']:
+            dict_copy.pop(slot, None)
+    return dict_copy
+
+
+_marker = object()
+
+
+def _check_generic(cls, parameters, elen=_marker):
```
Fidget-Spinner, line 91: I think Mathew pointed out in his documenting typing.py PR that `parameters` is confusing. Should we use `args` or `type_args` instead?

Also what's `elen`'s default for? When would we not want to pass in `__parameters__` ourselves?

typing_extensions/README.rst
```diff
@@ -116,9 +118,14 @@ These changes are _not_ backported to prevent subtle compatibility
...
 Certain types have incorrect runtime behavior due to limitations of older
-versions of the typing module.  For example, ``ParamSpec`` and ``Concatenate``
-will not work with ``get_args``, ``get_origin``. Certain PEP 612 special cases
-in user-defined ``Generic``\ s are also not available.
+versions of the typing module:
+
+- ``ParamSpec`` and ``Concatenate`` will not work with ``get_args`` and
+  ``get_origin``. Certain PEP 612 special cases in user-defined
+  ``Generic``\ s are also not available.
+- ``Unpack`` from PEP 646 does not work properly with user-defined
```
Fidget-Spinner, line 126: To clarify, this means even with our monkey patching, subscripting with arbitrary number of type arguments won't wrok right?

Fidget-Spinner (commented):
Looks good in general. Thanks Jelle! BTW, should I be reviewing this or Mathew's CPython PR first? I'd imagine any tweaks to his PR would affect yours.

typing_extensions/src/typing_extensions.py
JelleZijlstra: This is a modified copy of typing.py's `_check_generic`, and it's monkeypatched into typing later. We probably shouldn't modify the function's interface, because that makes the monkeypatching even riskier.

typing_extensions/src/test_typing_extensions.py
JelleZijlstra: I guess just that these don't fail. I can make them check `__args__` or something too.

typing_extensions/README.rst
JelleZijlstra: Defining the class doesn't even work. I'll make the note more precise.

```
Python 3.6.10 (default, Apr 28 2021, 18:14:25) 
[GCC Apple LLVM 12.0.0 (clang-1200.0.32.29)] on darwin
>>> from typing import Generic
>>> from typing_extensions import Unpack, TypeVarTuple
>>> Ts = TypeVarTuple("Ts")
>>> class X(Generic[Unpack[Ts]]): pass
... 
  File "/Users/jelle/.pyenv/versions/3.6.10/lib/python3.6/typing.py", line 965, in __new__
    ", ".join(str(g) for g in gvars)))
TypeError: Some type variables (Ts) are not listed in Generic[typing_extensions.Unpack[Ts]]
```

JelleZijlstra:
Thanks! I think it's better to focus on this first, since 3.11 has months of time left in any case, but with typing-extensions we can release at any time.

However, I'll look at his current PR for things to improve here; I think he wrote better docstrings.

typing_extensions/src/test_typing_extensions.py
JelleZijlstra: Done

srittau (approved):
I won't claim to really understand what is going on with typing internals, but I see no red flags.

mrahtz:
Thanks all - especially @JelleZijlstra!

Issue #961, opened 2021-11-23
typing-extensions 4.0 sdist issue


dtrodrigues:
`import typing_extensions` fails from the sdist for 4.0. This works for 3.x

```
[root@b2fff8624d4d ~]# python3 -m pip install  --no-binary :all: typing-extensions==4.0.0
Collecting typing-extensions==4.0.0
  Using cached typing_extensions-4.0.0.tar.gz (37 kB)
  Installing build dependencies ... done
  Getting requirements to build wheel ... done
  Preparing metadata (pyproject.toml) ... done
Building wheels for collected packages: typing-extensions
  Building wheel for typing-extensions (pyproject.toml) ... done
  Created wheel for typing-extensions: filename=typing_extensions-4.0.0-py3-none-any.whl size=22635 sha256=217b105e4382e579581f72d69a2390a59625aa55787f6f3c83f48e88e935c988
  Stored in directory: /root/.cache/pip/wheels/23/1c/b9/a6843f8ee0b06cae64a7fa6dcc1af022bc4fd3ba707147db2d
Successfully built typing-extensions
Installing collected packages: typing-extensions
Successfully installed typing-extensions-4.0.0
WARNING: Running pip as the 'root' user can result in broken permissions and conflicting behaviour with the system package manager. It is recommended to use a virtual environment instead: https://pip.pypa.io/warnings/venv
[root@b2fff8624d4d ~]# python3 -c 'import typing_extensions'
Traceback (most recent call last):
  File "<string>", line 1, in <module>
ModuleNotFoundError: No module named 'typing_extensions'
[root@b2fff8624d4d ~]# pip uninstall -y typing-extensions
Found existing installation: typing-extensions 4.0.0
Uninstalling typing-extensions-4.0.0:
  Successfully uninstalled typing-extensions-4.0.0
WARNING: Running pip as the 'root' user can result in broken permissions and conflicting behaviour with the system package manager. It is recommended to use a virtual environment instead: https://pip.pypa.io/warnings/venv
[root@b2fff8624d4d ~]# python3 -m pip install --no-binary :all: typing-extensions==3.10.0.2
Collecting typing-extensions==3.10.0.2
  Using cached typing_extensions-3.10.0.2.tar.gz (46 kB)
  Preparing metadata (setup.py) ... done
Skipping wheel build for typing-extensions, due to binaries being disabled for it.
Installing collected packages: typing-extensions
    Running setup.py install for typing-extensions ... done
[...]

PR #965, opened 2021-11-25
Rename src_py3 to src

AA-Turner:
As per #961; https://github.com/python/typing/issues/961#issuecomment-978814784

Also added a worflow file to test packaging and installing wheels and sdists to try to prevent future regressions.

A


dtrodrigues:
Will there be a post/patch version uploaded to PyPI in the short-term with an updated sdist?

JelleZijlstra:
#968

Issue #964, opened 2021-11-25
Self being a typing._SpecialForm prevents a case from PEP 673


CharString:
Among the [valid locations for Self](https://www.python.org/dev/peps/pep-0673/#valid-locations-for-self) in PEP 673 we find:
```python
TupleSelf = Tuple[Self, Self]
class Alias:
    def return_tuple(self) -> TupleSelf:
        return (self, self)
```

But `typing._type_check` regards `typing._SpecialForm` instances invalid type arguments
https://github.com/python/cpython/blob/e0f8a3e9b96bbc2597f61be96993ef0c768a19fe/Lib/typing.py#L179

So this use will result in a

    TypeError: Plain typing_extensions.Self is not valid as type argument

This also affects uses in linked lists and recursive dataclasses lke 
```python
@dataclasses.dataclass
class MenuItem:
    url: str
    title: str
    description: str = ""
    items: typing.List[typing_extensions.Self] = dataclasses.field(default_factory=list)
```

Gobot1234:
Hi, I'm not entirely sure what can be done here until the PEP is accepted and I can change https://github.com/python/cpython/blob/e0f8a3e9b96bbc2597f61be96993ef0c768a19fe/Lib/typing.py#L176
To 
```py
    if arg in (Any, NoReturn, Final, Self):
```
If you have any suggestions for how to fix this now, I'd be more than happy to implement them.

However, if you need this to work right now you you have a few options:
- use the PEP 585 builtin versions of the types which are much more lenient than the typing versions.
- avoid the annotations actually being executed at runtime (`typing_extensions.TypeAlias` and `__future__.annotations` or stringify the types).

JelleZijlstra:
We can probably fudge things in typing-extensions so that it's not an instance of `_SpecialForm`.

CharString:
@Gobot1234 Yes, for my use case I'm using a string as ForwardRef now. I just [wondered if I were doing something wrong](https://stackoverflow.com/q/70083371/383793).

@JelleZijlstra Like vendoring the _SpecialForm from the cpython 3.11 codebase? Or a more elegant fudge? Namespaces are one honking great idea :-)

JelleZijlstra:
Yes, vendoring would be the way to go. The place where this throws lets things it doesn't recognize through, so if we just make Self look like an unrelated object, it should pass.

CharString:
Something like this?

PR #966, opened 2021-11-26
Vendor typing._SpecialForm to fool typing._type_check

CharString:
Adds a local copy of _SpecialForm in our namespace, so
typing._type_check won't raise TypeError. (#964)

### DIFF
--- a/typing_extensions/src/typing_extensions.py
+++ b/typing_extensions/src/typing_extensions.py
@@ -2077,10 +2077,10 @@ if not hasattr(typing, "Self") and sys.version_info[:2] >= (3, 7):
             raise TypeError(f"Cannot instantiate {self!r}")
 
         def __or__(self, other):
-            return Union[self, other]
+            return typing.Union[self, other]
 
         def __ror__(self, other):
-            return Union[other, self]
+            return typing.Union[other, self]
 
         def __instancecheck__(self, obj):
             raise TypeError(f"{self} cannot be used with isinstance()")
--- a/typing_extensions/src/typing_extensions.py
+++ b/typing_extensions/src/typing_extensions.py
@@ -2048,12 +2048,55 @@ else:
 
     TypeGuard = _TypeGuard(_root=True)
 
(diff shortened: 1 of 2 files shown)


JelleZijlstra:
This has severe merge conflicts because we just renamed src_py3 to src, could you rebase?

CharString:
:) I was looking into that.

typing_extensions/src/typing_extensions.py
```diff
@@ -2048,40 +2048,61 @@ def __eq__(self, other):
...
+            raise AttributeError(item)
+
+        def __mro_entries__(self, bases):
+            raise TypeError(f"Cannot subclass {self!r}")
 
-elif sys.version_info[:2] >= (3, 9):
-    class _SelfForm(typing._SpecialForm, _root=True):
         def __repr__(self):
-            return 'typing_extensions.' + self._name
+            return 'typing.' + self._name
```
Gobot1234, line 2071:
```suggestion
            return f'typing_extensions.{self._name}'
```

typing_extensions/src/typing_extensions.py
```diff
@@ -2048,40 +2048,61 @@ def __eq__(self, other):
 
     TypeGuard = _TypeGuard(_root=True)
 
+if not hasattr(typing, "Self") and sys.version_info[:2] >= (3, 7):
+    # Vendored from cpython typing._SpecialFrom
```
JelleZijlstra, line 2052: It's a bit hard to follow the if chains here, but can't this just be together with the appropriate branch for 3.7-3.10 so we don't have to duplicate the conditions?

CharString: Should I refactor the uses of typing._SpecialForm to use this vendored  verion too?

JelleZijlstra: Let's change Self only for simplicity.

JelleZijlstra (approved):

PR #1006, opened 2022-01-04
Drop Python 3.6 in `typing_extensions`

AA-Turner:
Python 3.6 is now end of life (https://github.com/python/peps/commit/58b93bdfdb57a1ec3641983c6738341c5072d5b4)

This drops support and does various cleanup things (see commits for more detail).

Hopefully this would make e.g. #963 easier (xref https://github.com/python/typing/pull/963#issuecomment-983242050)

A

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -14,7 +14,7 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
-        python-version: ["3.6", "3.7", "3.8", "3.9", "3.10"]
+        python-version: ["3.7", "3.8", "3.9", "3.10"]
 
     runs-on: ubuntu-latest
 
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -18,7 +18,7 @@ standard library, so that users can experiment with them before they are added t
 standard library. Such features should ideally already be specified in a PEP or draft
 PEP.
 
-`typing_extensions` supports Python versions 3.6 and up.
+`typing_extensions` supports Python versions 3.7 and up.
 
 # Versioning scheme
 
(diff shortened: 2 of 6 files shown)


srittau:
+1 from me, but I know some people are in favor of supporting 3.6 for a while longer. On the other hand, typing-extensions 4 is still available.

hauntsaninja:
Is PEP 646 the only thing that will be included in the short-term future? If so, +1 from me. If not, might be nice to get those in, although as srittau says, we could always backport (but that's a hassle).

JelleZijlstra:
The only other thing that is on the horizon right now is `dataclass_transform` I believe. (There are also pending PEPs for `Literal[str]` and callable syntax, but those shouldn't affect typing-extensions. We already support `Self`.) But `dataclass_transform` will mostly be a stub-only thing, so it doesn't seem too important to support at runtime in typing-extensions.

srittau:
A minor wish: Could we split the changes to the workflow file (except for dropping 3.6) out into a separate PR for a cleaner history?

AA-Turner:
Only dropping 3.6 remains in this PR, see #1017 

A

JelleZijlstra:
I'd like to do one more release (4.1.0) with 3.6 support for the PEP 646 objects. After that we can drop 3.6. It should not require a major version change though, it can just be 4.2.0.

AA-Turner:
Updated the version number to 4.2.0 -- I can merge in master now to resolve conflicts, or wait until PEP 646 support &c has been released in v4.1.0. Which would you prefer?

A

JelleZijlstra:
Thanks! I'd suggest to defer dealing with the merge conflicts until we're ready, because you'll just get new ones if there are any additional changes.

srittau:
Converted to draft to signal that we are "waiting for stuff". Please unconvert once 4.1.0 has been released.

srittau:
Superceded  by #1104. Thanks!

PR #1007, opened 2022-01-04
Fix link

srittau:
Cc @sobolevn


sobolevn:
Oh, my bad! I added a `:ref:` for it. So, it should be: 

```
:ref:`recommended <private-definitions>`
```

PR #1014, opened 2022-01-07
Add a minimal setup.py file to typing_extensions

hoodmane:
For some purposes it's useful to have a `setup.py` file. I think this is the correct way to have `setup.py` defer to `pyproject.toml`.

### DIFF
--- a/typing_extensions/setup.py
+++ b/typing_extensions/setup.py
@@ -1,0 +1,4 @@
+import setuptools
+
+if __name__ == "__main__":
+    setuptools.setup()


the-knights-who-say-ni:

## CLA Missing

Our records indicate the following people have not signed the CLA:

@hoodmane

For legal reasons we need all the people listed to sign the CLA before we can look at your contribution. Please follow [the steps outlined in the CPython devguide](https://devguide.python.org/pullrequest/#licensing) to rectify this issue.

srittau:
What situations are those? Please also note that typing_extensions is not built using setuptools, but is using flit.

hoodmane:
Well I wanted to use it for building with Pyodide which is currently hard coded to try to use `setup.py`. In the case of typing-extensions, we don't need to patch it and it is pure Python, so we should just distribute the `typing_extensions` wheel from PyPI. More generally, we probably will have to switch to handling packages built with these different build systems. For now we can just inject this file into `typing_extensions` as the laziest possible fix.

I don't really understand the situation with Python packaging with `setup.py` and `setup.cfg` and `pyproject.toml` and `poetry` and `flit` and `meson`, etc, etc.

I guess I will close this, since it makes more sense for us to fix our build system downstream.

srittau:
> I don't really understand the situation with Python packaging with `setup.py` and `setup.cfg` and `pyproject.toml` and `poetry` and `flit` and `meson`, etc, etc.

No one does.

PR #1016, opened 2022-01-08
Add is_typeddict from Python 3.10

chrismoradi:
I'm a first time contributor. Still need to sign the PSF Contributor Agreement, but wanted to check that this is a contribution that you're open to accepting first. All changes are copied from Python 3.10.

### DIFF
--- a/typing_extensions/pyproject.toml
+++ b/typing_extensions/pyproject.toml
@@ -6,7 +6,7 @@ build-backend = "flit_core.buildapi"
 # Project metadata
 [project]
 name = "typing_extensions"
-version = "4.0.1"
+version = "4.1.0"
 description = "Backported and Experimental Type Hints for Python 3.6+"
 readme.text = """\
 Typing Extensions -- Backported and Experimental Type Hints for Python
--- a/typing_extensions/pyproject.toml
+++ b/typing_extensions/pyproject.toml
@@ -6,7 +6,7 @@ build-backend = "flit_core.buildapi"
 # Project metadata
 [project]
 name = "typing_extensions"
-version = "4.1.0"
+version = "4.0.1"
 description = "Backported and Experimental Type Hints for Python 3.6+"
 readme.text = """\
 Typing Extensions -- Backported and Experimental Type Hints for Python
--- a/typing_extensions/src/test_typing_extensions.py
+++ b/typing_extensions/src/test_typing_extensions.py
@@ -19,7 +19,7 @@ import typing_extensions
 from typing_extensions import NoReturn, ClassVar, Final, IntVar, Literal, Type, NewType, TypedDict, Self
 from typing_extensions import TypeAlias, ParamSpec, Concatenate, ParamSpecArgs, ParamSpecKwargs, TypeGuard
 from typing_extensions import Awaitable, AsyncIterator, AsyncContextManager, Required, NotRequired
-from typing_extensions import Protocol, runtime, runtime_checkable, Annotated, overload
+from typing_extensions import Protocol, runtime, runtime_checkable, Annotated, overload, is_typeddict
 try:
     from typing_extensions import get_type_hints
 except ImportError:
@@ -1680,6 +1680,13 @@ class TypedDictTests(BaseTestCase):
             'tail': bool,
             'voice': str,
         }
+
+    def test_is_typeddict(self):
+        assert is_typeddict(Point2D) is True
+        assert is_typeddict(Union[str, int]) is False
+        # classes, not instances
+        assert is_typeddict(Point2D()) is False
+
 
 
 class AnnotatedTests(BaseTestCase):
--- a/typing_extensions/src/test_typing_extensions.py
+++ b/typing_extensions/src/test_typing_extensions.py
@@ -1684,6 +1684,7 @@ class TypedDictTests(BaseTestCase):
 
     def test_is_typeddict(self):
         assert is_typeddict(Point2D) is True
+        assert is_typeddict(Point2Dor3D) is True
         assert is_typeddict(Union[str, int]) is False
         # classes, not instances
         assert is_typeddict(Point2D()) is False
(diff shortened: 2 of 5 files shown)


the-knights-who-say-ni:

## Recognized GitHub username

@chrismoradi

srittau:
I believe that having `is_typeddict()` in `typing_extensions` could be quite useful. But I think it should work with both `typing.TypedDict` and `typing_extensions.TypedDict`, in cases where they differ. (I think that's Python 3.8 and 3.9.0 and 3.9.1 if my brief glance at the code was correct.)

chrismoradi:
Thanks for the feedback! I'll try to resolve the edge cases around 3.8, 3.9.0 and 3.9.1.

hmc-cs-mdrissi:
In case it helps,

`isinstance(tp, type(TypedDict))`

is the implementation I use in 3.8 to detect typed dicts. Maybe that covers all versions.

chrismoradi:
Great suggestion, @hmc-cs-mdrissi! I'll give that a try.

chrismoradi:
Couldn't get `isinstance(tp, type(TypedDict))` to work across the versions.

@srittau, Tested this new version against 3.6.12, 3.7.11, 3.8.12, 3.9.0, 3.9.1, 3.9.7, and 3.10.1. Happy to test others if needed or fix style issues.

typing_extensions/src/typing_extensions.py
```diff
@@ -1123,6 +1125,26 @@ class Point2D(TypedDict):
...
+                year: int
+            is_typeddict(Film)  # => True
+            is_typeddict(Union[list, str])  # => False
+        """
+        typeddict_types = [_TypedDictMeta]
+        try:
+            typeddict_types.append(typing._TypedDictMeta)
+        except AttributeError:
+            pass
+        return isinstance(tp, tuple(typeddict_types))
```
Gobot1234, line 1145:
```suggestion
        TYPEDDICT_TYPES = (typing._TypedDictMeta, _TypedDictMeta) if hasattr(typing, "_TypedDictMeta") else (_TypedDictMeta,)


        def is_typeddict(tp):
         """Check if an annotation is a TypedDict class
         For example::
             class Film(TypedDict):
                 title: str
                 year: int
             is_typeddict(Film)  # => True
             is_typeddict(Union[list, str])  # => False
         """
         return isinstance(tp, TYPEDDICT_TYPES)
```
I don't see the point of not moving this into a constant.

chrismoradi: Totally happy to make this change. Waiting on @srittau to see if there are any other concerns.

typing_extensions/pyproject.toml
```diff
@@ -6,7 +6,7 @@ build-backend = "flit_core.buildapi"
 # Project metadata
 [project]
 name = "typing_extensions"
-version = "4.0.1"
+version = "4.1.0"
```
JelleZijlstra, line 9: Let's leave this out, we'll prepare a release separately. However, it would be good to add to the changelog.

typing_extensions/src/test_typing_extensions.py
```diff
@@ -1681,6 +1682,17 @@ def test_keys_inheritance(self):
...
 
+    def test_is_typeddict(self):
+        assert is_typeddict(Point2D) is True
+        assert is_typeddict(Union[str, int]) is False
+        # classes, not instances
+        assert is_typeddict(Point2D()) is False
+
+    @skipUnless(TYPING_3_8_0, "Python 3.8+ required")
+    def test_is_typeddict_against_typeddict_from_typing(self):
+        Point = typing.TypedDict('Point', {'x': int, 'y': int})
```
JelleZijlstra, line 1693: Please also test (for both the typing and typing_extensions versions):
- a class-based TypedDict
- a class inheriting from another TypedDict

typing_extensions/src/typing_extensions.py
JelleZijlstra: I agree with this suggestion, but please prefix TYPEDDICT_TYPES with an underscore to indicate that it's private (and don't use one-space indents :) ).

JelleZijlstra (changes requested):
Thanks! Please also add this to the README in the appropriate section.

typing_extensions/pyproject.toml
chrismoradi: Not sure how to add this to the changelog. Do I just add a line above: # Release 4.0.1 (November 30, 2021)
Should I make a section for "Release 4.x.x" instead?

JelleZijlstra: Something like https://github.com/python/typing/pull/963/files#diff-ad6d05e61a5c54be368726cc13431904e0e23882704fe4def8605c311c92bcbc

chrismoradi:
@JelleZijlstra, I think I covered all the changes you requested but happy to revisit if any are incomplete.

JelleZijlstra (approved):

typing_extensions/src/typing_extensions.py
```diff
@@ -1123,6 +1125,23 @@ class Point2D(TypedDict):
...
+    _TYPEDDICT_TYPES = (typing._TypedDictMeta, _TypedDictMeta) if hasattr(typing, "_TypedDictMeta") else (_TypedDictMeta,)
+
+    def is_typeddict(tp):
+        """Check if an annotation is a TypedDict class
+        For example::
+            class Film(TypedDict):
+                title: str
+                year: int
+            is_typeddict(Film)  # => True
+            is_typeddict(Union[list, str])  # => False
```
JelleZijlstra, line 1140:
```suggestion
        """Check if an annotation is a TypedDict class

        For example::
            class Film(TypedDict):
                title: str
                year: int

            is_typeddict(Film)  # => True
            is_typeddict(Union[list, str])  # => False
```
Restore spacing like in the CPython docstring

typing_extensions/src/typing_extensions.py
```diff
@@ -1123,6 +1125,25 @@ class Point2D(TypedDict):
         """
 
 
+if hasattr(typing, "is_typeddict"):
+    is_typeddict = typing.is_typeddict
+else:
+    _TYPEDDICT_TYPES = (typing._TypedDictMeta, _TypedDictMeta) if hasattr(typing, "_TypedDictMeta") else (_TypedDictMeta,)
```
JelleZijlstra, line 1131:
```suggestion
    if hasattr(typing, "_TypedDictMeta"):
        _TYPEDDICT_TYPES = (typing._TypedDictMeta, _TypedDictMeta)
    else:
        _TYPEDDICT_TYPES = (_TypedDictMeta,)
```
The line is too long

PR #1019, opened 2022-01-10
Add a note about modules that should not be included into stubs

sobolevn:
Refs https://github.com/python/typeshed/issues/6798

### DIFF
--- a/docs/source/stubs.rst
+++ b/docs/source/stubs.rst
@@ -533,7 +533,7 @@ It is recommended to exclude:
 
 1. Implementation details, with `multiprocessing/popen_spawn_win32.py <https://github.com/python/cpython/blob/main/Lib/multiprocessing/popen_spawn_win32.py>`_ as a notable example
 2. Modules that are not supposed to be imported, such as ``__main__.py``
-3. Protected modules that start with a single ``_`` char. However, when needed protected modules can still be added
+3. Protected modules that start with a single ``_`` char. However, when needed protected modules can still be added (see "Undocumented Objects" section below)
 
 Public Interface
 ----------------
--- a/docs/source/stubs.rst
+++ b/docs/source/stubs.rst
@@ -80,6 +80,17 @@ require separate installation. Stubs for third-party libraries are
 available on the `Python Package Index <https://pypi.org>`_. A stub package for
 a library called ``widget`` will be called ``types-widget``.
 
+Modules included in stubs
+=========================
+
+Not all modules should be included into stubs.
+
+It is recommended to exclude:
+
+1. Protected modules that start with a single ``_`` char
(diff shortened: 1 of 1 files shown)


srittau:
This section would best be places in the "Type Stub Content" section (formerly known as "what to include").

srittau (approved):

docs/source/stubs.rst
```diff
@@ -524,6 +524,17 @@ Type Stub Content
...
+Modules excluded fom stubs
+--------------------------
+
+Not all modules should be included into stubs.
+
+It is recommended to exclude:
+
+1. Implementation details, with `multiprocessing/popen_spawn_win32.py <https://github.com/python/cpython/blob/main/Lib/multiprocessing/popen_spawn_win32.py>`_ as a notable example
+2. Modules that are not supposed to be imported, such as ``__main__.py``
+3. Protected modules that start with a single ``_`` char. However, when needed protected modules can still be added
```
JelleZijlstra, line 536: What does "when needed" mean? We should probably refer to the "Undocumented Objects" section below.

srittau: Ping @sobolevn

sobolevn: Thanks! Updated!

docs/source/stubs.rst
```diff
@@ -533,7 +533,7 @@ It is recommended to exclude:
 
 1. Implementation details, with `multiprocessing/popen_spawn_win32.py <https://github.com/python/cpython/blob/main/Lib/multiprocessing/popen_spawn_win32.py>`_ as a notable example
 2. Modules that are not supposed to be imported, such as ``__main__.py``
-3. Protected modules that start with a single ``_`` char. However, when needed protected modules can still be added
+3. Protected modules that start with a single ``_`` char. However, when needed protected modules can still be added (see "Undocumented Objects" section below)
```
srittau, line 536: I hate RestructuredText, but the following should create a link to the section:

```suggestion
3. Protected modules that start with a single ``_`` char. However, when needed protected modules can still be added (see :ref:`undocumented-objects` section below)
```

This needs an anchor right above the section heading:

```
.. _undocumented-objects:
```

Did I mention that I'm not a fan of RestructedText?

sobolevn: Thanks!

> Did I mention that I'm not a fan of RestructedText?

Nope, but I learned to like it! It is way better for complex docs than any `md` extension out there (I've even used `mdx` 😒). For example, we use custom tags in most of my projects. Like:

- In: https://github.com/dry-python/returns/blame/1630e060a629082b2f9de62177d198a5b59e1929/docs/pages/container.rst#L42
- Out: https://returns.readthedocs.io/en/latest/pages/container.html#basics

LaTeX is also not very friendly.

But, I would still love to see some competition here 🙂

srittau: I agree that it's better suited for complex documents, which is why I rejected using markdown for the documentation. Especially the plugin and inter-document linking structure is better. Doesn't mean I like it. :)

srittau (approved):

PR #1026, opened 2022-01-15
@final: backport bpo-46342

JelleZijlstra:


typing_extensions/src/test_typing_extensions.py
```diff
@@ -2205,6 +2206,81 @@ class Alias:
             def return_tuple(self) -> TupleSelf:
                 return (self, self)
 
+
+class FinalDecoratorTests(BaseTestCase):
```
JelleZijlstra, line 2210: These tests were copied verbatim from typing

JelleZijlstra:
Seems like the tests are installing the wrong version of typing-extensions again

hauntsaninja (approved):
Looks good!

In general, feels safer to always use sys.version_info checks instead of hasattr if we don't always use the `typing` version of things. E.g., when looking at #1016 I had to lookup when `is_typeddict` got added to `typing` (since if it was added in 3.9, we'd do the wrong thing on 3.9.0 and 3.9.1)

JelleZijlstra:
Now it's just 3.11 that's failing, presumably because the version on GH Actions doesn't have my patch yet. I'll mark 3.11-dev as allowed to fail for now.

JelleZijlstra:
OK, CI is green now. @hauntsaninja could you take another look at the changes since your review?

.github/workflows/ci.yml
```diff
@@ -27,12 +27,20 @@ jobs:
           python-version: ${{ matrix.python-version }}
 
       - name: Test typing_extensions
+        if: matrix.python-version != "3.11-dev"
         run: |
```
hauntsaninja, line 31: I think the Github Actions incantation might be something like:
```
continue-on-error: ${{ matrix.python-version == '3.11-dev' }}
```

The advantage being that you get to see red crosses.

JelleZijlstra: Ah you're right, I was reading something that suggested `continue-on-error` wouldn't work properly

PR #1028, opened 2022-01-16
Backport tests from `test_typing.py`

sobolevn:
As promised in https://bugs.python.org/issue46386

Backported:
- https://github.com/python/cpython/pull/30619

Not backported:
- https://github.com/python/cpython/pull/30613 (because `ParamSpec` is already tested with `pickle` using all protocols)


JelleZijlstra (approved):
Thanks!

PR #1045, opened 2022-01-24
bpo-46396: remove extra checks from `Concatenate`

sobolevn:
Issue: https://bugs.python.org/issue46396

Please, let's way for my original PR to get merged first.
https://github.com/python/cpython/pull/30619


JelleZijlstra:
Thanks, closing since the CPython chance got rejected.

PR #1049, opened 2022-01-25
Annotated: backport bpo-46491

GBeauregard:
This allows `Annotated` to wrap `ClassVar` and `Final`.

https://bugs.python.org/issue46491

### DIFF
--- a/typing_extensions/CHANGELOG
+++ b/typing_extensions/CHANGELOG
@@ -1,5 +1,7 @@
 # Release 4.x.x
 
+- `Annotated` can now wrap `ClassVar` and `Final`. Backport from
+  bpo-46491. Patch by Gregory Beauregard (@GBeauregard)
 - Add missed `Required` and `NotRequired` to `__all__`. Patch by
   Yuri Karabas (@uriyyo).
 - The `@final` decorator now sets the `__final__` attribute on the
--- a/typing_extensions/CHANGELOG
+++ b/typing_extensions/CHANGELOG
@@ -1,7 +1,7 @@
 # Release 4.x.x
 
 - `Annotated` can now wrap `ClassVar` and `Final`. Backport from
-  bpo-46491. Patch by Gregory Beauregard (@GBeauregard)
+  bpo-46491. Patch by Gregory Beauregard (@GBeauregard).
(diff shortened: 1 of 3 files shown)


typing_extensions/src/typing_extensions.py
```diff
@@ -1377,8 +1381,11 @@ def __getitem__(self, params):
                                 "with at least two arguments (a type and an "
                                 "annotation).")
             else:
-                msg = "Annotated[t, ...]: t must be a type."
-                tp = typing._type_check(params[0], msg)
+                if type(params[0]).__name__ == "_ClassVar":
```
JelleZijlstra, line 1384: This should also check that params[0] is a `typing._TypingBase`.

typing_extensions/CHANGELOG
```diff
@@ -1,5 +1,7 @@
 # Release 4.x.x
 
+- `Annotated` can now wrap `ClassVar` and `Final`. Backport from
+  bpo-46491. Patch by Gregory Beauregard (@GBeauregard)
```
JelleZijlstra, line 4:
```suggestion
  bpo-46491. Patch by Gregory Beauregard (@GBeauregard).
```
(nit)

JelleZijlstra (changes requested):
Thanks! Just some small feedback

JelleZijlstra:
We could provide `typing_extensions.Annotated` on 3.9 and 3.10 to provide a version with the fix immediately, but I'm not sure that's worth it in this case. Alternatively, we can just skip the 3.9 and 3.10 tests until the next bugfix release.

typing_extensions/src/typing_extensions.py
GBeauregard: done, thanks!

GBeauregard:
> We could provide `typing_extensions.Annotated` on 3.9 and 3.10 to provide a version with the fix immediately, but I'm not sure that's worth it in this case. Alternatively, we can just skip the 3.9 and 3.10 tests until the next bugfix release.

Is this okay?
`@skipIf(sys.version_info[:2] in ((3, 9), (3, 10)), "Waiting for bugfix release.")`

JelleZijlstra:
Yes, let's do that

JelleZijlstra (approved):

PR #1053, opened 2022-02-02
add LiteralString

JelleZijlstra:
Based on the latest revision of PEP 675 (cc @pradeep90).

The implementation is copied from `Self`, so the only interesting part is the docstring.

### DIFF
--- a/typing_extensions/CHANGELOG
+++ b/typing_extensions/CHANGELOG
@@ -1,5 +1,6 @@
 # Release 4.x.x
 
+- Runtime support for PEP 675 and `typing.LiteralString`.
 - `Annotated` can now wrap `ClassVar` and `Final`. Backport from
   bpo-46491. Patch by Gregory Beauregard (@GBeauregard).
 - Add missed `Required` and `NotRequired` to `__all__`. Patch by
--- a/typing_extensions/CHANGELOG
+++ b/typing_extensions/CHANGELOG
@@ -1,6 +1,6 @@
 # Release 4.x.x
 
-- Runtime support for PEP 675 and `typing.LiteralString`.
+- Runtime support for PEP 675 and `typing_extensions.LiteralString`.
 - Add `Never` and `assert_never`. Backport from bpo-46475.
 - `ParamSpec` args and kwargs are now equal to themselves. Backport from
   bpo-46676. Patch by Gregory Beauregard (@GBeauregard).
--- a/typing_extensions/src/test_typing_extensions.py
+++ b/typing_extensions/src/test_typing_extensions.py
@@ -22,6 +22,8 @@ from typing_extensions import NoReturn, ClassVar, Final, IntVar, Literal, Type, 
 from typing_extensions import TypeAlias, ParamSpec, Concatenate, ParamSpecArgs, ParamSpecKwargs, TypeGuard
 from typing_extensions import Awaitable, AsyncIterator, AsyncContextManager, Required, NotRequired
 from typing_extensions import Protocol, runtime, runtime_checkable, Annotated, overload, final, is_typeddict
+from typing_extensions import LiteralString
+
 try:
     from typing_extensions import get_type_hints
 except ImportError:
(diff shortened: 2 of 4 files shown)


typing_extensions/src/test_typing_extensions.py
```diff
@@ -2229,6 +2231,48 @@ def test_no_isinstance(self):
...
+        with self.assertRaises(TypeError):
+            type(LiteralString)()
+
+    def test_no_isinstance(self):
+        with self.assertRaises(TypeError):
+            isinstance(1, LiteralString)
+        with self.assertRaises(TypeError):
+            issubclass(int, LiteralString)
+
+    def test_alias(self):
```
pradeep90, line 2269: Can we also test for `TypeVar`?

```
T = TypeVar("T", bound=LiteralString)
```

typing_extensions/src/test_typing_extensions.py
```diff
@@ -2229,6 +2231,48 @@ def test_no_isinstance(self):
...
+        StringTuple = Tuple[LiteralString, LiteralString]
+        class Alias:
+            def return_tuple(self) -> StringTuple:
+                return (self, self)
```
pradeep90, line 2273: This code doesn't make much sense for `LiteralString` since it won't type-check. Maybe the function could return `("foo", "bar" + "baz")`

pradeep90 (approved):
LGTM, but should we wait in case someone has last-minute, heavy objections to the `LiteralString` name?

JelleZijlstra:
Good idea, I'll hold off on merging for a while. I'm hoping to release typing-extensions 4.1 once we have PEP 646 support, dataclass_transform, all the functions I'm adding to typing, and this one all implemented.

typing_extensions/src/test_typing_extensions.py
JelleZijlstra: Done

typing_extensions/src/test_typing_extensions.py
JelleZijlstra: Good point, changed it

srittau:
I've converted this to a draft, just so I don't merge this by accident. Please unconvert (or merge) when ready.

srittau (approved):

typing_extensions/src/typing_extensions.py
```diff
@@ -2096,6 +2097,100 @@ def __eq__(self, other):
 
     TypeGuard = _TypeGuard(_root=True)
 
+
+if hasattr(typing, "LiteralString"):
+    LiteralString = typing.LiteralString
+elif sys.version_info[:2] >= (3, 7):
+    # Vendored from cpython typing._SpecialFrom
+    class _SpecialForm(typing._Final, _root=True):
```
sobolevn, line 2105: Note that it also copies `_SpecialForm` class.

typing_extensions/src/test_typing_extensions.py
```diff
@@ -2230,6 +2230,52 @@ def test_no_isinstance(self):
...
+            mod_name = 'typing_extensions'
+        self.assertEqual(repr(LiteralString), '{}.LiteralString'.format(mod_name))
+
+    def test_cannot_subscript(self):
+        with self.assertRaises(TypeError):
+            LiteralString[int]
+
+    def test_cannot_subclass(self):
+        with self.assertRaises(TypeError):
+            class C(type(LiteralString)):
```
sobolevn, line 2253: I would also add `class C(LiteralString): ...`

typing_extensions/src/test_typing_extensions.py
```diff
@@ -2230,6 +2230,52 @@ def test_no_isinstance(self):
             issubclass(int, TypeGuard)
 
 
+class LiteralStringTests(BaseTestCase):
+    def test_basics(self):
+        class Foo:
+            def bar(self) -> LiteralString: ...
```
sobolevn, line 2236: Let's also test `'LiteralString'` as well here.

typing_extensions/src/test_typing_extensions.py
```diff
@@ -2230,6 +2230,52 @@ def test_no_isinstance(self):
...
+
+        self.assertEqual(gth(Foo.bar), {'return': LiteralString})
```
sobolevn, line 2238:
```suggestion
        self.assertEqual(gth(Foo.bar), {'return': LiteralString})
        self.assertIsNone(get_origin(LiteralString), None))
```

typing_extensions/src/test_typing_extensions.py
```diff
@@ -2230,6 +2230,52 @@ def test_no_isinstance(self):
...
+
+    def test_alias(self):
+        StringTuple = Tuple[LiteralString, LiteralString]
+        class Alias:
+            def return_tuple(self) -> StringTuple:
+                return ("foo", "pep" + "675")
+
+    def test_typevar(self):
+        StrT = TypeVar("StrT", bound=LiteralString)
+        self.assertIs(StrT.__bound__, LiteralString)
```
sobolevn, line 2276: Do we test `pickle` in `typing_extensions`? 🤔

sobolevn (commented):
Thank you!

typing_extensions/src/typing_extensions.py
JelleZijlstra: Ah right, I knew I had copied it in some other place :)

typing_extensions/src/test_typing_extensions.py
sobolevn: Oups, `@skipUnless(PEP_560, "Python 3.7 required")` is needed!

typing_extensions/src/typing_extensions.py
JelleZijlstra: Oh wait, this is going to merge conflict in a bad way with the Never change. Maybe I should land that one first, then use the factored out _SpecialForm from that PR.

typing_extensions/src/test_typing_extensions.py
JelleZijlstra: And your suggestion had a syntax error :D

typing_extensions/src/test_typing_extensions.py
JelleZijlstra: Why not

typing_extensions/CHANGELOG
```diff
@@ -1,5 +1,6 @@
 # Release 4.x.x
 
+- Runtime support for PEP 675 and `typing.LiteralString`.
```
JelleZijlstra, line 3:
```suggestion
- Runtime support for PEP 675 and `typing_extensions.LiteralString`.
```

PR #1057, opened 2022-02-06
PEP 655 Add interaction with __required_keys__, __optional_keys__ and get_type_hints()

d-k-bo:
The current implementation of PEP 655 doesn't implement all specified features of the current draft:

- `Required` and `NotRequired` don't interact with `__required_keys__` and `__optional_keys__`
- `get_type_hints()` doesn't strip out `Required` nor `NotRequired` (by default)

Implementing the first aspect requires modifying `_TypedDictMeta`, which breaks backward compatibility when using typing's `is_typeddict()` on typing_extensions' `TypedDict`:

```python
import typing

import typing_extensions


class Movie(typing_extensions.TypedDict):
    title: str
    year: typing_extensions.NotRequired[int]


typing_extensions.is_typeddict(Movie)  # True
typing.is_typeddict(Movie)  # False
```

For example this PR breaks the current version of [typeguard](https://github.com/agronholm/typeguard/blob/de6ab051309ba74a0a27840f8172697c8778ae4f/src/typeguard/__init__.py#L64) when using `typing_extension.TypedDict` on a recent version of python.

This is my first contribution, so please point out if I missed any required steps before opening a PR.

### DIFF
--- a/typing_extensions/CHANGELOG
+++ b/typing_extensions/CHANGELOG
@@ -1,5 +1,7 @@
 # Release 4.x.x
 
+- Add interaction of `Required` and `NotRequired` with `__required_keys__`,
+  `__optional_keys__` and `get_type_hints()`. Patch by David Cabot (@d-k-bo).
 - Add `reveal_type`. Backport from bpo-46414.
 - Runtime support for PEP 681 and `typing_extensions.dataclass_transform`.
 - `Annotated` can now wrap `ClassVar` and `Final`. Backport from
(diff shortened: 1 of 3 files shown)


JelleZijlstra (approved):
Looks good, thank you!

cc @davidfstr in case you also want to take a look

JelleZijlstra (changes requested):
Could you add an entry to the CHANGELOG file? (Sorry, just noticed this.)

JelleZijlstra (approved):

PR #1060, opened 2022-02-09
Add Never and assert_never

JelleZijlstra:
Backport of python/cpython#30842, with additional tests from @sobolevn's python/cpython#31222.

### DIFF
--- a/typing_extensions/src/typing_extensions.py
+++ b/typing_extensions/src/typing_extensions.py
@@ -70,6 +70,7 @@ __all__ = [
 
     # One-off things.
     'Annotated',
+    'assert_never',
     'dataclass_transform',
     'final',
     'IntVar',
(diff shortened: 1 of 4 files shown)


typing_extensions/src/typing_extensions.py
```diff
@@ -2195,6 +2197,112 @@ def __subclasscheck__(self, cls):
     Self = _Self(_root=True)
 
 
+if hasattr(typing, "Never"):
+    Never = typing.Never
+elif sys.version_info[:2] >= (3, 7):
+    # Vendored from cpython typing._SpecialFrom
+    class _SpecialForm(typing._Final, _root=True):
```
sobolevn, line 2204: This looks like a duplicate of https://github.com/python/typing/blob/3d9d7d75ea4c9d414a343c7fda951ba578be2b90/typing_extensions/src/typing_extensions.py#L2116-L2156

Can we reuse it?

JelleZijlstra: Good point, done.

PR #1063, opened 2022-02-10
update list of PEPs

JelleZijlstra:
- PEP 673 was accepted
- Add 675, 677, and 681


hauntsaninja:
Oh oops, PEP 681 is still draft right?

JelleZijlstra:
Oops, yes. Fixed in #1064

PR #1065, opened 2022-02-10
Add guide to unreachable code

JelleZijlstra:
This adds a guide explaining how to use `assert_never()` and how it differs from `assert False`.

### DIFF
--- a/docs/source/unreachable.rst
+++ b/docs/source/unreachable.rst
@@ -98,8 +98,8 @@ You can also use ``assert_never()`` with a sequence of ``if`` statements:
        else:
            assert_never(op)
 
-Directly marking code as unreachable
-====================================
+Marking Code as Unreachable
+=======================
 
 Sometimes a piece of code is unreachable, but the type system is not
 powerful enough to recognize that. For example, consider a function that
--- a/docs/source/unreachable.rst
+++ b/docs/source/unreachable.rst
@@ -84,7 +84,7 @@ it is provided by the standard library as ``typing.assert_never``
 starting in Python 3.11,
 and is also present in ``typing_extensions`` starting at version 4.1.
 However, it is also possible to define a similar function in your own
-code, for example if you want to customize the error message.
+code, for example if you want to customize the runtime error message.
 
 You can also use ``assert_never()`` with a sequence of ``if`` statements:
 
--- a/docs/source/unreachable.rst
+++ b/docs/source/unreachable.rst
@@ -22,7 +22,8 @@ Type theory has a concept of a
 `bottom type <https://en.wikipedia.org/wiki/Bottom_type>`__,
 a type that has no values. Concretely, this can be used to represent
 the return type of a function that never returns, or the argument type
-of a function that may never be called.
+of a function that may never be called. You can also think of the
+bottom type as a union with no members.
 
 The Python type system has long provided a type called ``NoReturn``.
 While it was originally meant only for functions that never return,
--- a/docs/source/unreachable.rst
+++ b/docs/source/unreachable.rst
@@ -43,7 +43,7 @@ use it yet, you can always use ``typing.NoReturn`` instead.
 The ``Never`` type can be leveraged to perform static exhaustiveness checking,
 where we use the type checker to make sure that we covered all possible
 cases. For example, this can come up when code performs a separate action
-for each member of an enum, or for each type in a ``Union``.
+for each member of an enum, or for each type in a union.
 
 To have the type checker do exhaustiveness checking for us, we call a
 function with a parameter typed as ``Never``. The type checker will allow
--- a/docs/source/unreachable.rst
+++ b/docs/source/unreachable.rst
@@ -139,7 +139,7 @@ While ``assert False`` is the most idiomatic way to express this pattern,
 any statement that ends execution will do. For example, you could raise
 an exception or call a function that returns ``Never``.
 
-Detecting unexpectedly unreachable code
+Detecting Unexpectedly Unreachable Code
 =======================================
 
 Another possible problem is code that is supposed to execute, but that
--- a/docs/source/unreachable.rst
+++ b/docs/source/unreachable.rst
@@ -1,0 +1,147 @@
+.. _unreachable:
+
+********************************************
+Unreachable Code and Exhaustiveness Checking
+********************************************
+
+Sometimes it is necessary to write code that should never execute, and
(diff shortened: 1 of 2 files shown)


docs/source/unreachable.rst
```diff
@@ -0,0 +1,147 @@
...
+
+- ``Never``, the primitive type used for unreachable code
+- ``assert_never()``, a helper for exhaustiveness checking
+- Directly marking code as unreachable
+- Detecting unexpectedly unreachable code
+
+``Never`` and ``NoReturn``
+==========================
+
+Type theory has a concept of a
```
erictraut, line 21: This is a really good introduction to this concept — one of the best descriptions I've seen. One other way that I've found useful in describing this concept is "an empty union". I don't know if adding that phrase here would help some people better understand the concept or add confusion. I mention it in case you think it might be useful.

docs/source/unreachable.rst
```diff
@@ -0,0 +1,147 @@
...
+typing-extensions 4.1 add a new primitive, ``Never``. To type checkers,
+it has the same meaning as ``NoReturn``.
+
+In this guide, we'll use ``Never`` for the bottom type, but if you cannot
+use it yet, you can always use ``typing.NoReturn`` instead.
+
+``assert_never()`` and exhaustiveness checking
+==============================================
+
+The ``Never`` type can be leveraged to perform exhaustiveness checking,
```
erictraut, line 42: It might be useful to add "static" to the phrase "exhaustiveness checking". It's possible to add runtime assertions as a form of exhaustiveness checking, but this is a way to enforce exhaustiveness statically through static type checking.

erictraut (commented):
I added a couple of comments for consideration, but it looks great to me.

docs/source/unreachable.rst
JelleZijlstra:
```suggestion
The ``Never`` type can be leveraged to perform static exhaustiveness checking,
```
Good idea

docs/source/unreachable.rst
JelleZijlstra: Thank you! I'll see if I can work that in too.

JelleZijlstra: Added

docs/source/unreachable.rst
```diff
@@ -0,0 +1,148 @@
...
+
+Note how this is different from ``assert_never()``:
+
+- If we used ``assert_never()`` in the ``lowest_unused()`` function,
+  the type checker would produce an error, because the type checker
+  cannot prove that the line is unreachable.
+- If we used ``assert False`` instead of ``assert_never()`` in the
+  ``calculate()`` example above, we would not get the benefits of
+  exhaustiveness checking, because the type checker will not warn us
+  if the code is actually reachable.
```
hauntsaninja, line 136:
```suggestion
  exhaustiveness checking. If the code is actually reachable,
  the type checker will not warn us and we could hit the assertion
  at runtime.
```
trying to emphasise that `assert False` is the unsafe option

docs/source/unreachable.rst
```diff
@@ -0,0 +1,148 @@
...
+
+Because the ``assert_never()`` helper function is frequently useful,
+it is provided by the standard library as ``typing.assert_never``
+starting in Python 3.11,
+and is also present in ``typing_extensions`` starting at version 4.1.
+However, it is also possible to define a similar function in your own
+code, for example if you want to customize the error message.
+
+If you are not on Python 3.10 yet and cannot use the ``match`` statement,
+you can also use ``assert_never()`` with a sequence of ``if`` statements:
```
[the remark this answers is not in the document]
JelleZijlstra: Thanks for both suggestions!

docs/source/unreachable.rst
```diff
@@ -0,0 +1,148 @@
...
+enum (say, ``MULTIPLY``) but don't update the ``match`` statement,
+the type checker will give an error saying that you are not handling
+the ``MULTIPLY`` case.
+
+Because the ``assert_never()`` helper function is frequently useful,
+it is provided by the standard library as ``typing.assert_never``
+starting in Python 3.11,
+and is also present in ``typing_extensions`` starting at version 4.1.
+However, it is also possible to define a similar function in your own
+code, for example if you want to customize the error message.
```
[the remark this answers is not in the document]
JelleZijlstra: True, there's currently no way to customize the error that type checkers produce. Thanks!

docs/source/unreachable.rst
```diff
@@ -0,0 +1,148 @@
...
+In this guide, we'll use ``Never`` for the bottom type, but if you cannot
+use it yet, you can always use ``typing.NoReturn`` instead.
+
+``assert_never()`` and exhaustiveness checking
+==============================================
+
+The ``Never`` type can be leveraged to perform static exhaustiveness checking,
+where we use the type checker to make sure that we covered all possible
+cases. For example, this can come up when code performs a separate action
+for each member of an enum, or for each type in a ``Union``.
```
srittau, line 46: I'd use:

```suggestion
for each member of an enum, or for each type in a union.
```

Since `Union` is semi-deprecated and this also applies to PEP 604 unions.

docs/source/unreachable.rst
```diff
@@ -0,0 +1,148 @@
...
+   def calculate(left: int, op: Op, right: int) -> int:
+       if op is Op.ADD:
+           return left + right
+       elif op is Op.SUBTRACT:
+           return left - right
+       else:
+           assert_never(op)
+
+Directly marking code as unreachable
+====================================
```
srittau, line 102:
```suggestion
Marking Code as Unreachable
=======================
```

docs/source/unreachable.rst
```diff
@@ -0,0 +1,148 @@
...
+bottom type as a union with no members.
+
+The Python type system has long provided a type called ``NoReturn``.
+While it was originally meant only for functions that never return,
+this concept is naturally extended to the bottom type in general, and all
+type checkers treat ``NoReturn`` as a general bottom type.
+
+To make the meaning of this type more explicit, Python 3.11 and
+typing-extensions 4.1 add a new primitive, ``Never``. To type checkers,
+it has the same meaning as ``NoReturn``.
```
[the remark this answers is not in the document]
JelleZijlstra: Yes, I felt like it was risky to mention Never too much given that there's no released version of the language yet that provides it. I don't think the text would read too badly even if we don't update it in two years.

Issue #1067, opened 2022-02-10
Update the status of built-in generics usage in type stubs


q0w:
https://github.com/python/typeshed/issues/4820

PR #1068, opened 2022-02-11
Built-in generic now work without limitations

srittau:
Closes: #1067

### DIFF
--- a/docs/source/stubs.rst
+++ b/docs/source/stubs.rst
@@ -174,15 +174,8 @@ Type checkers support cyclic imports in stub files.
 Built-in Generics
 -----------------
 
-PEP 585 [#pep585]_ built-in generics are generally supported, with
-the following exceptions [#ts-4820]_:
-
-* Built-in generics don't work in type aliases.
-* Built-in generics don't work in base classes.
-* ``type`` is not supported.
-* Variable length tuples (``tuple[X, ...]``) are not supported.
-
-In these cases, the appropriate types from ``typing`` must be used.
+PEP 585 [#pep585]_ built-in generics are supported and should be used instead
+of the corresponding types from ``typing``.
 
 Using imports from ``collections.abc`` instead of ``typing`` is
 generally possible and recommended.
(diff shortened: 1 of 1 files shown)


docs/source/stubs.rst
```diff
@@ -174,15 +174,8 @@ Type checkers support cyclic imports in stub files.
...
-* Built-in generics don't work in base classes.
-* ``type`` is not supported.
-* Variable length tuples (``tuple[X, ...]``) are not supported.
-
-In these cases, the appropriate types from ``typing`` must be used.
+PEP 585 [#pep585]_ built-in generics are supported and should be used instead
+of the corresponding types from ``typing``.
 
 Using imports from ``collections.abc`` instead of ``typing`` is
 generally possible and recommended.
```
JelleZijlstra, line 181: Can we drop the "generally" here too? I don't think there are any exceptions left.

Also, it would be good to have some examples in this section. We can list bad examples like "from typing import Sequence" and "List[int]" and good ones like "from collections.abc import Sequence" and "list[int]".

srittau: I listed a few examples, although I only used positive examples as in other examples in this section.

docs/source/stubs.rst
```diff
@@ -174,18 +174,20 @@ Type checkers support cyclic imports in stub files.
...
 
-* Built-in generics don't work in type aliases.
-* Built-in generics don't work in base classes.
-* ``type`` is not supported.
-* Variable length tuples (``tuple[X, ...]``) are not supported.
+    from collections import defaultdict
 
-In these cases, the appropriate types from ``typing`` must be used.
+    def foo(t: type[MyClass]) -> list[int]: ...
+    x: defaultdict[int] = defaultdict(lambda: 42)
```
JelleZijlstra, line 183:
```suggestion
    x: defaultdict[int]
```
We're talking about stubs

JelleZijlstra (approved):
One nit

PR #1071, opened 2022-02-12
Document how to test type annotations

sirosen:
The goal of this PR is to resolve https://github.com/python/mypy/issues/11506 and document how to test type annotations.
In terms of presentation, the goal is to present guidance and offer up several options, many of which were proposed by contributors to that issue. I have attempted to present all of the information without bias, and everyone in that prior conversation deserves credit for their thoughts and input.

This is a broader refactor than just testing, though testing is the primary content.
It also consolidates information about checking type coverage/completeness.

Several of the goals from that thread were not achieved here, including documentation covering stubgen and monkeytype, stubtest, and potentially more. However, the document is written such that it should be possible to add a section on "Generating Annotations" as was planned earlier.

### DIFF
--- a/docs/source/reference.rst
+++ b/docs/source/reference.rst
@@ -7,7 +7,7 @@ Type System Reference
    :caption: Contents:
 
    stubs
-   tools
+   quality
    typing Module Documentation <https://docs.python.org/3/library/typing.html>
 
 .. The following pages are desired in a new TOC which will cover multiple
--- a/docs/source/reference.rst
+++ b/docs/source/reference.rst
@@ -7,6 +7,7 @@ Type System Reference
    :caption: Contents:
 
    stubs
+   tools
    typing Module Documentation <https://docs.python.org/3/library/typing.html>
 
 .. The following pages are desired in a new TOC which will cover multiple
--- a/docs/source/tools.rst
+++ b/docs/source/tools.rst
@@ -1,0 +1,207 @@
+.. _tools:
+
+***************************************
+Tools for Working With Type Annotations
+***************************************
+
+Testing Annotation Accuracy
+===========================
+
+When creating a package with type annotations, authors may want to validate
+that the annotations they publish meet their expectations.
+This is especially important for library authors, for whom the published
+annotations are part of the public interface to their package.
+
+There are several approaches to this problem, and this document will show
+a few of them.
+
+.. note::
+
+    For simplicity, we will assume that type-checking is done with ``mypy``.
+    Many of these strategies can be applied to other type-checkers as well.
+
+Testing Using ``mypy --warn-unused-ignores``
+--------------------------------------------
+
+Clever use of the ``--warn-unused-ignores`` can be used to check that certain
+expressions are or are not well-typed.
+
+The idea is to write normal python files which contain valid expressions along
+with invalid expressions annotated with ``type: ignore`` comments. When
+``mypy --warn-unused-ignores`` is run on these files, it should pass.
+A directory of test files, ``typing_tests/``, can be maintained.
+
+This strategy does not offer strong guarantees about the types under test, but
+it requires no additional tooling.
+
+If the following file is under test
+
+.. code-block:: python
+
+    # foo.py
+    def bar(x: int) -> str:
+        return str(x)
+
+Then the following file tests ``foo.py``:
+
+.. code-block:: python
+
+    bar(42)
+    bar("42")  # type: ignore [arg-type]
+    bar(y=42)  # type: ignore [call-arg]
+    r1: str = bar(42)
+    r2: int = bar(42)  # type: ignore [assignment]
+
+Checking ``reveal_type`` output from ``mypy.api.run``
+-----------------------------------------------------
+
+``mypy`` provides a subpackage named ``api`` for invoking ``mypy`` from a
+python process. In combination with ``reveal_type``, this can be used to write
+a function which gets the ``reveal_type`` output from an expression. Once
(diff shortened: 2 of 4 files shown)


JelleZijlstra:
Thanks! What you wrote looks good, but it's highly mypy-specific, so it may be a better fit for the mypy docs.

sirosen:
I agree. The original issue proposed an addition to the mypy docs, but I was asked (by a pyright maintainer, FWIW) to put it here instead.

I don't really care where the doc goes. I just know that when I realized my published types were incorrect, there was no guidance I could find on testing my fixes.

JelleZijlstra:
Eric did say "I guess it depends on how mypy-specific it is." :)

You might also be interested in https://mail.python.org/archives/list/typing-sig@python.org/thread/MITFQ6Z45RRMXY3HNM66IC3XXS3TA3JN/

sirosen:
> Eric did say "I guess it depends on how mypy-specific it is." :)

You're right. I may have taken a more forceful meaning from his comment than I should have. 😅 

It's mostly mypy-centric because that's what I know best. Does your feeling about where this belongs change if I were to add a 30-second intro to `monkeytype`?
I'm not qualified to write much about most of the other tools, but I remember using `monkeytype` a couple of years ago.

---

Aside: I find the `typing.assert_type` thread very interesting, thanks for sharing!
I could easily get into the naming debate (and shouldn't! 😹 ). One thing I'll note is that between runtime_checkable protocols and the annotation-driven libraries like pydantic, it's easy to imagine people believing that symbols from `typing` have runtime significance.

erictraut:
@sirosen, thanks for writing this. Although it is mypy-centric, I think it's still good to post on this site. Other contributors can then extend it to include tips based on functionality in other tools.

srittau:
I agree with @erictraut. I'd rather see this here and extended in the future than in the mypy docs. Are you okay with that, @JelleZijlstra?

JelleZijlstra:
Sure, let's do that.

docs/source/reference.rst
```diff
@@ -7,6 +7,7 @@ Type System Reference
    :caption: Contents:
 
    stubs
+   tools
```
srittau, line 10: I'm not 100% happy with the name "tools" as it doesn't seem to really fit the document. See below for more discussion.

docs/source/tools.rst
```diff
@@ -0,0 +1,207 @@
...
+
+.. note::
+
+    For simplicity, we will assume that type-checking is done with ``mypy``.
+    Many of these strategies can be applied to other type-checkers as well.
+
+Testing Using ``mypy --warn-unused-ignores``
+--------------------------------------------
+
+Clever use of the ``--warn-unused-ignores`` can be used to check that certain
```
srittau, line 26:
```suggestion
Clever use of ``--warn-unused-ignores`` can be used to check that certain
```

docs/source/tools.rst
```diff
@@ -0,0 +1,207 @@
+.. _tools:
+
+***************************************
+Tools for Working With Type Annotations
+***************************************
```
srittau, line 5: This document is less about tools and more about annotation quality. Maybe we could name it something along these lines? For example, "Testing and Ensuring Type Annotation Quality". The document could be named "quality" or maybe "annotation-quality" or something similar.

docs/source/tools.rst
```diff
@@ -0,0 +1,207 @@
...
+pyright documentation on verifying type completeness [#pyright-verifytypes]_.
+
+``mypy`` reports
+----------------
+
+``mypy`` offers several options options for generating reports on its analysis.
+See the mypy documentation on report generation [#mypy-report]_ for details.
+
+References
+==========
```
srittau, line 197: I'm not sure we should have a separate "References" section. Inline links seem more user-friendly to me. I think that even PEPs are moving away from a separate section.

srittau (changes requested):
Thanks, a great document! I especially love the `--warn-unused-ignores` trick. A few notes below.

docs/source/reference.rst
sirosen: 👍 I am 100% _not happy_ with it, but didn't come up with a good title. Let's figure it out!

docs/source/tools.rst
sirosen: I like the "quality" language, since testing is a form of QA.
I'll update to the title you suggested, `Testing and Ensuring Type Annotation Quality`, in `quality.rst`.

I considered dropping "Testing and" from the title, but it's a little vague without that. Let me know if you want any further adjustment than that.

docs/source/tools.rst
sirosen: Ah! And I had put in the references section in deference to the existing style of having PEPs listed at the end.
In that case, I'll switch to inline links. I think they're nicer too.

docs/source/tools.rst
srittau: Sounds fine to me. And nothing is set in stone, we can always change it later if we come up with something better or the focus of the document shifts.

srittau (approved):

sirosen:
I've applied the review suggestions. I think this is good to go now, but I'm happy to do more rounds of review if there are still issues.

> I especially love the `--warn-unused-ignores` trick

I also was very impressed by it, which is why I put it first in the doc!

Credit to @ikonst for [explaining it in the original mypy issue](https://github.com/python/mypy/issues/11506#issuecomment-965987842).

If I'd known about or thought of that approach first, I might never have written my elaborate `mypy.api` usage.

Issue #1073, opened 2022-02-13
4.1.0 error after relase in 2022.02.13 NameError: name '_UnpackAlias' is not defined


yt605155624:
File "/workspace/tools/venv/lib/python3.7/site-packages/typing_extensions.py", line 143, in _collect_type_vars
     not isinstance(t, _UnpackAlias)
 NameError: name '_UnpackAlias' is not defined

JelleZijlstra:
Thanks @yt605155624 for the report! Could you share the code that led to this error? I'd like to understand how our CI missed this case.

carlogrisetti:
@JelleZijlstra I have a TravisCI build failing due to this issue. It only happens on Python 3.7 and the stack starts at `from typing import Protocol` in `importlib_metadata` module
Another traceback is `from pytest import console_main`

JelleZijlstra:
Thanks! I see you're using 3.7.1 there; let me see if the version of 3.7 makes a difference.

carlogrisetti:
That was unintentional, TravisCI chose 3.7.1 when asked to test on 3.7 so I suppose this issue will arise in many other builds all over the GitHubverse :)
You could temporarily yank 4.1.0 waiting for a 4.1.1 release if times are not compact enough, otherwise I suppose that you'll receive a lot of complaints tomorrow with the start of the work week.

Regards!

JelleZijlstra:
typing.OrderedDict was new in 3.7.2, I suspect that has something to do with it. Unfortunately pyenv wasn't able to build 3.7.1 for me, but 3.7.10 is fine for me locally.

Sounds like Travis needs to update their 3.7 installation :)

But I'll release 4.1.1 later today to fix this.

JelleZijlstra:
We confirmed that it fails on 3.7.1. To make sure we don't get any other similar issues, I'm thinking of running CI on the .0 versions of all supported Python versions. That way, we run the earliest and latest bugfix release of each version, which will hopefully be enough to catch this sort of thing in general.

JelleZijlstra:
I just released https://pypi.org/project/typing-extensions/4.1.1/ with the fix. Thanks for reporting this bug!

yt605155624:
Thanks for your for your timely fix and release!

PR #1074, opened 2022-02-13
Fix `name '_UnpackAlias' is not defined`

sobolevn:
What is the best way to test this? 🤔 
Right now I can't find any reasonable example to do that

Closes https://github.com/python/typing/issues/1073

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -14,7 +14,7 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
-        python-version: ["3.6", "3.7.0", "3.7", "3.8", "3.9", "3.10", "3.11-dev"]
+        python-version: ["3.6", "3.7.1", "3.7", "3.8", "3.9", "3.10", "3.11-dev"]
 
     runs-on: ubuntu-latest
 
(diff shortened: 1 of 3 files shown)


JelleZijlstra:
Thanks for the quick fix! I asked the OP on the issue for a repro case. I can take a closer look myself too later today.

This change should also have an entry in the CHANGELOG, but we should first figure out what it actually fixes :)

.github/workflows/ci.yml
```diff
@@ -14,7 +14,7 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
-        python-version: ["3.6", "3.7", "3.8", "3.9", "3.10", "3.11-dev"]
+        python-version: ["3.6", "3.7.0", "3.7", "3.8", "3.9", "3.10", "3.11-dev"]
```
JelleZijlstra, line 17:
```suggestion
        python-version: ["3.6", "3.7.1", "3.7", "3.8", "3.9", "3.10", "3.11-dev"]
```
3.7.0 apparently doesn't exist on GH Actions

JelleZijlstra:
Unfortunately this doesn't fix 3.7.1 either.

JelleZijlstra:
I'm going to go with #1076

PR #1076, opened 2022-02-14
Fix Python 3.7.1 and run more versions in CI

### LINKED ISSUE #1073: 4.1.0 error after relase in 2022.02.13 NameError: name '_UnpackAlias' is not defined

JelleZijlstra:
Fixes #1073

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -14,7 +14,11 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
-        python-version: ["3.6", "3.7", "3.8", "3.9", "3.10", "3.11-dev"]
+        # We try to test on the earliest available bugfix release of each
+        # Python version, because typing sometimes changed between bugfix releases.
+        # For available versions, see:
+        # https://raw.githubusercontent.com/actions/python-versions/main/versions-manifest.json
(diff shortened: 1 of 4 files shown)


JelleZijlstra:
Looks like it's only 3.7.1 :D

JelleZijlstra:
The failing test is because python/cpython#9750 wasn't applied to 3.7.1 yet. I don't know how to fix that one without even more monkey patching, and nobody has complained about this directly yet, so I'm going to just skip that test on 3.7.1.

hauntsaninja (approved):
Thanks both for the fast fixes!

Interesting GA doesn't have some of the earlier versions. I assume that's just because they predate GA and not because they remove them.

hauntsaninja:
Btw, looks like tests are failing on 3.11, not sure if that's expected

JelleZijlstra:
> I assume that's just because they predate GA and not because they remove them.

That was my guess too. Thanks for the merge (and thanks @sobolevn for the initial PR)!

JelleZijlstra:
The 3.11 thing is because GH Actions's 3.11 is outdated, so @GBeauregard's fix for P.args equality isn't in yet. It should be OK with the next alpha.

PR #1078, opened 2022-02-14
Improve release instructions

JelleZijlstra:
This is what I've actually been doing, but better to document it.

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -30,11 +30,17 @@ backwards-incompatible changes.
 
 - Ensure that GitHub Actions reports no errors.
 
-- Update the version number in `pyproject.toml`.
+- Update the version number in `typing_extensions/pyproject.toml` and in
+  `typing_extensions/CHANGELOG`.
+
+- Make sure your environment is up to date
+
+ - `git checkout master`
+ - `git pull`
+ - `python -m pip install --upgrade build twine`
 
 - Build the source and wheel distributions:
 
-  - `python -m pip install --upgrade build`
   - `cd typing_extensions`
   - `rm -rf dist/`
   - `python -m build .`
@@ -42,4 +48,8 @@ backwards-incompatible changes.
 - Install the built distributions locally and test (if you were using `tox`, you already
   tested the source distribution).
 
-- Make sure twine is up to date, then run `twine upload dist/*`.
+- Run `twine upload dist/*`.
+
+- Tag the release. The tag should be just the version number, e.g. `4.1.1`.
+
+- `git push --tags`


JelleZijlstra:
cc @AA-Turner since I think you wrote the current version of these instructions

CONTRIBUTING.md
```diff
@@ -30,16 +30,26 @@ backwards-incompatible changes.
 
 - Ensure that GitHub Actions reports no errors.
 
-- Update the version number in `pyproject.toml`.
+- Update the version number in `typing_extensions/pyproject.toml` and in
```
hauntsaninja, line 33: nit: for releases I personally find it helpful to split up steps as much as possible and not use `and`

CONTRIBUTING.md
```diff
@@ -30,16 +30,26 @@ backwards-incompatible changes.
...
+ - `python -m pip install --upgrade build twine`
 
 - Build the source and wheel distributions:
 
-  - `python -m pip install --upgrade build`
   - `cd typing_extensions`
   - `rm -rf dist/`
   - `python -m build .`
 
 - Install the built distributions locally and test (if you were using `tox`, you already
```
hauntsaninja, line 48: Might be nice to provide the commands for this

hauntsaninja (approved):
Looks good, feel free to merge or wait for AA-Turner :-)

CONTRIBUTING.md
AA-Turner: I think here it is fine as the step is "update the version number", there are just two places to do it.

A

AA-Turner (approved):

JelleZijlstra:
Thanks for the reviews! I'd also prefer automatic releases, since I don't exactly trust myself to do it right. (It's a lot easier with `build` than it was before, though.) On Black and a couple of other projects I maintain we generate releases automatically on tags, which works well. Could you tell me more about the option to release on PR merge? I haven't seen that before.

CONTRIBUTING.md
JelleZijlstra: I tend to just install it and import it... I'd rather leave this out for now in the hope that we'll switch to fully automatic releases.

AA-Turner:
> Could you tell me more about the option to release on PR merge?

The idea is entirely stolen from Hypothesis, although that project may have stolen it from somewhere else. 

The precis is that each PR includes a changelog fragment file (`RELEASE.rst` or similar), which is detected by CI and used to bump the version (major/micro/patch), prepend to the changelog, and build/tag/release.

The versioning information comes from the marker file -- either a line with release level, or release level in the suffix (`RELEASE.minor.md`).

I have an implementation for another project that I'm happy to adapt for here -- the issue is testing, really, as you don't want to be releasing thousands of versions whilst tweaking the infrastructure. Test PyPI might be a good solution for this, though.

A

PR #1103, opened 2022-03-17
Add assert_type

JelleZijlstra:
See python/cpython#30843

### DIFF
--- a/typing_extensions/src/test_typing_extensions.py
+++ b/typing_extensions/src/test_typing_extensions.py
@@ -12,7 +12,7 @@ import types
 from unittest import TestCase, main, skipUnless, skipIf
 from test import ann_module, ann_module2, ann_module3
 import typing
-from typing import TypeVar, Optional, Union, Any
+from typing import TypeVar, Optional, Union, Any, AnyStr
 from typing import T, KT, VT  # Not in __all__.
 from typing import Tuple, List, Dict, Iterable, Iterator, Callable
 from typing import Generic, NamedTuple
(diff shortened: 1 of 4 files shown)


typing_extensions/src/typing_extensions.py
```diff
@@ -1231,6 +1231,27 @@ class Film(TypedDict):
         """
         return isinstance(tp, tuple(_TYPEDDICT_TYPES))
 
+
+if hasattr(typing, "assert_type"):
+    assert_type = typing.assert_type
+
+else:
+    def assert_type(__val, __typ):
+        """Assert (to the type checker) that the value is of the given type.
```
hauntsaninja, line 1240: More a critique of the original PR, but not sure that "Assert (to the type checker) that *val* has an inferred type of *typ*" is the least confusing phrasing.

You're not telling the type checker a fact, you're asking them to confirm it. Maybe "Ask a static type checker to confirm that the value is of the given type".

Or at least, we should do enough to do more to distinguish it from `assert isinstance(val, typ)`, which is what I would describe as "asserting something to the type checker"

JelleZijlstra: Could you bring this up on cpython instead? The docstring here is just a copy of the CPython one. We can sync it back later if we want.

typing_extensions/src/test_typing_extensions.py
```diff
@@ -23,6 +23,7 @@
 from typing_extensions import Awaitable, AsyncIterator, AsyncContextManager, Required, NotRequired
 from typing_extensions import Protocol, runtime, runtime_checkable, Annotated, overload, final, is_typeddict
 from typing_extensions import TypeVarTuple, Unpack, dataclass_transform, reveal_type, Never, assert_never, LiteralString
+from typing_extensions import assert_type
 from typing_extensions import get_type_hints, get_origin, get_args
```
srittau, line 27: Merge?

srittau (approved):
One nit, but LGTM.

typing_extensions/src/test_typing_extensions.py
JelleZijlstra, line 27:
```suggestion
from typing_extensions import assert_type, get_type_hints, get_origin, get_args
```

JelleZijlstra: Done. Thanks for the review!

typing_extensions/src/typing_extensions.py
hauntsaninja: Sure. Should I open a PR directly against the original BPO or would you prefer I file its own BPO

JelleZijlstra: We can keep the same one

PR #1111, opened 2022-03-21
test_typing_extensions: fix lint

hauntsaninja:
This is from a new check in a new version of flake8-bugbear


JelleZijlstra:
I'd rather we turn off the check (or even all of flake8-bugbear); I don't think it will be useful for this project.

JelleZijlstra:
But I don't care enough to change it :). Thanks for the fix!

PR #1116, opened 2022-03-30
TypeAlias can now be used, add guidance

srittau:


docs/source/stubs.rst
```diff
@@ -827,7 +833,7 @@ No::
 
 But the following is still necessary::
```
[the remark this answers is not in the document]
srittau: Well, in a way this doc must track bugs as it is a guide on how to write stubs. If it can't be used in stubs, we can't recommend to use it.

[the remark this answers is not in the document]
AlexWaygood: There are 53 files in typeshed where we still have to import `Union` and/or `Optional` due to mypy bugs. Considering how large a repository typeshed is, that isn't so many — we can probably say that PEP 604 syntax can be safely used in nearly all situations now, but that a few bugs still remain.

docs/source/stubs.rst
```diff
@@ -356,20 +356,28 @@ type stubs::
...
 -------------------
 
-Type checkers should accept module-level and class-level aliases, e.g.::
+Type checkers should accept module-level type aliases, optionally using
+``TypeAlias`` (PEP 613 [#pep613]_), e.g.::
 
   _IntList = list[int]
+  _StrList: TypeAlias = list[str]
+
+Type checkers should also accept regular module-level or class-level aliases,
```
AlexWaygood, line 365:
```suggestion
Type checkers should also accept module-level or class-level variable aliases,
```

docs/source/stubs.rst
```diff
@@ -847,6 +853,25 @@ No::
     z = 0  # type: int
     a = ...  # type: int
 
+Type Aliases
+------------
+
+Use ``TypeAlias`` for type aliases (but not for regular aliases).
```
AlexWaygood, line 859:
```suggestion
Use ``TypeAlias`` for type aliases (but not for variable aliases).
```

AlexWaygood (commented):
Maybe "variable aliases" instead of "regular aliases"?

docs/source/stubs.rst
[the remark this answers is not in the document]
srittau: Fair enough, I'll send a PR tomorrow, unless someone else wants to improve the wording? Please feel free to commit to the branch.

docs/source/stubs.rst
srittau: I agree that something more fitting than just "regular" would be a good idea, but "variable" doesn't really fit, in my opinion. For example, while class or function aliases could technically be considered variable aliases, it sounds a bit weird. Maybe we should just list the other cases? "Class, function, method, and (module- and class-level) attribute aliases".

docs/source/stubs.rst
```diff
@@ -356,20 +356,28 @@ type stubs::
...
+
+  def a() -> None: ...
+  b = a
 
   class C:
       def f(self) -> int: ...
       g = f
 
-An alias to a type may contain type variables. As per PEP 484 [#pep484]_,
+A type alias to a type may contain type variables. As per PEP 484 [#pep484]_,
```
hauntsaninja, line 375:
```suggestion
A type alias may contain type variables. As per PEP 484 [#pep484]_,
```

docs/source/stubs.rst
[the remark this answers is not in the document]
srittau: Actually, we should make that a separate PR, as this is not related to type aliases.

docs/source/stubs.rst
AlexWaygood: I see your point. I think I'd still go with "variable aliases", but I don't object massively if you stick with "regular aliases" (I think "regular aliases" is probably better than listing them all out)

srittau:
Pingety pong.

JelleZijlstra (approved):

PR #1128, opened 2022-04-10
Update the statuses of the typing PEPs

rednafi:
[PEP-675](https://peps.python.org/pep-0675/) has been accepted and will be implemented in Python 3.11. So, I removed the draft status.


JelleZijlstra:
Thanks! While you're at it, could you also update PEP 655 (accepted) and PEP 677 (rejected)?

rednafi:
Hi @JelleZijlstra. I went ahead and added the statuses of all the PEPs. I assume this is more complete.

JelleZijlstra:
I'd prefer to leave out "final" and "accepted", especially because the difference between the two isn't that useful for readers.

rednafi:
Fair point. I removed the `accepted` and `final` statuses.

JelleZijlstra (approved):

PR #1139, opened 2022-04-15
Fix tests on Python 3.11

JelleZijlstra:
- Defer to the PEP 646 implementation in typing.py on 3.11
- Adjust some tests accordingly. Noted a bug in
  https://github.com/python/cpython/pull/32341#issuecomment-1100466389
- typing._type_check() is more lenient in 3.11 and no longer rejects ints
- The representation of the empty tuple type changed

Tests pass for me on a 3.11 build from today now.

### DIFF
--- a/typing_extensions/src/test_typing_extensions.py
+++ b/typing_extensions/src/test_typing_extensions.py
@@ -29,6 +29,8 @@ from typing_extensions import assert_type, get_type_hints, get_origin, get_args
 # version of the typing module.
 TYPING_3_8_0 = sys.version_info[:3] >= (3, 8, 0)
 TYPING_3_10_0 = sys.version_info[:3] >= (3, 10, 0)
+
+# 3.11 makes runtime type checks (_type_check) more lenient.
 TYPING_3_11_0 = sys.version_info[:3] >= (3, 11, 0)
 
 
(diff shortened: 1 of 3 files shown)


srittau (approved):
LGTM, although I have one concern that you can ignore if you wish.

typing_extensions/src/test_typing_extensions.py
```diff
@@ -2425,7 +2437,10 @@ def test_basic_plain(self):
 
     def test_repr(self):
         Ts = TypeVarTuple('Ts')
-        self.assertEqual(repr(Unpack[Ts]), 'typing_extensions.Unpack[Ts]')
+        if TYPING_3_11_0:
+            self.assertEqual(repr(Unpack[Ts]), '*Ts')
+        else:
+            self.assertEqual(repr(Unpack[Ts]), 'typing_extensions.Unpack[Ts]')
```
srittau, line 2443: (Which I forgot to save.)

Wouldn't it make more sense to make the typing_extensions version match the one from Python 3.11?

JelleZijlstra: Yes, I was thinking about that too. On the other hand, `*Ts` is a syntax error before 3.11, so it may not be very helpful as `repr()` output.

JelleZijlstra: @mrahtz what do you think? Should we use `*Ts` in the repr() for Unpack even before 3.11?

mrahtz: I think I agree it should be `Unpack[Ts]` before 3.11 - isn't the `repr()` suppose to be as close as possible to something you can directly `eval()`?

mrahtz:
(Thanks, Jelle!)

typing_extensions/src/test_typing_extensions.py
JelleZijlstra: Thanks, then let's stick with the current behavior.

PR #1156, opened 2022-04-19
Add a README.rst file back temporarily.

### LINKED ISSUE #1152: Correct “Home” link on pypi `README.rst` ↦ `README.md`
On the pypi distributon page of typing-extensions https://pypi.org/project/typing-extensions/ there is an “Home” link within the “Project links” section on the left site bar.

This “Home” link currently point to 

    https://github.com/python/typing/blob/master/typing_extensions/README.rst

which is an invalid link resulting in an 404, because the file `README.rst` got renamed to `README.md` within the repository. Would be nice if the link could be corrected.

srittau:
Seems short, but I don't know what else to write.

Closes #1152

### DIFF
--- a/typing_extensions/README.rst
+++ b/typing_extensions/README.rst
@@ -1,1 +1,1 @@
-Please see ``README.md``.
+Please see `README.md <https://github.com/python/typing/blob/master/typing_extensions/README.md>`_ for the current version of the README file.
--- a/typing_extensions/README.rst
+++ b/typing_extensions/README.rst
@@ -1,0 +1,1 @@
+Please see ``README.md``.


typing_extensions/README.rst
```diff
@@ -0,0 +1 @@
+Please see ``README.md``.
```
JelleZijlstra, line 1: Could we make this a link? Not sure how smart GH's RST renderer is.

srittau: Seems to work fine: https://github.com/python/typing/blob/a326f37fbc78f25b988f7ecbbdc49b5900c4884a/typing_extensions/README.rst

JelleZijlstra:
Thanks!

PR #1164, opened 2022-04-27
Make library docs more FAQ-like

hauntsaninja:
In the discussion that ensued in [this thread](https://mail.python.org/archives/list/python-dev@python.org/thread/VIZEBX5EYMSYIJNDBF6DMUMZOCWHARSO/#LAR7R5MCAC74N7H7NQXJUGXG7LCOA4CB), there was feedback that we could use better docs for library authors looking to understand what it entails to add types to their package.

Some of this was based on https://mypy.readthedocs.io/en/stable/installed_packages.html

### DIFF
--- a/docs/source/libraries.rst
+++ b/docs/source/libraries.rst
@@ -4,79 +4,197 @@
 Typing Python Libraries
 ***********************
 
-Much of Python’s popularity can be attributed to the rich collection of
+Much of Python's popularity can be attributed to the rich collection of
 Python libraries available to developers. Authors of these libraries
 play an important role in improving the experience for Python
 developers. This document provides some recommendations and guidance for
 Python library authors.
 
-These recommendations are intended to provide the following benefits:
+Why provide type annotations?
+=============================
 
-1. Consumers of libraries should have a great coding experience with
-   fast and accurate completion suggestions, class and function
-   documentation, signature help (including parameter default values),
-   hover text, and auto-imports. This should happen by default without
-   needing to download extra packages and without any special
-   configuration. These features should be consistent across the Python
-   ecosystem regardless of a developer’s choice of editor, IDE, notebook
-   environment, etc.
-2. Consumers of libraries should be able to rely on complete and
-   accurate type information so static type checkers can detect and
-   report type inconsistencies and other violations of the interface
-   contract.
-3. Library authors should be able to specify a well-defined interface
-   contract that is enforced by tools. This allows a library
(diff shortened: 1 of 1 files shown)


erictraut (commented):
Thanks, these updates look great!

docs/source/libraries.rst
```diff
@@ -4,79 +4,197 @@
...
+      packages=["my_great_package"],
+   )
 
-In many existing type stubs (such as those found in typeshed), default
-parameter values are replaced with with ``...`` and all docstrings are
-removed. We recommend that default values and docstrings remain within
-the type stub file so language servers can display this information to
-developers.
+The presence of ``.pyi`` files does not affect the Python interpreter at runtime
+in any way. However, static type checkers will only look at ``.pyi`` file and
```
JelleZijlstra, line 138:
```suggestion
in any way. However, static type checkers will only look at the ``.pyi`` file and
```

docs/source/libraries.rst
```diff
@@ -4,79 +4,197 @@
...
-Library Interface
-=================
+Companion type stub package
+^^^^^^^^^^^^^^^^^^^^^^^^^^^
 
-`PEP 561 <https://www.python.org/dev/peps/pep-0561/>`__ indicates that a
-``py.typed`` marker file must be included in the package if the author
-wishes to support type checking of their code.
+These are often referred to as "stub-only" packages. The package should have a
+prefix of the runtime package name with a suffix of ``-stubs``. The py.typed
```
JelleZijlstra, line 145: Not a fan of this sentence:
```suggestion
These are often referred to as "stub-only" packages. The name of the stub package
should be the name of the runtime package suffixed with ``-stubs``. The ``py.typed``
```

docs/source/libraries.rst
```diff
@@ -4,79 +4,197 @@
...
+      version="0.1",
+      package_data={"my_great_package-stubs": ["__init__.pyi", "stuff.pyi"]},
+      packages=["my_great_package-stubs"]
+   )
+
+
+Users are then able to install the stubs-only package separately to provide
+types for the original library.
+
+Note that to ensure inclusion of ``.pyi`` and ``py.typed`` files in an sdist
```
JelleZijlstra, line 174: Weird that this is in the stub package section, while the one about inline stub files only talks about `setup.py`.

JelleZijlstra (commented):
Thanks for doing this!

PR #1189, opened 2022-05-19
Remove typing-extensions

JelleZijlstra:


README.md
```diff
@@ -1,10 +1,8 @@
 [![Chat at https://gitter.im/python/typing](https://badges.gitter.im/python/typing.svg)](https://gitter.im/python/typing?utm_source=badge&utm_medium=badge&utm_campaign=pr-badge&utm_content=badge)
 
-Static Typing for Python
-========================
+# Static Typing for Python
```
JelleZijlstra, line 3: Oops, I think this was some markdown autoformatter. Let me know if you'd like me to undo it.

.github/workflows/ci.yml
```diff
@@ -59,6 +29,3 @@ jobs:
 
       - name: Lint implementation
         run: flake8
```
srittau, line 31: Do we even need to run flake8? The only python code left is the typing-summary script in scripts.

srittau (approved):

.github/workflows/ci.yml
srittau: On the other hand, linting can't hurt, and who knows whether we add more scripts later. So I retract my comment.

PR #1192, opened 2022-05-21
Delete `CONTRIBUTING.md`

AlexWaygood:
It's all specific to `typing_extensions`


JelleZijlstra:
Maybe we should have a new one about contributing to the docs, but that can wait.

AlexWaygood:
> Maybe we should have a new one about contributing to the docs, but that can wait.

My thoughts too

PR #1198, opened 2022-05-25
Best Practices: Add note about object/Any

srittau:
Part of #851

### DIFF
--- a/docs/source/best_practices.rst
+++ b/docs/source/best_practices.rst
@@ -68,7 +68,7 @@ Using ``Any`` and ``object``
 Generally, use ``Any`` when a type cannot be expressed appropriately
 with the current type system or using the correct type is unergonomic.
 
-If a function accepts virtually any object as an argument, for example
+If a function accepts every possible object as an argument, for example
 because it's only passed to ``str()``, use ``object`` instead of ``Any`` as
 type annotation. Similarly, if the return value of a callback is ignored,
 annotate it with ``object``::
--- a/docs/source/best_practices.rst
+++ b/docs/source/best_practices.rst
@@ -22,24 +22,6 @@ Please see :ref:`contact` on how to join the discussion.
 Typing Features
 ===============
 
-Future Import
--------------
-
-We recommend to use ``from __future__ import annotations`` in source files
-that use type annotations. This allows you to use some newer syntax features —
-like built-in generics or new-style unions — in type annotations, even in
-Python versions that don't officially support these features yet.
-
-For example, the following code works with Python 3.7 and above::
-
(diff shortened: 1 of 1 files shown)


hauntsaninja (commented):
I'm not sure I'd recommend using `from __future__ import annotations` for new syntax without at least having a note mentioning that doing so breaks runtime uses of annotations (and e.g. type aliases won't work).

The note doesn't need to be lengthy or scary, e.g. it could just link to https://mypy.readthedocs.io/en/stable/runtime_troubles.html or something, but I do think we should acknowledge this.

srittau:
> I'm not sure I'd recommend using from __future__ import annotations for new syntax without at least having a note mentioning that doing so breaks runtime uses of annotations (and e.g. type aliases won't work).

I'm not sure I follow. What runtime uses break that work without the future import.

docs/source/best_practices.rst
```diff
@@ -44,12 +62,21 @@ No::
...
 
-Using `Any`
------------
+Using ``Any`` and ``object``
+----------------------------
 
 Generally, use ``Any`` when a type cannot be expressed appropriately
 with the current type system or using the correct type is unergonomic.
 
+If a function accepts virtually any object as an argument, for example
```
JelleZijlstra, line 71:
```suggestion
If a function accepts every possible object as an argument, for example
```
Avoid using the word "any" to describe how to use "object", it's going to be confusing.

JelleZijlstra:
I personally would not recommend using `from __future__ import annotations`:

- As Shantanu says, it can break some runtime use cases. See e.g. David Foster's message in https://discuss.python.org/t/finding-edge-cases-for-peps-484-563-and-649-type-annotations/14314 .
- It's still unclear whether the `__future__` import will ever become the default, so people using it may have to migrate later to another subtly different version.

hauntsaninja:
It's not just the edge cases, since we're currently recommending `from __future__ import annotations` as a way to backport new syntax (not just for forward references). So `typing.get_type_hints` and friends will break.

Suggested wording (based on mypy docs):

> Some code may have use cases for evaluation of annotations at runtime, for instance, through use of [`typing.get_type_hints`](https://docs.python.org/3/library/typing.html#typing.get_type_hints) or `eval`. If your annotation would raise an error when evaluated (say by using new-style unions on Python 3.9 and older), you may need to be careful. See [here](https://mypy.readthedocs.io/en/stable/runtime_troubles.html ) for more discussion on sharp edges.

srittau:
I have removed the future import section for now.

srittau:
Bumpity bump.

PR #1243, opened 2022-08-22
Add Aliasing Decorators section to libraries.rst

Drino:
I believe I'm not the only one who will struggle with behavior of `Callable` type aliases.
I hope if this is added to the documentation - some people will save an hour on figuring how `TypeAlias` work.

I'm not sure if it is worth introducing "decorator factory" term here, but it is used in [mypy documentation](https://mypy.readthedocs.io/en/stable/generics.html#decorator-factories).

Related discussion:
https://github.com/python/typing/discussions/1236

### DIFF
--- a/docs/source/libraries.rst
+++ b/docs/source/libraries.rst
@@ -490,6 +490,36 @@ original signature, thus blinding type checkers and other tools that
 provide signature assistance. As such, library authors are discouraged
 from creating decorators that mutate function signatures in this manner.
 
+Aliasing Decorators
+-------------------
+
+When writing a library with a couple of decorator factories 
+(i.e. functions returning decorators, like ``complex_decorator`` from 
+Annotating Decorators section) it may be tempting to create a shortcut 
+for a decorator. 
+
+Different type checkers handle ``TypeAlias`` involving ``Callable`` in a 
+different manner, so the most portable and easy way to create a shortcut 
+is to define a callable ``Protocol`` as described in `PEP 
+544 <https://peps.python.org/pep-0544/#callback-protocols>`_:
+
+.. code:: python
+
+   _F = TypeVar("_F", bound=Callable[..., Any])
+   
+   class PDecorator(Protocol):
+       def __call__(self, _: _F, /) -> _F:
+           ...
+
+   def decorator_factory(*, mode: str) -> PDecorator:
+       """
+        Decorator factory is invoked with arguments like this:
+          @decorator_factory(mode="easy")
+          def my_function(): ...
+        """
+      ...
+
+
 Generic Classes and Functions
 -----------------------------
 


JelleZijlstra (approved):
Thanks, this is a useful addition.

I'm not sure it should be in the "libraries" document, but there's already a TODO on moving it out.

docs/source/libraries.rst
```diff
@@ -490,6 +490,36 @@ original signature, thus blinding type checkers and other tools that
 provide signature assistance. As such, library authors are discouraged
 from creating decorators that mutate function signatures in this manner.
 
+Aliasing Decorators
+-------------------
+
+When writing a library with a couple of decorator factories 
+(i.e. functions returning decorators, like ``complex_decorator`` from 
+Annotating Decorators section) it may be tempting to create a shortcut
```
srittau, line 498: Please add the following line before the "Annotating Decorators" section heading:

```
.. _annotating-decorators:
```

(And a similar line above the "Aliasing Decorators" header.) See the "How much of my library needs types" section for an example.

You can then link the section like this:

```suggestion
:ref:`annotating-decorators` section) it may be tempting to create a shortcut 
```

srittau (changes requested):
Thanks, this makes sense. But I think we should rather recommend to use the `_typeshed.IdentityFunction` protocol, instead of recommending a custom protocol.

```python
from typing import TYPE_CHECKING
if TYPE_CHECKING:
    from _typeshed import IdentityFunction

def decorator_factory(*, mode: str) -> "IdentityFunction":
    """
    Decorator factory is invoked with arguments like this:
        @decorator_factory(mode="easy")
        def my_function(): ...
    """
    ...
```

AlexWaygood:
> But I think we should rather recommend to use the `_typeshed.IdentityFunction` protocol, instead of recommending a custom protocol.

I think it's important to note that this only works in stub files, if you have `from __future__ annotations` enabled, or if you use `-> "IdentityFunction"` instead of `-> IdentityFunction`.

srittau:
> I think it's important to note that this only works in stub files, if you have `from __future__ annotations` enabled, or if you use `-> "IdentityFunction"` instead of `-> IdentityFunction`.

Thanks, easy to forget when always using the future import. :)

Drino:
Thanks for the fast review and cool suggestion!

I've rewritten the example using `IdentityFunction`, though I'm a little bit concerned to show a "proper" example of decorator protocol definition, so I've added another one with slightly less trivial decorator.

docs/source/libraries.rst
```diff
@@ -490,6 +490,60 @@ original signature, thus blinding type checkers and other tools that
...
+(i.e. functions returning decorators, like ``complex_decorator`` from 
+:ref:`annotating-decorators` section) it may be tempting to create a shortcut 
+for a decorator. 
+
+Different type checkers handle ``TypeAlias`` involving ``Callable`` in a 
+different manner, so the most portable and easy way to create a shortcut 
+is to define a callable ``Protocol`` as described in `PEP 
+544 <https://peps.python.org/pep-0544/#callback-protocols>`_.
+
+There is already a ``Protcol`` called ``IndentityFunction`` defined in ``_typeshed``:
```
AlexWaygood, line 506:
```suggestion
There is already a ``Protocol`` called ``IdentityFunction`` defined in ``_typeshed``:
```

It might also be worth linking to the `_typeshed` README here: https://github.com/python/typeshed/blob/master/stdlib/_typeshed/README.md

Drino:
@srittau Hello, I've went through your comments and it seems that I've resolved the issues. Would you look at this PR one more time please?

srittau (approved):
I'm sorry. This completely to fell through the cracks. LGTM now, although I've changed a few links to refer to the – now existing – typing spec and to link to the Python docs.

PR #1352, opened 2023-02-19
Update docs for tools that inline pyi

hauntsaninja:
retype is no longer maintained. Since it's based on lib2to3, its corpse is not going to be useful for too long. merge_pyi now just wraps libCST, so mention that.

If someone seeing this is looking for things to do, I think libCST's support here could be improved. E.g., I filed https://github.com/Instagram/LibCST/pull/868 But maybe that should convert to typing.Union. I'm sure there are other things that could be improved!


AlexWaygood (approved):
LGTM

PR #1380, opened 2023-03-28
Link to mypy docs

hauntsaninja:


hauntsaninja:
(I'll incorporate any feedback on the similar CPython PR https://github.com/python/cpython/pull/103100/files)

srittau (approved):

PR #1416, opened 2023-06-19
Import the generics documentation from mypy

hauntsaninja:


docs/source/generics.rst
```diff
@@ -0,0 +1,984 @@
...
+
+.. code-block:: python
+
+   # Construct an empty Stack[int] instance
+   stack = Stack[int]()
+   stack.push(2)
+   stack.pop() + 1
+   stack.push('x')  # error: Argument 1 to "push" of "Stack" has incompatible type "str"; expected "int"
+
+When creating instances of generic classes, the type parameter can usually be
```
superbobry, line 53: Nit: s/type parameter/type argument/ here and below.

docs/source/generics.rst
```diff
@@ -0,0 +1,984 @@
...
+
+.. code-block:: python
+
+   class Box(Generic[T]):
+       def __init__(self, content: T) -> None:
+           self.content = content
+
+   Box(1)       # OK, inferred type is Box[int]
+   Box[int](1)  # Also OK
+   Box[int]('some string')  # error: Argument 1 to "Box" has incompatible type "str"; expected "int"
```
superbobry, line 65: Is it worth elaborating what "compatibility" means here?

My understanding is that class instantiations require argument types to be consistent subtypes of parameter types. So, here `str` is not a consistent subtype of `int`, thus the error. PEP-483 [calls](https://peps.python.org/pep-0483/#summary-of-gradual-typing) consistent subtyping just "consistency".

docs/source/generics.rst
```diff
@@ -0,0 +1,984 @@
...
+
+.. _generic-subclasses:
+
+Defining subclasses of generic classes
+**************************************
+
+User-defined generic classes and generic classes defined in :py:mod:`typing`
```
superbobry, line 72: What happens if the same generic class occurs multiple times in the hierarchy with different type arguments? For example,

```
class G(Generic[T]): ...
class A(G[int]): ...
class B(A, G[str]): ...
```

This is valid from the runtime MRO perspective due to type erasure, but perhaps this should be a static type error?

docs/source/generics.rst
```diff
@@ -0,0 +1,984 @@
...
+
+.. _variance-of-generics:
+
+Variance of generic types
+*************************
+
+There are three main kinds of generic types with respect to subtype
+relations between them: invariant, covariant, and contravariant.
+Assuming that we have a pair of types ``Animal`` and ``Bear``, and
+``Bear`` is a subtype of ``A``, these are defined as follows:
```
superbobry, line 342: You probably meant Animal instead of A here.

docs/source/generics.rst
```diff
@@ -0,0 +1,984 @@
...
+
+* A generic class ``MyCovGen[T]`` is called covariant in type variable
```
superbobry, line 344: Should we use "type parameter" instead of "type variable" to match the wording in the beginning of this document?

docs/source/generics.rst
```diff
@@ -0,0 +1,984 @@
...
+  ``T`` if ``MyCovGen[Bear]`` is always a subtype of ``MyCovGen[Animal]``.
```
superbobry, line 345: I think leaving out "always" here and below would make the wording more clear.

docs/source/generics.rst
```diff
@@ -0,0 +1,984 @@
...
+Type variables with upper bounds
+********************************
+
+By default, a type variable can be replaced with any type. This means that
+you can't do very much with an object of type ``T`` safely -- you don't
+know anything about it!
+
+It's therefore often useful to be able to limit the types that a type
+variable can take on, for instance, by restricting it to values that are
+subtypes of a specific type.
```
superbobry, line 461: Nit: "consistent subtypes", since the bound can include `Any`.

docs/source/generics.rst
```diff
@@ -0,0 +1,984 @@
...
+with the ``bound=...`` keyword argument to :py:class:`~typing.TypeVar`.
+
+.. code-block:: python
+
+    from typing import TypeVar, SupportsAbs
+
+    T = TypeVar('T', bound=SupportsAbs[float])
+
+In the definition of a generic function that uses such a type variable
+``T``, the type represented by ``T`` is assumed to be a subtype of
```
superbobry, line 473: Nit: as above, "consistent subtype".

docs/source/generics.rst
```diff
@@ -0,0 +1,984 @@
...
+    largest_in_absolute_value(-3.5, 2)   # OK, has type float
+    largest_in_absolute_value(5+6j, 7)   # OK, has type complex
+    largest_in_absolute_value('a', 'b')  # error: error: Value of type variable "T" of "largest_in_absolute_value" cannot be "str"
+
+Type parameters of generic classes may also have upper bounds, which
+restrict the valid values for the type parameter in the same way.
+
+.. _type-variable-value-restriction:
+
+Type variables with value restriction
```
superbobry, line 497: The term "value restriction" is slightly confusing here, because type variable constraints work on types, i.e. they restrict the set of types a type parameter can take.

I think PEP-484 and the `typing` docs use the term "constraints", so it's probably best to use it here as well.

docs/source/generics.rst
```diff
@@ -0,0 +1,984 @@
...
+.. _generic-type-aliases:
+
+Generic type aliases
+********************
+
+Type aliases can be generic. In this case they can be used in two ways:
+Subscripted aliases are equivalent to original types with substituted type
+variables, so the number of type arguments must match the number of free type variables
+in the generic type alias. Unsubscripted aliases are treated as original types with free
+variables replaced with ``Any``. Examples (following :pep:`PEP 484: Type aliases
```
superbobry, line 865: Do type checkers always use `Any` regardless of the type parameter bound?

docs/source/generics.rst
AlexWaygood: I actually think it's okay to be a little loose with the terminology here, since this is documentation aimed at beginners. The "consistent" qualifier doesn't mean much unless you've been brushing up on your type theory, and a beginner _probably_ won't have been :)

docs/source/generics.rst
AlexWaygood: Yes

docs/source/generics.rst
JelleZijlstra: Yes, we should use "TypeVar with constraints" for AnyStr-like TypeVars.

docs/source/generics.rst
hauntsaninja: Is this right? I meant in the case where there is no argument

docs/source/generics.rst
hauntsaninja: Like Alex says below, I think okay to have slightly lax terminology. This is meant to be approachable documentation, not necessarily a spec, and is pretty intuitive that str is not compatible with int

docs/source/generics.rst
hauntsaninja: Oh interesting. Both mypy and pyright accept this, but you're right that it is weird

docs/source/generics.rst
hauntsaninja: I did, thank you!

docs/source/generics.rst
hauntsaninja: Fixed

docs/source/generics.rst
hauntsaninja: Good point, fixed

docs/source/generics.rst
hauntsaninja: Thanks, fixed!

docs/source/generics.rst
superbobry: Yes, I think that is right. Type parameters are the type variables used in the definition of a generic class. Type arguments are the types a generic class is specialized with. So, below `T` is a type parameter of `Box` and `int` in `Box[int]()` is a type argument.

docs/source/generics.rst
superbobry: Thanks Alex, I agree that new typing users probably will not care too much about this detail. However, I expect them to be equally unfamiliar with both subtyping and consistent subtyping. So, we will not make this paragraph any harder to follow by using the right term.

docs/source/generics.rst
```diff
@@ -0,0 +1,986 @@
...
+            return x[0]
+
+        def pair_with_first(self, x: list[S]) -> tuple[S, T]:
+            return (x[0], self.content)
+
+    box = PairedBox("asdf")
+    reveal_type(box.first([1, 2, 3]))  # Revealed type is "builtins.int"
+    reveal_type(box.pair_with_first([1, 2, 3]))  # Revealed type is "tuple[builtins.int, builtins.str]"
+
+In particular, the ``self`` argument may also be generic, allowing a
```
JelleZijlstra, line 224: This example isn't great as it should be using `Self`. Not sure if there's a better example to give, though.

docs/source/generics.rst
```diff
@@ -0,0 +1,986 @@
+Generics
+========
+
+You may have seen type hints like ``list[str]`` or ``dict[str, int]`` in Python
```
AlexWaygood, line 4: General question: I'm not sure how backwards-compatible these docs are meant to be. In your last PR, I noticed you used `Optional[X]` instead of `X | None` throughout, but didn't mention it because I assumed you were trying to keep the code examples compatible with py38+. But this document uses PEP-585 syntax throughout, and doesn't mention the older way of writing these types until the last section right at the bottom.

I prefer making the new syntax more prominent than the old syntax, as this document does, but maybe we should have the note about how to do things in a pre-PEP-585 world slightly higher up?

JelleZijlstra: I think the ideal would be if we had some selector at the top of examples, similar to the one for Unix/Windows at https://packaging.python.org/en/latest/tutorials/packaging-projects/, where users could toggle between "Python 3.9+" and "Python 3.7-3.8" or so.

Setting that up is better left for a separate PR though.

AlexWaygood: Yes, that would be ideal!

docs/source/generics.rst
```diff
@@ -0,0 +1,986 @@
...
+
+   circle: Circle = Circle().set_scale(0.5).set_radius(2.7)
+   square: Square = Square().set_scale(0.5).set_width(3.2)
+
+Without using generic ``self``, the last two lines could not be type
+checked properly, since the return type of ``set_scale`` would be
+``Shape``, which doesn't define ``set_radius`` or ``set_width``.
+
+Other uses are factory methods, such as copy and deserialization.
+For class methods, you can also define generic ``cls``, using :py:class:`Type[T] <typing.Type>`:
```
hauntsaninja, line 258:
```suggestion
For class methods, you can also define generic ``cls``, using :py:class:`type`:
```

docs/source/generics.rst
hauntsaninja: For the mypy docs this is based on, I went by https://pypistats.org/packages/mypy minor version counts to determine what version should be emphasised. 3.9 is mode and median.

hauntsaninja:
Thank you everyone for reviews!

PR #1477, opened 2023-09-25
Add an anti-pitch for typing

hauntsaninja:


hauntsaninja:
@pfmoore Would the existence of this document help you? I've tried to represent things as fairly as possible, but I'm in quite deep and might be biased :-) Do you have any feedback?

docs/source/typing_anti_pitch.rst
```diff
@@ -0,0 +1,80 @@
...
+  the total cost of adding type annotations to a large extant codebase can be prohibitive.
+
+* Your library does enough dynamic things that type checking would be unlikely to help your users.
+  Either you would have to spend a lot of effort redesigning your library in ways that static type
+  checkers could better understand, or you'd have to invest a lot of effort into figuring out clever
+  type annotations to twist the arms of type checkers.
+
+* Your codebase has suffered at the hands of `Hyrum's Law <https://www.hyrumslaw.com/>`_
+  and all possible observable behaviour is depended on. In order to avoid false positives for your
+  users, all your types end up being either a) complicated ``Protocol``s that are hard to maintain,
```
AlexWaygood, line 53: I think you might need a backslash here 

```suggestion
  users, all your types end up being either a) complicated ``Protocol``\s that are hard to maintain,
```

docs/source/typing_anti_pitch.rst
```diff
@@ -0,0 +1,80 @@
...
+
+In the words of :pep:`484`::
+
+    It should also be emphasized that Python will remain a dynamically typed language, and the
+    authors have no desire to ever make type hints mandatory, even by convention.
+
+The idea that dynamism in Python is a strength of the language is reflected in the fact that
+Python's type system is gradual. See :pep:`483` for details, but the long and short of this is
+that you can add static types to your codebase only to the extent that you want to, and static
+type checkers are designed to be able to put up with this.
```
pfmoore, line 14: It’s not just static type checkers, it’s also IDEs, and maybe other tools as typing gains popularity.

And to cover those other tools, “are designed” should probably be changed to “should be designed”. That would allow directing complaints like “my IDE needs type hints to do autocompletion” towards the IDE to advise the user.

pfmoore (approved):
Overall, this seems good, and sufficiently balanced.

It still puts the onus on the library author to justify their choice to not use typing (even if it’s just to the extent of linking to this) - that’s the social side of the problem, though, and documentation won’t fix that.

docs/source/typing_anti_pitch.rst
```diff
@@ -0,0 +1,80 @@
...
+perhaps you'd still like to help your users who do use static type checking -- and maybe you have
+some enthusiastic would-be contributors willing to help with this.
+
+One option is encourage such contributors  to publish a :pep:`561` stub-only package that is
+maintained separately from your main project. You could also contribute these stubs to the
+`typeshed <https://github.com/python/typeshed>`_ project.
+
+If more users pester you about adding static types, feel free to link them to this document. And if
+you ever change your mind, make sure to check out some of the other guides in this documentation,
+and ask any questions you have over at `Python's typing discussions <https://github.com/python/typing/discussions>`_.
```
AlexWaygood, line 80: A lot of typing enthusiasts online have a slightly sneering attitude towards typeshed and/or the idea of stubs in general. Some of their criticisms are justified, but also, typeshed has made huge improvements in its CI checks in recent years. Possibly we could mention that there are tools available that can help to keep stubs accurate and up-to-date (maybe link to https://typing.readthedocs.io/en/latest/source/writing_stubs.html#tools-for-maintaining-stubs?)

docs/source/typing_anti_pitch.rst
```diff
@@ -0,0 +1,80 @@
...
+Python's type system is gradual. See :pep:`483` for details, but the long and short of this is
+that you can add static types to your codebase only to the extent that you want to, and static
+type checkers are designed to be able to put up with this.
+
+It's also worth noting that "static type checking" encompasses a spectrum of possible degrees of
+strictness. On the one hand, you can set yourself up so that your type checker does almost nothing.
+On the other -- well, I love type checking, but I would quit Python if I had to enable all
+possible strictness checks that type checkers offer.
+
+Anway, with all that said, here's a list of possible reasons to not use static type checking
```
hmc-cs-mdrissi, line 21: Could you add one point on readability/complexity? Readability is subjective but is one common reason that good type hints for a function may make signature too long/harder to understand.

AlexWaygood: Yeah. Readability can often be mitigated using type aliases, but that doesn't address the complexity issue.

docs/source/typing_anti_pitch.rst
```diff
@@ -0,0 +1,80 @@
...
+  instead of relying on explicit annotations, maybe they're too slow, maybe they don't integrate
+  well with your editor, maybe they're too hard to configure. Whatever the reason -- it just doesn't
+  work for your project.
+
+* Type checking in Python isn't actually strict enough, powerful enough or expressive enough for
+  you. Python type checkers end up making various decisions out of pragmatism, or due to limited
+  resources, and these decisions might not be the ones for you.
+
+Advice for maintainers of untyped libraries
+*******************************************
```
srittau, line 68: As first point I would mention the possibility to distribute a stub file directly with the library. This is the best solution for users, while not using type checking for the library itself. The other options are more suited if a maintainer doesn't want that burden either.

srittau (commented):
I think this makes a lot of sense. This feels a bit like the [You Might Not Need Redux](https://medium.com/@dan_abramov/you-might-not-need-redux-be46360cf367) article by the original author of Redux. A few general notes:

* Stylistic nit: I would lose the parentheses when discussing counterpoints. These are valid arguments and not just asides.
* While it's okay for now, I think putting this document into an "FAQ" section would be a better fit than the "Guides" section.

docs/source/typing_anti_pitch.rst
```diff
@@ -0,0 +1,80 @@
...
+possible strictness checks that type checkers offer.
+
+Anway, with all that said, here's a list of possible reasons to not use static type checking
+in Python::
+
+* You simply don't want to. Python is a tool that is meant to serve you. Python is a big tent,
+  multi-paradigm language that generally allows you to do things in the way that best suits your
+  needs, as best determined by you.
+
+* The cost-benefit ratio isn't good enough. Pleasing static type checkers requires a non-zero amount
```
Tinche, line 28: One interesting aspect of this is if your project is based on an ecosystem that simply doesn't work well with typing. For example, someone using FastAPI and SQLAlchemy will have a different (better) cost-benefit calculus here than someone using Django and the Django ORM. I personally view this as a reason to switch to better typed libraries but I bet the majority of the community would hold a different view ;)

docs/source/typing_anti_pitch.rst
pfmoore: No, I tried that with `editables` and apparently (I don't know whether this is still true) mypy doesn't typecheck the library code against the stubs. Which means there's literally zero benefit to the library maintainer for shipping a stubs file.

Also, it still puts the burden on the library maintainer and the point of this document is to *reduce* the burden on maintainers who don't want to add typing. The message should be "if all you (as a user) want is to have type annotations, you can write them yourself so you don't need the library maintainer to do *anything*".

docs/source/typing_anti_pitch.rst
pfmoore: If the phrase "you could contribute..." means the library maintainer, then again I see no benefit over just adding the type annotations, and the point here is that the library maintainer doesn't want to do that. If it means the end user who wants there to be type annotations, then yes, but will typeshed really accept annotations for a library when the submitter isn't the library maintainer? That seems somewhat risky - once again, the example of an over-restrictive type annotation that suggests the library requires particular input types, when actually the library maintainer prefes duck typing, and fully intends (without feeling the need to explicitly say so) to allow user-defined types that conform to the interface used by the library code.

As a library author, I'd be annoyed if I found out that typeshed had published annotations that didn't match my intent, and I had to ask for their removal or improvement. The public contract of my library should be defined by me, not by someone else who adds type annotations, no matter how well-intentioned.

docs/source/typing_anti_pitch.rst
pfmoore: > I personally view this as a reason to switch to better typed libraries

Unfortunately, and this is the root issue here, many people see it as a reason to ask[^1] library authors to add type annotations. Particularly in the case of specialised libraries where there may not *be* alternatives.

[^1]: And as is often the case with feature requests in open source, "ask" often ends up being more like "demand" or "pressure". This is especially true when dealing with people working on large commercial projects, which are often the projects that benefit most from type annotations 🙁

docs/source/typing_anti_pitch.rst
srittau: The idea is that the library code is not checked against the stubs. But it is shipped with the library, which has clear benefits to the library's users.

And of course this is more burden for the library maintainer than just handing off writing the stubs to other people, but it is a valid option, and the best one if a maintainer doesn't want to add type checking to the library, but is willing to maintain the stubs.

docs/source/typing_anti_pitch.rst
hmc-cs-mdrissi: https://github.com/python/typeshed/blob/main/CONTRIBUTING.md#third-party-library-stubs is the policy. I think most of stubs in typeshed are written by people separate from library maintainers.

A library can request it's stubs be changed/removed although I'm unsure that situation has happened in years typeshed existed. Stubs do sometimes leave typeshed but usually that's when library adopts the stubs and adds types.

AlexWaygood: > If the phrase "you could contribute..." means the library maintainer, then again I see no benefit over just adding the type annotations, and the point here is that the library maintainer doesn't want to do that

I understood the phrase "You could contribute..." as referring to the external contributors pressuring the maintainer to add typing to the project. I believe that's what @hauntsaninja intended it to mean, as well. I agree that the wording is currently slightly ambiguous, though, for sure!

> will typeshed really accept annotations for a library when the submitter isn't the library maintainer?

Yes, we do this all the time, since one of the primary reasons for typeshed's existence is to ease the burden on third-party maintainers who don't want to be bothered by this kind of thing :)

> That seems somewhat risky - once again, the example of an over-restrictive type annotation that suggests the library requires particular input types, when actually the library maintainer prefes duck typing, and fully intends (without feeling the need to explicitly say so) to allow user-defined types that conform to the interface used by the library code.

Well, we review any code that's contributed to us, and any bug reports are reported to us rather than the library maintainer. Our policy is to prefer duck-type-y protocols in situations like the ones you refer to.

> As a library author, I'd be annoyed if I found out that typeshed had published annotations that didn't match my intent, and I had to ask for their removal or improvement. The public contract of my library should be defined by me, not by someone else who adds type annotations, no matter how well-intentioned.

I'm sorry to hear that. We're generally pretty receptive to bug reports, whether they come from users or maintainers of libraries. We also generally remove stubs if a package has added a `py.typed` file to their package, indicating that they now either have inline types or ship their own stubs, rendering typeshed's stubs obsolete. I can't recall a situation where a maintainer has asked for us to remove stubs from typeshed without them adding a `py.typed` file first. If that happened, it would probably be tricky for us to balance the needs of our users (who might now be depending on a stubs package we're publishing to PyPI), and the desire of the maintainer.

Note, however, that even if typeshed didn't exist, it's still within anybody's power to publish a stubs package for any project they like! I think that typeshed is pretty clear about the fact that our stubs packages are contributed by third-party users, shouldn't be seen as "official" (except perhaps for the stdlib), and aren't necessarily blessed by the authors of the runtime package. Our README states:

> Typeshed contains external type annotations for the Python standard library and Python builtins, as well as third party packages as contributed by people external to those projects.

So, I think it's perhaps a stretch to say that typeshed claims to attempt to define the "public contract" of any external library. We're just doing the best we can to help people out in situations where users want typing but maintainers don't want to add it :)

pfmoore: > So, I think it's perhaps a stretch to say that typeshed claims to attempt to define the "public contract" of any external library.

That's fair. I'm very much talking theoretically at this point. At this point, the few libraries I publish personally have type annotations, and for the ones I contribute to adding types isn't my sole decision (and they mostly have adopted type annotations too).

The problem is that a lot of this is (as I've mentioned elsewhere, and you alluded to above) very much a social problem. It's about people's attitudes, and assumed responsibilities. And one of the attitudes that I'm bringing into this mix is a certain level of defensiveness about "ownership" of projects that I consider "mine". This comes from a history of me now having very limited volunteer time, and encountering too many users who feel they have a right to dictate my priorities. As a result, I'm taking a lot of this discussion too personally. Sorry for that.

In practice, all I really want here is for the document to be clearer that the suggestion is that *the user* can offer their suggested stubs to typeshed. That could be as simple as saying "*They* could also contribute...".

AlexWaygood: > The problem is that a lot of this is (as I've mentioned elsewhere, and you alluded to above) very much a social problem. It's about people's attitudes, and assumed responsibilities. And one of the attitudes that I'm bringing into this mix is a certain level of defensiveness about "ownership" of projects that I consider "mine". This comes from a history of me now having very limited volunteer time, and encountering too many users who feel they have a right to dictate my priorities. As a result, I'm taking a lot of this discussion too personally. Sorry for that.

Absolutely, I completely get that. As a typeshed maintainer, I view typeshed as part of the solution here, since one of the things we're trying hard to do is to reduce the burden on library maintainers like yourself. But we could also probably be even clearer about the fact that our third-party stubs are generally "unsactioned" :)

docs/source/typing_anti_pitch.rst
```diff
@@ -0,0 +1,80 @@
...
+
+* You're not opposed to type checking in theory, but you dislike Python type checkers in practice.
+  Maybe they don't understand enough of the idioms you use, maybe you'd like them to infer more
+  instead of relying on explicit annotations, maybe they're too slow, maybe they don't integrate
+  well with your editor, maybe they're too hard to configure. Whatever the reason -- it just doesn't
+  work for your project.
+
+* Type checking in Python isn't actually strict enough, powerful enough or expressive enough for
+  you. Python type checkers end up making various decisions out of pragmatism, or due to limited
+  resources, and these decisions might not be the ones for you.
```
mikeshardmind, line 65: While I find myself in this camp, I don't think this is a good point as an anti-pitch, if anything this pitcher to either 

a) use typing *in addition to* other checks
b) to not use Python for these tasks
or c) to create a type checker which is as strict as is desired.

Maybe there should be a section noting what type checking doesn't (currently) do? But then this is largely based on what implementations currently support. That's still somewhat Python's problem on a social level, having kicked this to external implementors and not having a reference implementation that others must conform with, but it's harder to have such a documentation section as a result as well.

either way though, I don't think this belongs, at least not without some modifications that point out that this isn't necessarily a "why not typing".

docs/source/typing_anti_pitch.rst
hauntsaninja: Added a commit for this: https://github.com/python/typing/pull/1477/commits/b2a059846aca1275ed9565f2006792590c10e7bc

docs/source/typing_anti_pitch.rst
hauntsaninja: Thanks, I added https://github.com/python/typing/pull/1477/commits/4a3a6ebd7044533d675173fde3562285331ab958 to better account for this

docs/source/typing_anti_pitch.rst
hauntsaninja: Thanks, I added https://github.com/python/typing/pull/1477/commits/292145126fb053fa77eb0d77d33601ad05c3e41c
While this is a "why not typing", it is implicitly a "why not typing in Python as is implemented today", so I think it still belongs, even if it's unlike the other items in this list.

docs/source/typing_anti_pitch.rst
AlexWaygood: @hauntsaninja, you marked this thread as "resolved", but none of the discussion on this thread has actually touched on my original suggestion in https://github.com/python/typing/pull/1477#discussion_r1335520420 that we could add a link here to the docs for best practices on testing stubs. The discussion has all been addressing Paul's comment in reply, which was really about a pretty different issue :)

(Though maybe you also think my suggestion of linking to best practices for testing stubs is a bad idea ;)

docs/source/typing_anti_pitch.rst
hauntsaninja: I added a separate mention of this and made sure to clarify the limits. This could be useful for maintainers whose main objection is readability

docs/source/typing_anti_pitch.rst
hauntsaninja: Sorry, missed that amongst the long replies in this thread, linked https://github.com/python/typing/pull/1477/commits/2836871dd559f0d7034407be36a9ba5fd2e0417d

hauntsaninja:
Thanks all! I think I got all the feedback; feel free to open a PR if there's something I missed :-)

PR #1482, opened 2023-10-04
Add .readthedocs.yaml

JelleZijlstra:


hauntsaninja (approved):
Ah, thanks, this sank a little low in my stack

JelleZijlstra:
I just made the same change to pyanalyze too, will confirm it works there before merging this PR and the one on typing-extensions. We might need this for mypy or Black too? I'll check.

hauntsaninja:
mypy's got sorted out, looks like Black might be fine too

PR #1512, opened 2023-11-26
Fix formatting in typing_anti_pitch.rst

MegaIng:
On readthedocs, the top quote has a scrollbar. I hope moving the `the` causes it to no longer be there, otherwise I would suggest not having this as a codeblock (it's kind of a misuse of codeblocks anyway). I also removed the python code highlighting.

The first line of the first bulletpoint is actually highlighted as code right now. This makes it very confusing to read.

### DIFF
--- a/docs/source/typing_anti_pitch.rst
+++ b/docs/source/typing_anti_pitch.rst
@@ -5,8 +5,8 @@ Reasons to avoid static type checking
 
 In the words of :pep:`484`::
 
-    It should also be emphasized that Python will remain a dynamically typed language, and the
-    authors have no desire to ever make type hints mandatory, even by convention.
+    It should also be emphasized that Python will remain a dynamically typed language, and
+    the authors have no desire to ever make type hints mandatory, even by convention.
 
 The idea that dynamism in Python is a strength of the language is reflected in the fact that
 Python's type system is gradual. See :pep:`483` for details, but the long and short of this is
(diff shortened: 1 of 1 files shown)


docs/source/typing_anti_pitch.rst
```diff
@@ -3,10 +3,12 @@
 Reasons to avoid static type checking
 =====================================
 
-In the words of :pep:`484`::
+In the words of :pep:`484`:
 
-    It should also be emphasized that Python will remain a dynamically typed language, and the
-    authors have no desire to ever make type hints mandatory, even by convention.
+.. code-block:: none
+
```
JelleZijlstra, line 9:
```suggestion
```
The correct formatting for block quotes is simply an indented block (https://docutils.sourceforge.io/docs/user/rst/quickref.html#block-quotes).

Looks like this:
Screenshot 2023-11-26 at 4 53 10 PM

MegaIng: Yeah, I am aware of that. But the original was using a codeblock, so I wanted to preserve this while fixing the scrollbar and syntax highlighting. If you guys want this changed, we can do that.

JelleZijlstra: I suppose @hauntsaninja can confirm but I doubt that was intentional. The text is a quote; we should use the appropriate formatting for quotes.

PR #1517, opened 2023-11-29
Add the typing spec

JelleZijlstra:
Copied over from https://github.com/JelleZijlstra/typing-spec

The text largely derives from the typing PEPs, though I added some
material. The organization is based on an outline by @erictraut.
@rchen152 and @daverball made contributions to the text.

This was previously discussed on https://discuss.python.org/t/proposed-initial-typing-spec/39389. I'll submit an issue to the Typing Council to request its imprimatur.


docs/spec/dataclasses.rst
```diff
@@ -0,0 +1,517 @@
...
+.. code-block:: python
+
+  _T = TypeVar("_T")
+  
+  def dataclass_transform(
+      *,
+      eq_default: bool = True,
+      order_default: bool = False,
+      kw_only_default: bool = False,
+      field_specifiers: tuple[type | Callable[..., Any], ...] = (),
```
sobolevn, line 154: New python versions also have `frozen_default`, should it be included?

docs/spec/directives.rst
```diff
@@ -0,0 +1,135 @@
...
+
+Type checkers are expected to understand simple version and platform
+checks, e.g.::
+
+  import sys
+
+  if sys.version_info[0] >= 3:
+      # Python 3 specific definitions
+  else:
+      # Python 2 specific definitions
```
sobolevn, line 127: Let's maybe use some more modern example with python3.12+ and 3.11-?

docs/spec/typeddict.rst
```diff
@@ -0,0 +1,637 @@
...
+* The class body should only contain lines with item definitions of the
+  form ``key: value_type``, optionally preceded by a docstring.  The
+  syntax for item definitions is identical to attribute annotations,
+  but there must be no initializer, and the key name actually refers
+  to the string value of the key instead of an attribute name.
+
+* Type comments cannot be used with the class-based syntax, for
+  consistency with the class-based ``NamedTuple`` syntax.  (Note that
+  it would not be sufficient to support type comments for backwards
+  compatibility with Python 2.7, since the class definition may have a
```
sobolevn, line 62: I think we can drop any 2.7 compatibility notes.

docs/spec/dataclasses.rst
JelleZijlstra: Good point, we should add that. (I'm also missing PEP 702, which I think got accepted after I started working on the spec).

JelleZijlstra:
Addressed @sobolevn's comments and added `@deprecated` from PEP-702.

hauntsaninja (commented):
One minor thing for later is we have a bunch of links to PEPs that should maybe become internal links

JelleZijlstra:
@hauntsaninja I agree, but won't have time right now to fix that. As you say, let's leave it for later.

docs/spec/directives.rst
```diff
@@ -0,0 +1,264 @@
...
+
+Don't expect a checker to understand obfuscations like
+``"".join(reversed(sys.platform)) == "xunil"``.
+
+``@deprecated``
+---------------
+
+(Originally specified in :pep:`702`.)
+
+The :py:func:`warnings.deprecated`
```
Viicos, line 142:
```suggestion
The :class:`warnings.deprecated`
```
As this is now implemented as a class?

JelleZijlstra: It's not documented as a class: https://docs.python.org/3.13/library/warnings.html#warnings.deprecated

Viicos: Ah yes makes sense, it isn't auto-generated docs

docs/spec/Makefile
```diff
@@ -0,0 +1,20 @@
+# Minimal makefile for Sphinx documentation
```
gvanrossum, line 1: Why is this here? I went into this directory and typed `make venv` but it complained about a missing `conf.py`.

docs/spec/concepts.rst
```diff
@@ -0,0 +1,95 @@
+Type system concepts
```
gvanrossum, line 1: This feels like a grab-bag section, and it's oddly placed at the start of the document (why start with unions, when nothing else has been described yet)?

docs/spec/historical.rst
```diff
@@ -0,0 +1,291 @@
+.. _historical:
```
gvanrossum, line 1: Maybe add `from __future__ import annotations` to this page? I was reminded of it in the section about forward references. I think it's within the scope of the typing spec, since it changes semantics: even if you don't use runtime introspection, enabling it changes what gets accepted.

gvanrossum (commented):
I'm not sure how precise a review you're expecting. I'm okay if this goes in right away (we can iterate on everything), but here are some thoughts that came up when I skimmed it.

docs/spec/Makefile
JelleZijlstra: Oops, probably a holdover from when it was in its own repo. I'll get rid of it. (You should `make` in the parent directory, which includes this spec and the other typing docs.)

docs/spec/concepts.rst
JelleZijlstra: This comes from Eric's outline in https://discuss.python.org/t/proposed-initial-typing-spec/39389/4, but I omitted the sections for which we don't have any text yet. I'm open to reorganizing this, but we can do that later too.

docs/spec/historical.rst
JelleZijlstra: I'm not sure it does; shouldn't type checkers infer the same types with and without the future import? Obviously the future import affects the runtime behavior, but this spec doesn't need to describe that.

docs/spec/Makefile
gvanrossum: Yeah, I figured that out eventually. It's a bit weird how clicking on the Table of Contents header at the top of the sidebar takes you to the global ToC, not the Spec ToC, even though there is no "Next topic" linked from spec/historical.html, the last topic in Spec (so I would expect that the ToC link would go to the Spec ToC). But that's a very minor detail at this point, and probably some Sphinx oddity.

JelleZijlstra:
> I'm not sure how precise a review you're expecting. I'm okay if this goes in right away (we can iterate on everything), but here are some thoughts that came up when I skimmed it.

Thanks! I think a general review is enough; this text is so big that we can't hope to get it all right, and I primarily want this in so we can start iterating on it and improving it.

docs/spec/concepts.rst
gvanrossum: Later is fine.

docs/spec/historical.rst
gvanrossum: Huh, I suppose. Does the spec not describe *any* runtime behavior?

JelleZijlstra: It does in a few cases, but in principle I think it shouldn't. docs.python.org describes how Python the language works; this spec should only describe the semantics of the type system.

PR #1540, opened 2023-12-12
Discussions moved to discourse

srittau:
Also clarify that the purpose of the issues


JelleZijlstra (approved):
We should also mention the spec somehow in the README, I'll send another PR for that.

PR #1541, opened 2023-12-12
Add "Modernizing Superseded Typing Features" guide

srittau:


srittau:
If this guide is accepted, we can also clean up the "Best Practices" document and cross-link to the modernizing guide.

docs/source/modernizing.rst
```diff
@@ -0,0 +1,319 @@
...
+    def f(x: str) -> str: ...
+
+.. _modernizing-typed-dict:
+
+``typing.TypedDict`` Legacy Forms
+=================================
+
+*Alternative available since:* Python 3.6
+
+:class:`TypedDict <typing.TypedDict>` supports two legacy forms for
```
JelleZijlstra, line 104: We can add that the first legacy form is still useful if you have keys that are not valid identifiers.

docs/source/modernizing.rst
```diff
@@ -0,0 +1,319 @@
...
+* :class:`typing.MappingView` (→ :class:`collections.abc.MappingView`)
+* :class:`typing.KeysView` (→ :class:`collections.abc.KeysView`)
+* :class:`typing.ItemsView` (→ :class:`collections.abc.ItemsView`)
+* :class:`typing.ValuesView` (→ :class:`collections.abc.ValuesView`)
+* :class:`typing.ContextManager` (→ :class:`contextlib.AbstractContextManager`)
+* :class:`typing.AsyncContextManager` (→ :class:`contextlib.AbstractAsyncContextManager`)
+* :class:`typing.Pattern` (→ :class:`re.Pattern`)
+* :class:`typing.Match` (→ :class:`re.Match`)
+
+The following types were made generic in Python 3.12:
```
JelleZijlstra, line 198: Not sure what change this refers to. It's not generic even on main:

```
% ./python.exe 
Python 3.13.0a2+ (heads/main:16448cab44, Dec  6 2023, 14:43:22) [Clang 14.0.0 (clang-1400.0.29.202)] on darwin
>>> from typing import Hashable
>>> Hashable[int]
    Hashable[int]
    ~~~~~~~~^^^^^
  File "/Users/jelle/py/cpython/Lib/typing.py", line 378, in inner
    return func(*args, **kwds)
           ~~~~^^^^^^^^^^^^^^^
  File "/Users/jelle/py/cpython/Lib/typing.py", line 1423, in __getitem__
    _check_generic(self, params, self._nparams)
    ~~~~~~~~~~~~~~^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/Users/jelle/py/cpython/Lib/typing.py", line 295, in _check_generic
    raise TypeError(f"{cls} is not a generic class")
TypeError: typing.Hashable is not a generic class
```

docs/source/modernizing.rst
```diff
@@ -0,0 +1,319 @@
...
+    class Reg[T](int): ...
+
+.. _modernizing-byte-string:
+
+``typing.ByteString``
+=====================
+
+*Alternative available since:* Python 3.0; Python 3.12, typing-extensions
+
+:class:`ByteString <typing.ByteString>` was originally intended to be a type
```
JelleZijlstra, line 308: What exactly this meant is very unclear. This text is probably fine though.

JelleZijlstra (commented):
Thanks! A few quick comments.

srittau:
Maybe it's a better idea to name this "Modernizing Superseded Typing Features", as many of those features are not technically deprecated.

docs/source/modernizing.rst
```diff
@@ -0,0 +1,336 @@
...
+**************************************
+Modernizing Superseded Typing Features
+**************************************
+
+Introduction
+============
+
+This guide helps to modernize your code by replacing older typing features
+with their modern equivalents. Not all features described here are obsolete,
+but they superseded by more modern alternatives, that are recommended to use.
```
erictraut, line 17: "they superseded" => "they are superseded"

docs/source/modernizing.rst
```diff
@@ -0,0 +1,336 @@
...
+
+    type IntList = list[int]
+
+Code supporting older Python versions should use
+:data:`TypeAlias <typing.TypeAlias>`, introduced in Python 3.10, but also
+available in typing-extensions, instead::
+
+    from typing import TypeAlias  # or typing_extensions.TypeAlias
+
+    IntList: TypeAlias = list[int]
```
TeamSpen210, line 273: This might want to also mention `typing_extensions.TypeAliasType`, which allows providing a name and explicit typevar order. That’s newer and not all checkers support it yet though.

srittau: @TeamSpen210 As I'm not terribly familiar with `TypeAliasType` and the new `type` syntax (yet), could you make a suggestion how we can word this?

Daverball:
```suggestion
    IntList: TypeAlias = list[int]
    
There are cases where `TypeAlias` is insufficient to express what you can express
using :keyword:`type`, such as when you want the order of type parameters to
be different from how they appear in the body of the alias.

In this case you can use `typing_extensions.TypeAliasType` to express this alias
in older versions of Python::

    from typing_extensions import TypeAliasType, TypeVar
    
    T = TypeVar("T", infer_variance=True)
    ListOrTuple = TypeAliasType("ListOrTuple", list[T] | tuple[T, ...], type_params=(T, ))
    
This is equivalent to::

    type ListOrTuple[T] = list[T] | tuple[T, ...]
    
Note however that `TypeAliasType` is newer and not yet supported by every type checker.
```

Maybe something like this? I couldn't think of a good example where you'd actually want to change the order of type parameters, but it at least conveys how to use `TypeAliasType` to be equivalent with a `type` keyword.

srittau: Thanks, I like your text, but I'd like to see an example where this could be useful. I want this document to be a practical guide, so keeping it concise is one of the goals.

Daverball: That's fair, until mypy supports `TypeAliasType` there might not be much point.

I think there's a potential use-case for using an unbound `TypeVar` as a slightly more strict gradual type compared to `Any`. With the old `TypeAlias` it's not possible to have unbound type params unless you pass them back in explicitly, so that could simplify some code, but pyright does not appear to allow unbound type params inside PEP695 type aliases and just treats it as a type parameter anyways if you ignore the error, so you still have to do it the old way and always pass in the type parameter into the alias to leave it unbound.

To illustrate what I mean:
```
from collections.abc import Callable
from typing import TypeVar

class Foo:
    pass

class Bar(Foo):
    pass

FooT = TypeVar("FooT", infer_variance=True, bound=Foo)

type AcceptsFoo = Callable[[FooT], object]

# I don't have to specify a type param here, because `AcceptsFoo` takes none
def foo(x: AcceptsFoo) -> None:
    return None
    
def accepts_foo(x: Foo) -> None:
    return None
    
def accepts_bar(x: Bar) -> None:
    return None

foo(accepts_foo)
# This would be rejected if FooT was actually bound
# since FooT should be contravariant, but since we
# leave it unbound it behaves like a gradual type
foo(accepts_bar)
# but this will be rejected since `int` is not a subclass of `Foo`
foo(int)
```

I think this is a better use-case than the explicit order of type params, but since there's no type checker support for it yet, I can't really use it as a motivating example.

TeamSpen210: Another reason to use TypeAliasType is for at runtime. Since it's an instance of that type and not just an assignment, using it in type annotations will preserve the alias when introspected.

erictraut: > ...but since there's no type checker support for [using an unbound type variable] yet...

Using a type variable that is not bound to a scope is expressly forbidden (both in PEP 484 and in PEP 695), so this will never be supported.

Daverball: @erictraut That's fair, but if you modify the above example to this:
```python
from collections.abc import Callable
from typing import TypeVar

class Foo:
    pass

class Bar(Foo):
    pass

FooT = TypeVar("FooT", infer_variance=True, bound=Foo)

type AcceptsFoo[FooT] = Callable[[FooT], object]

def foo(x: AcceptsFoo[FooT]) -> None:
    return None
    
def accepts_foo(x: Foo) -> None:
    return None
    
def accepts_bar(x: Bar) -> None:
    return None

foo(accepts_foo)
foo(accepts_bar)
foo(int)
```
It will work without raising any errors (beyond the last line) and will work in the way I described it, even though `FooT` remains unbound in `foo`, since it only appears in the function signature once.

To me both of those cases should be semantically equivalent. Using a singular `TypeVar` in functions is a common trick to achieve a gradual type like `Any` (i.e. it will match in either direction) but that will reject types that don't match the upper bound. But this trick tends to create quite verbose function signatures, which could be avoided if the above were allowed.

srittau:
Are there any more comments? Otherwise I'd like to merge this.

docs/source/modernizing.rst
```diff
@@ -0,0 +1,336 @@
...
+**************************************
+Modernizing Superseded Typing Features
+**************************************
+
+Introduction
+============
+
+This guide helps to modernize your code by replacing older typing features
+with their modern equivalents. Not all features described here are obsolete,
+but they are superseded by more modern alternatives, that are recommended to use.
```
JelleZijlstra, line 17:
```suggestion
but they are superseded by more modern alternatives, which are recommended to use.
```

docs/source/modernizing.rst
```diff
@@ -0,0 +1,336 @@
...
+.. note::
+
+    The latest version of typing-extensions is available for all Python
+    versions that have not reached their end of life, but not necessarily for
+    older versions.
+
+.. note::
+
+    :python:`from __future__ import annotations` is available since Python 3.7.
+    This will only work inside type annotations, while quoting is still
```
JelleZijlstra, line 36:
```suggestion
    This only has an effect inside type annotations, while quoting is still
```

docs/source/modernizing.rst
```diff
@@ -0,0 +1,336 @@
...
+* :class:`typing.AsyncIterator` (→ :class:`collections.abc.AsyncIterator`)
+* :class:`typing.AsyncGenerator` (→ :class:`collections.abc.AsyncGenerator`)
+* :class:`typing.Iterable` (→ :class:`collections.abc.Iterable`)
+* :class:`typing.Iterator` (→ :class:`collections.abc.Iterator`)
+* :class:`typing.Generator` (→ :class:`collections.abc.Generator`)
+* :class:`typing.Reversible` (→ :class:`collections.abc.Reversible`)
+* :class:`typing.Container` (→ :class:`collections.abc.Container`)
+* :class:`typing.Collection` (→ :class:`collections.abc.Collection`)
+* :data:`typing.Callable` (→ :class:`collections.abc.Callable`)
+* :class:`typing.AbstractSet` (→ :class:`collections.abc.Set`)
```
JelleZijlstra, line 187:
```suggestion
* :class:`typing.AbstractSet` (→ :class:`collections.abc.Set`), note the change in name
```

docs/source/modernizing.rst
```diff
@@ -0,0 +1,336 @@
...
+* :class:`typing.MutableMapping` (→ :class:`collections.abc.MutableMapping`)
+* :class:`typing.Sequence` (→ :class:`collections.abc.Sequence`)
+* :class:`typing.MutableSequence` (→ :class:`collections.abc.MutableSequence`)
+* :class:`typing.ByteString` (→ :class:`collections.abc.ByteString`), but see :ref:`modernizing-byte-string`
+* :class:`typing.MappingView` (→ :class:`collections.abc.MappingView`)
+* :class:`typing.KeysView` (→ :class:`collections.abc.KeysView`)
+* :class:`typing.ItemsView` (→ :class:`collections.abc.ItemsView`)
+* :class:`typing.ValuesView` (→ :class:`collections.abc.ValuesView`)
+* :class:`typing.ContextManager` (→ :class:`contextlib.AbstractContextManager`)
+* :class:`typing.AsyncContextManager` (→ :class:`contextlib.AbstractAsyncContextManager`)
```
JelleZijlstra, line 199:
```suggestion
* :class:`typing.ContextManager` (→ :class:`contextlib.AbstractContextManager`), note the change in name
* :class:`typing.AsyncContextManager` (→ :class:`contextlib.AbstractAsyncContextManager`), note the change in name
```

JelleZijlstra (approved):

PR #1550, opened 2023-12-25
Allow unary + in Literal integers

dmoisset:
As discussed in https://discuss.python.org/t/signed-values-in-literal-type/39919

### DIFF
--- a/docs/spec/literal.rst
+++ b/docs/spec/literal.rst
@@ -168,8 +168,8 @@ The following parameters are intentionally disallowed by design:
     ``Literal[4j]`` that contain just a single complex number are also
     prohibited.
 
-  - The only exception to this rule is the unary ``-`` (minus) for ints: types
-    like ``Literal[-5]`` are *accepted*.
+  - The only exceptions to this rule are the unary ``-`` (minus) and unary ``+`` (plus) for ints: types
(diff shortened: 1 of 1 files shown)


JelleZijlstra (approved):
Looks good, but it would be good to get an implementation in at least one type checker. It should be trivial, but as a general principle I'd like the spec to only specify behavior that has actually been implemented.

While reviewing the change I was thinking about whether we should specify what `+` means. My conclusion is that it's fine to leave it implicit that it means the same thing as it means at runtime.

JelleZijlstra:
I wrote an implementation for mypy, which was trivial as expected: python/mypy#16728.

erictraut (approved):

erictraut:
Oh, one thing I just thought about. Should a type checker explicitly reject redundant unary `+` or `-` operators in a literal type? Example: `Literal[+++++-----1]`. Do we care one way or another?

JelleZijlstra:
Good point. I'd say any number of `+`/`-` should be accepted as it's unambiguous, easy to implement, and matches the runtime behavior.

gvanrossum (approved):

rchen152 (approved):
Looks good to me. I think it's fine to leave the edge case of whether multiple `+`/`-` are allowed unspecified.

PR #1552, opened 2023-12-26
Initial check-in of type specification conformance suite.

erictraut:
This is a first cut at a conformance test suite for the typing specification. It builds on earlier work done by @hauntsaninja. The test suite is still very incomplete, but it provides a framework that should allow us to rapidly build out the missing tests.


conformance/README.md
```diff
@@ -0,0 +1,61 @@
...
+
+## Updating A Test Case
+
+If a test is updated (augmented or fixed), the process is similar to when adding a new test. Run the tool to generate new results and manually examine the output of each supported type checker. Then update the conformance status accordingly. Once the conformance status has been updated, re-run the test tool again to regenerate the summary report.
+
+## Updating a Type Checker
+
+If a new version of a type checker is released, re-run the test tool with the new version. If the type checker output has changed for any test cases, the tool will supply the old and new outputs. Examine these to determine whether the conformance status has changed. Once the conformance status has been updated, re-run the test tool again to regenerate the summary report.
```
debonte, line 57: Same
```suggestion
If a new version of a type checker is released, re-run the test tool with the new version. If the type checker output has changed for any test cases, the tool will supply the old and new outputs. Examine these to determine whether the conformance status has changed. Once the conformance status has been updated, re-run the test tool to regenerate the summary report.
```

conformance/src/type_checker.py
```diff
@@ -0,0 +1,213 @@
...
+        raise NotImplementedError
+
+
+class MypyTypeChecker(TypeChecker):
+    @property
+    def name(self) -> str:
+        return "mypy"
+
+    def install(self) -> None:
+        run("pip install mypy --upgrade", shell=True)
```
hauntsaninja, line 51: I find it's often more bulletproof to use `{sys.executable} -m pip`

conformance/src/type_checker.py
```diff
@@ -0,0 +1,213 @@
...
+    def install(self) -> None:
+        run("pip install pytype --upgrade", shell=True)
+
+    def get_version(self) -> str:
+        proc = run("pytype --version", stdout=PIPE, text=True, shell=True)
+        version = proc.stdout.strip()
+        return f'pytype {version}'
+
+    def run_tests(self, test_files: Sequence[str]) -> dict[str, str]:
+        command = "pytype -V 3.11 -k *.py"
```
hauntsaninja, line 164: Intentional that this isn't 3.12?

hauntsaninja (commented):
Thank you for working on this!

conformance/src/type_checker.py
erictraut: Thanks, I've made this change throughout.

conformance/tests/generics_self_advanced.py
```diff
@@ -0,0 +1,46 @@
...
+
+    @classmethod
+    def method1(cls) -> Self:
+        ...
+
+class ChildB(ParentB):
+    b: int = 0
+
+    def method2(self) -> None:
+        assert_type(self, Self)
```
hauntsaninja, line 35: Is this actually specified? The spec says:

> One implementation strategy is to simply desugar the former to the latter in a preprocessing step. If a method uses Self in its signature, the type of self within a method will be Self. In other cases, the type of self will remain the enclosing class.

Same question for the classmethod variation

hauntsaninja (commented):
Was reading through many of the test cases, had one question come up

conformance/tests/generics_self_advanced.py
erictraut: The spec may not say this explicitly (it probably should), but I don't know how you could correctly implement the `Self` type if an unannotated `self` doesn't have this type.

Consider this example, which I think should type check without errors given the spec.

```python
class Foo:
    def method1(self) -> Self:
        reveal_type(self)
        return self

    def method2(self) -> Any:
        reveal_type(self)
        x: Self = self
        return x
```

erictraut: If your reading of the spec is different, then let's ignore (or comment out) this part of the conformance test for now and queue the topic for discussion.

hauntsaninja: Your example should type check and I'm fine with specifying this behaviour. But yes, my reading of the current spec is different. In general, I think the construction of the conformance suite should be used as a way to iron out ambiguity in the spec.

https://github.com/python/mypy/issues/14075 is the relevant mypy issue

conformance/requirements.txt
```diff
@@ -0,0 +1,6 @@
+tomli
+tomlkit
+pyright
+mypy
+pyre-check
+pytype
```
AlexWaygood, line 6: To what extent is it a goal or non-goal for this conformance test suite to be runnable on Windows? I know pytype can't be run on Windows; I think maybe pyre can't either? Obviously Windows devs won't be able to run the conformance tests for pyre and pytype locally, but it would be nice to be able to at least run the mypy/pyright tests locally on Windows.

Having pytype in the `requirements.txt` file unqualified means `pip install -r conformance/requirements.txt` fails with a very long and ugly error message on Windows. It's pretty easy to stop pip from trying to install pytype on Windows; you'd just need to do this (but you'd also need to make sure the script doesn't try to run the pytype tests if pytype isn't installed):

```suggestion
pyre-check
pytype; platform_system != "Windows"
```

If it's just not a goal for this conformance test suite to be runnable on Windows at all, it might be nice to add a note to that effect in the documentation

AlexWaygood: Thanks!

hauntsaninja (approved):

hauntsaninja:
Merging, as discussed. Feedback on the structure or initial set of tests still very welcome :-)

conformance/.gitignore
```diff
@@ -0,0 +1,47 @@
+# Byte-compiled / optimized / DLL files
```
JelleZijlstra, line 1: This could probably go in a global .gitignore instead.

conformance/README.md
```diff
@@ -0,0 +1,71 @@
+# Type System Conformance
+
+## Motivation
+
+[PEP 729](https://peps.python.org/pep-0729/) provides a structured and documented way to specify and evolve the Python type system. In support of this effort, an official [Python typing spec](https://github.com/python/typing/tree/main/docs/spec) has been drafted. This spec consolidates details from various historical typing-related PEPs. The spec will be modified over time to clarify unspecified and under-specified parts of the type system. It will also be extended to cover new features of the type system.
```
JelleZijlstra, line 5: Link to the rendered version: https://typing.readthedocs.io/en/latest/spec/index.html

conformance/requirements.txt
```diff
@@ -0,0 +1,7 @@
+tomli
```
JelleZijlstra, line 1: I read earlier that we require 3.12, so shouldn't we be able to use the stdlib's tomllib module instead?

conformance/src/type_checker.py
```diff
@@ -0,0 +1,244 @@
...
+
+            # Generate a default config file.
+            pyre_config = (
+                '{"site_package_search_strategy": "pep561", "source_directories": ["."]}\n'
+            )
+            with open(".pyre_configuration", "w") as f:
+                f.write(pyre_config)
+
+            return True
+        except:
```
JelleZijlstra, line 148: Avoid raw except

JelleZijlstra (commented):
Thanks for putting this together! It's a great start.

I have a few small comments that I'll send a PR to fix when I get a chance.

One bigger area that doesn't seem ideal is that we're checking in the exact text of each type checker's output.

An idea for an alternative approach:
- In test cases, we write a magic comment "# E" or "# E: explanation for why this is an error" at every place where there should be an error
- For each type checker, we parse the output to extract the list of line numbers
- Then we compare the set of lines where there should be an error with the set where the type checker actually emitted an error.

This should remove some of the manual work in updating the .toml files.

conformance/.gitignore
erictraut: > One bigger area that doesn't seem ideal is that we're checking in the exact text of each type checker's output.

I agree it's not ideal, but I don't see a good alternative. Type checkers report the same error on different lines. Unless we mandate exactly where an error is reported (which is probably not viable), there will be many differences. Plus, type checkers sometimes report additional errors or warnings unrelated to the functionality being tested. I don't see a way around "manual scoring". That's why I think it's important to keep each test file relatively small and targeted.

rchen152:
Sorry for taking so long to look at this! It looks good to me, with the caveat that I didn't carefully go through all of the individual tests and conformance results - just read a couple and skimmed the rest. I'll see what I can do to speed up pytype. The main thing I can think of is that we should probably cache/reuse pieces so that pytype isn't doing things like reparsing all of typeshed every time it runs.

PR #1553, opened 2023-12-30
Add conformance tests for basic generic spec

hauntsaninja:


docs/spec/generics.rst
```diff
@@ -272,7 +272,7 @@ However, there are some special cases in the static typechecking context:
         # this is also an error
         an_attr: list[S] = []
 
-        def do_something(x: S) -> S:  # this is OK though
+        def do_something(self, x: S) -> S:  # this is OK though
```
erictraut, line 275: Good catch!

conformance/tests/generics_basic.py
```diff
@@ -0,0 +1,154 @@
...
+    concat(b, b)  # OK
+    concat(s, b)  # Type error
+    concat(b, s)  # Type error
+
+    concat(s, a)  # OK
+    concat(a, b)  # OK
+
+# > Specifying a single constraint is disallowed.
+
+BadConstraint1 = TypeVar('BadConstraint', str)  # Type error
```
erictraut, line 44: I recommend changing the quoted name (the first argument) to match the assigned name "BadConstraint1" to avoid an unnecessary (and unrelated) error here.

conformance/tests/generics_basic.py
```diff
@@ -0,0 +1,154 @@
...
+
+# > Note: those types cannot be parameterized by type variables
+
+BadConstraint2 = TypeVar('BadConstraint', str, T)  # Type error
```
erictraut, line 48: Same thing here — change the quoted name for consistency.

conformance/tests/generics_basic.py
```diff
@@ -0,0 +1,154 @@
...
+
+class MyStr(str): ...
+
+def test_concat_subtype(s: str, b: bytes, a: Any, m: MyStr) -> None:
+    assert_type(concat(m, m), str)
+    assert_type(concat(m, s), str)
+    concat(m, b)  # Type error
+
+    # TODO: should these be str or Any?
+    # reveal_type(concat(m, a))
```
erictraut, line 62: This gets into constraint-solving behavior which is underspecified, but I'm still surprised to see that mypy generates an `Any` in this case. That's not what I would expect, and it seems inconsistent with mypy's handling of non-constrained TypeVars.

conformance/tests/generics_basic.py
```diff
@@ -0,0 +1,154 @@
...
+class Pair2(Generic[T, T]):   # Type error
+      ...
+
+# > The ``Generic[T]`` base class is redundant in simple cases where you
+# > subclass some other generic class and specify type variables for its
+# > parameters.
+
+from collections.abc import Iterator
+
+class MyIter1(Iterator[T]):
```
erictraut, line 116: This results in a "derives from protocol but doesn't implement all methods" error, which is unrelated to the test case. To reduce the noise, I recommend fixing this error — either by choosing a non-protocol base class or adding a dummy implementation for `__next__`. 

I think that reducing noise is important in these tests. They're difficult enough to "score" even when there's no additional noise in the output.

conformance/tests/generics_basic.py
```diff
@@ -0,0 +1,154 @@
...
+    ...
+
+class MyIter2(Iterator[T], Generic[T]):
```
erictraut, line 119: Same here.

conformance/tests/generics_basic.py
```diff
@@ -0,0 +1,154 @@
...
+
+def test_my_iter(m1: MyIter1[int], m2: MyIter2[int]):
+    assert_type(next(m1), int)
+    assert_type(next(m2), int)
+
+# > You can use multiple inheritance with ``Generic``
+
+from collections.abc import Sized, Container
+
+class LinkedList(Sized, Generic[T]):
```
erictraut, line 130: Same here

conformance/tests/generics_basic.py
```diff
@@ -0,0 +1,154 @@
...
+    ...
+
+K = TypeVar('K')
+V = TypeVar('V')
+
+class MyMapping(Iterable[tuple[K, V]], Container[tuple[K, V]], Generic[K, V]):
```
erictraut, line 136: Same here

conformance/tests/generics_basic.py
```diff
@@ -0,0 +1,154 @@
...
+    ...
+
+# > Subclassing a generic class without specifying type parameters assumes
+# > ``Any`` for each position.  In the following example, ``MyIterable``
+# > is not generic but implicitly inherits from ``Iterable[Any]``::
+
+class MyIterableAny(Iterable):  # Same as Iterable[Any]
```
erictraut, line 143: Same here

conformance/tests/generics_basic.py
```diff
@@ -0,0 +1,154 @@
...
+
+def test_my_iterable_any(m: MyIterableAny):
+    assert_type(iter(m), Iterator[Any])
+
+# > Generic metaclasses are not supported
+
+class GenericMeta(type, Generic[T]): ...
+
+class GenericMetaInstance(metaclass=GenericMeta[T]):  # Type error
+    ...
```
erictraut, line 154: The spec doesn't explicitly state it, but it implies that the ordering of type parameters is normally dictated by the order in which they appear, but this can be overridden by an explicit ordering in a `Generic`. I recommend adding a test that validates this. I added a similar test case in the [protocols_generic.py](https://github.com/python/typing/blob/606fef43ca8df5818ca85278ab60549a82b44a40/conformance/tests/protocols_generic.py#L21) test.

conformance/tests/generics_scoping.py
```diff
@@ -0,0 +1,80 @@
+# Specification: https://typing.readthedocs.io/en/latest/spec/generics.html#scoping-rules-for-type-variables
+
+# TODO: write PEP 695 versions
```
erictraut, line 3: These are already written (see [here](https://github.com/python/typing/blob/main/conformance/tests/generics_syntax_scoping.py)), so you can remove this comment.

conformance/tests/generics_scoping.py
```diff
@@ -0,0 +1,80 @@
...
+# > A type variable used in a method of a generic class that coincides
+# > with one of the variables that parameterize this class is always bound
+# > to that variable.
+
+class MyClass(Generic[T]):
+    def meth_1(self, x: T) -> T: ...  # T here
+    def meth_2(self, x: T) -> T: ...  # and here are always the same
+
+a: MyClass[int] = MyClass()
+a.meth_1(1)
```
erictraut, line 26: Maybe add an "# OK" comment to this line. I find that it's easier to score the test if there are explicit "OK" and "Type error" comments.

conformance/tests/generics_scoping.py
```diff
@@ -0,0 +1,80 @@
...
+
+def fun_4(x: T) -> list[T]:
+    a_list: list[T] = []  # OK
+
+    class MyGeneric(Generic[T]):  # Type error
+        ...
+
+    return a_list
+
+# > A generic class nested in another generic class cannot use the same type
```
erictraut, line 68: The spec doesn't talk about generic type aliases, but they should follow the same rules here for scoping. I recommend adding tests for these even though the spec doesn't explicitly mention them. If you would prefer to augment the spec first and then add the tests, that's fine too, but it would be good in that case to record this as something that needs to be revisited.

conformance/tests/generics_scoping.py
```diff
@@ -0,0 +1,80 @@
...
+# > variables. The scope of the type variables of the outer class
+# > doesn't cover the inner one
+
+class Outer(Generic[T]):
+    class Bad(Iterable[T]):  # Type error
```
erictraut, line 73: `Bad` and `Inner` generate a "derives from protocol but doesn't implement all methods" error. Like above, I recommend fixing this to cut down on the noise.

conformance/tests/generics_scoping.py
```diff
@@ -0,0 +1,80 @@
...
+        ...
+    class AlsoBad:
+        x: list[T]  # Type error
+
+    class Inner(Iterable[S]):  # OK
+        ...
+    attr: Inner[T]  # OK
```
erictraut, line 80: Another good test for scoping is to ensure that a bar TypeVar used in the global scope generates an error.

```python
global_var1: T  # Type error
global_var2: list[T] = [] # Type error
list[T]() # Type error
```

conformance/tests/generics_basic.py
erictraut, line 154: It may also be useful to add tests for accessing a TypeVar as a runtime object — to make sure that this isn't disallowed.

```python
assert_type(T.__name__, str)
assert_type(T.__bound__, Any | None)
T.other # Type error

runtime_typevar: TypeVar = T # OK
```

hauntsaninja:
@erictraut I don't think we expect any "derives from protocol but doesn't implement all methods" errors here, or at least I don't see any.

On the following file pyright 1.1.345 has no complaints:
```
from typing import TypeVar
from collections.abc import Iterator
T = TypeVar('T')
class MyIter1(Iterator[T]): ...
MyIter1()
```
mypy, pyre and pytype are all fine with the class definition, but treat it as an abstract class and error when you attempt to instantiate.

I can still add the missing methods, but we should add this to the list of things to exactly specify.

erictraut:
> but treat it as an abstract class and error when you attempt to instantiate

Yeah, that makes sense. After reviewing the text in the [Explicitly Declaring Implementation](https://typing.readthedocs.io/en/latest/spec/protocol.html#explicitly-declaring-implementation) section, I agree that pyright should emit those errors only if it's marked `@final`. I'll make that change.

hauntsaninja:
Okay, this is mergeable. Remaining items:
- https://github.com/python/typing/pull/1553#discussion_r1450680832 I agree mypy's behaviour here is surprising. I'll look into changing it and report back
- https://github.com/python/typing/pull/1553#discussion_r1450705833 These proved controversial, I'll follow up separately on this one

PR #1559, opened 2024-01-02
Make pytype conformance test a little faster.

rchen152:
This change gets the pytype test down to ~30s. It's still considerably slower than the others, but any further improvements will likely require untangling the sad state of affairs that is
https://github.com/google/pytype/issues/1501, which will take some time.

* Runs pytype's 'check' option through pytype.io.check_py to skip the costly type inference that it does by default.
* Adds a progress bar so that it doesn't look like pytype is hanging.

This generates a few changes to results/pytype/, mostly due to error messages being formatted and parsed differently.

### DIFF
--- a/conformance/requirements.txt
+++ b/conformance/requirements.txt
@@ -1,5 +1,6 @@
 tomli
 tomlkit
+tqdm
 pyright
 mypy
 pyre-check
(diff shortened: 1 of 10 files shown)


JelleZijlstra (approved):
Thank you!

PR #1561, opened 2024-01-05
Added more type system conformance tests

erictraut:
* Added conformance tests for `@final` decorator.
* Added conformance tests for `Final` annotation.
* Added conformance tests for @override decorator.
* Added conformance tests for "Type checker directives" section of spec. 
* Added conformance tests for `@overload`.
* Added conformance tests for the "Special types in annotations" section of the typing spec. 
* Added conformance tests for the "Protocols" section of the type specification. 
* Added conformance tests for PEP 695 type parameter functionality.
* Added conformance tests for TypeVarTuple.
* Added conformance tests for ParamSpec.
* Added conformance tests for "Variance" section of Generics chapter.

Added link to conformance results.

Improved error recovery and reporting for test tool if it cannot decode ".toml" file.

Updated pyright test results for latest version (1.1.344). Added tests for Annotated.


conformance/tests/generics_paramspec_semantics.py
```diff
@@ -0,0 +1,143 @@
...
+from typing import Callable, Concatenate, Generic, ParamSpec, TypeVar, assert_type
+
+P = ParamSpec("P")
+
+
+def changes_return_type_to_str(x: Callable[P, int]) -> Callable[P, str]:
+    ...
+
+
+def returns_int(a: str, b: bool, /) -> int:
```
hauntsaninja, line 17: It's a little confusing that use of positional-only means this example doesn't match the spec. Should we remove that / change the spec / just add a comment?

conformance/tests/generics_paramspec_semantics.py
```diff
@@ -0,0 +1,143 @@
...
+        self.f = f
+        self.prop = prop
+
+
+def callback_a(q: int, /) -> str:
+    ...
+
+
+y1 = Y(callback_a, 1)
+assert_type(y1, Y[int, [int]])
```
hauntsaninja, line 81: Spec says `# Should resolve to Y[(q: int), int]` which I think has the wrong ordering

conformance/tests/generics_syntax_declarations.py
```diff
@@ -0,0 +1,82 @@
...
+    assert_type(T, TypeVar)
+    assert_type(Ts, TypeVarTuple)
+    assert_type(P, ParamSpec)
+
+
+class ClassB[T](Generic[T]):  # Runtime error
+    ...
+
+
+class ClassC[S, T](Protocol[S, T]):  # Type error
```
hauntsaninja, line 23: we lost the `class ClassA[S, T](Protocol)  # OK` example

conformance/tests/generics_syntax_infer_variance.py
```diff
@@ -0,0 +1,142 @@
...
+        ...
+
+
+vco3_1: ShouldBeCovariant3[float] = ShouldBeCovariant3[int]()  # OK
+vco3_2: ShouldBeCovariant3[int] = ShouldBeCovariant3[float]()  # Type error
+
+
+@dataclass(frozen=True)
+class ShouldBeCovariant4(Generic[T]):
+    x: T
```
hauntsaninja, line 55: could have a similar test for Final

hauntsaninja (commented):
Looks good, but haven't yet looked at TypeVarTuple stuff

conformance/tests/generics_paramspec_semantics.py
erictraut: The exact signature isn't important for purposes of this test. I used position-only parameters so I could use an `assert_type` with a `Callable` type to verify the results.

conformance/tests/generics_paramspec_basic.py
```diff
@@ -0,0 +1,40 @@
...
+
+def func2(x: Concatenate[int, P]) -> int:  # Type error
+    ...
+
+
+def func3(x: list[P]) -> None:  # Type error
+    ...
+
+
+def func3(x: Callable[[int, str], P]) -> None:  # Type error
```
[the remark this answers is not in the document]
erictraut: Fixed

conformance/tests/generics_paramspec_semantics.py
erictraut: I agree that's a typo in the spec. I included a fix.

conformance/tests/generics_syntax_infer_variance.py
erictraut: Added test with Final.

conformance/tests/generics_syntax_declarations.py
erictraut: Added the missing test.

hauntsaninja (approved):

PR #1570, opened 2024-01-10
Made proposed change to the typing spec with regard to `NoReturn` and…

erictraut:
… where it can be used.

### DIFF
--- a/docs/spec/special-types.rst
+++ b/docs/spec/special-types.rst
@@ -85,25 +85,16 @@ is unreachable and will behave accordingly::
       return 'whatever works'  # Error might be not reported by some checkers
                                # that ignore errors in unreachable blocks
 
-The ``NoReturn`` type is only valid as a return annotation of functions,
-and considered an error if it appears in other positions::
-
-  from typing import NoReturn
-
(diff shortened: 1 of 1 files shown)


sobolevn (approved):
I agree, I have many existing use-cases for `NoReturn` in different contexts. I think that making it not-the-same-as `Never` is aсtually an unneeded complexity.

gvanrossum (approved):

JelleZijlstra (approved):

rchen152 (approved):

hauntsaninja (approved):
Thanks! The phrasing might be read as implying some sort of stylistic preference, curious if that was intentional

erictraut:
> The phrasing might be read as implying some sort of stylistic preference, curious if that was intentional

Not really. The intention was to try to give some background explanation for why there are two ways to spell the same type — and where each has typically been used. If you have a suggestion for how to improve the wording, let me know.

JelleZijlstra:
Merged as this has been approved by all Council members. I think the wording is fine, but we can tweak it in a follow-up PR if desired.

PR #1571, opened 2024-01-10
[spec] Update typeshed language to conform to reality

srittau:
* Describe typeshed as a "project", not a "repository".
* While typeshed still contains third-party stubs, those are usually not distributed with type checkers, but available from PyPI. Update the language accordingly.
* This also means that step 5 of the import resolution ordering should reference the stdlib only.
* Fix the reference to another section.
* Fix typo in section name.

### DIFF
--- a/docs/spec/distributing.rst
+++ b/docs/spec/distributing.rst
@@ -77,8 +77,9 @@ The Typeshed Project
 ^^^^^^^^^^^^^^^^^^^^
 
 The `typeshed project <https://github.com/python/typeshed>`_ contains type
-annotations for the standard library and a selection of third-party libraries
-that don't ship their own type annotations or stubs. Policies regarding the
+stubs for the standard library (vendored or handled specially by type checkers)
+and type stubs for third-party libraries that don't ship their own type information
+(typically distributed via PyPI). Policies regarding the
 stubs collected there are decided separately and described in the project's
 documentation.
 
@@ -239,7 +240,7 @@ resolve modules containing type information:
    bundled with the package SHOULD be used (be they in ``.pyi`` type
    stub files or inline in ``.py`` files).
 
-5. Typeshed (if used) - Provides the stdlib types.
+5. Typeshed - only for modules in the standard library.
 
 If typecheckers identify a stub-only namespace package without the desired module
 in step 3, they should continue to step 4/5. Typecheckers should identify namespace packages
--- a/docs/spec/distributing.rst
+++ b/docs/spec/distributing.rst
@@ -79,7 +79,7 @@ The Typeshed Project
 The `typeshed project <https://github.com/python/typeshed>`_ contains type
 annotations for the standard library and a selection of third-party libraries
 that don't ship their own type annotations or stubs. Policies regarding the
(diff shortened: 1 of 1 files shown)


erictraut (approved):

JelleZijlstra (approved):
Thanks! Let's leave this open for a few days in case other Council members have a different opinion (cc @hauntsaninja @gvanrossum @rchen152), but I don't think this needs to go through the formal decision-making procedure.

There is one typeshed-related issue that maybe we *should* add to the spec: the format of the VERSIONS file. I think that's important to specify because it affects type checker interoperability. However, that should be discussed separately.

rchen152 (approved):
These all look like good changes to me.

docs/spec/distributing.rst
```diff
@@ -73,13 +73,14 @@ Additional notes on stub files:
...
-There is a `shared repository <https://github.com/python/typeshed>`_ where useful stubs are being
-collected.  Policies regarding the stubs collected here are
-decided separately and reported in the repo's documentation.
+The Typeshed Project
+^^^^^^^^^^^^^^^^^^^^
 
+The `typeshed project <https://github.com/python/typeshed>`_ contains type
+annotations for the standard library and a selection of third-party libraries
+that don't ship their own type annotations or stubs. Policies regarding the
+stubs collected there are decided separately and reported in the project's
```
gvanrossum, line 82: reported -> described? "Reporting a policy" somehow doesn't sound right.

gvanrossum (approved):
LG, one wordsmithing nit.

docs/spec/distributing.rst
```diff
@@ -73,13 +73,14 @@ Additional notes on stub files:
...
-
-There is a `shared repository <https://github.com/python/typeshed>`_ where useful stubs are being
-collected.  Policies regarding the stubs collected here are
-decided separately and reported in the repo's documentation.
+The Typeshed Project
+^^^^^^^^^^^^^^^^^^^^
 
+The `typeshed project <https://github.com/python/typeshed>`_ contains type
+annotations for the standard library and a selection of third-party libraries
+that don't ship their own type annotations or stubs. Policies regarding the
```
hauntsaninja, line 81:
```suggestion
The `typeshed project <https://github.com/python/typeshed>`_ contains type
stubs for the standard library (vendored or handled specially by type checkers)
and type stubs for third-party libraries that don't ship their own type information
(typically distributed via PyPI). Policies regarding the
```

docs/spec/distributing.rst
```diff
@@ -238,8 +239,7 @@ resolve modules containing type information:
    bundled with the package SHOULD be used (be they in ``.pyi`` type
    stub files or inline in ``.py`` files).
 
-5. Typeshed (if used) - Provides the stdlib types and several third party
-   libraries.
+5. Typeshed (if used) - Provides the stdlib types.
```
hauntsaninja, line 242:
```suggestion
5. Typeshed - only for modules in the standard library.
```

hauntsaninja (approved):
Thank you, had some wording nits, but don't feel strongly

PR #1573, opened 2024-01-11
Spec: Add "meta" file describing process for changing the spec

JelleZijlstra:
I think this isn't currently very clearly documented. Hopefully this
will be a good place to find it.

### DIFF
--- a/docs/spec/meta.rst
+++ b/docs/spec/meta.rst
@@ -1,0 +1,48 @@
+Meta-topics
+===========
+
+About this specification
+------------------------
+
+This document was created following the acceptance of :pep:`729`
+to serve as a specification for the Python type system. The
+initial text consists of the "Specification" sections of :pep:`484`
+and subsequent typing-related PEPs, pasted together and reorganized.
+This creates a document that encompasses all aspects of the type
+system that have been specified in PEPs, but not necessarily a
+coherent whole. The hope is that incremental improvements will
+be made to this document to make it more coherent and complete.
+
+Changing the specification
+--------------------------
+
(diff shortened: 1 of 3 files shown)


docs/spec/meta.rst
```diff
@@ -0,0 +1,48 @@
...
+- Minor, non-substantive changes can simply be proposed as PRs to
+  the `python/typing <https://github.com/python/typing>`__ repository,
+  and may be merged by anyone with commit access. Such changes may
+  include formatting fixes, linking improvements, etc.
+- Major changes should go through the PEP process, as described in
+  :pep:`1`. What counts as a major change is not precisely defined,
+  but it would generally include any change of a similar magnitude
+  to `previous typing PEPs <https://peps.python.org/topic/typing/>`__.
+- Substantive changes that do not rise to the level of a PEP must
+  be approved by the Typing Council. The procedure is described below.
```
gvanrossum, line 30: The ordering here is a bit weird -- it feels like counting 1, 3, 2.

I noticed because there's a gray area between minor and substantive where the TC needs to approve but no conformance test suite changes are needed (e.g. because we're merely confirming that what happens to work is in fact meant to work rather than undefined behavior where the current crop of checkers all happen to agree).

JelleZijlstra: You're right, I will reorder the list. (I had a reason for this order but it's not worth the confusion.)

gvanrossum (approved):

rchen152 (approved):

hauntsaninja (approved):

PR #1575, opened 2024-01-12
Updated conformance test results to reflect recent change to the rule…

erictraut:
…s about `NoReturn`. The previous version of the spec indicates that type checkers should reject the use of `NoReturn` outside of a return type annotation.


hauntsaninja (approved):

hauntsaninja:
Linking https://github.com/python/typing-council/issues/5

PR #1576, opened 2024-01-12
Updated spec and conformance test to reflect that a type variable tha…

erictraut:
…t is defined as covariant or contravariant should not generate an error when bound to a generic function or type alias. Variance has no meaning in this context, so it should be ignored. This brings the spec in alignment with the behavior of the major type checkers.


hauntsaninja:
Linking https://github.com/python/typing-council/issues/8

conformance/tests/generics_variance.py
```diff
@@ -52,10 +52,7 @@ def dump_employee(e: E) -> E:
 B_co = TypeVar("B_co", covariant=True)
 
 
-# > Variance is only applicable to generic types; generic functions do not
-# > have this property. The latter should be defined using only type variables
-# > without covariant or contravariant keyword arguments.
-def bad_func(x: list[B_co]) -> B_co:  # Type checker error
+def func(x: list[B_co]) -> B_co:  # OK
```
JelleZijlstra, line 55:
```suggestion
# > Variance has no meaning, and should therefore be ignored by type checkers, if
# > a type variable is bound to a generic function or type alias.
def func(x: list[B_co]) -> B_co:  # OK
```
Quoting the new wording in the spec.

Also, could you add some test cases demonstrating the desired behavior around aliases?

erictraut: Added the quote and reran tests.

I think we have these cases covered in the compliance test suite already. They're further down in the same file.

hauntsaninja:
@erictraut should this have been merged before https://github.com/python/typing-council/issues/8 got resolved?

erictraut:
Oops, I mixed up this PR with another one. I didn't mean to merge this one yet. I'll revert it.

erictraut:
Note to self: Use the "Draft" feature so it's obvious that a PR isn't yet ready to merge.

erictraut:
Here's the PR that reverts this one: https://github.com/python/typing/pull/1586. I'll create a new PR that re-applies this one and make sure to mark it as Draft this time.

hauntsaninja:
Thanks! :-) I added some labels so hopefully we confuse ourselves less in the future

PR #1577, opened 2024-01-12
Modernise more description in stubs distribution

hauntsaninja:
A common confusion in the Python ecosystem is the difference between distribution names and package names. I've seen users be slightly confused by types-* and *-stubs, so it can't hurt to be more explicit

Remove mentions of ancient pip and distutils

One maybe controversial thing is the loosening of a SHOULD to a MAY. This is meant to reflect the reality of how e.g. typeshed operates, which I think is pretty well thought through and I would be opposed to changing.

### DIFF
--- a/docs/spec/distributing.rst
+++ b/docs/spec/distributing.rst
@@ -113,19 +113,11 @@ frameworks.
 Package maintainers who wish to support type checking of their code MUST add
 a marker file named ``py.typed`` to their package supporting typing. This marker applies
 recursively: if a top-level package includes it, all its sub-packages MUST support
-type checking as well. To have this file installed with the package,
-maintainers can use existing packaging options such as ``package_data`` in
-distutils, shown below.
+type checking as well.
 
-Distutils option example::
-
-    setup(
-        ...,
-        package_data = {
-            'foopkg': ['py.typed'],
-        },
-        ...,
-        )
+To have this file including with the package, maintainers can use existing packaging
+options such as ``package_data`` in ``setuptools``. For more details, see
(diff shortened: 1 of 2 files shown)


docs/spec/distributing.rst
```diff
@@ -113,19 +113,11 @@ frameworks.
...
-    setup(
-        ...,
-        package_data = {
-            'foopkg': ['py.typed'],
-        },
-        ...,
-        )
+To have this file including with the package, maintainers can use existing packaging
+options such as ``package_data`` in ``setuptools``. For more details, see
+`the guide to providing type annotations <https://typing.readthedocs.io/en/latest/source/libraries.html#how-to-provide-type-annotations>`_
```
srittau, line 120: This should be an internal Sphinx link. Needs an anchor in the target document, though.

docs/spec/distributing.rst
```diff
@@ -146,7 +138,10 @@ corresponding ``*.py`` files. However, the stubs can also be put in a separate
 package and distributed separately. Third parties can also find this method
 useful if they wish to distribute stub files. The name of the stub package
 MUST follow the scheme ``foopkg-stubs`` for type stubs for the package named
-``foopkg``. Note that for stub-only packages adding a ``py.typed`` marker is not
+``foopkg`` (note the name of the distribution containing the package may be
+different than the mandated ``*-stubs`` package name).
```
srittau, line 142: It would be great to recommend the `-stubs` name, though, and discourage using `types-`, which is used by typescript. See pypi/warehouse#4967 for a discussion about that.

JelleZijlstra (commented):
Looks good, but i agree with Sebastian's comments.

srittau (approved):

PR #1578, opened 2024-01-13
Added language to reflect the intended behavior for triple-quoted type annotation expressions.

erictraut:
This is related to [this discussion](https://discuss.python.org/t/newlines-within-triple-quoted-type-expressions/42833).


gvanrossum (approved):

docs/spec/annotations.rst
```diff
@@ -114,6 +114,16 @@ be evaluated.
...
 
+If a triple quote is used, the string should be parsed as though it is
+implicitly surrounded by parentheses. This allows newline characters to be 
+used within the string literal.
+
+  value: """
+      int |
+      str |
+      list[Any]
+    """
```
gvanrossum, line 125: Can you double-check that this renders correctly as a text block? I'm no Sphinx expert, but IIRC text blocks are normally introduced by a double colon, as in the example a few lines down.

erictraut: Oh, good point. I'm not very experienced with Sphinx notation, so I may have gotten this wrong. Perhaps someone more knowledgeable could help here.

srittau:
```suggestion
used within the string literal::

    value: """
        int |
        str |
        list[Any]
    """
```

gvanrossum:
LGTM now.

(OT: We should have a workflow to preview the Sphinx-generated HTML, as we have for CPython.)

gvanrossum:
Preview [here](https://typing--1578.org.readthedocs.build/en/1578/spec/annotations.html) looks good.

hauntsaninja:
Linking https://github.com/python/typing-council/issues/9

rchen152 (approved):
Looks like a good change.

erictraut:
@rchen152, if that looks good to you, can you sign off on the TC issue [here](https://github.com/python/typing-council/issues/9). Thanks!

rchen152:
> @rchen152, if that looks good to you, can you sign off on the TC issue [here](https://github.com/python/typing-council/issues/9). Thanks!

Ah, I didn't realize that that was what the checkboxes were for. Done!

gvanrossum (approved):
LGTM, but try not to force push (I have no idea if you made a subtle change since the last commit).

PR #1582, opened 2024-01-14
Nit: typing.Text was only available since 3.5

srittau:


srittau:
But `str` was available since 3.0 ...

PR #1589, opened 2024-01-16
Conformance tests and spec change for generic type erasure

hauntsaninja:


docs/spec/generics.rst
```diff
@@ -333,15 +333,15 @@ argument(s) is substituted.  Otherwise, ``Any`` is assumed.  Example::
 
   class Node(Generic[T]):
```
hauntsaninja, line 334: The paragraph above could probably be improved. For instance, without the change I make here, pyright would infer `Node()` to be `Node[None]`, and I think that should be legal.

That said, handling of default values for generics in general is underspecified so maybe we tackle this later (mypy handles them poorly, but in part because it wasn't clear what behaviour should be https://github.com/python/mypy/issues/3737)

docs/spec/generics.rst
```diff
@@ -373,8 +373,8 @@ class instance that does not have an instance attribute with the same name::
   Node.x = 1       # Error
   Node.x           # Error
   type(p).x        # Error
-  p.x              # Ok (evaluates to None)
-  Node[int]().x    # Ok (evaluates to None)
+  p.x              # Ok (evaluates to int)
+  Node[int]().x    # Ok (evaluates to int)
   p.x = 1          # Ok, but assigning to instance attribute
```
hauntsaninja, line 378: These just seem straight up incorrect to me

docs/spec/generics.rst
erictraut: When this part of the spec was written, mypy defaulted to "implicit-optional", which would give you the same effect as adding `| None`. Since then, the spec was updated to recommend against "implicit-optional", but this code sample wasn't updated. I'd be in favor of just updating the spec to fix the sample.

gvanrossum (commented):
Changes to spec/generics.rst LGTM.

erictraut:
@hauntsaninja, is this PR ready to merge, or do you have additional work to do on it?

hauntsaninja:
This is ready to merge, I just wanted to leave open for a few days since I change the meaning of the spec.

(And I still owe tests for "Abstract generic types" and "Type variables with an upper bound")

PR #1591, opened 2024-01-18
Added draft chapter to typing spec for enumerations.

erictraut:


docs/spec/enums.rst
```diff
@@ -0,0 +1,308 @@
...
+
+    reveal_type(Color["RED"]) # Revealed type is 'Literal[Color.RED]' (or 'Color')
+    reveal_type(Color(3)) # Revealed type is 'Literal[Color.BLUE]' (or 'Color')
+
+Calling an enum class does not invoke its constructor, as is normal
+in most Python classes. The defined ``__new__`` method of an enum class is
+replaced with a custom implementation that performs a value-based lookup of
+an enum member.
+
+Enum members are immutable. Type checkers should generate an error if an enum
```
AlexWaygood, line 67: Strictly speaking, the enumeration itself is immutable, but the members within the enum are mutable: while you cannot add or delete members from the enumeration, the members themselves support the setting of arbitrary attributes in the same way as the vast majority of "normal" Python classes

AlexWaygood:
```pycon
>>> from enum import Enum
>>> class Foo(Enum):
...   BAR = 'bar'
...   
>>> Foo.BAR.whatever = 42
>>> Foo.BAR.whatever
42
```

docs/spec/enums.rst
```diff
@@ -0,0 +1,308 @@
...
+
+    def func2(any_color: Color):
+        reveal_type(any_color.value)  # Revealed type is Literal[1, 2, 3] (or int or object or Any)
+
+
+The value of ``_value_`` can be assigned in a constructor method. This technique
+is sometimes used to initialize both the member value and non-member attributes.
+If the value assigned in the class body is a tuple, the unpacked tuple value is
+passed to the constructor. Type checkers may validate consistency between assigned
+tuple values and the constructor signature.
```
AlexWaygood, line 234:
```suggestion
tuple values and the constructor signature::
```

docs/spec/enums.rst
```diff
@@ -0,0 +1,308 @@
...
+        RED = auto()
+        GREEN = auto()
+        BLUE = auto()
+
+    reveal_type(Color.RED.value) # Revealed type is Literal[1] (or int or object or Any)
+
+
+If an enum class provides an explicit type annotation for ``_value_``, type
+checkers should enforce this declared type when values are assigned to
+``_value__`::
```
AlexWaygood, line 265:
```suggestion
``_value_`::
```

docs/spec/enums.rst
```diff
@@ -0,0 +1,319 @@
+Enumerations
+============
+
+The ``Enum`` class behaves differently from other Python classes in several
```
srittau, line 4: Maybe too nitpicky, but maybe make it explicit that you are referring to stdlib's `enum.Enum`, and not just any old `Enum` class?

docs/spec/enums.rst
```diff
@@ -0,0 +1,319 @@
...
+    class Color1(Enum): # Supported
+        RED = 1
+        GREEN = 2
+        BLUE = 3
+
+    Color2 = Enum('Color2', 'RED', 'GREEN', 'BLUE') # Optional
+    Color3 = Enum('Color3', ['RED', 'GREEN', 'BLUE']) # Optional
+    Color4 = Enum('Color4', ('RED', 'GREEN', 'BLUE')) # Optional
+    Color5 = Enum('Color5', 'RED, GREEN, BLUE') # Optional
+    Color6 = Enum('Color6', 'RED GREEN BLUE') # Optional
```
beauxq, line 32: Should these line comments follow this specification from pep 8?

> Inline comments should be separated by at least two spaces from the statement.

Most of the later line comments in this document follow this, but not all of them.

docs/spec/enums.rst
erictraut: Yeah, agreed. I should also make it clear that this applies to subclasses of `enum.Enum` and any class that uses `enum.EnumType` (or a subclass thereof) as a metaclass.

docs/spec/enums.rst
```diff
@@ -0,0 +1,342 @@
...
+        match c:
+            case Color.RED | Color.BLUE:
+                print("red or blue")
+            case Color.GREEN:
+                print("green")
+            case _:
+                reveal_type(c)  # Revealed type is Never
+
+
+Likewise, a type checker should treat a complete union of all literal members
```
debonte, line 331: Is "should" here a typo? In the first paragraph, type checkers "**may** therefore expand an enum type into a union," and then here you say "Likewise, a type checker **should**...". If "should" is correct here, the word "Likewise" seems wrong.

docs/spec/enums.rst
```diff
@@ -0,0 +1,342 @@
...
+    reveal_type(Color(3))  # Revealed type is 'Literal[Color.BLUE]' (or 'Color')
+
+Calling an enum class does not invoke its constructor, as is normal
+in most Python classes. The defined ``__new__`` method of an enum class is
+replaced with a custom implementation that performs a value-based lookup of
+an enum member.
+
+Enum classes are immutable. Type checkers should generate an error if an enum
+is modified or deleted because these attempts will result in runtime
+errors::
```
ethanfurman, line 89:
```suggestion
Enum classes are immutable with regards to members. Type checkers should
generate an error if an enum member is modified or deleted because these attempts
will result in runtime errors::
```

docs/spec/enums.rst
```diff
@@ -0,0 +1,342 @@
...
+    class Shape(EnumWithNoMembers):  # OK (because no members are defined)
+        SQUARE = 1
+        CIRCLE = 2
+
+    class ExtendedShape(Shape):  # Type checker error: Shape is implicitly final
+        TRIANGLE = 3
+
+
+Members and Non-Member Attributes
+---------------------------------
```
ethanfurman, line 109:
```suggestion
Member and Non-Member Attributes
--------------------------------
```

docs/spec/enums.rst
```diff
@@ -0,0 +1,342 @@
...
+  A type checker should report an error if a type annotation is used
+  for an enum member because this type will be incorrect and misleading
+  to readers of the code::
+
+    class Pet(Enum):
+        CAT = 1  # OK
+        DOG: int = 2  # Type checker error
+
+    reveal_type(Pet.CAT)  # Revealed type is Literal[Pet.CAT]
+    reveal_type(Pet.DOG)  # Revealed type is Literal[Pet.DOG]
```
ethanfurman, line 150: Above, with `reveal_type(Color(3))` `Color` was allowed -- should `Pet` be allowed here?

docs/spec/enums.rst
```diff
@@ -0,0 +1,342 @@
...
+
+    reveal_type(Example.a)  # Revealed type is Literal[Example.a]
+    reveal_type(Example.b)  # Revealed type is int
+    reveal_type(Example.c)  # Revealed type is Literal[Example.c]
+
+
+* An enum class can define a class symbol named ``_ignore_``. This can be a list
+  of names or a string containing a space-delimited list of names that are
+  excluded from the list of members at runtime. Type checkers may support this
+  mechanism::
```
ethanfurman, line 193:
```suggestion
* An enum class can define a class symbol named ``_ignore_``. This can be a list
  of names or a string containing a space-delimited list of names that are
  removed from the enum class at runtime. Type checkers may support this
  mechanism::
```

docs/spec/enums.rst
```diff
@@ -0,0 +1,342 @@
...
+
+    class Pet(Enum):
+        CAT = 1  # Member attribute
+        DOG = 2  # Non-member attribute
+        FISH = 3  # Non-member attribute
+        _ignore_ = "DOG FISH"
+
+    reveal_type(Pet.CAT)  # Revealed type is Literal[Pet.CAT]
+    reveal_type(Pet.DOG)  # Revealed type is int (if _ignore_ is supported)
+    reveal_type(Pet.FISH)  # Revealed type is int (if _ignore_ is supported)
```
ethanfurman, line 203: These two will both produce `AttributeError` an runtime.

docs/spec/enums.rst
```diff
@@ -0,0 +1,342 @@
...
+
+    reveal_type(Color.RED._name_)  # Revealed type is Literal["RED"] (or str)
+    reveal_type(Color.RED.name)  # Revealed type is Literal["RED"] (or str)
+
+    def func1(red_or_blue: Literal[Color.RED, Color.BLUE]):
+        reveal_type(red_or_blue.name)  # Revealed type is Literal["RED", "BLUE"] (or str)
+
+    def func2(any_color: Color):
+        reveal_type(any_color.name)  # Revealed type is Literal["RED", "BLUE", "GREEN"] (or str)
+
```
ethanfurman, line 226: I'm a complete novice at typing, but it seems to me that the revealed type should be `str`, because it always will be a `str`, and using it directly will never produce an enum member.

docs/spec/enums.rst
```diff
@@ -0,0 +1,342 @@
...
+Likewise, a type checker should treat a complete union of all literal members
+as compatible with the enum type::
+
+    class Answer(Enum):
+        Yes = 1
+        No = 2
+
+    def func(val: object) -> Answer:
+        if val is not Answer.Yes and val is not Answer.No:
```
ethanfurman, line 339: You could also say:
```suggestion
        if val not in (Answer.Yes, Answer.No):
```
or even
```suggestion
        if val not in Answer:
```
although with newer Pythons, `1` and `2` would also pass the test.

ethanfurman (changes requested):
Added some corrections and suggestions.

docs/spec/enums.rst
erictraut: I think the existing wording is correct. The Enum documentation uses the term "members". The term "non-member attributes" describes all other attributes.

docs/spec/enums.rst
JelleZijlstra: `Literal["RED", "BLUE", "GREEN"]` means one of the exact strings "RED", "BLUE", or "GREEN", which is what the value would be at runtime. The draft spec allows both this precise type and the more general type `str`.

erictraut: At runtime, it's a `str` value, but it's a specific string. A type checker can determine this statically, so it's better for it to evaluate a string literal type here.

docs/spec/enums.rst
ethanfurman: Ah, right -- I keep getting tripped up by that phrasing.  I may have to change to `enum class attributes` instead of `non-member attributes`.

docs/spec/enums.rst
erictraut: The reason I didn't use those more complex forms is that mypy's type narrowing logic doesn't support them. Pyright does, but this spec is type checker agnostic. It's therefore best to pick examples that are likely to work on all type checkers.

docs/spec/enums.rst
```diff
@@ -0,0 +1,338 @@
...
+* An enum class can define a class symbol named ``_ignore_``. This can be a list
+  of names or a string containing a space-delimited list of names that are
+  deleted from the enum class at runtime. Type checkers may support this
+  mechanism::
+
+    class Pet(Enum):
+        _ignore_ = "DOG FISH"
+        CAT = 1  # Member attribute
+        DOG = 2  # Non-member attribute
+        FISH = 3  # Non-member attribute
```
ethanfurman, line 199:
```suggestion
        DOG = 2  # temporary variable, will be removed from the final enum class
        FISH = 3  # temporary variable, will be removed from the final enum class
```

docs/spec/enums.rst
```diff
@@ -0,0 +1,332 @@
...
+---------------
+
+Enum classes can be defined using a "class syntax" or a "function syntax".
+The function syntax offers several ways to specify enum members: names passed
+as individual arguments, a list or tuple of names, a string of
+comma-delimited or space-delimited names, a list or tuple of tuples that contain
+name/value pairs, and a dictionary of name/value items.
+
+Type checkers should support the class syntax, but the function syntax (in
+is various forms) is optional::
```
JelleZijlstra, line 22:
```suggestion
its various forms) is optional::
```

docs/spec/enums.rst
```diff
@@ -0,0 +1,332 @@
...
+    Color6 = Enum('Color6', 'RED GREEN BLUE')  # Optional
+    Color7 = Enum('Color7', [('RED': 1), ('GREEN': 2), ('BLUE': 3)])  # Optional
+    Color8 = Enum('Color8', (('RED': 1), ('GREEN': 2), ('BLUE': 3)))  # Optional
+    Color9 = Enum('Color9', {'RED': 1, 'GREEN': 2, 'BLUE': 3})  # Optional
+
+If a type checker supports the functional syntax, it should enforce name
+consistency. That is, if the type is assigned to a variable, the name of
+the variable must match the name of the enum class::
+
+  WrongName = Enum('Color', 'RED GREEN BLUE')  # Type checker error
```
JelleZijlstra, line 42: Not sure this should be an error, commented on Discuss

docs/spec/enums.rst
```diff
@@ -0,0 +1,332 @@
...
+
+    class ExtendedShape(Shape):  # Type checker error: Shape is implicitly final
+        TRIANGLE = 3
+
+
+Defining Members
+----------------
+
+When using the "class syntax", enum classes can define both members and
+other (non-member) attributes. The ``EnumType`` metaclass (named ``EnumMeta``
```
JelleZijlstra, line 109: Editorial suggestion: Move the note about EnumMeta/EnumType a few paragraphs up where the metaclass is first mentioned.

erictraut: Good suggestion. I've updated the draft accordingly.

docs/spec/enums.rst
```diff
@@ -0,0 +1,333 @@
...
+        GREEN = 2
+        BLUE = 3
+
+    for color in Color:
+        reveal_type(color)  # Revealed type is 'Color'
+
+    reveal_type(Color["RED"])  # Revealed type is 'Literal[Color.RED]' (or 'Color')
+    reveal_type(Color(3))  # Revealed type is 'Literal[Color.BLUE]' (or 'Color')
+
+Calling an enum class does not invoke its constructor, as is normal
```
rchen152, line 87: I find the wording of this paragraph a little confusing, especially the second sentence. After reading it a few times, I think the second sentence is describing enum's runtime behavior? But at first I thought it was directing type checkers on how to implement `enum.__new__`. What do you think of rewording the paragraph like so, to focus more on the effect of calling an enum class and less on what's happening under the hood?

> Unlike most Python classes, calling an enum class does not invoke its constructor. Instead, the call performs a value-based lookup of an enum member.

docs/spec/enums.rst
```diff
@@ -0,0 +1,333 @@
...
+  Within a type stub, members can be defined using the actual runtime values,
+  or a placeholder of ``...`` can be used::
+
+    class Pet(Enum):
+        genus: str  # Non-member attribute
+        species: str  # Non-member attribute
+
+        CAT = ...  # Member attribute
+        DOG = ...  # Member attribute
+
```
rchen152, line 136: I think it would be worth adding the example from https://discuss.python.org/t/draft-of-typing-spec-chapter-for-enums/43496/4 of how to declare the type with `_value_` when using `...` , since it's useful and non-obvious.

docs/spec/enums.rst
erictraut: Thanks, I like your proposed wording better than my original — definitely clearer.

docs/spec/enums.rst
erictraut: I think it would be confusing to add `_value_` to this example because the concept hasn't been introduced yet. It is provided as an example later in the document when `_value_` is discussed.

rchen152: Ah, true, this may not be a good place in the doc for this example.

But what I took away from the later example:
```
class Color(Enum):
        _value_: int
        RED = 1 # OK
        GREEN = "green"  # Type error
```
was that member values should be type-checked against the declared type of `_value_`; I didn't make the connection to how `_value_` could be used in combination with `...` in stubs until I read the Discuss thread.

erictraut: Ah, that's a good point. I've added another example to make this clear.

erictraut:
@JelleZijlstra, did mypy 1.9 fix the issue in stubtest that was blocking us from making the necessary changes in typeshed for enums? If not, is that something that could be prioritized for mypy 1.10?

docs/spec/enums.rst
```diff
@@ -0,0 +1,350 @@
...
+        IONIC = ...
+        CORINTHIAN = ...
+    
+    reveal_type(ColumnType.DORIC.value)  # Revealed type is int (or object or Any)
+
+
+Enum Literal Expansion
+----------------------
+
+From the perspective of the type system, an most enum classes are equivalent
```
Hnasar, line 309:
```suggestion
From the perspective of the type system, most enum classes are equivalent
```
typo

Hnasar (commented):
Very well written, thank you!

stroxler (commented):
Thanks for writing this, I've never actually known enum behavior as well as I should and learned quite a few things from reading.

docs/spec/enums.rst
```diff
@@ -0,0 +1,363 @@
...
+    class Pet(Enum):
+        CAT = 1  # OK
+        DOG: int = 2  # Type checker error
+
+* Methods, callables, and descriptors (including properties), and nested classes
+  that are defined in the class are not treated as enum members by the
+  ``EnumType`` metaclass and should likewise not be treated as enum members by
+  a type checker::
+
+    def identity(__x): return __x
```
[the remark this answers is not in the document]
erictraut: Yep, good point. I made the change.

docs/spec/enums.rst
```diff
@@ -0,0 +1,363 @@
...
+        GREEN = 2
+        BLUE = 3
+
+    for color in Color:
+        reveal_type(color)  # Revealed type is 'Color'
+
+    reveal_type(Color["RED"])  # Revealed type is 'Literal[Color.RED]' (or 'Color')
+    reveal_type(Color(3))  # Revealed type is 'Literal[Color.BLUE]' (or 'Color')
+
+Unlike most Python classes, Calling an enum class does not invoke its constructor.
```
JelleZijlstra, line 81:
```suggestion
Unlike most Python classes, calling an enum class does not invoke its constructor.
```

docs/spec/enums.rst
```diff
@@ -0,0 +1,363 @@
...
+
+    class Pet(Enum):
+        genus: str  # Non-member attribute
+        species: str  # Non-member attribute
+
+        CAT = ...  # Member attribute
+        DOG = ...  # Member attribute
+
+* Members defined within an enum class should not include explicit type
+  annotations. Type checkers should infer a literal type for all members.
```
carljm, line 130: At runtime, `Enum` allows member values to be of any type. The type system only specifies literal types for a limited set of types. How should type checkers handle enums whose values are of an unusual type, which the type checker may not normally support literals for?

docs/spec/enums.rst
```diff
@@ -0,0 +1,363 @@
...
+    class Example(Enum):
+        a = member(1)  # Member attribute
+        b = nonmember(2)  # Non-member attribute
+
+        @member
+        def c(self) -> None:  # Member method
+            pass
+
+    reveal_type(Example.a)  # Revealed type is Literal[Example.a]
+    reveal_type(Example.b)  # Revealed type is int
```
carljm, line 187: It doesn't seem wrong for this to be `Literal[2]`, either. (It should of course not be `Literal[Example.b]`)

carljm (approved):
Looks good to me, thanks for writing this.

docs/spec/enums.rst
```diff
@@ -0,0 +1,363 @@
...
+
+* Members defined within an enum class should not include explicit type
+  annotations. Type checkers should infer a literal type for all members.
+  A type checker should report an error if a type annotation is used
+  for an enum member because this type will be incorrect and misleading
+  to readers of the code::
+
+    class Pet(Enum):
+        CAT = 1  # OK
+        DOG: int = 2  # Type checker error
```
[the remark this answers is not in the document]
erictraut: Yes, this should be a type checker error in that case as well. Any annotation should be flagged as an error. 

There's a fundamental problem with annotating symbols corresponding to enum member types: their types change dynamically due to the magic of the EnumMeta class. Initially, when executing the class body, these symbols are assigned a value based on the assignment expression (e.g. `int` or `Literal[2]` in the example above). After the class body is executed, these symbols are overwritten by the EnumMeta class with a different value that is an instance of the enum class. Because of this dynamic runtime behavior, there's no annotation that accurately describes the type of an enum member.

docs/spec/enums.rst
erictraut: This statement is talking about the type of the enum member, not the type of its corresponding value. You are correct that enum member _values_ can have arbitrary types, but enum members cannot. They are always literal enum values.

```python
class Foo(Enum):
    A = [1, 2, 3]

reveal_type(Foo.A)  # Literal[Foo.A]
reveal_type(Foo.A.value)  # list[int]
```

docs/spec/enums.rst
erictraut: Yes, I agree. I'll update accordingly.

docs/spec/enums.rst
carljm: I see. I wonder if there is a wording clarification that would have helped me avoid this mistake, but I don't have any immediate ideas. I agree that if you are clearly keeping mind the distinction between a member type and a member value type, it is clear.

erictraut:
I'm not sure how to fix the CLA Signing issue with this PR, so I'm going to close it and submit a fresh PR. See #1759.

PR #1599, opened 2024-01-20
Added draft chapter to typing spec for tuples. 

erictraut:
This consolidates and augments existing information about tuples within the type system.


docs/spec/tuples.rst
```diff
@@ -0,0 +1,154 @@
+Tuples
+======
+
+The ``tuple`` class has some special behaviors and properties that make it
+different from other classes from a typing perspective. The most obvious
+difference is that ``tuple`` is variadic -- it supports an arbitrary number
+of type arguments. Each type argument represents a single element, and those
+elements are immutable at runtime. These properties affect subtyping rules and
```
JelleZijlstra, line 8:
```suggestion
of type arguments. In fixed-sized tuple types, each type argument represents
a single element. These properties affect subtyping rules and
```

It's not strictly accurate that the elements are immutable, since a tuple can contain mutable objects. The statement seems unnecessary anyway, as the immutability of tuples is a property of the language and doesn't have to be repeated here.

It's also not accurate that each type arguments represents a single element (in `tuple[int, ...]`, `int` represents an arbitrary number of elements). The suggested wording fixes that issue but is still concise.

docs/spec/tuples.rst
```diff
@@ -0,0 +1,154 @@
...
+    def func[*Ts](t: tuple[*Ts]):
+        t5: tuple[*tuple[str], *Ts]  # OK
+        t6: tuple[*tuple[str, ...], *Ts]  # Type error
+
+The ``*`` syntax requires Python 3.11 or newer. For older versions of Python,
+the ``typing.Unpack`` special form can be used:
+``tuple[int, Unpack[tuple[str, ...]], int]``.
+
+Unpacked tuples can also be used for ``*args`` parameters in a function
+signature. For example, ``def f(*args: *tuple[int, str]): ...``. Unpacked tuples
```
JelleZijlstra, line 82: Incomplete sentence, maybe add that this is equivalent to `def f(arg1: int, arg2: str, /)`.

docs/spec/tuples.rst
```diff
@@ -0,0 +1,154 @@
...
+can also be used for specializing generic classes or type variables that are
+parameterized using a ``TypeVarTuple``. For more details, see "``*args`` as a
+Type Variable Tuple" in the Generics section. [TODO: not sure how to create a
```
JelleZijlstra, line 85: You have to add a link target. Example from elsewhere in this repo: there is a line `.. _best-practices:` that creates a link target, and elsewhere we link to that target with `` :ref:`best-practices` ``

docs/spec/tuples.rst
```diff
@@ -0,0 +1,154 @@
...
+Type Compatibility Rules
+------------------------
+
+Because tuple contents are immutable, the element types of a tuple are covariant.
+For example, ``tuple[int, int]`` is a subtype of ``tuple[float, complex]``.
+
+As discussed above, a homogeneous tuple of arbitrary length is equivalent
+to a union of tuples of different lengths. That means ``tuple[()]``, 
+``tuple[int]`` and ``tuple[int, *tuple[int, ...]]`` are all subtypes of
+``tuple[int, ...]``.
```
JelleZijlstra, line 98:
```suggestion
``tuple[int, ...]``, but the converse is not true: `tuple[int, ...]` is not a subtype of `tuple[int]`.
```

docs/spec/tuples.rst
```diff
@@ -0,0 +1,155 @@
...
+---------------
+
+The type of a tuple can be expressed by listing the element types. For
+example, ``tuple[int, int, str]`` is a tuple containing an ``int``, another
+``int``, and a ``str``.
+
+The empty tuple can be annotated as ``tuple[()]``.
+
+Arbitrary-length homogeneous tuples can be expressed using one type and
+ellipsis, for example ``tuple[int, ...]``. This type is equivalent to a union
```
superbobry, line 23: Equivalent wrt subtyping or consistent subtyping?

docs/spec/tuples.rst
```diff
@@ -0,0 +1,155 @@
...
+of tuples containing zero or more ``int`` elements (``tuple[()] |
+tuple[int] | tuple[int, int] | tuple[int, int, int] | ...``).
+Arbitrary-length homogeneous tuples are sometimes referred to as "unbounded
+tuples". Both of these terms appear within the typing spec, and they refer to
+the same concept.
+
+The type ``tuple[Any, ...]`` is special in that it is bidirectionally
```
superbobry, line 30: What makes this case special? Doesn't this follow from consistent subtyping rules for any and the equivalence you gave above?

erictraut: This is special because it's a gradual type with respect to the length of the tuple. It is bidirectionally compatible with any tuple. The use of `tuple[T, ...]` where `T` is any type other than `Any` is _not_ a gradual type. It is shorthand for a union type. A value with this type is not assignable to a tuple with a known fixed length. The next section contains some specific examples that help demonstrate this difference.

docs/spec/tuples.rst
erictraut: I mean "equivalent with respect to subtyping". We may need more precise terminology here. I tend to use "equivalent" to mean that the two types describe _exactly_ the same set of values (from a set theory standpoint), and I use the term "bidirectionally compatible" to describe the concept of two types that are consistent subtypes of each other but not necessarily equivalent. If you have a better suggestion for terminology, let me know.

docs/spec/tuples.rst
superbobry: Thanks for the clarification, Eric! I assume `T` does not have to be exactly `Any` to `tuple[T, ...]` to be a gradual type. It just needs to contain `Any`, right?

docs/spec/tuples.rst
superbobry: My preference would be to spell out "equivalent wrt ..." explicitly, but having a glossary where "equivalent" and "bidirectionally compatible" are defined would also work.

docs/spec/tuples.rst
```diff
@@ -0,0 +1,155 @@
...
+    def func(subj: tuple[int | str, int | str]):
+        match subj:
+            case x, str():
+                reveal_type(subj)  # tuple[int | str, str]
+            case y:
+                reveal_type(subj)  # tuple[int | str, int]
+
+The ``tuple`` class derives from ``Sequence[T_co]`` where ``T_co`` is a covariant
+(non-variadic) type variable. The specialized type of ``T_co`` should be computed
+by a type checker based on the union of element types within the tuple.
```
superbobry, line 151: Does this mean that given

```
x: Any
y: int
```

the type of `(x, y)` is `Sequence[Any | int]` and not `Sequence[Any]`?

docs/spec/tuples.rst
erictraut: > I assume T does not have to be exactly Any to tuple[T, ...] to be a gradual type.

No, it needs to be exactly `Any`. I think the text and examples are already pretty clear on this account, but if you can think of a way to make it clearer, let me know.

docs/spec/tuples.rst
erictraut: > the type of (x, y) is Sequence[Any | int] and not Sequence[Any]?

Yes, the type would be `Sequence[Any | int]` based on the current draft text.

It looks like mypy uses a join operation here rather than a union, so it produces `Sequence[Any]` rather than `Sequence[Any | int]`. Maybe the spec needs to be less prescriptive here to allow for both unions and joins (or any other operation that produces some type that is a supertype of all element types). I'll post that question in the discussion to get feedback from the broader community.

(There are many reasons why a union is better than a join here. There are [63 active bug reports](https://github.com/python/mypy/labels/topic-join-v-union) in the mypy issue tracker that are labeled "topic-join-v-union".)

docs/spec/tuples.rst
```diff
@@ -0,0 +1,155 @@
...
+first element is guaranteed to be of type ``int``, the last element is
+guaranteed to be of type ``str``, and the elements in the middle are zero or
+more elements of type ``str``. The type ``tuple[*tuple[int, ...]]`` is
+equivalent to ``tuple[int, ...]``.
+
+If an unpacked ``*tuple[Any, ...]`` is embedded within another tuple, that
+portion of the tuple is bidirectionally type compatible with any tuple of
+any length.
+
+Only one unbounded tuple can be used with another tuple::
```
rchen152, line 65: nit: I think this is supposed to say "within," not "with."

docs/spec/tuples.rst
```diff
@@ -0,0 +1,155 @@
...
+``int``, and a ``str``.
+
+The empty tuple can be annotated as ``tuple[()]``.
+
+Arbitrary-length homogeneous tuples can be expressed using one type and
+ellipsis, for example ``tuple[int, ...]``. This type is equivalent to a union
+of tuples containing zero or more ``int`` elements (``tuple[()] |
+tuple[int] | tuple[int, int] | tuple[int, int, int] | ...``).
+Arbitrary-length homogeneous tuples are sometimes referred to as "unbounded
+tuples". Both of these terms appear within the typing spec, and they refer to
```
JelleZijlstra, line 27: We can change this later, but I'd prefer to use only a single term throughout the spec. "Arbitrary-length homogeneous tuples" is longer but more clear.

JelleZijlstra (approved):

docs/spec/tuples.rst
gvanrossum, line 23: I think technically you have to say "and **an** ellipsis".

gvanrossum (approved):
I'll sign off. I have one English grammar nit.

I also wonder if we ought to say something about sequence unpacking? I recall being confused for a while when trying out some tuple examples in the mypy sandbox until I realized that `x: int; y: int; x, y = blah()` is very different than `xy: tuple[int, int] = blah()`. It's probably not on-topic for this section though.

docs/spec/tuples.rst
erictraut: Good catch. I've pushed a fix.

This text was copied directly from the [original language in PEP 484](https://peps.python.org/pep-0484/#the-typing-module). DIdn't you write much of that PEP? :)

erictraut: > I also wonder if we ought to say something about sequence unpacking?

I agree the spec should eventually say something about tuple unpacking. Note that tuple unpacking is different from general sequence unpacking, especially in cases where a tuple has a known length.

I started to write a section about tuple unpacking behaviors, but I decided to defer that discussion for now. I think the topic will be somewhat controversial. It's also intertwined with an even more controversial topic: how to treat an unpacked tuple in a function call (e.g. `func(*my_tuple)`) when the length of the tuple isn't known or the function includes an `*args` parameter. The behaviors here differ significantly across type checkers.

This is on a long list of issues I'm tracking. We'll get to it eventually.

erictraut:
I've written conformance tests for the new "tuples" section to see where the four major type checkers deviate from the spec. Results are in [this PR](https://github.com/python/typing/pull/1609).

docs/spec/tuples.rst
```diff
@@ -0,0 +1,155 @@
...
+The length of a tuple at runtime is immutable, so it is safe for type checkers
+to use length checks to narrow the type of a tuple::
+
+    def func(val: tuple[int] | tuple[str, str] | tuple[int, *tuple[str], int]):
+        if len(val) == 1:
+            # Type can be narrowed to tuple[int].
+            reveal_type(val)  # tuple[int]
+        
+        if len(val) == 2:
+            # Type can be narrowed to tuple[str, str] | tuple[int, int].
```
hauntsaninja, line 123: This seems incorrect, how do you get `tuple[int, int]`? Did you mean `tuple[int, *tuple[str, ...], int]` as the last element in the union?

docs/spec/tuples.rst
```diff
@@ -0,0 +1,155 @@
...
+            # Type can be narrowed to tuple[int, str, int].
+            reveal_type(val)  # tuple[int, str, int]
+
+This property may also be used to safely narrow tuple types within a ``match``
+statement that uses sequence patterns.
+
+If a tuple element is a union type, the tuple can be safely expanded into a
+union of tuples. For example, ``tuple[int | str]`` is equivalent to 
+``tuple[int] | tuple[str]``. If multiple elements are union types, the expansion
+must consider all combinations. For example, ``tuple[int | str, int | str]`` is
```
hauntsaninja, line 136: "must" is a little strong here for behaviour that is exponential

docs/spec/tuples.rst
JelleZijlstra: I think this is right. The `*tuple[str, ...]` expands to 0 or more elements. If the tuple is of length 2, that means the variadic part must be empty (because there must be an int at the start and one at the end).

hauntsaninja: Yes that's what I'm saying it should be. The type of the argument is currently `tuple[int, *tuple[str], int]`, not `tuple[int, *tuple[str, ...], int]`

JelleZijlstra: Oh you're right, good catch.

docs/spec/tuples.rst
```diff
@@ -0,0 +1,155 @@
...
+        v4: tuple[Any, ...] = t2  # OK
+
+        v5: tuple[float, float] = t3  # OK
+        v6: tuple[int, *tuple[str, ...]] = t3  # OK
+
+
+The length of a tuple at runtime is immutable, so it is safe for type checkers
+to use length checks to narrow the type of a tuple::
+
+    def func(val: tuple[int] | tuple[str, str] | tuple[int, *tuple[str], int]):
```
JelleZijlstra, line 117:
```suggestion
    def func(val: tuple[int] | tuple[str, str] | tuple[int, *tuple[str, ...], int]):
```
Pretty sure this was meant, see Shantanu's comments below.

docs/spec/tuples.rst
erictraut: I'll change the text to read "...full expansion must consider...". You're correct that partial expansion wouldn't require this. Pyright's current implementation performs partial expansion only (i.e. a single tuple element is expanded rather than multiple), and that's how I avoid the combinatorial explosion.

docs/spec/tuples.rst
erictraut: Yes, good catch. I caught the problem in the test suite, but I forgot to update the bug in the spec. Consider it fixed.

PR #1604, opened 2024-01-24
Link directly to spec and guide on distribution

### LINKED ISSUE #1603: Outdated stub package guidance
The guidance for distributing stub packages on the "Type Stubs" page ([link](https://typing.readthedocs.io/en/latest/source/stubs.html#distribution)) still recommends using the `types-<package>` naming convention. If I've understood recent updates (see #1577) correctly, this is (now?) only preferred for typeshed stub distributions. This section is why I published the stub package I maintain with the `types-` prefix, so, it should probably be changed if that's not the way things are meant to be done.

hauntsaninja:
Fixes #1603

### DIFF
--- a/docs/source/stubs.rst
+++ b/docs/source/stubs.rst
@@ -72,14 +72,8 @@ Distribution
 ============
 
 Type stubs can be distributed with or separately from the implementation;
-see :pep:`561` for more information. The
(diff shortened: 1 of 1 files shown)


srittau (approved):
Good for now, see also #1605.

PR #1614, opened 2024-01-30
Added draft chapter to typing spec for named tuples

erictraut:


docs/spec/namedtuples.rst
```diff
@@ -0,0 +1,147 @@
...
+    Point1 = namedtuple('Point1', ['x', 'y'])
+    Point2 = namedtuple('Point2', ('x', 'y'))
+    Point3 = namedtuple('Point3', 'x y')
+    Point4 = namedtuple('Point4', 'x, y')
+
+    Point5 = NamedTuple('Point5', [('x', int), ('y', int)])
+    Point6 = NamedTuple('Point6', (('x', int), ('y', int)))
+
+At runtime, the ``namedtuple`` function disallows field names that are
+illegal Python identifiers and either rases an exception or replaces these
```
JelleZijlstra, line 80:
```suggestion
illegal Python identifiers and either raises an exception or replaces these
```

docs/spec/namedtuples.rst
```diff
@@ -0,0 +1,147 @@
...
+
+The fields within a named tuple instance can be accessed by name using an 
+attribute access (``.``) operator. Type checkers should support this::
+
+    p = Point(1, 2)
+    assert_type(p.x, int)
+    assert_type(p.units, str)
+
+Like normal tuples, elements of a named tuple can also be accessed by index,
+and type checkers should support this.
```
JelleZijlstra, line 110:
```suggestion
and type checkers should support this::
```
(Or add a `::` alone on the next line; currently the block isn't recognized as a code block.)

JelleZijlstra (commented):
Looks good, a few editorial issues

docs/spec/namedtuples.rst
```diff
@@ -0,0 +1,147 @@
...
+in the base class::
+
+    class PointWithName(Point):
+        name: str  # OK
+        x: int  # Type error (invalid override of named tuple field)
+
+In Python 3.11 and newer, the class syntax supports generic named tuple classes.
+Type checkers should support this::
+
+    class Property[T](NamedTuple):
```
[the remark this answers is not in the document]
erictraut: We're trying to stick with modern syntax in the typing spec when adding new sections.

docs/spec/namedtuples.rst
```diff
@@ -0,0 +1,147 @@
...
+illegal Python identifiers and either raises an exception or replaces these
+fields with a parameter name of the form ``_N``. The behavior depends on
+the value of the ``rename`` argument. Type checkers may replicate this
+behavior statically::
+
+    NT1 = namedtuple("NT1", ["a", "b"])  # Type error (duplicate field name)
+    NT2 = namedtuple("NT2", ["abc", "def"], rename=False)  # Type error (illegal field name)
+
+    NT3 = namedtuple("NT3", ["abc", "def"], rename=True)  # OK
+    NT3(abc="", _1="")  # OK
```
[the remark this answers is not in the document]
erictraut: `def` is not a valid identifier in Python because it's a reserved keyword. If you try running this code, you'll see that a runtime exception is raised.

docs/spec/namedtuples.rst
```diff
@@ -0,0 +1,147 @@
...
+        x: int
+        y: int
+        units: str = "meters"
+
+Regardless of whether the class syntax or function syntax is used to define
+a named tuple, type checkers should synthesize a ``__new__`` method based on
+the named tuple fields. This mirrors the runtime behavior. In the example
+above, the synthesized ``__new__`` method would look like the following::
+
+    def __new__(cls, x: int, y: int, units: str = "meters") -> None:
```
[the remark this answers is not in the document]
erictraut: Good catch. I've changed it to `Self`.

docs/spec/namedtuples.rst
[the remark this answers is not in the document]
TeamSpen210: Makes sense.

docs/spec/namedtuples.rst
```diff
@@ -0,0 +1,147 @@
...
+    Point5 = NamedTuple('Point5', [('x', int), ('y', int)])
+    Point6 = NamedTuple('Point6', (('x', int), ('y', int)))
+
+At runtime, the ``namedtuple`` function disallows field names that are
+illegal Python identifiers and either raises an exception or replaces these
+fields with a parameter name of the form ``_N``. The behavior depends on
+the value of the ``rename`` argument. Type checkers may replicate this
+behavior statically::
+
+    NT1 = namedtuple("NT1", ["a", "b"])  # Type error (duplicate field name)
```
sobolevn, line 85: Sorry, did you mean `NT1 = namedtuple("NT1", ["a", "a"]) # Type error (duplicate field name)`?

Right now it has two different fields.
```suggestion
    NT1 = namedtuple("NT1", ["a", "a"])  # Type error (duplicate field name)
```

sobolevn (commented):
Question: should `Unpack[]` be mentioned here?

docs/spec/namedtuples.rst
erictraut: Good catch. Fixed.

erictraut:
@sobolevn, you asked about `Unpack[]`. Can you elaborate on that? What do you have in mind?

sobolevn:
@erictraut `typing.py` mentions that you can unpack tuples: https://github.com/python/cpython/blob/1aec0644447e69e981d582449849761b23702ec8/Lib/typing.py#L1587-L1597

And since `namedtuple` is compatible with `tuple`, I think that it should be explicit that you cannot unpack namedtuples.

One more related thing: https://peps.python.org/pep-0646/ does not mention `Unpack[tuple]` at all 🤔

erictraut:
@sobolevn, "unpacking a tuple" is different from using the special form `Unpack`. The `Unpack` special form can be used only in certain type annotation contexts and only with an unpacked TypeVarTuple or `tuple` according to PEP 646. 

For example, this is allowed:
```python
from typing import Unpack

def func[*Ts](x: tuple[int, Unpack[tuple[str, ...]]], *args: Unpack[Ts]) -> tuple[Unpack[Ts]]: ...
```

This is equivalent to:
```python
def func[*Ts](x: tuple[int, *tuple[str, ...]], *args: *Ts) -> tuple[*Ts]: ...
```

The `Unpack` special form doesn't have anything to do with named tuples.

Named tuples, like regular tuples, *can* be unpacked using the unpack operator or through a destructuring assignment.

```python
NT = namedtuple("NT", "x y z")

my_tuple = (1, 2, 3)
my_named_tuple = NT(1, 2, 3)

t1 = [*my_tuple]
t2 = [*my_named_tuple]

a, b, c, = my_tuple
x, y, z = my_named_tuple
```

sobolevn:
@erictraut yes, this is exactly why I am asking.

> A named tuple is a subtype of a ``tuple``

You can unpack `tuple`, some people might think that if you can unpack a tuple, you might also unpack its subtypes, for example, a namedtuple: `Unpack[NT]`.

Is it worth mentioning in the docs that you cannot do that, what you think?

erictraut:
As I said, "unpacking a tuple" isn't the same as using the `Unpack` special form in a type annotation. You seem to be confusing the two concepts. The `Unpack` special form cannot be used with anything other than a `TypeVarTuple` or a `tuple` as explained in PEP 646 (and in the TypeVarTuple and tuple chapters in the typing spec). I think it would be odd for the "named tuple" chapter to say anything about the `Unpack` special form.

JelleZijlstra (approved):

docs/spec/namedtuples.rst
```diff
@@ -0,0 +1,147 @@
...
+============
+
+As with tuples, named tuple classes require some special behaviors for type
+checking.
+
+
+Defining Named Tuples
+---------------------
+
+A named tuple class can be defined using a class syntax or a function syntax.
```
hamdanal, line 11: Thank you for writing this up. I have a tiny nit about the terminology "function syntax" used throughout this page. For me this term is associated with the `def` statement, I feel it describes the syntax used to write functions in python. Here however named tuples are created with a function call. Would "function-call syntax" or maybe "functional syntax" be a better term here?

erictraut: That's a good point. The typing docs use the term "functional syntax" and the `namedtuple` documentation calls it a "factory function call". I think either of these work, but I've opted for the latter.

hauntsaninja (approved):

gvanrossum (approved):

PR #1618, opened 2024-02-04
Added clarifications about illegal uses of `Annotated` special form.

erictraut:
This is a draft proposal related to [this discussion](https://discuss.python.org/t/is-annotated-compatible-with-type-t/43898/17).


docs/spec/qualifiers.rst
```diff
@@ -227,6 +227,24 @@ details of the syntax:
...
+    v2: type[Any] = SmallInt  # Type error
+
+* An attempt to call ``Annotated`` (whether parameterized or not) should be
+  treated as a type error by type checkers::
+
+    Annotated()  # Type error
+    Annotated[int, ""](0)  # Type error
+
+    SmallInt = Annotated[int, ValueRange(0, 100)]
+    SmallInt(1)  # Type error
```
adriangb, line 245: This sounds good. I was surprised when I learned this worked at runtime. Pydantic doesn't use this feature. I don't see any reason not to prohibit it at the type-checker level.

JelleZijlstra (approved):

hauntsaninja (approved):

gvanrossum (approved):

PR #1619, opened 2024-02-05
Added clarification about pre-3.8 way to specify positional-only parameters.

erictraut:
This is a draft proposal related to [this discussion](https://discuss.python.org/t/ambiguities-about-positional-only-parameters/42328).

### DIFF
--- a/docs/spec/historical.rst
+++ b/docs/spec/historical.rst
@@ -188,25 +188,30 @@ When checking Python 2.7 code, type checkers should treat the ``int`` and
 type ``str`` as well as ``unicode`` should be acceptable.
 
 
-Positional-only arguments
--------------------------
+Positional-only parameters
+--------------------------
 
 Some functions are designed to take their arguments only positionally,
 and expect their callers never to use the argument's name to provide
 that argument by keyword. Before Python 3.8 (:pep:`570`), Python did
-not provide a way to declare positional-only arguments.
+not provide a way to declare positional-only parameters.
 
-To support positional-only arguments on older Python versions, type
-checkers support the following special case:
-all arguments with names beginning with
-``__`` are assumed to be positional-only, except if their names also
-end with ``__``::
+To support positional-only parameters on older Python versions, type
+checkers should support the following special case: all parameters with
+names beginning with ``__`` are assumed to be positional-only unless their
+names also end with ``__``::
 
-  def quux(__x: int, __y__: int = 0) -> None: ...
+  def f(__x: int, __y__: int = 0) -> None: ...
 
-  quux(3, __y__=1)  # This call is fine.
+  f(3, __y__=1)  # This call is fine.
 
-  quux(__x=3)  # This call is an error.
+  f(__x=3)  # This call is an error.
+
+Consistent with :pep:`570` syntax, positional-only parameters cannot appear
+after parameters that accept keyword arguments. Type checkers should enforce
+this requirement::
+
+  def g(x: int, __y: int) -> None: ...  # Type error
 
 
 Generics in standard collections
--- a/docs/spec/historical.rst
+++ b/docs/spec/historical.rst
@@ -196,7 +196,7 @@ and expect their callers never to use the argument's name to provide
 that argument by keyword. Before Python 3.8 (:pep:`570`), Python did
 not provide a way to declare positional-only parameters.
 
-To support positional-only parameters for code that is intended to work
+To support positional-only parameters in code that must remain compatible
 with older versions of Python, type checkers should support the following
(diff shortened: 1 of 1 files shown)


docs/spec/historical.rst
```diff
@@ -188,25 +188,30 @@ When checking Python 2.7 code, type checkers should treat the ``int`` and
...
+Positional-only parameters
+--------------------------
 
 Some functions are designed to take their arguments only positionally,
 and expect their callers never to use the argument's name to provide
 that argument by keyword. Before Python 3.8 (:pep:`570`), Python did
-not provide a way to declare positional-only arguments.
+not provide a way to declare positional-only parameters.
+
+To support positional-only parameters on older Python versions, type
```
JelleZijlstra, line 199: This reasoning arguably doesn't apply any more, because 3.7 is no longer supported. Still, the old syntax should remain supported for compatibility with old code. Perhaps we could add a statement like "While Python 3.7 and lower are no longer supported, type checkers should continue to support this syntax to support code written for earlier versions of Python."

JelleZijlstra (approved):
Looks good, one optional suggestion

hauntsaninja (approved):

docs/spec/historical.rst
```diff
@@ -188,25 +188,30 @@ When checking Python 2.7 code, type checkers should treat the ``int`` and
...
-Positional-only arguments
--------------------------
+Positional-only parameters
+--------------------------
 
 Some functions are designed to take their arguments only positionally,
 and expect their callers never to use the argument's name to provide
 that argument by keyword. Before Python 3.8 (:pep:`570`), Python did
-not provide a way to declare positional-only arguments.
+not provide a way to declare positional-only parameters.
```
gvanrossum, line 197: Nit: Do we really want to draw the distinction between arguments and parameters here? It's a little odd to see the paragraph switch from "argument" to "parameter" suddenly. I believe I saw a proposal once suggesting to use "arguments" for functions, and reserve "parameters" for generic types (and type forms)?

docs/spec/historical.rst
```diff
@@ -188,25 +188,30 @@ When checking Python 2.7 code, type checkers should treat the ``int`` and
...
+
+To support positional-only parameters in code that must remain compatible
+with older versions of Python, type checkers should support the following
+special case: all parameters with names beginning with ``__`` are assumed to 
+be positional-only unless their names also end with ``__``::
```
gvanrossum, line 202: Nit: The phrasing here seems to draw attention to the trailing `__`. Suggestion:
```suggestion
special case: all parameters with names that begin but don't end with ``__``
are assumed to be positional-only::
```

gvanrossum (approved):
Two optional nits.

docs/spec/historical.rst
erictraut: Arguments refers to the values you pass to a function when you call it, whereas parameters defined as part of a function signature and are assigned the arguments at call time. I don't know what "positional-only arguments" would mean, so I find this phrase completely confusing. [PEP 570](https://peps.python.org/pep-0570/) consistently (and correctly, IMO) uses the phrase "positional-only parameters". That's also the phrase we use elsewhere in the typing spec. I don't think we should deviate from that phrase.

docs/spec/historical.rst
erictraut: I didn't modify this text from the earlier version, but I agree that your new phrasing is easier to understand. I'm OK with updating it for clarity.

gvanrossum (approved):

docs/spec/historical.rst
```diff
@@ -188,25 +188,30 @@ When checking Python 2.7 code, type checkers should treat the ``int`` and
...
-  def quux(__x: int, __y__: int = 0) -> None: ...
+  f(__x=3)  # This call is an error.
 
-  quux(3, __y__=1)  # This call is fine.
+Consistent with :pep:`570` syntax, positional-only parameters cannot appear
+after parameters that accept keyword arguments. Type checkers should enforce
+this requirement::
 
-  quux(__x=3)  # This call is an error.
+  def g(x: int, __y: int) -> None: ...  # Type error
```
nineteendo, line 214: How about overloads? I don't see a reason to make `func` positional only here:

```python
@overload
def map2(func: Callable[[_T1], _S], __iter1: Iterable[_T1]) -> map: ...
@overload
def map2(func: Callable[[_T1, _T2], _S], __iter1: Iterable[_T1], __iter2: Iterable[_T2]) -> map: ...

def map2(func: Callable[..., _S], *iters: Iterable[Any]) -> map:
    return map(func, iters)
```

PR #1620, opened 2024-02-07
Spec: add `Self` reference to link from PEP 673

hugovk:
Like https://github.com/python/typing/pull/1566, so we can mark PEP 673 as Final and refer to the typing spec as the canonical docs.


srittau (approved):

hugovk:
Thanks! PEPs PR for reference: https://github.com/python/peps/pull/3644

PR #1621, opened 2024-02-07
Spec: add Protocols reference to link from PEP 544

hugovk:
Like https://github.com/python/typing/pull/1566, to mark https://peps.python.org/pep-0544/ as final.

### DIFF
--- a/docs/spec/protocol.rst
+++ b/docs/spec/protocol.rst
@@ -1,3 +1,5 @@
+.. _protocols:
+
 Protocols
 ---------
 


JelleZijlstra:
Thanks! We should probably just do a pass of adding some reasonable link targets throughout the spec.

hugovk:
Yep, not a bad idea.

I'm doing them one-by-one because I don't know which PEP I'm going to look at next (or even when!).

There's about 18 typing PEPs to finish up, so here are the ones to target if it helps:

* https://peps.python.org/topic/typing/#accepted-peps-accepted-may-not-be-implemented-yet

PR #1622, opened 2024-02-07
Bump GitHub Actions

hugovk:
To fix some deprecation warings:

image

Would Dependabot be useful here? To make it less spammy, we can set it to monthly and group updates into a single PR.

https://github.com/python/typing/actions/runs/7818405828?pr=1621

### DIFF
--- a/.github/workflows/build-docs.yml
+++ b/.github/workflows/build-docs.yml
@@ -13,9 +13,9 @@ jobs:
     runs-on: ubuntu-latest
 
     steps:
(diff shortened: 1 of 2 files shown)


AlexWaygood:
I'd be open to using Dependabot to update GitHub Actions iff all updates were bundled together (but I can't speak for the other maintainers of this repo :)

hugovk:
Here you go! https://github.com/python/typing/pull/1623

PR #1623, opened 2024-02-07
Add Dependabot to keep GitHub Actions up-to-date

hugovk:
Monthly grouped updates to keep it less spammy.


AlexWaygood (approved):
It's a shame there isn't a "quarterly" option like there is with pre-commit

PR #1625, opened 2024-02-10
Spec: add TypeGuard reference to link from PEP 647

hugovk:
Like https://github.com/python/typing/pull/1566, so we can mark PEP 647 as Final and refer to the typing spec as the canonical docs.

(I'll re-open https://github.com/python/peps/pull/3577)


JelleZijlstra:
I'll make a PR today to add link targets throughout the spec.

PR #1626, opened 2024-02-10
Add reference targets throughout the spec

JelleZijlstra:
To facilitate cross-linking from the PEPs and the CPython documentation.

Also fix a few instances of "Type[]" that I noticed. My editor removed some trailing whitespace.


docs/spec/protocol.rst
```diff
@@ -458,7 +457,7 @@ Example::
   cached_func((1, 2, 3)) # OK, tuple is both hashable and iterable
 
 
-``Type[]`` and class objects vs protocols
+``type[]`` and class objects vs protocols
 ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
 
 Variables and parameters annotated with ``Type[Proto]`` accept only concrete
```
srittau, line 463:
```suggestion
Variables and parameters annotated with ``type[Proto]`` accept only concrete
```

srittau (approved):
LGTM, but spotted one more instance of `Type`.

PR #1637, opened 2024-02-21
Proposed spec change about TypeGuard/bool compatibility

JelleZijlstra:
See https://discuss.python.org/t/compatibility-between-typeguard-and-bool/46364


docs/spec/narrowing.rst
```diff
@@ -19,9 +19,9 @@ user-defined type guard function. Return statements within a type guard function
 should return bool values, and type checkers should verify that all return paths
 return a bool.
 
-In all other respects, ``TypeGuard`` is a distinct type from bool. It is not a
-subtype of bool. Therefore, ``Callable[..., TypeGuard[int]]`` is not assignable
-to ``Callable[..., bool]``.
+``TypeGuard`` is also valid as the return type of a ``Callable`` type. In that
+context, it is treated as a subtype of bool. For example, ``Callable[..., TypeGuard[int]]``
+is assignable to ``Callable[..., bool]``.
```
srittau, line 24: Two questions:

1. What about return types of "`Callable`-like" types, e.g. callable protocols?
2. While previously the wording stated explicitly that `TypeGuard` is not a subtype of `bool`, now it just says that in the callable case it is. But what about other contexts? Is `def foo(x) -> Typeguard[int]` a subtype of `def foo(x) -> bool`?

erictraut: Yeah, I think it would be clearer to avoid using the special form `Callable` here and instead say "callable" to refer to all callable types. The special form is only one way to specify a callable type, and this statement applies to all callables. With that in mind, it might be good to include at least one other case (e.g. a callback protocol) in the conformance tests.

JelleZijlstra: Good points. I pushed an update to address your concerns.

gvanrossum (approved):
Spec update LGTM.

Issue #1643, opened 2024-02-29
Spec: Add PEP 705 (Read-only TypedDict items)


JelleZijlstra:
PEP-705 was just accepted. We should incorporate it into the spec and conformance test suite.

erictraut:
This is now complete.

JelleZijlstra:
We also need additions to the test suite, don't we?

JelleZijlstra:
Oh, you just did (#1654). Thanks!

PR #1650, opened 2024-03-08
Added the spec part of PEP 705 to the typing spec.

erictraut:
This is related to https://github.com/python/typing/issues/1643.

### DIFF
--- a/docs/spec/typeddict.rst
+++ b/docs/spec/typeddict.rst
@@ -579,7 +579,7 @@ may also enforce this restriction.
 
 The :pep:`alternative functional syntax <589#alternative-syntax>`
 for TypedDict also supports
-``Required[]`` and ``NotRequired[]``:
+``Required[]``, ``NotRequired[]``, and ``ReadOnly[]``:
 
 ::
 
@@ -660,13 +660,6 @@ The ``typing.ReadOnly`` type qualifier is used to indicate that an item declared
     blur["members"] = ["Damon Albarn"]  # Type check error: "members" is read-only
     blur["members"].append("Damon Albarn")  # OK: list is mutable
 
-
-Alternative functional syntax
------------------------------
-
-The :pep:`alternative functional syntax <589#alternative-syntax>` for TypedDict also supports the new type qualifier::
-
-   Band = TypedDict("Band", {"name": str, "members": ReadOnly[list[str]]})
 
 Interaction with other special types
 ------------------------------------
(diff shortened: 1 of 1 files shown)


docs/spec/typeddict.rst
```diff
@@ -637,3 +637,193 @@ In particular allowing ``Annotated[]`` to be the outermost annotation
...
+    blur: Band = {"name": "blur", "members": []}
+    blur["name"] = "Blur"  # OK: "name" is not read-only
+    blur["members"] = ["Damon Albarn"]  # Type check error: "members" is read-only
+    blur["members"].append("Damon Albarn")  # OK: list is mutable
+
+
+Alternative functional syntax
+-----------------------------
+
+The :pep:`alternative functional syntax <589#alternative-syntax>` for TypedDict also supports the new type qualifier::
```
JelleZijlstra, line 667: Can you link this to https://typing.readthedocs.io/en/latest/spec/typeddict.html#alternative-syntax instead? It would be even better if that section directly mentioned `ReadOnly`.

docs/spec/typeddict.rst
```diff
@@ -637,3 +637,193 @@ In particular allowing ``Annotated[]`` to be the outermost annotation
...
+
+    def update_a(a: A, b: B) -> None:
+        a.update(b)  # Accepted by type checker: "x" cannot be set on b
+
+Note: Nothing will ever match the ``Never`` type, so an item annotated with it must be absent.
+
+Keyword argument typing
+-----------------------
+
+:pep:`692` introduced ``Unpack`` to annotate ``**kwargs`` with a ``TypedDict``. Marking one or more of the items of a ``TypedDict`` used in this way as read-only will have no effect on the type signature of the method. However, it *will* prevent the item from being modified in the body of the function::
```
JelleZijlstra, line 812: Should ideally link to https://typing.readthedocs.io/en/latest/spec/callables.html#unpack-for-keyword-arguments, and perhaps this section should be incorporated there instead.

JelleZijlstra (commented):
Thanks! Some suggestions for better organization of the spec.

erictraut:
@JelleZijlstra, do you think these new sections should go through formal TC review? I'm thinking that if they are based solely on an approved spec and are not introducing any changes, we can skip the full review and sign-off process. Do you agree?

JelleZijlstra:
I agree, we already approved the PEP, so we don't need to go through the whole process again to approve the implementation.

JelleZijlstra (approved):

PR #1652, opened 2024-03-08
spec: Separate "Glossary" section

JelleZijlstra:
I'd like to add a glossary section for https://discuss.python.org/t/basic-terminology-for-types-and-type-forms/46741,
but felt it was better to do the housekeeping in a separate PR.

This PR makes no substantive changes; it just moves text around and uses the Sphinx glossary format. Therefore, I don't think it needs Typing Council approval.

### DIFF
--- a/docs/spec/distributing.rst
+++ b/docs/spec/distributing.rst
@@ -3,6 +3,8 @@
 Distributing type information
 =============================
 
+.. _stub-files:
+
 Stub files
 ----------
 
@@ -119,7 +121,7 @@ type checking as well.
 
 To have this file including with the package, maintainers can use existing packaging
 options such as ``package_data`` in ``setuptools``. For more details, see
-:ref:`the guide to providing type annotations <providing-type-annotations>`_.
+:ref:`the guide to providing type annotations <providing-type-annotations>`.
 
 For namespace packages (see :pep:`420`), the ``py.typed`` file should be in the
 submodules of the namespace, to avoid conflicts and for clarity.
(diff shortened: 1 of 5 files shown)


docs/spec/protocol.rst
```diff
@@ -234,7 +234,7 @@ Now the protocol ``SizedAndClosable`` is a protocol with two methods,
 ``__len__`` and ``close``. If one omits ``Protocol`` in the base class list,
 this would be a regular (non-protocol) class that must implement ``Sized``.
 Alternatively, one can implement ``SizedAndClosable`` protocol by merging
-the ``SupportsClose`` protocol from the example in the `definition`_ section
+the ``SupportsClose`` protocol from the example in the `protocol-definition`_ section
```
JelleZijlstra, line 237: This link was wrong

docs/spec/distributing.rst
```diff
@@ -119,7 +121,7 @@ type checking as well.
 
 To have this file including with the package, maintainers can use existing packaging
 options such as ``package_data`` in ``setuptools``. For more details, see
-:ref:`the guide to providing type annotations <providing-type-annotations>`_.
+:ref:`the guide to providing type annotations <providing-type-annotations>`.
```
JelleZijlstra, line 124: This link was broken (https://typing.readthedocs.io/en/latest/spec/distributing.html#packaging-type-information)

docs/spec/glossary.rst
```diff
@@ -0,0 +1,33 @@
+.. _`glossary`:
+
+Glossary
+========
+
+This section defines a few terms that may be used elsewhere in the specification.
+
+The definition of "MAY", "MUST", and "SHOULD", and "SHOULD NOT" are
+to be interpreted as described in :rfc:`2119`.
```
srittau, line 9: I'd probably leave this at the top of the spec. These are not really typing-related glossary terms, but relate to how the spec is to be interpreted.

JelleZijlstra: Good point, done

srittau (approved):
LGTM

Issue #1642, opened 2024-02-29
Spec: Add PEP 696 (Type parameter defaults)


JelleZijlstra:
PEP-696 was just accepted. We should incorporate it into the spec and the conformance test suite.

JelleZijlstra:
(FYI I set up autolinking for this repo so PEP-696 links to the PEP.)

erictraut:
The typing spec and conformance tests have now been updated, so marking this as closed.

PR #1655, opened 2024-03-09
Incorporated PEP 696 into the typing spec.

erictraut:
I mostly copy-and-pasted from the PEP. I made a few adjustments to the existing typing spec in places where it said that TypeVars default to `Any`, since that's no longer strictly true. I also adjusted certain word spellings from British to American for consistency with the rest of the typing spec.

This is part of https://github.com/python/typing/issues/1642.

### DIFF
--- a/docs/spec/generics.rst
+++ b/docs/spec/generics.rst
@@ -1964,11 +1964,11 @@ Type parameters are valid as parameters to generics inside of a
    class Bar(Generic[T, ListDefaultT]):
        def __init__(self, x: T, y: ListDefaultT): ...
 
-   reveal_type(Bar)                    # type is type[Bar[T, ListDefaultT = list[T]]]
-   reveal_type(Bar[int])               # type is type[Bar[int, list[int]]]
-   reveal_type(Bar[int]())             # type is Bar[int, list[int]]
-   reveal_type(Bar[int, list[str]]())  # type is Bar[int, list[str]]
-   reveal_type(Bar[int, str]())        # type is Bar[int, str]
+   reveal_type(Bar)                         # type is type[Bar[T, ListDefaultT = list[T]]]
+   reveal_type(Bar[int])                    # type is type[Bar[int, list[int]]]
+   reveal_type(Bar[int](0, []))             # type is Bar[int, list[int]]
+   reveal_type(Bar[int, list[str]](0, []))  # type is Bar[int, list[str]]
+   reveal_type(Bar[int, str](0, ""))        # type is Bar[int, str]
 
 Specialization Rules
 ~~~~~~~~~~~~~~~~~~~~
--- a/docs/spec/generics.rst
+++ b/docs/spec/generics.rst
@@ -1990,7 +1990,7 @@ Subclassing
 '''''''''''
 
 Generic classes with type parameters that have defaults behave similarly
-generic type aliases. That is, subclasses can be further subscripted following
+to generic type aliases. That is, subclasses can be further subscripted following
 normal subscription rules, non-overridden defaults should be substituted.
 
 ::
--- a/docs/spec/generics.rst
+++ b/docs/spec/generics.rst
@@ -1777,9 +1777,8 @@ Default Ordering and Subscription Rules
 
 The order for defaults should follow the standard function parameter
 rules, so a type parameter with no ``default`` cannot follow one with
-a ``default`` value. Doing so should ideally raise a ``TypeError`` in
-``typing._GenericAlias``/``types.GenericAlias``, and a type checker
-should flag this as an error.
+a ``default`` value. Doing so may raise a ``TypeError`` at runtime,
+and a type checker should flag this as an error.
 
 ::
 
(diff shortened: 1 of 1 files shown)


docs/spec/generics.rst
```diff
@@ -1761,6 +1763,352 @@ overloads for each possible rank is, of course, a rather cumbersome
...
+
+Default values can be provided for a TypeVar, ParamSpec, or TypeVarTuple.
+
+Default Ordering and Subscription Rules
+'''''''''''''''''''''''''''''''''''''''
+
+The order for defaults should follow the standard function parameter
+rules, so a type parameter with no ``default`` cannot follow one with
+a ``default`` value. Doing so should ideally raise a ``TypeError`` in
+``typing._GenericAlias``/``types.GenericAlias``, and a type checker
```
JelleZijlstra, line 1781:
```suggestion
a ``default`` value. Doing so may raise a ``TypeError`` at runtime, and a type checker
```
It's not the typing spec's job to say what the runtime should do.

docs/spec/generics.rst
```diff
@@ -1761,6 +1763,352 @@ overloads for each possible rank is, of course, a rather cumbersome
...
+   reveal_type(Bar[int, list[str]]())  # type is Bar[int, list[str]]
+   reveal_type(Bar[int, str]())        # type is Bar[int, str]
+
+Specialization Rules
+~~~~~~~~~~~~~~~~~~~~
+
+``Generic`` ``TypeAlias``\ es
+'''''''''''''''''''''''''''''
+
+``Generic`` ``TypeAlias``\ es can be further subscripted
```
JelleZijlstra, line 1979:
```suggestion
Generic type aliases
'''''''''''''''''''''''''

Generic type aliases can be further subscripted
```
"Generic" here refers to the concept of a generic, not the class `typing.Generic`, so it should not be monospaced. Similarly, this section should apply to all type aliases, whether or not they're created with `typing.TypeAlias`, so "TypeAlias" should also not be monospaced.

docs/spec/generics.rst
```diff
@@ -1761,6 +1763,352 @@ overloads for each possible rank is, of course, a rather cumbersome
...
+   reveal_type(MyAlias)          # type is type[SomethingWithNoDefaults[int, DefaultStrT]]
+   reveal_type(MyAlias[bool]())  # type is SomethingWithNoDefaults[int, bool]
+
+   MyAlias[bool, int]  # Invalid: too many arguments passed to MyAlias
+
+Subclassing
+'''''''''''
+
+Subclasses of ``Generic``\ s with type parameters that have defaults
+behave similarly to ``Generic`` ``TypeAlias``\ es. That is, subclasses can be
```
JelleZijlstra, line 1999:
```suggestion
Subclasses of generic classes with type parameters that have defaults
behave similarly to generic type aliases. That is, subclasses can be
```

docs/spec/generics.rst
```diff
@@ -1761,6 +1763,352 @@ overloads for each possible rank is, of course, a rather cumbersome
...
+Specialization Rules
+~~~~~~~~~~~~~~~~~~~~
+
+``Generic`` ``TypeAlias``\ es
+'''''''''''''''''''''''''''''
+
+``Generic`` ``TypeAlias``\ es can be further subscripted
+following normal subscription rules. If a type parameter has a default
+that hasn't been overridden, it should be treated like it was
+substituted into the ``TypeAlias``. However, it can be specialized
```
JelleZijlstra, line 1982:
```suggestion
substituted into the type alias. However, it can be specialized
```

JelleZijlstra (commented):
Thanks, some editorial feedback

docs/spec/generics.rst
erictraut: Good catch. I caught (and removed) one such statement, but I missed this one.

docs/spec/generics.rst
```diff
@@ -1761,6 +1763,347 @@ overloads for each possible rank is, of course, a rather cumbersome
...
+   MyAlias: TypeAlias = SomethingWithNoDefaults[int, DefaultStrT]  # Valid
+   reveal_type(MyAlias)          # type is type[SomethingWithNoDefaults[int, DefaultStrT]]
+   reveal_type(MyAlias[bool]())  # type is SomethingWithNoDefaults[int, bool]
+
+   MyAlias[bool, int]  # Invalid: too many arguments passed to MyAlias
+
+Subclassing
+'''''''''''
+
+Generic classes with type parameters that have defaults behave similarly
```
JelleZijlstra, line 1995:
```suggestion
Generic classes with type parameters that have defaults behave similarly to
```

docs/spec/generics.rst
```diff
@@ -1761,6 +1763,348 @@ overloads for each possible rank is, of course, a rather cumbersome
...
+
+This allows for a value to be used again when the type parameter to a
+generic is missing but another type parameter is specified.
+
+To use another type parameter as a default the ``default`` and the
+type parameter must be the same type (a ``TypeVar``'s default must be
+a ``TypeVar``, etc.).
+
+`This could be used on builtins.slice <https://github.com/python/typing/issues/159>`__
+where the ``start`` parameter should default to ``int``, ``stop``
```
hauntsaninja, line 1876:
```suggestion
For instance, this could be used to type something like builtins.slice,
where the ``start`` parameter should default to ``int``, ``stop``
```
nit: original language was more suited for proposal, sounds dated if typeshed starts using 696 on slice

hauntsaninja (approved):

PR #1660, opened 2024-03-15
Stubs now support more features

srittau:
* The `Union` shorthand is now fully supported.
* PEP 570 syntax for positional-only parameter is now supported.

### DIFF
--- a/docs/source/stubs.rst
+++ b/docs/source/stubs.rst
@@ -187,9 +187,8 @@ generally possible and recommended::
 Unions
 ------
 
-Declaring unions with ``Union`` and ``Optional`` is supported by all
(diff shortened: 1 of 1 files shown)


AlexWaygood:
Thanks!

henribru:
There is a leftover reference to this:

> On the other hand, type stubs should not use the positional-only syntax from [PEP 570](http://www.python.org/dev/peps/pep-0570), introduced in Python 3.8, although type checker authors are encouraged to support it.

srittau:
Thanks for pointing that out. The whole "Syntax" section is outdated.

srittau:
#1661

PR #1682, opened 2024-04-05
Conformance tests: Add mechanism for ignoring errors

### LINKED ISSUE #1681: Conformance tests: Add mechanism to ignore errors
In the conformance test suite, sometimes type checkers produce additional errors unrelated to the topic of the test. Ideally in such cases, we should write the test in a way that avoids the unrelated error, but that's not always possible. For example, we have a test `X: TypeAlias = list or set`, which tests that type checkers reject incorrect type aliases. But mypy also produces an error saying that the `or` is useless because `list` is always true. This error is correct but it's not what we're testing for.

To handle such cases, I propose adding a new key to the `.toml` output files, containing a list of error messages from that type checker that are ignored for conformance checking:
```toml
ignore_errors = [
    "Function \"list\" could always be true in boolean context",
]
```

JelleZijlstra:
Fixes #1681


conformance/results/pyre/aliases_explicit.toml
```diff
@@ -9,52 +9,52 @@ Incorrectly rejects import alias of `TypeAlias` when used to define type alias.
 Does not report invalid specialization of already-specialized generic type alias.
 """
 output = """
-aliases_explicit.py:24:0 Incompatible variable type [9]: GoodTypeAlias9 is declared to have type `TA` but is used as type `Type[typing.Callable[..., Variable[$synthetic_attribute_resolution_variable]]]`.
```
JelleZijlstra, line 12: All this is just because the line numbers got shifted by one.

PR #1683, opened 2024-04-05
Conformance tests: Add mechanism for marking that an error should appear on one of several lines

### LINKED ISSUE #1680: Conformance tests: Add mechanism to allow errors on one of multiple lines
Sometimes type checkers differ on the exact line where an error is reported, e.g. issues with decorated functions may be reported on either the decorator line or the `def` line. For the automated conformance checking that I just added in #1678, we need some way to annotate such expected errors.

For example, this incorrect code produces a mypy error on line 3 and a pyright error on line 4:

```python
from typing import final

@final
def f():
    pass
```

My proposed syntax is to add "# E or +1", meaning that an error should be reported either on the comment line or the line after. We would write the test case like this:

```python
from typing import final

@final  # E or +1: @final must be used on a method, not a global function
def f():
    pass
```

Or alternatively:
```python
from typing import final

@final
def f():  # E or -1: @final must be used on a method, not a global function
    pass
```

This syntax isn't the greatest but I haven't thought of anything better; suggestions welcome.

JelleZijlstra:
Fixes #1680


erictraut (approved):
Looks like this works quite well. Thanks!

PR #1693, opened 2024-04-06
Spec: Define "type expression", "annotation expression", "type qualifier"

JelleZijlstra:
As discussed in https://discuss.python.org/t/basic-terminology-for-types-and-type-forms/46741

This is not meant to prescribe any concrete change in specified behavior, but it makes the spec more precise and explicit. I added several crosslinks where appropriate and added glossary entries for the new terms.


docs/spec/annotations.rst
```diff
@@ -53,43 +53,154 @@ decorators ``@property``, ``@staticmethod`` and ``@classmethod``.
...
+* In a type annotation (always as part of an annotation expression)
+* The first argument to :ref:`cast() <cast>`
+* The second argument to :ref:`assert_type() <assert-type>`
+* The bounds and constraints of a ``TypeVar`` (whether created through the
+  old syntax or the native syntax in Python 3.12)
+* The definition of a type alias (whether created through the ``type`` statement,
+  the old assignment syntax, or the ``TypeAliasType`` constructor)
+* The type arguments of a generic class (which may appear in a base class
+  or in a constructor call)
+* The definitions of fields in the class-based forms for creating
```
erictraut, line 77: I think you mean the functional forms, right? The class-based forms use proper variable annotations.

docs/spec/annotations.rst
```diff
@@ -53,43 +53,154 @@ decorators ``@property``, ``@staticmethod`` and ``@classmethod``.
...
+  :ref:`TypedDict <typeddict>` and :ref:`NamedTuple <namedtuple>` types
+* The base type in the definition of a :ref:`NewType <newtype>`
+
+.. _`annotation-expression`:
+
+An *annotation expression* is an expression that is acceptable to use in
+an annotation context (a function parameter annotation, function return
+annotation, or variable annotation). Generally, an annotation expression
+is a type expression, optionally surrounded by one or more :term:`type qualifiers <type qualifier>`
+or by `Annotated`. Each type qualifier is only valid in some contexts. Note
```
erictraut, line 87: Grammar nit: "is only valid..." should be "is valid only..."

docs/spec/annotations.rst
```diff
@@ -53,43 +53,154 @@ decorators ``@property``, ``@staticmethod`` and ``@classmethod``.
...
+.. productionlist:: expression-grammar
+    annotation_expression: <Required> '[' `annotation_expression` ']'
+                         : | <NotRequired> '[' `annotation_expression` ']'
+                         : | <ReadOnly> '[' `annotation_expression`']'
+                         : | <ClassVar> '[' `annotation_expression`']'
+                         : | <Final> '[' `annotation_expression`']'
+                         : | <InitVar> '[' `annotation_expression` ']'
+                         : | <Annotated> '[' `annotation_expression` ',' expression (',' expression)* ']'
+                         : | <TypeAlias>
+                         :       (only valid in variable annotations)
```
erictraut, line 108: Grammar nit: "only valid..." should be "valid only...". See below also.

docs/spec/class-compat.rst
```diff
@@ -10,7 +10,7 @@ Class type compatibility
 
 (Originally specified in :pep:`526`.)
 
-A covariant type ``ClassVar[T_co]`` exists in the ``typing``
+A covariant :term:`type qualifier` ``ClassVar[T_co]`` exists in the ``typing``
```
erictraut, line 13: I don't think it makes sense to talk about variance in the context of `ClassVar`. In most languages, this would be a keyword (like `static` in TypeScript).

docs/spec/historical.rst
```diff
@@ -280,10 +280,10 @@ and type checkers may warn if they are used.
 --------------------------
 
 Before Python 3.10 (:pep:`604`), Python did not support the ``|`` operator
-for creating unions of types. Therefore, the ``typing.Union`` special form can also
+for creating unions of types. Therefore, the ``typing.Union`` :term:`special form` can also
 be used to create union types. Type checkers should treat the two forms as equivalent.
 
-In addition, the ``Optional`` special form provides a shortcut for a union with ``None``.
+In addition, the ``Optional`` :term:`special form` provides a shortcut for a union with ``None``.
```
erictraut, line 286: I'm not sure I'd call this a "shortcut" since `Optional[Foo]` is longer than `Foo | None`. :)

erictraut (commented):
This is a great addition to the typing spec. It provides significant clarification on a bunch of issues that have caused confusion for a long time, and it gives us a much stronger foundation to build on in the future. Thanks for doing this, Jelle! 

Overall, it looks good to me. I added a couple of minor suggestions.

docs/spec/historical.rst
Daverball: It is shorter than `Union[Foo, None]` though, so it should be sufficient to clarify `typing.Union` instead of "union". It is part of the historical section and `typing.Union` was mentioned in the paragraph just before. I'd also clarify `typing.Optional`, or drop the prefix in both paragraphs, it's weird to use a prefix in the paragraph above but not below.

docs/spec/annotations.rst
JelleZijlstra: Yes, thanks

docs/spec/annotations.rst
JelleZijlstra: Changed all instances of "only valid" to "valid only"

docs/spec/class-compat.rst
JelleZijlstra: I noticed this too but left it alone because it seemed unrelated. However, we might as well change it now too.

docs/spec/historical.rst
JelleZijlstra: Yes, I'm pretty sure this was originally meant as a "shortcut" relative to `Union[T, None]`, though even that is only a little longer than `Optional[T]`. I'll revise the sentence to merely note the equivalence.

docs/spec/annotations.rst
```diff
@@ -53,43 +53,154 @@ decorators ``@property``, ``@staticmethod`` and ``@classmethod``.
...
 
-Forward references
+String annotations
 ------------------
 
-When a type hint contains names that have not been defined yet, that
+When a type hint cannot be evaluated at runtime, that
 definition may be expressed as a string literal, to be resolved later.
+Such annotations must be plain string literals, not raw strings, byte
+strings or f-strings.
```
JelleZijlstra, line 203:
```suggestion
```

Saving this for https://discuss.python.org/t/exotic-kinds-of-string-annotations/50527

hauntsaninja (approved):

docs/spec/annotations.rst
```diff
@@ -53,42 +53,151 @@ decorators ``@property``, ``@staticmethod`` and ``@classmethod``.
...
-
-While annotations are normally the best format for type hints,
-there are times when it is more appropriate to represent them
-by a special comment, or in a separately distributed stub
-file.  (See below for examples.)
+Type and annotation expressions
+-------------------------------
+
+The terms *type expression* and *annotation expression* denote specific
+subsets of Python expressions that are used in the type system.
```
gvanrossum, line 60: Would it make sense to clarify that all type expressions are annotation expressions, but some annotation expressions aren't type expressions? (E.g. `ClassDef[...]`.)

docs/spec/annotations.rst
```diff
@@ -53,42 +53,151 @@ decorators ``@property``, ``@staticmethod`` and ``@classmethod``.
...
-``Callable``, all ABCs and stand-ins for concrete classes exported
-from ``typing`` (e.g. ``Sequence`` and ``Dict``), type variables, and
-type aliases.
+exceptions at the time the function is defined (but see :ref:`forward-references`).
+
+.. _`expression-grammar`:
+
+The following grammar describes the allowed elements of type and annotation expressions:
+
+.. productionlist:: expression-grammar
```
gvanrossum, line 99: The rendered version of this block requires a lot of horizontal scrolling. Maybe manually break up the longer lines?

docs/spec/annotations.rst
```diff
@@ -53,42 +53,151 @@ decorators ``@property``, ``@staticmethod`` and ``@classmethod``.
...
+
+Notes:
+
+* The grammar assumes the code has already been parsed as Python code, and
+  loosely follows the structure of the AST. Syntactic details like comments
+  and whitespace are ignored.
+
+* ``<Name>`` refers to a :term:`special form`. Most special forms must be imported
+  from :py:mod:`typing` or ``typing_extensions``, except for ``None``, ``type``,
+  ``tuple``, and ``InitVar``. The latter two have aliases in :py:mod:`typing`: :py:class:`typing.Type`
```
gvanrossum, line 179: I think "the latter two" no longer refers to `type` and `tuple`.

docs/spec/annotations.rst
```diff
@@ -53,42 +53,151 @@ decorators ``@property``, ``@staticmethod`` and ``@classmethod``.
...
+  and :py:class:`typing.Tuple`.  ``InitVar`` must be imported from :py:mod:`dataclasses`.
+  ``Callable`` may be imported from either :py:mod:`typing` or :py:mod:`collections.abc`.
+  Special forms may be aliased
+  (e.g., ``from typing import Literal as L``), and they may be referred to by a
+  qualified name (e.g., ``typing.Literal``). There are other special forms that are not
+  acceptable in any annotation or type expression, including ``Generic``, ``Protocol``,
+  and ``TypedDict``.
+
+* Any leaf denoted as ``name`` may also be a qualified name (i.e., ``module '.' name``).
```
gvanrossum, line 188: Or multiply-qualified, e.g. `m1.m2.name`.

docs/spec/annotations.rst
```diff
@@ -53,42 +53,151 @@ decorators ``@property``, ``@staticmethod`` and ``@classmethod``.
...
+                   : | `tuple_type_expression`
+                   : | <Annotated> '[' `type_expression` ',' expression (',' expression)* ']'
+                   : | <TypeGuard> '[' `type_expression` ']'
+                   :       (valid only in some contexts)
+                   : | <TypeIs> '[' `type_expression` ']'
+                   :       (valid only in some contexts)
+                   : | `string_annotation`
+                   :       (must evaluate to a valid `type_expression`)
+    maybe_unpacked: `type_expression` | `unpacked`
+    unpacked: '*' `unpackable``
```
gvanrossum, line 156: Stray extra backtick.

gvanrossum (commented):
Found a small number of nits, and one suggestion (breaking lines) that you can take or leave.

docs/spec/annotations.rst
JelleZijlstra: Added

docs/spec/annotations.rst
JelleZijlstra: Broke up the lines so that it no longer scrolls horizontally on my laptop.

docs/spec/annotations.rst
JelleZijlstra: Added an extra example

docs/spec/annotations.rst
JelleZijlstra: Fixed

gvanrossum (approved):

Bigdog000 (approved):

JelleZijlstra:
I think more tests would be helpful. I probably won't have time soon to work on them.

erictraut:
I reviewed the new section with test cases in mind. I think our coverage is really high already. I didn't see anything obvious that was missing. Some of the coverage is spread out among various tests. The basic type forms are tested in `annotations_typeexpr.py`, `Annotated` is covered in `qualifiers_annotated.py`, `ClassVar` is covered in `classes_classvar.py`, `Required` and `NotRequired` are covered in `typeddicts_required.py`, etc.

PR #1694, opened 2024-04-07
Spec: Replace various direct references to PEPs

JelleZijlstra:


docs/spec/literal.rst
```diff
@@ -463,19 +463,7 @@ values of the ``Status`` enum have already been exhausted::
             # 's' must be of type 'str' since all other options are exhausted
             print("Got custom status: " + s)
 
-The interaction described above is not new: it's already
```
JelleZijlstra, line 466: I removed these paragraphs since they seem to consist only of historical context, not actual specification.

srittau (approved):

PR #1695, opened 2024-04-07
Update PEPs list

JelleZijlstra:


srittau (approved):
LGTM, although I wonder whether we shouldn't use the official PEP titles.

JelleZijlstra:
For PEPs that introduce a new typing construct (e.g. TypeGuard in PEP 647), I find the name of the new construct more intuitive than the PEP's formal title (in this case, "User-defined Type Guards"). I wouldn't mind changing it, though.

We could also replace this section with a reference to https://peps.python.org/topic/typing/.

srittau:
Replacing it with a link sounds like a good idea to me, so we don't need to maintain redundant information.

PR #1697, opened 2024-04-07
Fix heading levels and a link

srittau:


srittau:
The wrong heading levels prevented the headings from showing up at all.

hauntsaninja (approved):
Thanks!

Issue #1692, opened 2024-04-06
Conformance tests: Align automated and manual scoring


JelleZijlstra:
In #1678 I added an automated tool for scoring the conformance test suite, and #1682 and #1683 added some features to support edge cases.

Now I am working on going through the remaining files where the automated conformance judgment on a file differs from the manual one (open PRs: #1684, #1685, #1687, #1688, #1689, #1690, #1691).

I'll use this issue to track the remaining discrepancies (from `conformance/src/unexpected_fails.py`):

- [x] mypy/dataclasses_usage.toml: Pass vs. Fail (#1684)
- [x] mypy/directives_version_platform.toml: Pass vs. Fail (#1685)
- [ ] mypy/namedtuples_define_functional.toml: Pass vs. Fail
- [x] mypy/overloads_basic.toml: Pass vs. Fail (#1687)
- [x] mypy/specialtypes_never.toml: Pass vs. Fail (#1688)
- [x] mypy/typeddicts_alt_syntax.toml: Pass vs. Fail (#1689)
- [x] mypy/typeddicts_type_consistency.toml: Pass vs. Fail (#1690)
- [ ] pyre/directives_reveal_type.toml: Unsupported vs. Pass
- [ ] pyre/namedtuples_define_functional.toml: Pass vs. Fail
- [x] pyright/classes_override.toml: Pass vs. Fail (#1691)
- [ ] pyright/dataclasses_frozen.toml: Pass vs. Fail
- [ ] pyright/dataclasses_slots.toml: Pass vs. Fail
- [ ] pyright/dataclasses_transform_field.toml: Pass vs. Fail
- [ ] pyright/dataclasses_transform_func.toml: Pass vs. Fail
- [ ] pyright/generics_defaults.toml: Pass vs. Fail
- [ ] pyright/literals_literalstring.toml: Pass vs. Fail
- [x] pyright/overloads_basic.toml: Pass vs. Fail (#1687)
- [ ] pyright/protocols_variance.toml: Pass vs. Fail
- [x] pyright/specialtypes_never.toml: Pass vs. Fail (#1688)
- [ ] pyright/tuples_unpacked.toml: Pass vs. Fail
- [x] pytype/dataclasses_usage.toml: Pass vs. Fail (#1684)
- [ ] pytype/directives_assert_type.toml: Pass vs. Fail
- [x] pytype/directives_version_platform.toml: Pass vs. Fail (#1685)
- [ ] pytype/generics_scoping.toml: Pass vs. Fail
[...]

PR #1701, opened 2024-04-09
Conformance tests: Ignore overlapping overload errors in LiteralString

JelleZijlstra:
Part of #1692


erictraut:
I don't have a strong opinion. I'm fine with changing the test to eliminate the condition that leads to this error. I'm also fine suppressing the error. If you prefer, we could suppress the error by adding a `# pyright: reportOverlappingOverload=false` at the top of the file or a `# pyright: ignore[reportOverlappingOverload]` on that line. (Technically, mypy should be generating an "overlapping overload" error here as well. The fact that it doesn't is presumably because it doesn't yet have full support for `LiteralString`.) Whatever you prefer is fine with me.

JelleZijlstra:
Thanks! I thought of a simple way to write the test that avoids overlapping overloads, so I'll push that in a minute.

> Technically, mypy should be generating an "overlapping overload" error here as well. The fact that it doesn't is presumably because it doesn't yet have full support for `LiteralString`.

It does detect it; in the current version of the PR I made it ignore the mypy error as well. Mypy currently treats LiteralString as a synonym of str.

PR #1702, opened 2024-04-09
Conformance tests: Fix pyright scoring for protocols_variance

JelleZijlstra:
Pyright produces warnings instead of errors for many issues in this test case.

Part of #1692

### DIFF
--- a/conformance/results/pyright/qualifiers_final_decorator.toml
+++ b/conformance/results/pyright/qualifiers_final_decorator.toml
@@ -18,3 +18,4 @@ qualifiers_final_decorator.py:126:5 - error: Function "func1" cannot be marked @
 conformance_automated = "Pass"
 errors_diff = """
 """
+ignore_errors = ["reportMissingModuleSource"]
--- a/conformance/results/pyright/protocols_definition.toml
+++ b/conformance/results/pyright/protocols_definition.toml
@@ -96,3 +96,4 @@ protocols_definition.py:341:22 - error: Expression of type "Concrete6_Bad3" cann
 conformance_automated = "Pass"
 errors_diff = """
 """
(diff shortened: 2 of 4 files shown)


conformance/results/pyright/qualifiers_final_decorator.toml
```diff
@@ -18,3 +18,4 @@ qualifiers_final_decorator.py:126:5 - error: Function "func1" cannot be marked @
 conformance_automated = "Pass"
 errors_diff = """
 """
+ignore_errors = ["reportMissingModuleSource"]
```
JelleZijlstra, line 21: Pyright (and several others) don't actually pick up the `_qualifiers_final_decorator.pyi` stub file. Leaving that for another day.

erictraut: I'm not sure what you mean by "don't actually pick up the stub file". Can you elaborate? I'm seeing pyright resolve the import with the stub file when I run the test. Are you seeing something different?

erictraut (approved):

conformance/results/pyright/qualifiers_final_decorator.toml
JelleZijlstra: I think I misinterpreted the pyright error; I thought it meant pyright couldn't find the file at all, but it seems like it means only that pyright couldn't find the `.py` file, but it picked up the `.pyi` correctly.

It does appear that pytype can't resolve the import (`File "qualifiers_final_decorator.py", line 8, in <module>: Couldn't import pyi for '_qualifiers_final_decorator' [pyi-error]`), but all other type checkers are fine. Pytype also fails this test for other reasons, so we don't need to worry about that yet.

PR #1711, opened 2024-04-11
scripts: Remove obsolete script

JelleZijlstra:
We no longer use the typing-sig mailing list.


srittau:
The pre-commit error seems unrelated, but should probably be fixed first, just to be sure.

JelleZijlstra:
I think that error was just because I had already authorized pre-commit on the repo but the config hadn't been merged in yet.

PR #1718, opened 2024-04-15
Added typing spec chapter focused on exception behavior

erictraut:
Added typing spec chapter focused on exception behavior — in particular, context managers and whether they suppress exceptions.

### DIFF
--- a/docs/spec/exceptions.rst
+++ b/docs/spec/exceptions.rst
@@ -36,7 +36,27 @@ between these two cases. This is done by examining the return type
 annotation of the ``__exit__`` method of the context manager.
 
 If the return type of the ``__exit__`` method is specifically ``bool`` or
-``Literal[True]``, a type checker should assume that exceptions *are*
+``Literal[True]``, a type checker should assume that exceptions *can be*
 suppressed. For any other return type, a type checker should assume that
 exceptions *are not* suppressed. Examples include: ``Any``, ``Literal[False]``,
 ``None``, and ``bool | None``.
+
+This convention was chosen because most context managers do not suppress
+exceptions, and it is common for their ``__exit__`` method to be annotated as
+returning ``bool | None``. Context managers that suppress exceptions are
+relatively rare, so they are considered a special case.
+
+For example, the following context manager suppresses exceptions::
+
+    class Suppress:
+        def __enter__(self) -> None:
+            pass
+
+        def __exit__(self, exc_type, exc_value, traceback) -> bool:
+            return True
+
+    with Suppress():
+        raise ValueError("This exception is suppressed")
+
+    # The exception is suppressed, so this line is reachable.
+    print("Code is reachable")
--- a/docs/spec/exceptions.rst
+++ b/docs/spec/exceptions.rst
@@ -11,7 +11,7 @@ statements affect code flow and therefore can affect type evaluation. For exampl
     try:
         some_function()
         x = 1
-    except:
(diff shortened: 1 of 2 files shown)


docs/spec/exceptions.rst
```diff
@@ -0,0 +1,42 @@
+Exceptions
+==========
+
+Some type checking behaviors, such as type narrowing and reachability analysis,
```
erictraut, line 4: I'm not sure if this introductory section is needed for context. I could be convinced to simply drop it for brevity.

hauntsaninja (approved):

maxfischer2781:
> I wonder if it's also worth having one sentence explaining why `bool | None` does not admit the possibility of exception suppression.

Yes please. Looking at this without knowing its motivation this is extremely confusing otherwise. Starting from the datamodel, one must assume that ``-> Literal[True, ...]`` is the important part *no matter what the `...` is*. So `-> bool` and `-> bool | None` would behave the same as they both include the case of `True`.  
Explaining that this rule prefers convenience (only the less common suppressing context managers need special handling) over strictness (the more common non-suppressing context managers need precise annotations) would make it clear that this is not an error but by design.

docs/spec/exceptions.rst
Hnasar: I find the concrete example helpful, as a newcomer.

I also wouldn't mind an example for the ContextManager section; I had to look up the original mypy bug report to see what this was addressing.

Maybe something like

> For example, type checkers that employ code flow analysis will be able to identify whether `return default` is reachable, depending on whether `context.__exit__` "suppresses" exceptions:
> ```
> with context():
>     return d["foo"]
> return default
> ```

docs/spec/exceptions.rst
```diff
@@ -0,0 +1,62 @@
...
+require a type checker to understand code flow. Code flow normally proceeds
+from one statement to the next, but some statements such as ``for``, ``while``
+and ``return`` can change the code flow. Similarly, ``try``/``except``/``finally``
+statements affect code flow and therefore can affect type evaluation. For example::
+
+    x = None
+    try:
+        some_function()
+        x = 1
+    except:
```
gvanrossum, line 14: Do we really want to use a bare `except` here? It's hard to prove that *no* exception can be raised, which seems necessary to prove that we always reach `x = 1`. It might be better to use a concrete exception (e.g. `ValueError` or `RuntimeError`) in the example, to make it believable that the type checker can carry out the needed analysis -- and to avoid using constructs that are not recommended as best practice.

docs/spec/exceptions.rst
```diff
@@ -0,0 +1,62 @@
...
+    # The type of `x` at this point could be None if `some_function` raises
+    # an exception or `Literal[1]` if it doesn't, so a type checker may
+    # choose to narrow its type based on this analysis.
+    reveal_type(x)  # Literal[1] | None
+
+
+Context Managers
+----------------
+
+Context managers may optionally "suppress" exceptions. When such a context
```
gvanrossum, line 26: Maybe briefly remind the reader *how* they do that? (I.e., by returning `True`.)

gvanrossum (commented):
Nits.

JelleZijlstra (approved):

erictraut:
The TC has [signed off](https://github.com/python/typing-council/issues/27) on this change, so I'm proceeding to merge it.

PR #1723, opened 2024-04-20
Expanded "Callables" chapter to include additional details.

erictraut:


docs/spec/callables.rst
```diff
@@ -347,60 +403,355 @@ generate errors in such cases.
...
+
+    class Standard(Protocol):
+        def __call__(self, a: int, b: str) -> None: ...
+
+    def func(int_args: IntKwargs, int_str_kwargs: IntStrKwargs, str_kwargs: StrKwargs):
+        f1: KwOnly = int_args  # Error: str is not subtype of int
+        f2: KwOnly = int_str_kwargs  # OK
+        f3: KwOnly = str_kwargs  # OK
+        f4: IntStrKwargs = str_kwargs  # Error: int | str is not subtype of str
+        f5: IntStrKwargs = int_str_kwargs  # OK
```
JelleZijlstra, line 662: Redundant since `int_str_kwargs` is already of type `IntStrKwargs`.

docs/spec/callables.rst
```diff
@@ -347,60 +403,355 @@ generate errors in such cases.
...
+    class IntStrKwargs(Protocol):
+        def __call__(self, **kwargs: int | str) -> None: ...
+
+    class StrKwargs(Protocol):
+        def __call__(self, *, a: int, **kwargs: str) -> None: ...
+
+    class Standard(Protocol):
+        def __call__(self, a: int, b: str) -> None: ...
+
+    def func(int_args: IntKwargs, int_str_kwargs: IntStrKwargs, str_kwargs: StrKwargs):
```
JelleZijlstra, line 657:
```suggestion
    def func(int_kwargs: IntKwargs, int_str_kwargs: IntStrKwargs, str_kwargs: StrKwargs):
```
(And also below)

JelleZijlstra (commented):
Thank you! A few minor comments.

docs/spec/callables.rst
```diff
@@ -3,47 +3,107 @@
...
------------------
+Terminology
+-----------
+
+In this section, and throughout this specification, the term  "parameter"
+refers to a named symbol associated with a function that accepts the value of
+an argument (or multiple arguments) passed to the function. The term
+"argument" refers to a value passed to a function when it is called.
+
+In Python, parameters fall into one of three categories: positional-only,
```
JelleZijlstra, line 14: I would say there are five, aligning with https://docs.python.org/3.10/library/inspect.html#inspect.Parameter.kind. It's mostly a matter of semantics, but I think it's better if we use the same definitions as the runtime inspect module. Below you already have separate sections for `*args` and `**kwargs`, consistent with seeing them as separate kinds.

docs/spec/callables.rst
```diff
@@ -347,60 +403,355 @@ generate errors in such cases.
...
+
+If ``...`` is used with signature concatenation, the ``...`` portion continues
+to mean "any conceivable set of parameters that could be compatible"::
+
+    type CallbackWithInt[**P] = Callable[Concatenate[int, P], str]
+    type CallbackWithStr[**P] = Callable[Concatenate[str, P], str]
+
+    def func(cb: Callable[[int, str], str]) -> None:
+        f1: CallbackWithInt[...] = cb  # OK
+        f2: CallbackWithStr[...] = cb  # Error
```
Daverball, line 461: I think you should also provide a simple example which just directly uses `Concatenate` without an intermediary `ParamSpec`. I had to read this example a couple of times, to convince myself that these examples are sufficient to illustrate that `Concatenate[int, ...]` also behaves in this way.
```suggestion
If ``...`` is used with signature concatenation, the ``...`` portion continues
to mean "any conceivable set of parameters that could be compatible"::

    type CallbackWithInt[**P] = Callable[Concatenate[int, P], str]
    type CallbackWithStr[**P] = Callable[Concatenate[str, P], str]

    def func(cb: Callable[[int, str], str]) -> None:
        f1: Callable[Concatenate[int, ...], str] = cb # OK
        f2: Callable[Concatenate[str, ...], str] = cb # Error
        f3: CallbackWithInt[...] = cb  # OK
        f4: CallbackWithStr[...] = cb  # Error
```

hauntsaninja (approved):

docs/spec/callables.rst
```diff
@@ -347,60 +403,394 @@ generate errors in such cases.
...
+  def bad_cb(*args: bytes, max_items: int | None) -> list[bytes]:
       ...
 
   comb: Combiner = good_cb  # OK
   comb = bad_cb  # Error! Argument 2 has incompatible type because of
-                 # different name and kind in the callback
+                 # different parameter name and kind in the callback
+
+Callback protocols and ``Callable[...]`` types can generally be used
+interchangeably.
```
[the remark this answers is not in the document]
erictraut: Yes, I agree this is a topic that needs discussion. I don't talk about method binding at all in this group of changes. I think the topic of methods (instance, class, static) and how they are meant to bind deserves its own section — probably an entire chapter. It's on my list.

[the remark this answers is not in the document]
erictraut: Here's another area that is completely unspecified currently that leads to subtle differences between callback protocols and `Callable`: https://github.com/microsoft/pyright/issues/7782.

docs/spec/callables.rst
```diff
@@ -3,47 +3,107 @@
...
-It may be useful to declare an argument as having a default
-without specifying the actual default value.  For example::
+    func(1, 2)  # OK
+    func(1, b=2)  # OK
+    func(a=1, b=2)  # Error
 
-  def foo(x: AnyStr, y: AnyStr = ...) -> AnyStr: ...
+Support for the ``/`` delimiter was introduced in Python 3.8 (:pep:`570`).
+Prior to Python 3.8, positional parameters were specified using a :ref:`naming
+convention <pos-only-double-underscore>`.
```
JelleZijlstra, line 60:
```suggestion
For compatibility with earlier versions of Python, the type system also
supports specifying positional-only parameters using a :ref:`naming
convention <pos-only-double-underscore>`.
```
Nitpick: The convention is still supported, and the convention is not part of Python itself but only of the type system.

docs/spec/callables.rst
```diff
@@ -347,60 +403,394 @@ generate errors in such cases.
...
 
-  def feeder(get_next_item: Callable[[], str]) -> None:
-      # Body
+    class Proto1[**P](Protocol):
+        def __call__(self, a: int, *args: P.args, **kwargs: P.kwargs) -> None: ...
 
-  def async_query(on_success: Callable[[int], None],
-                  on_error: Callable[[int, Exception], None]) -> None:
-      # Body
+    assert_type(func1, Callable[..., None])  # OK
```
JelleZijlstra, line 466: I think a reasonable type checker could reject this call on the grounds that it knows `func1` is an instance of the concrete function type, rather than a generic callable. Function objects have several attributes that callables in general don't.

Safer (but clunkier) ways to demonstrate the behavior could include the override example below, and defining a callback protocol that takes `*args: Any, **kwargs: Any`.

docs/spec/callables.rst
```diff
@@ -347,60 +403,394 @@ generate errors in such cases.
...
+        f5: IntStrKwargs = int_kwargs  # Error: int | str is not subtype of int
+        f6: StrKwargs = int_str_kwargs  # OK
+        f7: StrKwargs = int_kwargs  # Error: str is not subtype of int
+        f8: IntKwargs = int_str_kwargs  # OK
+        f9: IntKwargs = str_kwargs  # Error: int is not subtype of str
+        f10: Standard = int_str_kwargs  # Error: Does not accept positional arguments
+        f11: Standard = str_kwargs  # Error: Does not accept positional arguments
+
+Subtyping relationships for callable signatures that contain a ``**kwargs``
+with an unpacked ``TypedDict`` are described in the section :ref:`above <unpack-kwargs>`.
```
JelleZijlstra, line 710: Perhaps part of that section should be moved here, but that can wait for another day.

JelleZijlstra (approved):

docs/spec/callables.rst
```diff
@@ -424,22 +424,59 @@ Meaning of ``...`` in ``Callable``
...
+be compatible in kind and type, but any additional parameters are permitted::
+
+    cb3: Callable[Concatenate[int, ...], str]
+    cb3 = lambda x: str(x)  # OK
+    cb3 = lambda a, b, c: str(a)  # OK
+    cb3 = lambda : ""  # Error
+    cb3 = lambda *, a: str(a)  # Error
+
+
+If the input signature in a def statement includes both a ``*args`` and
```
gvanrossum, line 449: def statement -> function definition ?

gvanrossum (approved):
Good work! Here are some editorial nits -- take them or leave them. I will check the box in the issue as well.

docs/spec/callables.rst
JelleZijlstra: That would be better since one could argue that an `async def` is not a def statement.
```suggestion
If the input signature in a function definition includes both a ``*args`` and
```

erictraut:
Thanks for the review @gvanrossum. I incorporated all of your suggested changes.

PR #1733, opened 2024-05-09
Updated conformance results for pyright 1.1.362

erictraut:
Updated conformance results for pyright 1.1.362, which fixed two non-conformance issues but introduced a regression in another conformance test.

Improved callables_annotation test to cover a few missing cases and make the test more resilient to differences in interpretation for `assert_type`.

### DIFF
--- a/conformance/tests/callables_annotation.py
+++ b/conformance/tests/callables_annotation.py
@@ -129,14 +129,34 @@ class Proto7(Protocol):
         pass
 
 
-def func(p1: Proto1, p2: Proto2, p3: Proto3, p7: Proto7):
-    assert_type(p1, Callable[..., None])  # OK
-    assert_type(p1, Proto5[Any])  # E
-    assert_type(p2, Callable[Concatenate[int, ...], None])  # OK
-    assert_type(p3, Callable[..., None])  # E
-    assert_type(p3, Proto4[...])  # OK
+class Proto8(Protocol):
+    def __call__(self) -> None: ...
 
-    f1: Proto6 = p7  # OK
+
+def func5(
+    p1: Proto1,
+    p2: Proto2,
+    p3: Proto3,
+    p4: Proto4[...],
+    p5: Proto5[Any],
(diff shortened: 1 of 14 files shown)


conformance/tests/callables_annotation.py
```diff
@@ -129,14 +129,34 @@ def __call__(self, a: float, /, b: int, *, k: str, m: str) -> None:
...
+    ok4: Proto5[Any] = c1  # OK
+    ok5: Callable[Concatenate[int, ...], None] = p2  # OK
+    ok6: Proto2 = c2  # OK
+    ok7: Callable[..., None] = p3  # OK
+    ok8: Proto3 = c1  # OK
+    ok9: Proto4[...] = p3  # OK
+    ok10: Proto3 = p4  # OK
+    ok11: Proto6 = p7  # OK
+
+    err1: Proto5[Any] = p8  # E
```
davidhalter, line 159: @erictraut Can you explain why this has to be an error? I don't understand why Proto5[Any] cannot be assigned by p8 if ok11 can be assigned by p7. Mypy works the same way as Zuban here. I can see how Proto6 and Proto7 should be assignable when reading the spec. However I wouldn't even know how to implement that in Mypy or Zuban. Can you let me know what your thought processes were here?

erictraut: This is an error because the `__call__` method in `Proto5` requires that the callable accept `*args` and `**kwargs`, but `p8` does not meet this requirement. The spec contains a special case for signatures that explicitly specify `*args: Any, **kwargs: Any`. This is equivalent to `...` in a `Callable` annotation, and it indicates a gradual form. The special case does not apply in this case because the parameter annotations in `Proto5.__call__` are not `Any` but are instead `T_contra`. The fact that `T_contra` happens to be specialized to `Any` in this case does not matter for purposes of this special case. Refer to [this part of the spec](https://typing.python.org/en/latest/spec/callables.html#meaning-of-in-callable) for details.

PR #1737, opened 2024-05-16
Add example with overlap to overload

nineteendo:
- Issue: https://github.com/python/cpython/issues/118706

### DIFF
--- a/docs/spec/overload.rst
+++ b/docs/spec/overload.rst
@@ -77,6 +77,31 @@ that can't easily be expressed using a union or a type variable::
   def utf8(value):
       <actual implementation>
 
+You should make sure that the overloads don't overlap with a
+different return type::
+
+  @overload
+  def int2str(x: int = ...) -> str:
+      ...
+  @overload
+  def int2str(x: None = ...) -> None:
+      ...
+  def int2str(x=None):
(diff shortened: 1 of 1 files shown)


srittau:
I'm not part of the typing council, but while this is good advice, I don't think the spec is the correct place for it. This would probably be best placed in a tutorial (once we got one) or maybe one of the guides.

hauntsaninja:
This specific change reads like user-facing documentation, so I agree it doesn't fit in the spec. The guides and user facing documentation are in: https://github.com/python/typing/tree/main/docs/source (type checkers also have their own version of these, e.g. the mypy user facing docs are usually pretty good)

To my knowledge the spec does not mandate linting overlapping overloads at all. As a hard prerequisite for this, the spec would need to mandate the overload resolution ordering, which we don't currently do.

nineteendo:
As far as I can tell, that would require a new section, so I would need help with that. And mypy doesn't mention defaults either: https://mypy.readthedocs.io/en/stable/more_types.html#type-checking-the-variants
For some reason I thought `...` should also be used if the default was incompatible with the type.

nineteendo:
Do we fix it here, or should this be addressed in the mypy documentation?

PR #1741, opened 2024-05-20
Add conformance tests for Final dataclass fields

carljm:
Add conformance tests for the spec changes merged in https://github.com/python/typing/pull/1669

None of the tested type checkers support nesting Final within ClassVar in dataclasses yet, so they all fail the automated conformance check. Mypy also requires a Final assignment in a dataclass body to be "initialized" (thus effectively requiring final dataclass fields to have a default). But all support the core behavior of defining a final dataclass field, so I marked them all as Partial.

### DIFF
--- a/conformance/results/mypy/dataclasses_final.toml
+++ b/conformance/results/mypy/dataclasses_final.toml
@@ -1,4 +1,4 @@
-conformant = "Fail"
+conformant = "Partial"
 notes = """
 Wrongly requires a Final dataclass field to be initialized at class level.
 Doesn't support Final nested inside ClassVar.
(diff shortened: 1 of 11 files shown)


conformance/results/mypy/dataclasses_final.toml
```diff
@@ -0,0 +1,22 @@
+conformant = "Fail"
```
erictraut, line 1: This should be either "Unsupported" or "Partial". The summary report generator won't know what to do with "Fail".

erictraut (commented):
Looks good. One minor comment.

conformance/results/mypy/dataclasses_final.toml
carljm: I guess on second thought maybe I'll go with Partial for mypy, too. It does support the core functionality of defining a final dataclass field and preventing reassignment of it, although forcing a default value is kind of a major limitation.

PR #1748, opened 2024-05-23
Proposed clarification of spec for int/float/complex promotion

### LINKED ISSUE #1746: Clarify the float/int/complex special case
The typing spec currently says this:

> Python’s numeric types complex, float and int are not subtypes of each other, but to support common use cases, the type system contains a straightforward shortcut: when an argument is annotated as having type float, an argument of type int is acceptable; similar, for an argument annotated as having type complex, arguments of type float or int are acceptable.

This is helpful in that it clarifies that there is not a subtype relationship here. It remains unclear (to me, at least) in the following ways:

1. It implies that this special case applies only to function argument annotations, and not to any other annotation (e.g. an annotation of the type of an attribute of a class). I find this surprising, but perhaps it is intended? I think it is surprising enough that if it's intended, the wording should be even clearer, and explicitly show examples where the special case does not apply outside an argument annotation.
2. The wording "is acceptable" in the absence of a subtype relationship does not clarify how this special case actually fits into the type system; we can accept an `int` to a `float` typed argument, but then how do we type that name within the function body? If what we actually mean here is that `float` should be interpreted as `float | int`, then we should say that clearly.

There was some discussion of this on https://github.com/python/typing/issues/1663, which is about the different but related question of whether to mention the numeric tower and PEP 3141.

(This issue should be tagged as typing spec, but I don't have permissions to set labels.)

JelleZijlstra:
Fixes #1746.


conformance/tests/specialtypes_promotions.py
```diff
@@ -10,7 +12,20 @@
 
 
 def func1(f: float):
-    f.numerator  # E
+    f.numerator  # E: attribute exists on int but not float
+
+    if isinstance(f, float):
+        assert_type(f, float)
```
hauntsaninja, line 18: It's not actually clear to me what the behaviour of assert_type should be

JelleZijlstra: Oh right, because the second argument to `assert_type()` is itself a type expression. I'll adjust this test.

conformance/tests/specialtypes_promotions.py
```diff
@@ -2,6 +2,8 @@
 Tests "type promotions" for float and complex when they appear in annotations.
 """
 
+from typing import assert_type
```
erictraut, line 5: It appears this import is no longer necessary.

erictraut (approved):
Looks good to me. One minor suggested improvement in the test case, but it's fine without this change.

conformance/tests/specialtypes_promotions.py
JelleZijlstra: No, there is still an `assert_type()` call (line 20).

erictraut: Never mind! :)

conformance/tests/specialtypes_promotions.py
```diff
@@ -2,6 +2,8 @@
...
+
 # Specification: https://typing.readthedocs.io/en/latest/spec/special-types.html#special-cases-for-float-and-complex
 
 v1: float = 1
```
carljm, line 9: This line implies that a `float` variable annotation also means `float | int`, but it's not super clear (it could also be interpreted as just a coercion of literals based on the type context). I would slightly prefer to see an explicit test of the type system rules, in the form `x: int = 1; y: float = x`.

JelleZijlstra: Good point, added some more variants.

conformance/tests/specialtypes_promotions.py
```diff
@@ -2,15 +2,32 @@
...
 
 
 def func1(f: float):
-    f.numerator  # E
+    f.numerator  # E: attribute exists on int but not float
+
+    if isinstance(f, float):
+        f.hex()  # OK (attribute exists on float but not int)
+    else:
+        assert_type(f, int)
```
erictraut, line 22: Mypy doesn't pass this test, but that's not reflected in the automated test results (or the manual scoring, currently). The problem is that mypy treats this branch as unreachable, and it doesn't check `assert_type` statements in code that it considers unreachable. That means you can replace this line with `assert_type(f, <any other type>)`, and mypy will not emit an error here. Any ideas for how to write the test in a way that catches this issue in an automated way?

JelleZijlstra: Good catch. We could introduce a type error here and assert that type checkers show the error.

conformance/tests/specialtypes_promotions.py
```diff
@@ -2,15 +2,36 @@
...
-v2 = 1
+v1: int = 1
+v2: float = 1
+v3: float = v1
+v4: complex = 1.2
+v4 = 1
+
+
+def func1(f: float) -> int:
+    f.numerator  # E: attribute exists on int but not float
```
hauntsaninja, line 17: I think worth having an unguarded `f.hex()` as well

carljm (approved):
Looks good to me.

conformance/tests/specialtypes_promotions.py
carljm: This one is interesting, because it's not totally clear to me what behavior we should specify for an unguarded `.hex()` call.

The proposed spec wording, as is, would mean that this should be an error, right? Because "member `int` of `float | int` has no method `hex`." `builtins.int` in typeshed doesn't have the `hex` method. So if you want to call `hex` you have to guard it with `isinstance(float)`.

But this doesn't seem to be an error in either [mypy](https://mypy-play.net/?mypy=latest&python=3.12&gist=39ea003c9f2d03abc3e13a4f94dee708) or [pyright](https://pyright-play.net/?code=CYUwZgBGAUAeBcUA2B7AhgFwJTwFAQIlgDoALEWaLXIA). The latter is particularly interesting, since my understanding was that pyright already used the `float | int` interpretation of `float` annotations.

@erictraut What's the explanation of pyright's behavior here? Is this special-cased in some way?

erictraut: In pyright's current implementation, I have a function called [expandPromotionTypes](https://github.com/microsoft/pyright/blob/2e397e9a0a84818342712af7639114b48e274a7a/packages/pyright-internal/src/analyzer/typeEvaluator.ts#L3746) that is responsible for taking the type `float` and expanding it to `Float | int` and `complex` into `Complex | Float | int`. (I use a capitalized name here to indicate the "real" type.) 

There are three places where I initially added calls to this function:
1) `isinstance` type narrowing
2) Class pattern matching (which is the `match` statement equivalent of `isinstance` checks)
3) Attribute access ("dot") expression forms

I also 4) changed the inference logic for float literals to be inferred as `Float` rather than `float`.

I ended up backing out 3 and 4 because these two cases produced too much noise, including some false positive errors. See [this issue](https://github.com/microsoft/pyright/issues/6032), which shows some of the pain this caused pyright users.

I think it's reasonable to add 3 back, but doing so requires an additional change to avoid some of the noise. Namely, a call to `float()` or `complex()` constructors needs to evaluate to `Float` and `Complex`, respectively.

I think that adding 4 back would be problematic, especially in situations where float literals are used in list, set and dict expressions. These are problematic because the types are invariant. Consider the following code:

```python
x = [3.1] # Should the type of `[3.1]` be inferred as `list[float]` or `list[Float]`?
x.append(1) # Should this be allowed? Most devs would expect it to be!
```

I think the expression `[3.1]` should continue to be inferred as `list[float]`, which probably means that `3.1` should continue to be inferred as `float` and not `Float`. I think that's OK, but it does lead to an apparent inconsistency because `[float(3.1)]` is evaluated as `list[Float]`. The typing spec doesn't dictate type inference rules (currently), so these are not in scope for the spec, but this issue is something that type checker maintainers / authors will need to consider.

Here's [a PR](https://github.com/microsoft/pyright/pull/8024) (and "mypy_primer" run) that adds back 3 from my list above. This makes pyright conformant with the proposed language in this typing spec update (i.e. it now generates an error for `f.numerator` in the conformance test case). The "mypy_primer" run shows one change in the `dd_trace` library. It's in code that looks like this:

```python
x = {"a": float(0)} # Now evaluates to `dict[str, Float]` rather than `dict[str, float]`
x["b"] = 1 # Now a type error because `int` cannot be assigned to this `dict`
```

This change is not without consequences, but I think the impact is relatively minimal and new type errors are straightforward to fix.

carljm: I see, so pyright's implementation doesn't actually expand `float` to `float | int` when it's encountered in an annotation, but rather at certain points where the type is used.

I guess with #3 added back, pyright would also error on `f: float; f.hex()`. The `hex/fromhex` case is similar to the `is_integer` case encountered in https://github.com/microsoft/pyright/issues/6032, but I expect `is_integer` is more commonly used. And `is_integer` was added to the `int` type in Python 3.12 to avoid this problem. It doesn't seem like your mypy primer run encountered any issues with calls to `.hex` or `.fromhex`, which doesn't surprise me.

One thing that does surprise me on that mypy primer run is that the two new errors don't appear to involve attribute access (your (3)), but rather inferred types for containers (your (4)). E.g. one of the errors is on this line: https://github.com/DataDog/dd-trace-py/blob/main/ddtrace/llmobs/_integrations/bedrock.py#L43

erictraut: > so pyright's implementation doesn't actually expand float to float | int when it's encountered in an annotation

That's correct. It would be confusing for pyright & pylance users to see `float | int` in hover text, inlined type annotations, completion suggestions, `reveal_type` text, etc. if they use `float` in a type annotation. And likewise, if they specify `float | int` in a type annotation, it would be confusing to reduce that to just `float`. I try to retain the same form used in the annotation where possible.

docs/spec/special-types.rst
```diff
@@ -113,11 +113,25 @@ Special cases for ``float`` and ``complex``
...
+When a reference to the built-in type ``float`` appears in a :term:`type expression`,
+it is interpreted as if it were a union of the built-in types ``float`` and ``int``.
+Similarly, when a reference to the type ``complex`` appears, it is interpreted as
+a union of the built-in types ``complex``, ``float`` and ``int``.
+These implicit unions behave exactly like the corresponding explicit union types,
+but type checkers may choose to display them differently in user-visible output
+for clarity.
+
+Type checkers should support narrowing the type of a variable to exactly ``float``
+or ``int``, without the implicit union, through a call to ``isinstance()``::
```
grievejia, line 127: IIUC this is basically saying that if `float` appears as an argument to `isinstance()` then we should interpret it differently from the case when `float` appears in function annotations?

In that case, I'm wondering about what should happen in other cases, e.g. `cast()`, `assert_type()`, or other kinds of type forms -- should we interpret `float` to be the "real" float in those places as well? Does it make sense to have the spec discuss about the "scope" of this kind of special-casing interpretations?

JelleZijlstra: The proposed wording says that `float` should be interpreted as `float | int` in type expressions. The first argument to `cast()` is a type expression, but the second argument to `isinstance()` is not.

carljm (approved):

DiscordLiz (changes requested):
IF this is actually going to be pushed through as a "Clarification", I'd appreciate if it was clear on what the behavior with `TypeIs[float]`, and Type[T], where the parameter passed is `float`.

Both of these are *runtime* use of float, but are in an annotation context. `TypeIs` has a case where it either contradicts the guidance to treat `TypeIs` with the same narrowing rules as isinstance, or conflicts with the statement that `float` means `float | int` in an annotation context.

DiscordLiz:
There's also a case that should have an explicit ruling with overload compatability, shown here: https://discuss.python.org/t/clarifying-the-float-int-complex-special-case/54018/44 as the changes do have other effects that break existing cases.

jorenham:
In NumPy the `np.float64` scalar type is a subclass of both `np.floating` and `float`. And currently both pyright and mypy reject the assignment of `np.float64` to `int` at the moment:

```pyi
import numpy as np

x: np.float64

i: int = x  # ❌
f: float = x  # ✅
```

Will this change?

JukkaL:
I'm still in the process of figuring out the implications to mypy (which uses a differerent approach currently), in case this would surface some additional edge cases worth documenting. I'm generally positive about clarifying the spec here though.

JelleZijlstra:
Yes, this approach would imply changes to how mypy currently implements the special case. An alternative could be to specify mypy's behavior (which is essentially to pretend that int is a subclass of float), but I think that behavior is harder to understand and harder to specify precisely.

conformance/tests/specialtypes_promotions.py
```diff
@@ -2,15 +2,36 @@
 Tests "type promotions" for float and complex when they appear in annotations.
 """
 
+from typing import assert_type
```
beauxq, line 5: It's not clear to me which of these should report errors, and which shouldn't.
```py
def foo(x: float):
    assert_type(x, float | int)


def goo(x: float):
    assert_type(x, float)


def hoo(x: float | int):
    assert_type(x, float)
```

JelleZijlstra: This follows from the definitions in https://typing.python.org/en/latest/spec/annotations.html#type-and-annotation-expressions . Every time we're in a type expression, float means `float | int`.

carljm: (So to spell it out, the answer is that none of those should report an error, because "parameter annotation" and "second argument to `assert_type`" are both type expressions, so all of those examples are equivalent to each other, and every occurrence of `| int` is redundant.)

llucax:
Any news on this? It looks like it got a few approvals but it didn't have any activity for a long time.

JelleZijlstra:
I'm not sure we'll get consensus on this. Which is a shame since the current text in the spec seems unsatisfactory too.

beauxq:
I think the best road forward is for type checkers to just start following what the spec currently says.

I recognize that there are downsides to that, but I don't see a better path than that, towards getting rid of the special case.

carljm:
If we have consensus that "getting rid of the special case" is the best long-term approach, then the first necessary practical step is a re-annotation of typeshed to turn a lot of `float` argument annotations into `float | int`. This should be a no-op for type checkers that implement the special case (which is all of them, in some form), but is a pre-requisite to enable any type checker to even experiment with an option to disable the special case.

We have the same problem with all existing third-party library annotations in the world, and it's less clear to me how we can ever do a similar migration there.

I don't know if we have consensus on "remove the special case" as the ultimately desirable outcome. Personally I definitely wish the special case did not exist, but I'm not sure I see any realistic path to eliminating it.

jorenham:
> but I'm not sure I see any realistic path to eliminating it.

Maybe we could add a simple configuration option in `py.typed` that allows disabling these promotion rules on a per-project basis?

And then we might even be able to switch the default (i.e. disable int/float/complex promotions by default) after a long time (Python 3.20 or something). But that should probably depend on on how the community has adapted at that point.

davidhalter:
> > but I'm not sure I see any realistic path to eliminating it.
> 
> Maybe we could add a simple configuration option in `py.typed` that allows disabling these promotion rules on a per-project basis?
> 
> And then we might even be able to switch the default (i.e. disable int/float/complex promotions by default) after a long time (Python 3.20 or something). But that should probably depend on on how the community has adapted at that point.

I think that's a pretty sensible solution to this problem. We might even want to introduce something like [Rust editions](https://doc.rust-lang.org/edition-guide/editions/) where a specific year (e.g. the Python typing edition 2027) would break with some backwards compatibility of the older (default) edition. But that edition would need to be opt-in.

srittau:
> If we have consensus that "getting rid of the special case" is the best long-term approach, then the first necessary practical step is a re-annotation of typeshed to turn a lot of `float` argument annotations into `float | int`.

I actually have considered proposing this for typeshed. We could just replace `float` with `float | int`. This would introduce a small number of false ~~positives~~ negatives, but I wouldn't be too worried about that. Another alternative would be to introduce a temporary `Float` type alias to `int | float` that marks types that haven't been checked, yet. New code should not use that.

Overall I think we should start rather sooner than later with removing the special case, even if it will take a long time to remove.

jorenham:
> This would introduce a small number of false positives, but I wouldn't be too worried about that.

When I tried replacing all `float` with `float | int` and `complex` with `complex | float` in scipy-stubs a while back (https://github.com/scipy/scipy-stubs/pull/492), Pyright became significantly slower. But that could very well have been specific to scipy-stubs, or something that the latest pyright version already resolved, so it's certainly worth trying out in typeshed.

Anyway, I guess my point is that performance hits are also something to be mindful of when doing this.

carljm:
> I actually have considered proposing this for typeshed. We could just replace `float` with `float | int`. This would introduce a small number of false positives, but I wouldn't be too worried about that.

Replacing parameter types should never result in false positives. Function return types should ideally be modified depending on the actual behavior of the function: can it return an `int` or a `float`, or can it actually return only a `float`? False positives could occur if we replace all return types without considering the actual function behavior.

Mutable attributes are the most difficult case, but if all type checkers allow setting them to an `int` already, then it's not really a false positive to start reporting that they could have an `int` value.

srittau:
> Replacing parameter types should never result in false positives.

Sorry, I meant false negatives.

llucax:
> I actually have considered proposing this for typeshed. We could just replace float with float | int. This would introduce a small number of false positives negatives, but I wouldn't be too worried about that. Another alternative would be to introduce a temporary Float type alias to int | float that marks types that haven't been checked, yet. New code should not use that.

This is what we decided to do for our libraries, it seems to be the only option to avoid confusing behaviour. It is ugly but it works. It is a shame that this hack is needed though. We will name the type alias `FloatInt` for extra clarity (as ridiculous as it sounds), otherwise it still feels extremely weird to have this failing an exhaustiveness check in `mypy`:

```py
def f() -> Float: pass

match f():
    case float():
        pass
```

srittau:
I've opened python/typeshed#16059 with a concrete proposal for typeshed.

PR #1760, opened 2024-06-03
Typo fix: Replace `TypeVar` with `TypeVarTuple` in a docs example

Sachaa-Thanasius:
Current text from a subsection within 'TypeVarTuple':
```py
Ts1 = TypeVar('Ts1')
Ts2 = TypeVar('Ts2')

IntTuple = Tuple[int, *Ts1]
IntFloatTuple = IntTuple[float, *Ts2]  # Valid
```

It's a small thing, but those first two variables should `TypeVarTuple` instances, I think; the example doesn't really make sense otherwise. Seems to be a typo from the original PEP.

### DIFF
--- a/docs/spec/generics.rst
+++ b/docs/spec/generics.rst
@@ -1574,8 +1574,8 @@ First, type arguments to generic aliases can be variadic. For example, a
 
 ::
 
-    Ts1 = TypeVar('Ts1')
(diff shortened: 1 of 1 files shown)


erictraut (approved):
Yes, good catch.

PR #1764, opened 2024-06-03
Implemented conformance tests for Enums chapter.

erictraut:


erictraut:
Most of these tests are direct copies of the sample code in the spec. The only part that was a little tricky is that many of the type checker behaviors in the Enums chapter are optional, so many of the error conditions needed to be marked as `# E?`, and manual inspection was required to compile the "notes" section.

erictraut:
@srittau, there appears to be a general problem with the pre-commit bot. The "CLA signing" rule thinks that it is a human who hasn't signed the CLA. Any thoughts on how we can fix this?

JelleZijlstra:
I guess that's a bug in the bot, cc @ambv. It should ignore the pre-commit bot's email address somehow. We probably haven't run into this before because the main other repo where we use the bot is CPython, which doesn't use pre-commit autofixers.

I'm an admin on this repo and I can merge through the CLA bot failure, so I'll merge this (after I've reviewed it, which might take some time).

conformance/tests/enums_expansion.py
```diff
@@ -0,0 +1,78 @@
...
+        case Color.RED | Color.BLUE:
+            print("red or blue")
+        case Color.GREEN:
+            print("green")
+        case _:
+            assert_type(c, Never)  # E?
+
+
+# > This rule does not apply to classes that derive from enum.Flag because
+# > these enums allow flags to be combined in arbitrary ways.)
```
JelleZijlstra, line 39:
```suggestion
# > these enums allow flags to be combined in arbitrary ways.
```
Since the opening parentheses was omitted

conformance/tests/enums_expansion.py
```diff
@@ -0,0 +1,78 @@
...
+
+
+class CustomFlags(Flag):
+    FLAG1 = 1
+    FLAG2 = 2
+    FLAG3 = 4
+
+
+def test1(f: CustomFlags):
+    if f is CustomFlags.FLAG1 | CustomFlags.FLAG2:
```
JelleZijlstra, line 49:
```suggestion
    if f is CustomFlags.FLAG1 or f is CustomFlags.FLAG2:
```
I feel this makes the test a bit more interesting, since now a type checker that implements Flags naively should actually fail the test.

conformance/tests/enums_members.py
```diff
@@ -0,0 +1,144 @@
...
+# > a double underscore) is treated as a non-member.
+
+
+class Example2(Enum):
+    A = 1  # Member attribute
+    __B = 2  # Non-member attribute
+
+
+assert_type(Example2.A, Literal[Example2.A])
+Example.__B  # E
```
JelleZijlstra, line 126: I assume this is meant to be `Example2`, not `Example`.
```suggestion
Example2.__B  # E
```

conformance/tests/enums_expansion.py
JelleZijlstra: Oh, but apparently mypy does perform type narrowing on the code as you wrote it? That's bad.

erictraut:
@JelleZijlstra, thanks for the review. I incorporated changes based on your suggestions.

I'm not able to override the "CLA Signing" rule, so I can't merge this. If/when you feel it's ready, please merge.

JelleZijlstra (approved):

PR #1769, opened 2024-06-06
Annotated: propose better terminology

JelleZijlstra:
Arising out of: https://discuss.python.org/t/pep-746-typedmetadata-for-type-checking-of-pep-593-annotated/53834/50

Dedicated discussion thread: https://discuss.python.org/t/proposed-spec-changes-for-annotated/55807

### DIFF
--- a/docs/spec/qualifiers.rst
+++ b/docs/spec/qualifiers.rst
@@ -227,8 +227,8 @@ Here are the specific details of the syntax:
 
     * If ``Annotated`` is used in a place where arbitrary
       :term:`annotation expressions <annotation expression>` are allowed,
-      the base type may also be an annotation expression
-    * Otherwise, the base type must be a valid :term:`type expression`.
+      the base expression may also be an annotation expression
+    * Otherwise, the base expression must be a valid :term:`type expression`.
 
 * Multiple metadata objects are supported (``Annotated`` supports variadic
   arguments)::
@@ -294,13 +294,13 @@ Meaning
 The metadata provided by ``Annotated`` can be used for either static
 analysis or at runtime. If a library (or tool) encounters an instance of
 ``Annotated[T, x]`` and has no special logic for metadata ``x``, it
-should ignore it and simply treat the type as ``T``. Thus, in general,
+should ignore it and simply treat the annotation as ``T``. Thus, in general,
 any :term:`type expression` or :term:`annotation expression` may be
 wrapped in ``Annotated`` without changing its meaning. However, type
 checkers may choose to recognize particular metadata objects and use
 them to implement extensions to the standard type system.
 
-``Annotated`` metadata may apply either to the type or to the symbol
+``Annotated`` metadata may apply either to the base expression or to the symbol
 being annotated, or even to some other aspect of the program.
 
 Consuming metadata
@@ -314,20 +314,20 @@ can scan through the metadata to determine if they are of interest
 
 **Unknown metadata:** When a tool or a library does not support
 metadata or encounters an unknown metadata object it should ignore it
-and treat the annotated type as the base type. For example, when encountering
+and treat the annotation as the base expression. For example, when encountering
 a metadata object that is not an instance of ``struct2.ctype`` in the metadata
 for a name (e.g., ``Annotated[str, 'foo', struct2.ctype("<10s")]``), the
 ``struct2`` unpack method should ignore it.
 
 **Namespacing metadata:** Namespaces are not needed for metadata since
-the class used by the metadata object acts as a namespace.
+the class of the metadata object acts as a namespace.
 
 **Multiple metadata objects:** It's up to the tool consuming the metadata
 to decide whether the client is allowed to have several metadata objects on
-one type and how to merge those objects.
+one annotation and how to merge those objects.
 
 Since the ``Annotated`` type allows you to put several metadata objects of
-the same (or different) type(s) on any type, the tools or libraries
+the same (or different) type(s) on any annotation, the tools or libraries
 consuming the metadata are in charge of dealing with potential
 duplicates. For example, if you are doing value range analysis you might
 allow this::
--- a/docs/spec/qualifiers.rst
+++ b/docs/spec/qualifiers.rst
@@ -213,8 +213,8 @@ following should be allowed::
 Syntax
 ^^^^^^
 
-``Annotated`` is parameterized with a *base expression* and an arbitrary list of
-Python values that represent associated *metadata*::
+``Annotated`` is parameterized with a *base expression* and at least one
+Python values representing associated *metadata*::
 
     from typing import Annotated
 
--- a/docs/spec/qualifiers.rst
+++ b/docs/spec/qualifiers.rst
@@ -213,16 +213,17 @@ following should be allowed::
 Syntax
 ^^^^^^
 
-``Annotated`` is parameterized with a *base type* and an arbitrary list of
-Python values that represent *metadata* for the type::
+``Annotated`` is parameterized with a *base expression* and an arbitrary list of
+Python values that represent associated *metadata*::
 
     from typing import Annotated
 
-    Annotated[BaseType, Metadata1, Metadata2, ...]
+    Annotated[BaseExpr, Metadata1, Metadata2, ...]
 
 Here are the specific details of the syntax:
 
-* The base type (the first argument to ``Annotated``) must be a valid type
+* The base expression (the first argument to ``Annotated``) must be valid
+  in the context where it is being used:
 
     * If ``Annotated`` is used in a place where arbitrary
       :term:`annotation expressions <annotation expression>` are allowed,
@@ -286,6 +287,21 @@ Here are the specific details of the syntax:
 ``Annotated``. The term "annotations" is deprecated to avoid confusion
 with the parameter, return, and variable annotations that are part of
 the Python syntax.
+
+Meaning
+^^^^^^^
+
+The metadata provided by ``Annotated`` can be used for either static
+analysis or at runtime. If a library (or tool) encounters an instance of
+``Annotated[T, x]`` and has no special logic for metadata ``x``, it
+should ignore it and simply treat the type as ``T``. Thus, in general,
+any :term:`type expression` or :term:`annotation expression` may be
+wrapped in ``Annotated`` without changing its meaning. However, type
+checkers may choose to recognize particular metadata objects and use
+them to implement extensions to the standard type system.
+
+``Annotated`` metadata may apply either to the type or to the symbol
+being annotated, or even to some other aspect of the program.
 
 Consuming metadata
 ^^^^^^^^^^^^^^^^^^
--- a/docs/spec/qualifiers.rst
+++ b/docs/spec/qualifiers.rst
@@ -230,12 +230,12 @@ Here are the specific details of the syntax:
       the base expression may also be an annotation expression
     * Otherwise, the base expression must be a valid :term:`type expression`.
 
-* Multiple metadata objects are supported (``Annotated`` supports variadic
+* Multiple metadata elements are supported (``Annotated`` supports variadic
   arguments)::
 
     Annotated[int, ValueRange(3, 10), ctype("char")]
 
-* There must be at least one metadata object (``Annotated[int]`` is not valid)
+* There must be at least one metadata element (``Annotated[int]`` is not valid)
 
 * The order of the metadata is preserved and matters for equality
   checks::
@@ -251,13 +251,14 @@ Here are the specific details of the syntax:
         int, ValueRange(3, 10), ctype("char")
     ]
 
-* Duplicated metadata is not removed::
+* Duplicated metadata elements are not removed::
 
     Annotated[int, ValueRange(3, 10)] != Annotated[
         int, ValueRange(3, 10), ValueRange(3, 10)
     ]
 
-* ``Annotated`` can be used with nested and generic aliases::
+* ``Annotated`` can be used with nested and generic aliases, but only if it
+  wraps a type expression::
 
     T = TypeVar("T")
     Vec = Annotated[list[tuple[T, T]], MaxLen(10)]
@@ -293,11 +294,11 @@ Meaning
 
 The metadata provided by ``Annotated`` can be used for either static
 analysis or at runtime. If a library (or tool) encounters an instance of
-``Annotated[T, x]`` and has no special logic for metadata ``x``, it
-should ignore it and simply treat the annotation as ``T``. Thus, in general,
+``Annotated[T, x]`` and has no special logic for metadata element ``x``, it
+should ignore it and treat the expression as equivalent to ``T``. Thus, in general,
 any :term:`type expression` or :term:`annotation expression` may be
 wrapped in ``Annotated`` without changing its meaning. However, type
-checkers may choose to recognize particular metadata objects and use
+checkers may choose to recognize particular metadata elements and use
 them to implement extensions to the standard type system.
 
 ``Annotated`` metadata may apply either to the base expression or to the symbol
@@ -313,7 +314,7 @@ can scan through the metadata to determine if they are of interest
 (e.g., using ``isinstance()``).
 
 **Unknown metadata:** When a tool or a library does not support
-metadata or encounters an unknown metadata object it should ignore it
+metadata or encounters an unknown metadata element, it should ignore it
 and treat the annotation as the base expression. For example, when encountering
 a metadata object that is not an instance of ``struct2.ctype`` in the metadata
 for a name (e.g., ``Annotated[str, 'foo', struct2.ctype("<10s")]``), the
@@ -322,11 +323,11 @@ for a name (e.g., ``Annotated[str, 'foo', struct2.ctype("<10s")]``), the
 **Namespacing metadata:** Namespaces are not needed for metadata since
 the class of the metadata object acts as a namespace.
 
-**Multiple metadata objects:** It's up to the tool consuming the metadata
-to decide whether the client is allowed to have several metadata objects on
-one annotation and how to merge those objects.
+**Multiple metadata elements:** It's up to the tool consuming the metadata
+to decide whether the client is allowed to have several metadata elements on
+one annotation and how to merge those elements.
 
-Since the ``Annotated`` type allows you to put several metadata objects of
+Since the ``Annotated`` type allows you to put several metadata elements of
 the same (or different) type(s) on any annotation, the tools or libraries
 consuming the metadata are in charge of dealing with potential
 duplicates. For example, if you are doing value range analysis you might
@@ -346,9 +347,8 @@ Writing ``typing.Annotated`` everywhere can be quite verbose;
 fortunately, the ability to alias types means that in practice we
 don't expect clients to have to write lots of boilerplate code::
 
-    T = TypeVar('T')
-    Const = Annotated[T, my_annotations.CONST]
+    type Const[T] = Annotated[T, my_annotations.CONST]
 
     class C:
-        def const_method(self: Const[List[int]]) -> int:
+        def const_method(self, x: Const[list[int]]) -> int:
             ...
--- a/docs/spec/qualifiers.rst
+++ b/docs/spec/qualifiers.rst
@@ -306,7 +306,7 @@ for a name (e.g., ``Annotated[str, 'foo', struct2.ctype("<10s")]``), the
 **Namespacing metadata:** Namespaces are not needed for metadata since
 the class used by the metadata object acts as a namespace.
 
-**Multiple metadata object:** It's up to the tool consuming the metadata
+**Multiple metadata objects:** It's up to the tool consuming the metadata
 to decide whether the client is allowed to have several metadata objects on
 one type and how to merge those objects.
 
--- a/docs/spec/qualifiers.rst
+++ b/docs/spec/qualifiers.rst
@@ -308,7 +308,7 @@ being annotated, or even to some other aspect of the program.
 Consuming metadata
 ^^^^^^^^^^^^^^^^^^
 
-Ultimately, the responsibility of how to interpret the metadata (if
+Ultimately, deciding how to interpret the metadata (if
 at all) is the responsibility of the tool or library encountering the
 ``Annotated`` type. A tool or library encountering an ``Annotated`` type
 can scan through the metadata to determine if they are of interest
--- a/docs/spec/qualifiers.rst
+++ b/docs/spec/qualifiers.rst
@@ -214,7 +214,7 @@ Syntax
 ^^^^^^
 
 ``Annotated`` is parameterized with a *base expression* and at least one
-Python values representing associated *metadata*::
+Python value representing associated *metadata*::
 
     from typing import Annotated
 
--- a/docs/spec/qualifiers.rst
+++ b/docs/spec/qualifiers.rst
@@ -213,21 +213,30 @@ following should be allowed::
 Syntax
 ^^^^^^
 
-``Annotated`` is parameterized with a type and an arbitrary list of
-Python values that represent the annotations. Here are the specific
-details of the syntax:
+``Annotated`` is parameterized with a *base type* and an arbitrary list of
+Python values that represent *metadata* for the type::
 
-* The first argument to ``Annotated`` must be a valid type
+    from typing import Annotated
 
-* Multiple type annotations are supported (``Annotated`` supports variadic
+    Annotated[BaseType, Metadata1, Metadata2, ...]
+
+Here are the specific details of the syntax:
+
+* The base type (the first argument to ``Annotated``) must be a valid type
+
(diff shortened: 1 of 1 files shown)


docs/spec/qualifiers.rst
```diff
@@ -272,33 +281,38 @@ details of the syntax:
...
 
-**Namespacing annotations:** Namespaces are not needed for annotations since
-the class used by the annotations acts as a namespace.
+**Namespacing metadata:** Namespaces are not needed for metadata since
+the class used by the metadata object acts as a namespace.
 
-**Multiple annotations:** It's up to the tool consuming the annotations
-to decide whether the client is allowed to have several annotations on
-one type and how to merge those annotations.
+**Multiple metadata object:** It's up to the tool consuming the metadata
```
erictraut, line 309: object -> objects

docs/spec/qualifiers.rst
```diff
@@ -213,35 +213,44 @@ following should be allowed::
...
-Python values that represent the annotations. Here are the specific
-details of the syntax:
+``Annotated`` is parameterized with a *base type* and an arbitrary list of
+Python values that represent *metadata* for the type::
 
-* The first argument to ``Annotated`` must be a valid type
+    from typing import Annotated
 
-* Multiple type annotations are supported (``Annotated`` supports variadic
+    Annotated[BaseType, Metadata1, Metadata2, ...]
```
erictraut, line 221: As I mentioned in discourse, I have misgivings about the term "base type" here. I'd prefer if we simply used the term "type". The problem with "base type" is that it implies that the type is modified by the metadata. First, this isn't the case for static type checkers. Second, metadata doesn't need to (and often doesn't in practice) refer to the type; it can apply to the symbol that the annotation is applied to — or something else completely.

erictraut (commented):
Thanks for doing this. I think the term "metadata" is useful here, and it improves the readability of the spec.

docs/spec/qualifiers.rst
JelleZijlstra:
```suggestion
**Multiple metadata objects:** It's up to the tool consuming the metadata
```

docs/spec/qualifiers.rst
JelleZijlstra: I think it's important for clarity to have some term that specifically refers to what I call the base type. Would using "inner type" address your concern?

docs/spec/qualifiers.rst
```diff
@@ -213,35 +213,44 @@ following should be allowed::
 Syntax
 ^^^^^^
 
-``Annotated`` is parameterized with a type and an arbitrary list of
-Python values that represent the annotations. Here are the specific
-details of the syntax:
+``Annotated`` is parameterized with a *base type* and an arbitrary list of
+Python values that represent *metadata* for the type::
```
erictraut, line 217: The phrase "for the type" is problematic here. It implies that the metadata must apply to the *type*. That hasn't previously been specified, and some existing use cases for `Annotated` treat the metadata as applying to the *symbol*, not the *type*. This phrasing therefore constrains the use cases of `Annotated`. Is this your intent? If so, I think this deserves a much broader discussion before we make such a change.

docs/spec/qualifiers.rst
erictraut: I don't think "inner type" is much better than "base type". I think it reads fine without giving it an adjective. `Annotated` allows one to specify a "type" and some metadata that may or may not have anything to do with the type.

docs/spec/qualifiers.rst
JelleZijlstra:
```suggestion
Python values that represent *metadata*::
```
I definitely don't intend to narrow the possible use cases for Annotated. Would simply removing "for the type" work?

docs/spec/qualifiers.rst
JelleZijlstra: I would find it difficult to explain how PEP 746 works without some term specifically for the base type.

JelleZijlstra: Another option: We could call it the "base expression". That actually works better than type because the "base type" may be an annotation expression, not a type expression (e.g., `x: Annotated[Final[int], "some metadata"]` is legal).

docs/spec/qualifiers.rst
erictraut: Yes, I think that works.

docs/spec/qualifiers.rst
erictraut: Yeah, I think "base expression" works well.

erictraut: Or "type expression", since this type argument needs to be a valid type expression, whereas the metadata expressions do not.

JelleZijlstra: > since this type argument needs to be a valid type expression

It doesn't; annotation expressions can be wrapped in Annotated[] according to the grammar I wrote. 

However, neither mypy nor pyright handle this the way I'd expect: https://pyright-play.net/?strict=true&code=GYJw9gtgBALgngBwJYDsDmUkQWEMoCCKKYMAhjAKYAmANFAGKpkA2AUGwB4BchxpFGgG0mKVkNQwAuvQBEESuVlSoAXigBGNnF6jxkleq1c1UAEzbTFoA

```
from typing import Annotated, Final

x: Annotated[Final[int], "meta"] = 1
y: Final[int] = 1

x = 2  # no pyright error
y = 2  # pyright error
```

Pyright appears to simply ignore the Final, and mypy gives an error saying Final needs to be the outermost qualifier (https://mypy-play.net/?mypy=latest&python=3.12&gist=1069ad21b4b074710886b35744c9deaf). This is another discrepancy we'll have to resolve eventually.

docs/spec/qualifiers.rst
AlexWaygood: You could also consider something like

```suggestion
Python values that represent *associated metadata*::
```

which would leave open the possibility that it could be the _type_ that the metadata is associated with, but it could also be the _symbol_ that the metadata is associated with

docs/spec/qualifiers.rst
AlexWaygood: I like "inner type" or "wrapped type" for this. I agree with Eric in not particularly liking "base type"

docs/spec/qualifiers.rst
```diff
@@ -213,35 +213,44 @@ following should be allowed::
...
   arguments)::
 
     Annotated[int, ValueRange(3, 10), ctype("char")]
 
-* ``Annotated`` must be called with at least two arguments (
-  ``Annotated[int]`` is not valid)
+* There must be at least one metadata object (``Annotated[int]`` is not valid)
 
-* The order of the annotations is preserved and matters for equality
+* The order of the metadata is preserved and matters for equality
```
AlexWaygood, line 239: Could refer to these as metadata "elements" rather than "objects"

```suggestion
* The order of the metadata elements is preserved and matters for equality
```

JelleZijlstra: I think in this line it's fine to just say "The order of the metadata".

docs/spec/qualifiers.rst
JelleZijlstra: Thanks, I added "associated" (but outside the italics because the term being introduced is "metadata", not "associated metadata").

docs/spec/qualifiers.rst
AlexWaygood: I don't mind this line too much, but I was intending to also make a broader point -- you use "metadata objects" quite a bit; I personally prefer "metadata elements" as a term.

docs/spec/qualifiers.rst
erictraut: > annotation expressions can be wrapped in Annotated[] according to the grammar I wrote

Oh, I didn't catch that. I think that's a mistake. The first argument to `Annotated` must be a valid type. This is spelled out clearly later in the spec. The expression `Final[int]` is not a valid type, so it doesn't make sense for it to be used in `Annotated`. I think should be fixed this in the grammar, and type checkers should enforce this rule.

rchen152: I think the part of the spec that says that the first argument to `Annotated` must be a valid type is out-of-date. It looks like that restriction was relaxed in [this issue](https://bugs.python.org/issue46491). Elsewhere, the spec says that `Final` may be wrapped by `Annotated` (near the bottom of [this section](https://typing.readthedocs.io/en/latest/spec/qualifiers.html#semantics-and-examples)) and that `Required`, `NotRequired`, and `Annotated` can appear in any nesting order ([here](https://typing.readthedocs.io/en/latest/spec/typeddict.html#interaction-with-annotated)).

erictraut: Ah yes, I forgot about those provisions elsewhere in the spec. So yes, it sounds like we have an inconsistency here. I think it would be cleaner to allow only type expressions as the first type argument to `Annotated`, but I guess that ship has sailed, and it would be disruptive to undo that decision.

erictraut: I've updated pyright to honor `Final` when it's nested in `Annotated`: https://github.com/microsoft/pyright/issues/8097.

docs/spec/qualifiers.rst
```diff
@@ -272,33 +282,53 @@ details of the syntax:
...
+with the parameter, return, and variable annotations that are part of
+the Python syntax.
+
+Meaning
+^^^^^^^
+
+The metadata provided by ``Annotated`` can be used for either static
+analysis or at runtime. If a library (or tool) encounters an instance of
+``Annotated[T, x]`` and has no special logic for metadata ``x``, it
+should ignore it and simply treat the type as ``T``. Thus, in general,
```
carljm, line 297: Nit, just to clarify that T may not be precisely a type (i.e. a type expression): 
```suggestion
should ignore it and simply treat the annotation as ``T``. Thus, in general,
```

docs/spec/qualifiers.rst
```diff
@@ -272,33 +282,53 @@ details of the syntax:
...
 
-**Unknown annotations:** When a tool or a library does not support
-annotations or encounters an unknown annotation it should just ignore it
-and treat annotated type as the underlying type. For example, when encountering
-an annotation that is not an instance of ``struct2.ctype`` to the annotations
-for name (e.g., ``Annotated[str, 'foo', struct2.ctype("<10s")]``), the unpack
-method should ignore it.
+**Unknown metadata:** When a tool or a library does not support
+metadata or encounters an unknown metadata object it should ignore it
+and treat the annotated type as the base type. For example, when encountering
```
carljm, line 317:
```suggestion
metadata or encounters an unknown metadata object it should ignore it
and treat the annotation as the base expression. For example, when encountering
```

docs/spec/qualifiers.rst
```diff
@@ -272,33 +282,53 @@ details of the syntax:
...
+a metadata object that is not an instance of ``struct2.ctype`` in the metadata
+for a name (e.g., ``Annotated[str, 'foo', struct2.ctype("<10s")]``), the
+``struct2`` unpack method should ignore it.
 
-**Namespacing annotations:** Namespaces are not needed for annotations since
-the class used by the annotations acts as a namespace.
+**Namespacing metadata:** Namespaces are not needed for metadata since
+the class used by the metadata object acts as a namespace.
```
carljm, line 323:
```suggestion
the class of the metadata object acts as a namespace.
```

docs/spec/qualifiers.rst
```diff
@@ -272,33 +282,53 @@ details of the syntax:
...
 
-**Multiple annotations:** It's up to the tool consuming the annotations
-to decide whether the client is allowed to have several annotations on
-one type and how to merge those annotations.
+**Multiple metadata objects:** It's up to the tool consuming the metadata
+to decide whether the client is allowed to have several metadata objects on
+one type and how to merge those objects.
```
carljm, line 327:
```suggestion
one annotation and how to merge those objects.
```

docs/spec/qualifiers.rst
```diff
@@ -272,33 +282,53 @@ details of the syntax:
...
 
-Since the ``Annotated`` type allows you to put several annotations of
-the same (or different) type(s) on any node, the tools or libraries
-consuming those annotations are in charge of dealing with potential
+Since the ``Annotated`` type allows you to put several metadata objects of
+the same (or different) type(s) on any type, the tools or libraries
```
carljm, line 330:
```suggestion
the same (or different) type(s) on any annotation, the tools or libraries
```

docs/spec/qualifiers.rst
```diff
@@ -213,35 +213,45 @@ following should be allowed::
...
+    Annotated[BaseExpr, Metadata1, Metadata2, ...]
+
+Here are the specific details of the syntax:
+
+* The base expression (the first argument to ``Annotated``) must be valid
+  in the context where it is being used:
+
+    * If ``Annotated`` is used in a place where arbitrary
+      :term:`annotation expressions <annotation expression>` are allowed,
+      the base type may also be an annotation expression
```
carljm, line 230:
```suggestion
      the base expression may also be an annotation expression
```

docs/spec/qualifiers.rst
```diff
@@ -213,35 +213,45 @@ following should be allowed::
...
+    * Otherwise, the base type must be a valid :term:`type expression`.
```
carljm, line 231:
```suggestion
    * Otherwise, the base expression must be a valid :term:`type expression`.
```

docs/spec/qualifiers.rst
```diff
@@ -272,33 +282,53 @@ details of the syntax:
...
+The metadata provided by ``Annotated`` can be used for either static
+analysis or at runtime. If a library (or tool) encounters an instance of
+``Annotated[T, x]`` and has no special logic for metadata ``x``, it
+should ignore it and simply treat the type as ``T``. Thus, in general,
+any :term:`type expression` or :term:`annotation expression` may be
+wrapped in ``Annotated`` without changing its meaning. However, type
+checkers may choose to recognize particular metadata objects and use
+them to implement extensions to the standard type system.
+
+``Annotated`` metadata may apply either to the type or to the symbol
```
carljm, line 303:
```suggestion
``Annotated`` metadata may apply either to the base expression or to the symbol
```

docs/spec/qualifiers.rst
```diff
@@ -313,7 +343,7 @@ Aliases & Concerns over verbosity
 ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
 
 Writing ``typing.Annotated`` everywhere can be quite verbose;
-fortunately, the ability to alias annotations means that in practice we
+fortunately, the ability to alias types means that in practice we
```
carljm, line 346: I actually think this is a place where the previous wording was preferable; "annotation" here was not being used to refer to a metadata object.

carljm (commented):
This is definitely clearer, thank you!

I would prefer to be a bit more precise in our usage of the word "type" in the typing spec. A type specifies a set of possible runtime values. A type expression specifies a type. An annotation expression specifies more than just a type. `int` is a type; `Final[int]` is not a type, it's an annotation (though it does include a type, as well as additional qualifiers). So I would not prefer to use the term "base type" for the leftmost thing in `Annotated`, or to casually refer to `Annotated` as wrapping a "type".

docs/spec/qualifiers.rst
JelleZijlstra: There is no ability to alias annotations, though, only to alias types. For example, this is illegal:

```
FinalInt = Final[int]
x: FinalInt = 1
```

docs/spec/qualifiers.rst
JelleZijlstra: Edited to say "elements"

docs/spec/qualifiers.rst
```diff
@@ -213,41 +213,52 @@ following should be allowed::
 Syntax
 ^^^^^^
 
-``Annotated`` is parameterized with a type and an arbitrary list of
-Python values that represent the annotations. Here are the specific
-details of the syntax:
+``Annotated`` is parameterized with a *base expression* and an arbitrary list of
```
AlexWaygood, line 216:
```suggestion
``Annotated`` is parameterized with a *base expression* and an arbitrary number of
```

Though "arbitrary" isn't quite right either, in fact, since there has to be at least one metadata element. (But that's stated explicitly lower down, so maybe it's okay to have slightly imprecise language here.)

Anyway, you don't have to deal with these nits now if you don't want to, since the issues pre-exist this PR

docs/spec/qualifiers.rst
```diff
@@ -213,41 +213,52 @@ following should be allowed::
...
+    Annotated[BaseExpr, Metadata1, Metadata2, ...]
+
+Here are the specific details of the syntax:
+
+* The base expression (the first argument to ``Annotated``) must be valid
+  in the context where it is being used:
+
+    * If ``Annotated`` is used in a place where arbitrary
+      :term:`annotation expressions <annotation expression>` are allowed,
+      the base expression may also be an annotation expression
```
AlexWaygood, line 230:
```suggestion
      the base expression may be an annotation expression.
```

docs/spec/qualifiers.rst
```diff
@@ -213,41 +213,52 @@ following should be allowed::
...
+* The order of the metadata is preserved and matters for equality
   checks::
 
     Annotated[int, ValueRange(3, 10), ctype("char")] != Annotated[
         int, ctype("char"), ValueRange(3, 10)
     ]
 
 * Nested ``Annotated`` types are flattened, with metadata ordered
-  starting with the innermost annotation::
+  starting with the innermost::
```
AlexWaygood, line 248:
```suggestion
  starting with the innermost ``Annotated`` expression::
```

docs/spec/qualifiers.rst
```diff
@@ -213,41 +213,52 @@ following should be allowed::
...
-* Duplicated annotations are not removed::
+* Duplicated metadata elements are not removed::
 
     Annotated[int, ValueRange(3, 10)] != Annotated[
         int, ValueRange(3, 10), ValueRange(3, 10)
     ]
 
-* ``Annotated`` can be used with nested and generic aliases::
+* ``Annotated`` can be used with nested and generic aliases, but only if it
+  wraps a type expression::
```
AlexWaygood, line 261:
```suggestion
  wraps a :term:`type expression`::
```

docs/spec/qualifiers.rst
```diff
@@ -272,33 +283,53 @@ details of the syntax:
...
+"annotations" instead of "metadata" for the extra arguments to
+``Annotated``. The term "annotations" is deprecated to avoid confusion
+with the parameter, return, and variable annotations that are part of
+the Python syntax.
+
+Meaning
+^^^^^^^
+
+The metadata provided by ``Annotated`` can be used for either static
+analysis or at runtime. If a library (or tool) encounters an instance of
```
AlexWaygood, line 296:
```suggestion
or runtime analysis. If a library (or tool) encounters an instance of
```

docs/spec/qualifiers.rst
```diff
@@ -272,33 +283,53 @@ details of the syntax:
...
+``Annotated[T, x]`` and has no special logic for metadata element ``x``, it
+should ignore it and treat the expression as equivalent to ``T``. Thus, in general,
+any :term:`type expression` or :term:`annotation expression` may be
+wrapped in ``Annotated`` without changing its meaning. However, type
```
AlexWaygood, line 300:
```suggestion
wrapped in ``Annotated`` without changing the meaning of the
wrapped expression. However, type
```

docs/spec/qualifiers.rst
```diff
@@ -272,33 +283,53 @@ details of the syntax:
...
+checkers may choose to recognize particular metadata elements and use
```
AlexWaygood, line 301:
```suggestion
checkers may additionally choose to recognize particular metadata elements and use
```

AlexWaygood (commented):
Some minor nitpicks below, but basically this looks great to me. I still prefer "wrapped expression" or "underlying expression" to "base expression", but I think "base expression" is also fine

docs/spec/qualifiers.rst
JelleZijlstra: How is this new wording? 

> Anyway, you don't have to deal with these nits now if you don't want to, since the issues pre-exist this PR

I ended up rewriting most of the Annotated section so I consider any additional clarity issues to be in scope.

AlexWaygood: > How is this new wording?

Much improved, thanks!

AlexWaygood (approved):

docs/spec/qualifiers.rst
```diff
@@ -272,33 +283,54 @@ details of the syntax:
...
+``Annotated`` metadata may apply either to the base expression or to the symbol
+being annotated, or even to some other aspect of the program.
 
-Consuming annotations
-^^^^^^^^^^^^^^^^^^^^^
+Consuming metadata
+^^^^^^^^^^^^^^^^^^
 
-Ultimately, the responsibility of how to interpret the annotations (if
+Ultimately, the responsibility of how to interpret the metadata (if
```
carljm, line 311: pre-existing wording nit: redundant and awkward "responsibility of" here:
```suggestion
Ultimately, how to interpret the metadata (if
```

carljm (approved):
This looks good to me!

docs/spec/qualifiers.rst
JelleZijlstra: That reads awkward to me too, how about:
```suggestion
Ultimately, deciding how to interpret the metadata (if
```

carljm: Sure, that's fine.

docs/spec/qualifiers.rst
```diff
@@ -213,41 +213,52 @@ following should be allowed::
...
 
-* Duplicated annotations are not removed::
+* Duplicated metadata elements are not removed::
 
     Annotated[int, ValueRange(3, 10)] != Annotated[
         int, ValueRange(3, 10), ValueRange(3, 10)
     ]
 
-* ``Annotated`` can be used with nested and generic aliases::
+* ``Annotated`` can be used with nested and generic aliases, but only if it
```
erictraut, line 260: Perhaps this should clarify "...in the definition of nested and generic type aliases...". 

After all, it can be use "with aliases" (i.e. type aliases can be used within the base expression) even if the base expression is not a type expression. For example:

```python
class Foo[T](TypedDict):
    x: Annotated[Required[MyTypeAlias[T]], 1]
```

docs/spec/qualifiers.rst
```diff
@@ -272,33 +283,54 @@ details of the syntax:
     SmallInt = Annotated[int, ValueRange(0, 100)]
     SmallInt(1)  # Type error
 
+:pep:`593` and an earlier version of this specification used the term
+"annotations" instead of "metadata" for the extra arguments to
+``Annotated``. The term "annotations" is deprecated to avoid confusion
```
erictraut, line 288: I find it a bit odd to say that the "term is deprecated". The typing spec is the source of truth here, so is it necessary to clarify that it deviates in terminology from the PEP or from earlier versions of the spec? We don't do this elsewhere in the spec where we have clarified or updated terminology.

docs/spec/qualifiers.rst
```diff
@@ -272,33 +283,54 @@ details of the syntax:
...
-**Unknown annotations:** When a tool or a library does not support
-annotations or encounters an unknown annotation it should just ignore it
-and treat annotated type as the underlying type. For example, when encountering
-an annotation that is not an instance of ``struct2.ctype`` to the annotations
-for name (e.g., ``Annotated[str, 'foo', struct2.ctype("<10s")]``), the unpack
-method should ignore it.
+**Unknown metadata:** When a tool or a library does not support
+metadata or encounters an unknown metadata element, it should ignore it
+and treat the annotation as the base expression. For example, when encountering
+a metadata object that is not an instance of ``struct2.ctype`` in the metadata
```
erictraut, line 320: I find this example (including its wording) really confusing, but that's probably because it requires knowledge of `struct2`. Since this was in the previous version of the spec, it's probably fine to leave it, but maybe we can clean it up in a subsequent round of changes.

erictraut (commented):
I left a few minor comments. I don't feel that strongly about any of them. If you decide to ignore them for this PR, that's OK with me.

docs/spec/qualifiers.rst
JelleZijlstra: I think it's helpful in cases where another term was previously in use. I'm not sure there is any comparable situation elsewhere in the spec.

docs/spec/qualifiers.rst
JelleZijlstra: I'll just delete the example; I feel the text of the section is clear enough without it.

JelleZijlstra:
The Typing Council has signed off, so I'm merging this.

erictraut:
Thanks @JelleZijlstra. I scanned through the changes again to see if they required any modifications or additions to the conformance tests, but I didn't see any.

PR #1774, opened 2024-06-11
Adds links to Python Discourse and ruff to typing.readthedocs.io.

rchen152:
I noticed that the "Discussions and Support" section still linked to the typing-sig mailing list. While touching this page, I figured I might as well add ruff, which is used by typeshed, to the list of linters, and make the items in each list grammatically consistent.


hauntsaninja (approved):
Thanks!

PR #1783, opened 2024-06-21
Add PEP 712 converters to the `dataclasses` spec

thejcannon:
Picking up from https://discuss.python.org/t/would-adding-converter-to-dataclass-transform-even-need-a-pep/55125

Which stems from https://discuss.python.org/t/pep-712-adding-a-converter-parameter-to-dataclasses-field/26126 (already discussed) this PR adds `converter` as another field specifier parameter.

### DIFF
--- a/docs/spec/dataclasses.rst
+++ b/docs/spec/dataclasses.rst
@@ -365,6 +365,12 @@ positional and may use other names.
 * ``alias`` is an optional str parameter that provides an alternative
   name for the field. This alternative name is used in the synthesized
   ``__init__`` method.
+* ``converter`` is an optional parameter that specifies a single-argument 
+  callable used to convert values when assigning to the associated 
+  attribute (Including the assignment in in ``__init__``). The parameter 
+  type of this single argument provides the type of the synthesized 
+  ``__init__`` parameter associated with the field. The return type of 
+  the callable must be compatible with the field's declared type.
 
 It is an error to specify more than one of ``default``,
 ``default_factory`` and ``factory``.
@@ -525,6 +531,56 @@ This includes, but is not limited to, the following semantics:
   of ``3`` in the generated ``__init__`` method. A final class variable on a
   dataclass must be explicitly annotated as e.g. ``x: ClassVar[Final[int]] =
   3``.
+
+Converters
+^^^^^^^^^^
+
+The ``converter`` parameter can be specified in a field definition to provide a single-argument callable 
+used to convert values when assigning to the associated attribute. This feature allows for automatic type 
+conversion and validation during object initialization and attribute assignment.
+
+Converter behavior:
+
+* For non-frozen dataclasses, the converter is used for all attribute assignment, including assignment
+  of default values and direct attribute setting (e.g., ``obj.attr = value``).
+* For frozen dataclasses, the converter is only used inside a ``dataclass``-synthesized ``__init__`` 
+  when setting the attribute.
+* The converter is not used when reading attributes, as the attributes should already have been converted.
+
+Typing rules for converters:
+
+* The ``converter`` must be a callable that accepts a single positional argument.
+* The parameter type of this single argument provides the type of the synthesized ``__init__`` parameter 
+  associated with the field.
+* The return type of the callable must be compatible with the field's declared type.
+
+When used with ``default`` or ``default_factory``:
+
+* If a ``converter`` is provided alongside ``default`` or ``default_factory``, the type of the default 
+  value should be compatible with the type of the single argument to the ``converter`` callable.
+* Default values are unconditionally converted using the ``converter`` if provided.
+
+Example usage:
+
+.. code-block:: python
+
+    def str_or_none(x: Any) -> str | None:
+        return str(x) if x is not None else None
+
+    @dataclasses.dataclass
+    class Example:
+        int_field: int = dataclasses.field(converter=int)
+        str_field: str | None = dataclasses.field(converter=str_or_none)
+        path_field: pathlib.Path = dataclasses.field(
+            converter=pathlib.Path,
+            default="default/path.txt"
+        )
+
+    # Usage
+    example = Example("123", None, "some/path")
+    # example.int_field == 123
+    # example.str_field == None
+    # example.path_field == pathlib.Path("some/path")
 
 
 Undefined behavior
--- a/docs/spec/dataclasses.rst
+++ b/docs/spec/dataclasses.rst
@@ -541,10 +541,8 @@ conversion and validation during object initialization and attribute assignment.
 
 Converter behavior:
 
-* For non-frozen dataclasses, the converter is used for all attribute assignment, including assignment
+* The converter is used for all attribute assignment, including: assignment
   of default values and direct attribute setting (e.g., ``obj.attr = value``).
-* For frozen dataclasses, the converter is only used inside a ``dataclass``-synthesized ``__init__``
-  when setting the attribute.
 * The converter is not used when reading attributes, as the attributes should already have been converted.
 
 Typing rules for converters:
@@ -556,9 +554,9 @@ Typing rules for converters:
 
 When used with ``default`` or ``default_factory``:
 
-* If a ``converter`` is provided alongside ``default`` or ``default_factory``, the type of the default
-  value should be assignable to the the ``converter`` parameter.
-* Default values are unconditionally converted using the ``converter`` if provided.
+* If a ``converter`` is provided alongside ``default`` or ``default_factory``, 
+  default values are converted using the ``converter``.
+* The type of the default value should be assignable to the single-parameter of the ``converter``
 
 Example usage:
 
@@ -567,11 +565,11 @@ Example usage:
     def str_or_none(x: Any) -> str | None:
         return str(x) if x is not None else None
 
-    @dataclasses.dataclass
+    @custom_dataclass
     class Example:
-        int_field: int = dataclasses.field(converter=int)
-        str_field: str | None = dataclasses.field(converter=str_or_none)
-        path_field: pathlib.Path = dataclasses.field(
+        int_field: int = field(converter=int)
+        str_field: str | None = field(converter=str_or_none)
+        path_field: pathlib.Path = field(
             converter=pathlib.Path,
             default="default/path.txt"
         )
--- a/docs/spec/dataclasses.rst
+++ b/docs/spec/dataclasses.rst
@@ -534,7 +534,7 @@ Converters
 
 The ``converter`` parameter can be specified in a field definition to provide a callable
 used to convert values when assigning to the associated attribute. This feature allows for automatic type
-conversion and validation attribute assignment.
+conversion and validation during attribute assignment.
 
 Converter behavior:
 
--- a/docs/spec/dataclasses.rst
+++ b/docs/spec/dataclasses.rst
@@ -365,12 +365,9 @@ positional and may use other names.
 * ``alias`` is an optional str parameter that provides an alternative
   name for the field. This alternative name is used in the synthesized
   ``__init__`` method.
-* ``converter`` is an optional parameter that specifies a single-parameter
-* ``converter`` is an optional parameter that specifies a single-parameter
-  attribute (including the assignment in ``__init__``). The type
-  of this argument provides the type of the synthesized
-  ``__init__`` parameter associated with the field. The return type of
-  the callable must be assignable to the field's declared type.
+* ``converter`` is an optional parameter that specifies a
+  callable used to convert values when assigning to the associated
+  attribute.
 
 It is an error to specify more than one of ``default``,
 ``default_factory`` and ``factory``.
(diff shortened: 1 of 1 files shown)


docs/spec/dataclasses.rst
```diff
@@ -526,6 +532,56 @@ This includes, but is not limited to, the following semantics:
...
+* The converter is not used when reading attributes, as the attributes should already have been converted.
+
+Typing rules for converters:
+
+* The ``converter`` must be a callable that accepts a single positional argument.
+* The parameter type of this single argument provides the type of the synthesized ``__init__`` parameter
+  associated with the field.
+* The return type of the callable must be compatible with the field's declared type.
+
+When used with ``default`` or ``default_factory``:
```
carljm, line 557: This seems redundant with the text of the bullet point itself; could just eliminate this and include the next bullet point in the above list of typing rules directly?

docs/spec/dataclasses.rst
```diff
@@ -526,6 +532,56 @@ This includes, but is not limited to, the following semantics:
...
+Converters
+^^^^^^^^^^
+
+The ``converter`` parameter can be specified in a field definition to provide a single-argument callable
+used to convert values when assigning to the associated attribute. This feature allows for automatic type
+conversion and validation during object initialization and attribute assignment.
+
+Converter behavior:
+
+* For non-frozen dataclasses, the converter is used for all attribute assignment, including assignment
```
rchen152, line 544: I find it a little confusing for this section to refer to "non-frozen dataclasses," since that sounds to me like you're talking about stdlib dataclasses. But stdlib dataclasses don't have the `converter` parameter, right?

Maybe you could write this as "non-frozen classes impacted by `dataclass_transform`" and then just "frozen classes" or "non-frozen classes" in the rest of this section.

docs/spec/dataclasses.rst
```diff
@@ -526,6 +532,56 @@ This includes, but is not limited to, the following semantics:
...
+* Default values are unconditionally converted using the ``converter`` if provided.
+
+Example usage:
+
+.. code-block:: python
+
+    def str_or_none(x: Any) -> str | None:
+        return str(x) if x is not None else None
+
+    @dataclasses.dataclass
```
rchen152, line 570: Similar comment to above - rather than using stdlib dataclasses, I think it would be better to cook up a little example using `dataclass_transform`.

thejcannon: Oh this was just a copy paste mistake 😅

docs/spec/dataclasses.rst
```diff
@@ -365,6 +365,12 @@ positional and may use other names.
 * ``alias`` is an optional str parameter that provides an alternative
   name for the field. This alternative name is used in the synthesized
   ``__init__`` method.
+* ``converter`` is an optional parameter that specifies a single-argument
+  callable used to convert values when assigning to the associated
```
erictraut, line 369:
```suggestion
* ``converter`` is an optional parameter that specifies a single-parameter
```

docs/spec/dataclasses.rst
```diff
@@ -365,6 +365,12 @@ positional and may use other names.
 * ``alias`` is an optional str parameter that provides an alternative
   name for the field. This alternative name is used in the synthesized
   ``__init__`` method.
+* ``converter`` is an optional parameter that specifies a single-argument
```
erictraut, line 368: Should say "single-parameter", not "single-argument". Callables have parameters; argument are passed to the callable.

docs/spec/dataclasses.rst
```diff
@@ -365,6 +365,12 @@ positional and may use other names.
...
+  callable used to convert values when assigning to the associated
+  attribute (Including the assignment in ``__init__``). The parameter
```
erictraut, line 370: Nit: "Including" should not be capitalized here.

docs/spec/dataclasses.rst
```diff
@@ -365,6 +365,12 @@ positional and may use other names.
...
+  type of this single argument provides the type of the synthesized
```
erictraut, line 371: "The parameter type of this single argument" is confusing. Recommend simplifying to "The the of this parameter".

docs/spec/dataclasses.rst
```diff
@@ -526,6 +532,56 @@ This includes, but is not limited to, the following semantics:
   dataclass must be explicitly annotated as e.g. ``x: ClassVar[Final[int]] =
   3``.
 
+Converters
+^^^^^^^^^^
+
+The ``converter`` parameter can be specified in a field definition to provide a single-argument callable
```
erictraut, line 538: "single-argument" -> "single-parameter"

docs/spec/dataclasses.rst
erictraut, line 544: I agree with rchen152 that this section is confusing. Why does the behavior for frozen classes need to be discussed at all here? A converter should work anywhere that assignment works — including initial assignment via the synthesized `__init__`. If a class is frozen, then subsequent assignments will generate an exception. That's how frozen classes work, so I don't think there's a need to explain that here.

docs/spec/dataclasses.rst
```diff
@@ -526,6 +532,56 @@ This includes, but is not limited to, the following semantics:
...
+
+* For non-frozen dataclasses, the converter is used for all attribute assignment, including assignment
+  of default values and direct attribute setting (e.g., ``obj.attr = value``).
+* For frozen dataclasses, the converter is only used inside a ``dataclass``-synthesized ``__init__``
+  when setting the attribute.
+* The converter is not used when reading attributes, as the attributes should already have been converted.
+
+Typing rules for converters:
+
+* The ``converter`` must be a callable that accepts a single positional argument.
```
erictraut, line 552: "callable that accepts a single positional argument" => "a callable with one positional-only parameter"

docs/spec/dataclasses.rst
```diff
@@ -526,6 +532,56 @@ This includes, but is not limited to, the following semantics:
...
+* The parameter type of this single argument provides the type of the synthesized ``__init__`` parameter
```
erictraut, line 553: "The parameter type of this single argument" => "The type of the parameter"

docs/spec/dataclasses.rst
```diff
@@ -526,6 +532,56 @@ This includes, but is not limited to, the following semantics:
...
+  associated with the field.
+* The return type of the callable must be assignable to the field's declared type.
+
+When used with ``default`` or ``default_factory``:
+
+* If a ``converter`` is provided alongside ``default`` or ``default_factory``, the type of the default
+  value should be assignable to the type of the single argument to the ``converter`` callable.
```
erictraut, line 560: "the type of the single argument to the ``converter`` callable" => "the ``converter`` parameter"

docs/spec/dataclasses.rst
```diff
@@ -526,6 +532,56 @@ This includes, but is not limited to, the following semantics:
...
+* Default values are unconditionally converted using the ``converter`` if provided.
```
erictraut, line 561: I recommend moving this point up to the first sentence, as in:

"If a ``converter`` is provided alongside ``default`` or ``default_factory``, default values are converted using the ``converter``".

Then you can explain how a type checker should validate this behavior.

erictraut, line 561: What does "unconditionally" mean here? Perhaps just delete that word if it's not providing any meaning.

docs/spec/dataclasses.rst
```diff
@@ -526,6 +532,56 @@ This includes, but is not limited to, the following semantics:
...
+Example usage:
+
+.. code-block:: python
+
+    def str_or_none(x: Any) -> str | None:
+        return str(x) if x is not None else None
+
+    @dataclasses.dataclass
+    class Example:
+        int_field: int = dataclasses.field(converter=int)
```
erictraut, line 572: The `dataclasses.field` doesn't support converters. I recommend changing this to simply `field`, which the reader can assume is some field specifier used with a dataclass transform.

erictraut (changes requested):

docs/spec/dataclasses.rst
thejcannon:
```suggestion
* ``converter`` is an optional parameter that specifies a single-parameter
```

docs/spec/dataclasses.rst
thejcannon, line 371:
```suggestion
  attribute (including the assignment in ``__init__``). The type
  of this argument provides the type of the synthesized
```

docs/spec/dataclasses.rst
thejcannon:
```suggestion
The ``converter`` parameter can be specified in a field definition to provide a single-parameter callable
```

docs/spec/dataclasses.rst
thejcannon: Actually I'll push back on this one, since the callable can be more than a one-positional-parameter-only callable. E.g. `foo(*args)` or `foo(x, y=None)` should be compatible.

docs/spec/dataclasses.rst
thejcannon:
```suggestion
* The parameter type of the parameter provides the type of the synthesized ``__init__`` parameter
```

docs/spec/dataclasses.rst
thejcannon:
```suggestion
  value should be assignable to the the ``converter`` parameter.
```

docs/spec/dataclasses.rst
thejcannon: The intent was to ensure there's no ambiguity on frozen dataclasses, since one could mistakenly assume they don't "assign" their attributes.
I think here we'll just say "both frozen and non-frozen" and move on...

docs/spec/dataclasses.rst
thejcannon: (although I think this is more of a problem _most_ places we say single-parameter. Maybe it's worth clarifying once at the top, that by "single-parameter" we mean "single-parameter-compatible"? (or just use that slightly verbose wording throughout?)

docs/spec/dataclasses.rst
thejcannon: Good eye

docs/spec/dataclasses.rst
carljm: Let's stay away from "compatible" -- that word isn't defined to have any particular meaning in the spec.

I think the current wording here ("callable that accepts a single positional argument") is OK, and accurately describes the requirements, without placing additional unnecessary restrictions on the callable.

Unless @erictraut you were intending to suggest a more substantive change, and you believe the callable really should be restricted to only accepting one parameter, and that parameter must be positional-only? I'm not sure I see the rationale for those limitations.

carljm (changes requested):

docs/spec/dataclasses.rst
erictraut: Above, the spec talks about a "single-parameter callable", so I was trying to make the language consistent here and use "parameter" rather than "argument", but in the process I unintentionally added a new limitation. I agree that limiting it to a positional-only parameter is unnecessary. The current wording is fine.

thejcannon: Yeah this is exactly the kind of ambiguity I'm hoping to avoid by being explicit 😂

I'm happy to tweak the wording though. Would "assignable" be a correct thing to say?

Alternatively something akin to "must accept a single positional argument" (which allows any parameter spec that matches the criteria)

erictraut (approved):
LGTM

docs/spec/dataclasses.rst
```diff
@@ -526,6 +529,52 @@ This includes, but is not limited to, the following semantics:
...
+
+.. code-block:: python
+
+    def str_or_none(x: Any) -> str | None:
+        return str(x) if x is not None else None
+
+    @custom_dataclass
+    class Example:
+        int_field: int = field(converter=int)
+        str_field: str | None = field(converter=str_or_none)
```
Hnasar, line 566: since `dataclasses.field` isn't going to gain the `converter=` kwarg: in your example can you show the definition of `field` as a `field_specifier` of `custom_dataclass` to show that this is not `dataclasses.field`?

thejcannon: How about `custom_field`?

docs/spec/dataclasses.rst
```diff
@@ -526,6 +529,52 @@ This includes, but is not limited to, the following semantics:
...
+
+Typing rules for converters:
+
+* The ``converter`` must be a callable that must accept a single positional argument
+  (but may optionally accept other arguments, which are ignored for typing purposes).
+* The type of the first positional parameter provides the type of the synthesized ``__init__`` parameter
+  associated with the field.
+* The return type of the callable must be assignable to the field's declared type.
+* If ``default`` or ``default_factory`` are provided, the type of the default value should be
+  assignable to the first positional parameter of the ``converter``
```
rchen152, line 554: grammar nit: missing period at the end of the sentence

rchen152 (approved):

docs/spec/dataclasses.rst
thejcannon:
```suggestion
  assignable to the first positional parameter of the ``converter``.
```

docs/spec/dataclasses.rst
```diff
@@ -365,6 +365,9 @@ positional and may use other names.
 * ``alias`` is an optional str parameter that provides an alternative
   name for the field. This alternative name is used in the synthesized
   ``__init__`` method.
+* ``converter`` is an optional parameter that specifies a
+  callable used to convert values when assigning to the associated
+  attribute.
```
carljm, line 370: Wording nit: "associated attribute" feels like very specific terminology, but it's not used in describing any of the other field specifier parameters. If there is something particularly unique to `converter` that we are trying to communicate with this wording, maybe we should communicate it more explicitly? Otherwise, it looks more consistent to just use "field" rather than "associated attribute."
```suggestion
  callable used to convert values when assigning to the field.
```

docs/spec/dataclasses.rst
```diff
@@ -526,6 +529,52 @@ This includes, but is not limited to, the following semantics:
   dataclass must be explicitly annotated as e.g. ``x: ClassVar[Final[int]] =
   3``.
 
+Converters
+^^^^^^^^^^
+
+The ``converter`` parameter can be specified in a field definition to provide a callable
+used to convert values when assigning to the associated attribute. This feature allows for automatic type
```
carljm, line 536: Here again, is "associated attribute" meaningfully different from just saying "field"?

carljm (approved):
One wording nit comment, but the substance of this change looks good to me. Thanks for your work on this!

docs/spec/dataclasses.rst
```diff
@@ -526,6 +529,52 @@ This includes, but is not limited to, the following semantics:
...
+
+* The converter is used for all attribute assignment, including assignment
+  of default values, assignment in synthesized ``__init__`` methods
+  and direct attribute setting (e.g., ``obj.attr = value``).
+* The converter is not used when reading attributes, as the attributes should already have been converted.
+
+Typing rules for converters:
+
+* The ``converter`` must be a callable that must accept a single positional argument
+  (but may optionally accept other arguments, which are ignored for typing purposes).
```
JelleZijlstra, line 549: Provided the other arguments are optional (i.e., they have a default or are variadic). Hopefully that's clear but maybe we can use a wording that makes it more explicit.

JelleZijlstra (approved):
Looks good, one nitpick.

thejcannon:
Screenshot 2024-07-20 at 6 59 05 AM

🫡 This is a high point in my career 😂

PR #1798, opened 2024-07-14
Add guide to type narrowing

JelleZijlstra:
Some of the text derives from the "How to teach this" section of PEP 742.
The asyncio example is based on a sample written by Liz King in
https://discuss.python.org/t/problems-with-typeis/55410.


erictraut (commented):
Thanks for writing this! Looks good overall. I added a couple of minor suggestions. Feel free to use or ignore as you see fit.

docs/source/type_narrowing.rst
```diff
@@ -0,0 +1,324 @@
...
+
+    def is_point(x: object) -> TypeIs[Point]:
+        return (
+            isinstance(x, dict)
+            and all(isinstance(key, str) for key in x)
+            and "x" in x
+            and "y" in x
+            and isinstance(x["x"], int)
+            and isinstance(x["y"], int)
+        )
```
AlexWaygood, line 137:
```suggestion
    def is_point(obj: object) -> TypeIs[Point]:
        return (
            isinstance(obj, dict)
            and all(isinstance(key, str) for key in obj)
            and isinstance(obj.get("x"), int)
            and isinstance(obj.get("y"), int)
        )
```

docs/source/type_narrowing.rst
```diff
@@ -0,0 +1,324 @@
...
+``TypeIs`` and ``TypeGuard``
+----------------------------
+
+:py:data:`typing.TypeIs` and :py:data:`typing.TypeGuard` are both tools for narrowing the type of a variable
+based on a user-defined function. Both can be used to annotate functions that take an
+argument and return a boolean depending on whether the input argument is compatible with
+the narrowed type. These function can then be used in ``if`` checks to narrow the type
+of a variable.
+
+``TypeIs`` usually has the most intuitive behavior, but it
```
AlexWaygood, line 150:
```suggestion
``TypeIs`` usually has the more intuitive behavior, but it
```

docs/source/type_narrowing.rst
```diff
@@ -0,0 +1,324 @@
...
+  only for positive integers.
+
+``TypeIs`` and ``TypeGuard`` differ in the following ways:
+
+* ``TypeIs`` requires the narrowed type to be a subtype of the input type, while
+  ``TypeGuard`` does not.
+* When a ``TypeGuard`` function returns ``True``, type checkers narrow the type of the
+  variable to exactly the ``TypeGuard`` type. When a ``TypeIs`` function returns ``True``,
+  type checkers can infer a more precise type combining the previously known type of the
+  variable with the ``TypeIs`` type. (Technically, this is known as an intersection type.)
```
AlexWaygood, line 167: Not sure "technically" adds much here: the whole document is technical :)

```suggestion
  variable with the ``TypeIs`` type. (This is known as an "intersection type".)
```

You could possibly say

```suggestion
  variable with the ``TypeIs`` type. (In type theory, this is known as an
  "intersection type".)
```

docs/source/type_narrowing.rst
```diff
@@ -0,0 +1,324 @@
...
+
+    def takes_bool_list(x: list[bool]) -> None:
+        maybe_mutate_list(x)
+        reveal_type(x)  # list[bool]
+        assert all(isinstance(i, bool) for i in x)  # fails at runtime
+
+    takes_bool_list([True, False])
+
+To avoid this problem, use ``TypeIs`` instead of ``TypeGuard`` where possible.
+If you must use ``TypeGuard``, avoid narrowing across incompatible types.
```
AlexWaygood, line 252: Possibly some better (and certainly more general) advice is to try (though it's not always possible) to use covariant, immutable types for parameter annotations wherever possible. If you do so, there will be many more cases where you'll be able to use `TypeIs`, and so the problem here just arises much less in the first place

docs/source/type_narrowing.rst
```diff
@@ -0,0 +1,324 @@
+**************
+Type Narrowing
+**************
+
+Python programs often contain values that can take on multiple types and
+that are distinguished by a conditional check at runtime. For example, here
+the variable *name* can be either a ``str`` or ``None``, and the
+``if name is not None`` narrows it down to just ``str``::
```
AlexWaygood, line 8:
```suggestion
Python programs often contain symbols that take on multiple types within a
single given scope and that are distinguished by a conditional check at
runtime. For example, here the variable *name* can be either a ``str`` or
``None``, and the ``if name is not None`` narrows it down to just ``str``::
```

docs/source/type_narrowing.rst
```diff
@@ -0,0 +1,324 @@
...
+patterns include:
+
+* ``if x is not None``
+* ``if x``
+* ``if isinstance(x, SomeType)``
+* ``if callable(x)``
+
+In addition to narrowing local variables, type checkers usually also support
+narrowing instance attributes and sequence members, such as
+``if x.some_attribute is not None`` or ``if x[0] is not None``.
```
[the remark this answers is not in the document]
JelleZijlstra: Pyre seems to do type narrowing here too: https://pyre-check.org/play?input=%23%20pyre-strict%0A%0Afrom%20typing%20import%20*%0A%0Afrom%20dataclasses%20import%20dataclass%0A%0A%40dataclass%0Aclass%20X%3A%0A%20%20%20%20a%3A%20int%20%7C%20None%0A%0Adef%20f(x%3A%20X)%20-%3E%20None%3A%0A%20%20%20%20if%20x.a%20is%20not%20None%3A%0A%20%20%20%20%20%20%20%20reveal_type(x.a)%0A%20%20%20%20%20%20%20%20print(x.a%20%2B%201)%0A%20%20%20%20print(x.a%20%2B%201)

I'm not sure why this shouldn't be mentioned. It's something users commonly need and ask for. While it is somewhat more obviously unsafe than some other narrowing patterns, there's few kinds of type narrowing that are 100% safe.

docs/source/type_narrowing.rst
JelleZijlstra: Thanks, I think the first suggestion is best (when in doubt, I like to go with the less verbose version).

docs/source/type_narrowing.rst
JelleZijlstra: Maybe this?
```suggestion
To avoid this problem, use ``TypeIs`` instead of ``TypeGuard`` where possible.
If you must use ``TypeGuard``, avoid narrowing across incompatible types.
Prefer using covariant, immutable types in parameter annotations (e.g.,
``Sequence`` or ``Iterable`` instead of ``list``. If you do this, it is more likely
that you'll be able to use ``TypeIs`` to implement your type narrowing functions.
```

docs/source/type_narrowing.rst
```diff
@@ -0,0 +1,324 @@
...
+
+    def maybe_greet(name: str | None) -> None:
+        if name is not None:
+            print("Hello, " + name)
+
+This technique is called *type narrowing*.
+In order to avoid false positives on such code, type checkers understand
+various kinds of conditionals that are used to narrow types in Python code.
+The exact set of type narrowing constructs that a type checker understands
+is not specified and varies across type checkers. Commonly understood type
```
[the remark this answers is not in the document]
JelleZijlstra: I went with just "patterns"; that seems clear in context.

docs/source/type_narrowing.rst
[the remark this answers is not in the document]
AlexWaygood: > Pyre seems to do type narrowing here too:

Correct, but if you insert a single intermediate function call, it decides that it can no longer assume that the attribute wasn't modified in between the two accesses by the intermediate function call, and no longer does the type narrowing: https://pyre-check.org/play?input=%23%20pyre-strict%0A%0Afrom%20typing%20import%20*%0A%0Afrom%20dataclasses%20import%20dataclass%0A%0A%40dataclass%0Aclass%20X%3A%0A%20%20%20%20a%3A%20int%20%7C%20None%0A%0Adef%20do_something_else()%20-%3E%20None%3A%0A%20%20%20%20pass%0A%0Adef%20f(x%3A%20X)%20-%3E%20None%3A%0A%20%20%20%20if%20x.a%20is%20not%20None%3A%0A%20%20%20%20%20%20%20%20do_something_else()%0A%20%20%20%20%20%20%20%20reveal_type(x.a)%0A%20%20%20%20%20%20%20%20print(x.a%20%2B%201)%0A%20%20%20%20print(x.a%20%2B%201)

> I'm not sure why this shouldn't be mentioned.

Well, I just worry about documenting behaviour here that isn't standardised between all major type checkers. It might be pretty confusing for users of type checkers with differing behaviour.

docs/source/type_narrowing.rst
AlexWaygood: That looks good to me! Though you need a closing parenthesis after "instead of `list`"

docs/source/type_narrowing.rst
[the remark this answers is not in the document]
JelleZijlstra: The text is already pretty clear that the exact type narrowing behaviors differ among type checkers, but I'll add another line to clarify that.

I think it's important to mention this behavior because it's a common pattern in practice, so people are likely to run into this.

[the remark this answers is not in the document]
AlexWaygood: Thanks, I think the extra line definitely helps

docs/source/type_narrowing.rst
```diff
@@ -0,0 +1,323 @@
...
+
+Consult your type checker's documentation for more information on the type
+narrowing constructs it supports.
+
+The type system also includes two ways to create *user-defined* type narrowing
+functions: :py:data:`typing.TypeIs` and :py:data:`typing.TypeGuard`. These
+are useful if you want to reuse a more complicated check in multiple places, or
+you use a check that the type checker doesn't understand. In these cases, you
+can define a ``TypeIs`` or ``TypeGuard`` function to perform the check and allow type checkers
+to use it to narrow the type of a variable. Among the two, ``TypeIs`` usually
```
carljm, line 39: nit: for only two items, I would usually prefer "Between" over "Among"
```suggestion
to use it to narrow the type of a variable. Between the two, ``TypeIs`` usually
```

docs/source/type_narrowing.rst
```diff
@@ -0,0 +1,323 @@
...
+
+``TypeIs`` usually has the more intuitive behavior, but it
+introduces more restrictions. ``TypeGuard`` is the right tool to use if:
+
+* You want to narrow to a type that is not compatible with the input type, for example
+  from ``list[object]`` to ``list[int]``.  ``TypeIs`` only allows narrowing between
+  compatible types.
+* Your function does not return ``True`` for all input values that are compatible with
+  the narrowed type. For example, you could have a ``TypeGuard[int]`` that returns ``True``
+  only for positive integers.
```
carljm, line 157: Since you already use "assignable to" below (and link to the glossary), maybe prefer "assignable to" over "compatible with" in these points as well?

carljm (approved):
This looks good to me! Thanks for writing it up.

docs/source/type_narrowing.rst
AlexWaygood: It looks like you marked this as resolved but didn't push the change, was that deliberate? :)

carljm:
Oh, it might also be useful to mention `__subclasscheck__` and `__instancecheck__` in combination with `isinstance` guards as another source of unsafety in narrowing?

docs/source/type_narrowing.rst
```diff
@@ -0,0 +1,323 @@
...
+
+    def maybe_greet(name: str | None) -> None:
+        if name is not None:
+            print("Hello, " + name)
+
+This technique is called *type narrowing*.
+To avoid false positives on such code, type checkers understand
+various kinds of conditional checks that are used to narrow types in Python code.
+The exact set of type narrowing constructs that a type checker understands
+is not specified and varies across type checkers. Commonly understood
```
superbobry, line 18: Should we have a minimum required set of narrowing constructs? The one bellow might as well be one with a few additions, e.g. `issubclass`.

docs/source/type_narrowing.rst
```diff
@@ -0,0 +1,323 @@
...
+patterns include:
+
+* ``if x is not None``
+* ``if x``
+* ``if isinstance(x, SomeType)``
+* ``if callable(x)``
+
+In addition to narrowing local variables, type checkers usually also support
+narrowing instance attributes and sequence members, such as
+``if x.some_attribute is not None`` or ``if x[0] is not None``, though the exact
```
superbobry, line 28: Is it worth saying that such narrowing is potentially unsound but still useful in practice?

docs/source/type_narrowing.rst
```diff
@@ -0,0 +1,323 @@
...
+This technique is called *type narrowing*.
+To avoid false positives on such code, type checkers understand
+various kinds of conditional checks that are used to narrow types in Python code.
+The exact set of type narrowing constructs that a type checker understands
+is not specified and varies across type checkers. Commonly understood
+patterns include:
+
+* ``if x is not None``
+* ``if x``
+* ``if isinstance(x, SomeType)``
```
superbobry, line 23: Would it be useful to have an example of narrowing to a subclass? E.g.

```
def f(x: Base):
  if isinstance(x, SomeImpl): ...
```

docs/source/type_narrowing.rst
AlexWaygood: > Should we have a minimum required set of narrowing constructs?

That would require a change to the spec, which would have to go through the typing council. This PR doesn't propose a change to the spec, it just adds a new page of user-facing documentation (it's descriptive, not prescriptive).

docs/source/type_narrowing.rst
```diff
@@ -0,0 +1,323 @@
...
+        return isinstance(x, int)
+
+    # Incorrect: does not return True for all ints
+    def is_positive_int(x: object) -> TypeIs[int]:
+        return isinstance(x, int) and x > 0
+
+    # Incorrect: returns True for some non-ints
+    def is_real_number(x: object) -> TypeIs[int]:
+        return isinstance(x, (int, float))
+
```
srittau, line 105: These two examples would also be good to have in the `TypeIs` documentation at docs.python.org. Also, the `TypeIs` and `TypeGuard` docs could link to this document when it gets merged.

srittau (approved):
LGTM, with or without the changes suggested by others.

srittau:
I'd like to get #1802 in first, so this guide can start off at the right URL.

docs/source/type_narrowing.rst
JelleZijlstra: As the document explains later, virtually all narrowing is potentially unsound.

docs/source/type_narrowing.rst
JelleZijlstra: Since this behavior is unspecified and `isinstance()` behavior does vary across type checkers, I'd prefer not to go into too much detail.

docs/source/type_narrowing.rst
JelleZijlstra: Thanks, good point. I wrote that section of the PEP before your spec change :)

JelleZijlstra: There is also a usage of "compatible" on line 87, but there it refers to runtime compatibility so I don't think "assignable" is right; there I'll simply say "is of type T".

JelleZijlstra:
@carljm good call, added a section on `isinstance()`.

srittau:
Please move to the guides directory when resolving conflicts!

docs/source/type_narrowing.rst
superbobry: Sure, I guess this is a request for a spec update then :)

docs/source/type_narrowing.rst
superbobry: It might still be useful to explain where the unsoundness comes from. E.g. I think something like `if x is not None` can only be unsound in the presence of dynamic features, such as `globals()["x"] = ...`.

superbobry: Thanks, I somehow missed it during my initial review. The section looks good.

docs/source/type_narrowing.rst
JelleZijlstra: Covering this behavior in the spec is probably useful but it's a much larger project. This PR is meant as user-facing documentation, not a full spec.

PR #1799, opened 2024-07-14
Glossary: Improve definition of "gradual form"

JelleZijlstra:
The current wording "A gradual form is a type expression" is not strictly correct, because `...` is not a type expression. cc @carljm


hauntsaninja (approved):

carljm (approved):
Thank you!

Issue #845, opened 2021-08-23
Document Structure and Wishlist


srittau:
Here is how I envision the typing documentation to be structured:

## Tutorials

* [ ] Typing tutorial – from the ground up
* [ ] Type stub tutorial – writing a type stub for an existing library, stubgen, contributing to typeshed

## Howto Guides

* [ ] How to add typing to an existing code base

## Reference

* [ ] Link to the typing module documentation
    + Probably needs a review
* [ ] Best practices
* [ ] Type stub syntax and supported constructs
* [ ] Type stub style guide

## Links

* [ ] Links to all typing-related PEPs
* [ ] Links to type checkers and other typing tools

----

Additions, corrections, criticism welcome!

sobolevn:
Might be helpful, quick eco-system overview: https://github.com/typeddjango/awesome-python-typing/

PR #1801, opened 2024-07-15
Remove empty documents

srittau:
These documents were part of a proposed structure. In the meantime, the structure has diverged from the proposal. We also have issue #845 to discuss a structure if we find time in the future to work on this.


srittau:
This also fixes the remaining build warnings, which can potentially obscure real warnings.

AlexWaygood (approved):

PR #1802, opened 2024-07-15
Split source directory into guides and reference

srittau:
This brings order into both the repository, but also the URL structure by replacing the meaningless "source" component in the URL. This also means that e.g. https://typing.readthedocs.io/en/latest/guides/ will be a guide index, instead of https://typing.readthedocs.io/en/latest/source/guides.html which then links to other documents in the "source" directory.

----

Depends on #1801.

----

DON'T MERGE, just approve please. If this gets approved, I will merge it and add redirects in the readthedocs config from the old paths to the new paths.


JelleZijlstra (approved):
Thanks! Every time I look at this repo I am confused about what "source" means.

srittau:
Done and redirect added.

PR #1804, opened 2024-07-15
Update conformance test sources to initialize uninitialized variables

yangdanny97:
This PR adds constructors that initialize the attributes for the classes defined in each conformance test, per @erictraut comment on another PR (linked below).

https://github.com/python/typing/pull/1734#issuecomment-2106525424

Please let me know if you'd prefer things to be initialized in a different way.

cc @stroxler

### DIFF
--- a/conformance/tests/annotations_forward_refs.py
+++ b/conformance/tests/annotations_forward_refs.py
@@ -88,6 +88,9 @@ class ClassD:
 
     y: int = 0  # E: Refers to local int, which isn't a legal type expression
 
+    def __init__(self, ClassC: "ClassC") -> None:
+        self.ClassC = ClassC
+
 
 assert_type(ClassD.str, str)
 assert_type(ClassD.x, int)
--- a/conformance/tests/annotations_forward_refs.py
+++ b/conformance/tests/annotations_forward_refs.py
@@ -88,8 +88,8 @@ class ClassD:
 
     y: int = 0  # E: Refers to local int, which isn't a legal type expression
 
-    def __init__(self, ClassC: "ClassC") -> None:
-        self.ClassC = ClassC
+    def __init__(self) -> None:
+        self.ClassC = ClassC()
 
 
 assert_type(ClassD.str, str)
--- a/conformance/tests/dataclasses_transform_func.py
+++ b/conformance/tests/dataclasses_transform_func.py
@@ -35,27 +35,16 @@ class Customer1:
     id: int
     name: str
 
-    def __init__(self, id: int, name: str) -> None:
-        self.id = id
-        self.name = name
-
 
 @create_model(frozen=True)
 class Customer2:
     id: int
     name: str
 
-    def __init__(self, id: int, name: str) -> None:
-        self.id = id
-        self.name = name
-
 
 @create_model(frozen=True)
 class Customer2Subclass(Customer2):
     salary: float
-
-    def __init__(self, salary: float) -> None:
-        self.salary = salary
 
 
 c1_1 = Customer1(id=3, name="Sue")
@@ -100,9 +89,6 @@ class Customer3:
 @create_model  # E[Customer3Subclass]
 class Customer3Subclass(Customer3):  # E[Customer3Subclass]
     age: int
-
-    def __init__(self, age: int) -> None:
-        self.age = age
 
 
 c3_1 = Customer3(id=2, name="hi")
--- a/conformance/tests/dataclasses_transform_func.py
+++ b/conformance/tests/dataclasses_transform_func.py
@@ -35,16 +35,27 @@ class Customer1:
     id: int
     name: str
 
+    def __init__(self, id: int, name: str) -> None:
+        self.id = id
+        self.name = name
+
 
 @create_model(frozen=True)
 class Customer2:
     id: int
     name: str
 
+    def __init__(self, id: int, name: str) -> None:
+        self.id = id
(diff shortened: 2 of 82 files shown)


conformance/tests/dataclasses_transform_func.py
```diff
@@ -35,17 +35,28 @@ class Customer1:
     id: int
     name: str
 
+    def __init__(self, id: int, name: str) -> None:
```
erictraut, line 38: Adding an `__init__` in this case is not needed, and it incorrectly changes the nature of this test. This class is supposed to be a dataclass-like class, so its instance methods are initialized via a synthesized `__init__` method. If you add an explicit `__init__` method, this will cause the dataclass mechanism not to synthesize an `__init__`.

conformance/tests/dataclasses_transform_func.py
```diff
@@ -35,17 +35,28 @@ class Customer1:
...
+        self.id = id
+        self.name = name
+
 
 @create_model(frozen=True)
 class Customer2:
     id: int
     name: str
 
+    def __init__(self, id: int, name: str) -> None:
```
erictraut, line 48: See above. This `__init__` method shouldn't be added.

conformance/tests/dataclasses_transform_func.py
```diff
@@ -35,17 +35,28 @@ class Customer1:
...
+        self.id = id
+        self.name = name
+
 
 @create_model(frozen=True)
 class Customer2Subclass(Customer2):
     salary: float
 
+    def __init__(self, salary: float) -> None:
```
erictraut, line 57: See above. This `__init__` method shouldn't be added.

conformance/tests/dataclasses_transform_func.py
```diff
@@ -90,6 +101,9 @@ class Customer3:
 class Customer3Subclass(Customer3):  # E[Customer3Subclass]
     age: int
 
+    def __init__(self, age: int) -> None:
```
erictraut, line 104: See above. This `__init__` method shouldn't be added.

conformance/tests/dataclasses_usage.py
```diff
@@ -172,8 +173,12 @@ class DC13:
     x: int
     x_squared: int
 
+    def __init__(self, x: int, x_squared: int):
```
erictraut, line 176: Introducing an `__init__` method here breaks the test. The purpose of this test is to verify that no `__init__` method is synthesized because of the `init=False` argument. This `__init__` should be removed.

conformance/tests/enums_member_values.py
```diff
@@ -77,6 +77,9 @@ class Color3(Enum):
     RED = 1  # OK
     GREEN = "green"  # E
 
+    def __init__(self, value: int) -> None:
```
erictraut, line 80: This is incorrect. The `_value_` field is special for enums. It is filled in by the "magic" logic of the EnumMeta metaclass. It should never be initialized directly by a user-provided `__init__` method. I think the `__init__` should be removed in this case. Pyre should include special-cased handling of `_value_` and avoid emitting a diagnostic in this case.

conformance/tests/typeddicts_class_syntax.py
```diff
@@ -58,6 +58,10 @@ class GenericTypedDict(Generic[T]):
     name: str
     value: T
 
+    def __init__(self, name: str, value: T) -> None:
```
erictraut, line 61: The `GenericTypedDict` was intended to be a `TypedDict`, but there's a bug in the test's class definition where it doesn't actually include `TypedDict` in its base class list. Please fix the bug in the class definition and remove the `__init__`, which isn't needed for `TypedDict` classes.

erictraut (changes requested):
Thanks for this change. Most of it looks good to me, but there are a few places where the introduction of an `__init__` method is inappropriate because it changes what the test case was attempting to test.

erictraut (approved):
LGTM

erictraut:
@yangdanny97, this draft of the PR looks good to me. If you think it's ready to go, I'm happy to merge it.

yangdanny97:
Sounds good, thanks!

conformance/tests/annotations_forward_refs.py
```diff
@@ -88,6 +88,9 @@ def int(self) -> None:  # OK
 
     y: int = 0  # E: Refers to local int, which isn't a legal type expression
 
+    def __init__(self, ClassC: "ClassC") -> None:
```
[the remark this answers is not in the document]
erictraut: This looks to me like a bug in mypy. This is a deferred evaluation because the annotation is in quotes. The `ClassC` declared in the class body isn't in scope when the evaluation is performed.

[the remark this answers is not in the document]
erictraut: That said, this test isn't intended to check for this behavior, so we could trivially change it to avoid the mypy bug:

```python
    def __init__(self) -> None:
        self.ClassC = ClassC()
```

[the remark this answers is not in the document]
erictraut: @yangdanny97, if you're OK with this proposed modification, please push an update.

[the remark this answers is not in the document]
erictraut: @JelleZijlstra, are you still reviewing the PR, or are you done? If you're good with it, please merge.

PR #1808, opened 2024-07-17
Replace `Annotation` by `Annotated`

### LINKED ISSUE #1805: Clarify signification of `Annotation`
https://github.com/python/typing/pull/1669 Clarified usage of `Final` in dataclasses. The following section was added:

https://github.com/python/typing/blob/42e9118068944e63f4c321f26e4f37f5d77a6d59/docs/spec/qualifiers.rst?plain=1#L179-L181

I believe `Annotation` means `Annotated`? If so, I can open a PR fixing this.

---

Ideally we could also update the [conformance tests](https://github.com/python/typing/pull/1741) to add an extra `final_annotated: Annotated[Final[int], ...]`. It will be interesting to see the output, as mypy seems [to reject it](https://mypy-play.net/?mypy=latest&python=3.12).

Viicos:
Fixes #1805

### DIFF
--- a/docs/spec/qualifiers.rst
+++ b/docs/spec/qualifiers.rst
@@ -177,7 +177,7 @@ annotations for function arguments::
        ...
 
 ``Final`` may be wrapped only by other type qualifiers (e.g. ``ClassVar`` or
-``Annotation``). It cannot be used in a type parameter (e.g.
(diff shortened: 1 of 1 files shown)


srittau (approved):
Thanks!

PR #1809, opened 2024-07-17
Turn Sphinx warnings into errors

srittau:
This will – for example – prevent accidental "floating" documents that aren't part of the TOC structure.


JelleZijlstra:
Thanks! Yes, I didn't do this before because we had all those empty documents, glad we're rid of those now.

PR #1811, opened 2024-07-19
Fix link to Python docs in dataclasses spec

Viicos:


docs/spec/dataclasses.rst
```diff
@@ -488,7 +488,7 @@ This includes, but is not limited to, the following semantics:
         wheel_count: int
 
 * Field ordering and inheritance is assumed to follow the rules
-  specified in `the Python docs <https://docs.python.org/3/library/dataclasses.html#inheritance>`. This includes the effects of
+  specified in `the Python docs <https://docs.python.org/3/library/dataclasses.html#inheritance>`_. This includes the effects of
```
JelleZijlstra, line 491:
```suggestion
  specified in `the Python docs <https://docs.python.org/3/library/dataclasses.html#inheritance>`__. This includes the effects of
```
I think this should fix the CI `/home/runner/work/typing/typing/docs/spec/dataclasses.rst:4:Duplicate explicit target name: "the python docs".`.

Issue #1605, opened 2024-01-24
Integrate the Type Stubs document into the spec


srittau:
The [Type Stubs document](https://typing.readthedocs.io/en/latest/source/stubs.html) (source/stubs.rst) is based on a proposed PEP by @rchen152, @JelleZijlstra, @Solumin, and myself to describe the allowed elements of type stubs. It was turned into kind of a proto-spec document, instead of a PEP. I haven't reviewed it in detail, but I think at least the "spec" parts of this document should be moved to the spec: chapters "Syntax" and "Supported Constructs". They might need updating, though. The "Distribution" chapter is mostly redundant with the spec anyway. The other chapters should probably moved to other documents, like "[Writing and Maintaining Stub Files](https://typing.readthedocs.io/en/latest/source/writing_stubs.html)".

JelleZijlstra:
Agree that the sections you flagged should be incorporated into the spec, with some updates.

For example, the "Syntax" section says stubs should follow 3.7 syntax, but 3.7 is dead now; we should instead say stubs should follow the syntax of the earliest Python version that is not EOL yet.

rchen152:
I'd like to take a stab at this, if that's okay.

erictraut:
Thanks for volunteering, @rchen152.

There's overlap between the [Type Stubs](https://typing.readthedocs.io/en/latest/source/stubs.html) document and the [Typing Python Libraries](https://github.com/python/typing/blob/main/docs/source/libraries.rst) document. I'd love to see these unified when this content is added to the typing spec.

rchen152:
The contents of this doc have been merged into the spec and the guide on writing stubs.

PR #1815, opened 2024-07-22
Merge the specification portion of the "Type Stubs" doc into the spec

rchen152:
The [Type Stubs](https://typing.readthedocs.io/en/latest/reference/stubs.html) doc describes allowed elements for stub files. As a standalone doc, it doesn't really fit in with the rest of the typing documentation. This PR merges its content into the spec:
* The "Introduction," "Syntax," and "Supported Constructs" sections have been moved into the typing specification.
* "Distribution" has been deleted, as the entirety of this section was basically "See [link]."
* Some content that was repeated across this doc, the spec, and the "Typing Python Libraries" guide has been deduplicated.

For https://github.com/python/typing/issues/1605.


docs/spec/distributing.rst
```diff
@@ -10,72 +10,428 @@ Stub files
...
+in the ``typing`` module, plus these additional ones:
+
+ * ``classmethod``
+ * ``staticmethod``
+ * ``property`` (including ``.setter``)
+ * ``abc.abstractmethod``
+ * ``dataclasses.dataclass``
+ * functions decorated with ``@typing.dataclass_transform``
+
+The behavior of other decorators should instead be incorporated into the types.
```
srittau, line 351: Future note: This advice should probably be moved to the guide.

docs/spec/distributing.rst
```diff
@@ -10,72 +10,428 @@ Stub files
...
+  from contextlib import AbstractContextManager
+  def f() -> AbstractContextManager[int]: ...
+
+Version and Platform Checks
+"""""""""""""""""""""""""""
+
+Stub files for libraries that support multiple Python versions can use version
+checks to supply version-specific type hints. Stubs for different Python
+versions should still conform to the most recent supported Python version's
+syntax, as explain in the :ref:`stub-file-syntax` section above.
```
srittau, line 370:
```suggestion
syntax, as explained in the :ref:`stub-file-syntax` section above.
```

srittau (changes requested):
Thanks for splitting those docs! A few notes below.

I also think that the spec part of this needs more love in the future (not in this PR) to remove redundant information or replace it with links to other sections. Some of the text also probably better belongs in the guide, or should at least be duplicated there.

We should also not completely remove the old stubs document, because that leads to 404 errors. Instead I recommend a short document with a text similar to this: "The contents of this document have been moved here: [...]". I think by adding the [`:orphan:` directive](https://www.sphinx-doc.org/en/master/usage/restructuredtext/field-lists.html#metadata) we can prevent this document to be linked in the TOC and suppress the corresponding, but I haven't tried this myself.

I've added a few comments marked "Future note". I don't think they should be included in this PR (to keep it focused), but we should address these points in a followup PR. (I can work on that.)

docs/spec/distributing.rst
```diff
@@ -10,72 +10,428 @@ Stub files
...
+    def foo(x: Literal[""]) -> int: ...
+
+Comments
+""""""""
+
+Standard Python comments are accepted everywhere Python syntax allows them.
+
+Two kinds of structured comments are accepted:
+
+* A ``# type: X`` comment at the end of a line that defines a variable,
```
erictraut, line 97: It's a bit odd to discuss `# type: X` and `# type: ignore` comments in the stubs section (other than perhaps to mention that they are supported in stubs). Describing how they work is redundant, since these are covered in other places within the typing spec. I recommend either removing the "Comments" section or paring it back to simply mention that `# type: X` and `# type: ignore` comments are supported in stubs.

docs/spec/distributing.rst
```diff
@@ -10,72 +10,428 @@ Stub files
...
+particular, built-in generics, pipe union syntax (``X | Y``), and forward
+references can be used.
+
+Type checkers should have a configurable search path for stub files. If a stub
+file is found, the type checker should not read the corresponding "real" module.
+See :ref:`mro` for more information.
+
+.. _stub-file-supported-constructs:
+
+Supported Constructs
```
erictraut, line 67: I find the current contents of this section to be a bit confusing because it vacillates between "here's what a conformant type checker must support" and "here's guidance for a stub author". I think the typing spec should include only the former. The latter should be included in guides.

docs/spec/distributing.rst
```diff
@@ -10,72 +10,428 @@ Stub files
...
+    public_attr: int
+    _private_looking_public_attr: int
+    private_attr: int
+
+Type checkers support cyclic imports in stub files.
+
+Built-in Generics
+"""""""""""""""""
+
+:pep:`585` built-in generics are supported and should be used instead
```
erictraut, line 135: This is a good example of something that should be moved to a guide, not included in the typing spec.

docs/spec/distributing.rst
```diff
@@ -10,72 +10,428 @@ Stub files
...
 
-* Third-party modules whose authors have not yet added type hints
+    def foo(iter: Iterable[int]) -> None: ...
 
-* Standard library modules for which type hints have not yet been
-  written
+Unions
+""""""
+
+Declaring unions with the shorthand syntax or ``Union`` and ``Optional`` is
```
erictraut, line 153: Likewise, I don't think this belongs in the typing spec. It's fine content for a guide.

docs/spec/distributing.rst
```diff
@@ -10,72 +10,428 @@ Stub files
...
+    x = ...  # type: int
+
+The ellipsis literal can stand in for any value::
+
+    x: int = ...  # type is int
+
+A variable annotated as ``Final`` and assigned a literal value has the
+corresponding ``Literal`` type::
+
+    x: Final = 0  # type is Literal[0]
```
erictraut, line 177: What types of value expressions should a type checker be expected to evaluate here? For example, should simple unary and binary arithmetic operators be supported? What about bitwise operators? What about the use of other `Final` variables? Mypy supports all of these, I think. Pyright recently filled out this full matrix.

```python
a: Final = -1
b: Final = 1 + 2
c: Final = 1 << 3
d: Final = 1 | 2 | 64
e: Final = (1 + c) ** 2
```

docs/spec/distributing.rst
```diff
@@ -10,72 +10,428 @@ Stub files
...
+
+In all other cases, the type of a variable is unspecified when the variable is
+unannotated or when the annotation and the assigned value disagree::
+
+    x = 0  # type is unspecified
+    x = ...  # type is unspecified
+    x: int = ""  # type is unspecified
```
erictraut, line 184: I don't think the typing spec should mandate type checker behaviors for error conditions. If the stub author has included bugs in their stub definitions, the behavior should be undefined. 

If we are going to mandate a behavior for this case, I think the specified type annotation should be honored rather than treating the type as unspecified.

docs/spec/distributing.rst
```diff
@@ -10,72 +10,428 @@ Stub files
...
+
+Classes
+"""""""
+
+Class definition syntax follows general Python syntax, but type checkers
+are only expected to understand the following constructs in class bodies:
```
erictraut, line 190: Grammar nit: "only expected to understand" should be "expected to understand only"

docs/spec/distributing.rst
```diff
@@ -10,72 +10,428 @@ Stub files
...
+
+* The ellipsis literal ``...`` is ignored and used for empty
+  class bodies. Using ``pass`` in class bodies is undefined.
+* Instance attributes follow the same rules as module level attributes
+  (see above).
+* Method definitions (see below) and properties.
+* Method aliases.
```
erictraut, line 197: I think that class-scoped type aliases should be included in this list.

docs/spec/distributing.rst
```diff
@@ -10,72 +10,428 @@ Stub files
...
+        read_write: int
+        @property
+        def read_only(self) -> int: ...
+        def do_stuff(self, y: str) -> None: ...
+        doStuff = do_stuff
+        class Inner: ...
+
+More complex statements don't need to be supported.
+
+The type of generic classes can be narrowed by annotating the ``self``
```
erictraut, line 214: This section is redundant (already covered in the "Callables" section) and should be removed.

docs/spec/distributing.rst
```diff
@@ -10,72 +10,428 @@ Stub files
...
+
+    class Complex(Base):
+        read_write: int
+        @property
+        def read_only(self) -> int: ...
+        def do_stuff(self, y: str) -> None: ...
+        doStuff = do_stuff
+        class Inner: ...
+
+More complex statements don't need to be supported.
```
erictraut, line 212: I'm not sure what "more complex statements" means here. This language is fine for a guide, but it's not specific enough for the typing spec. I wouldn't know how to write conformance tests based on this statement, for example.

docs/spec/distributing.rst
```diff
@@ -10,72 +10,428 @@ Stub files
...
+The class must match the class in which it is declared. Using other classes,
+including sub or super classes, will not work. In addition, the ``self``
+annotation cannot contain type variables.
+
+Functions and Methods
+"""""""""""""""""""""
 
-* Modules that must be compatible with Python 2 and 3
+Function and method definition syntax follows general Python syntax.
+For backwards compatibility, positional-only parameters can also be marked by
```
erictraut, line 233: The information about positional-only parameters is redundant (already covered elsewhere in the spec) and should not be repeated in this section.

docs/spec/distributing.rst
```diff
@@ -10,72 +10,428 @@ Stub files
...
+    # x is positional-only
+    # y can be used positionally or as keyword argument
+    # z is keyword-only
+    def foo(x, /, y, *, z): ...  # recommended
+    def foo(__x, y, *, z): ...  # backwards compatible syntax
 
-Stub files have the same syntax as regular Python modules.  There is one
-feature of the ``typing`` module that is different in stub files:
-the ``@overload`` decorator described below.
+If an argument or return type is unannotated, per :pep:`484` its
```
erictraut, line 243: The typing spec shouldn't be quoting PEP 484. It should refer to itself.

In this case, the wording for return types has been updated since PEP 484 to allow for return type inference, so this section should be careful not to contradict other sections of the typing spec.

docs/spec/distributing.rst
```diff
@@ -10,72 +10,428 @@ Stub files
...
+type is assumed to be ``Any``. It is preferred to leave unknown
```
erictraut, line 244: "It is preferred...": This is guidance to stub authors, not something that should be included in the typing spec.

docs/spec/distributing.rst
```diff
@@ -10,72 +10,428 @@ Stub files
...
+types unannotated rather than explicitly marking them as ``Any``, as some
+type checkers can optionally warn about unannotated arguments.
 
-The type checker should only check function signatures in stub files;
-It is recommended that function bodies in stub files just be a single
-ellipsis (``...``).
+If an argument has a literal or constant default value, it must match the implementation
```
erictraut, line 248: This should say "parameter", not "argument". The two terms are defined in the "Callables" section of this spec. In any case, this probably doesn't belong in the typing spec; it's more appropriate for a guide.

docs/spec/distributing.rst
```diff
@@ -10,72 +10,428 @@ Stub files
...
+    # The following arguments all have type int.
+    def annotated(a: int, b: int = 42, c: int = ...): ...
+    # The following default values are invalid and the types are unspecified.
+    def invalid(a: int = "", b: Foo = Foo()): ...
 
-While stub files are syntactically valid Python modules, they use the
-``.pyi`` extension to make it possible to maintain stub files in the
-same directory as the corresponding real module.  This also reinforces
-the notion that no runtime behavior should be expected of stub files.
+For a class ``C``, the type of the first argument to a classmethod is
```
erictraut, line 259: This doesn't belong in this section. The inferred type of `self` and `cls` should be covered in other parts of the typing spec, not in the stubs section. (It's not currently covered, but I hope to fix that in the near future.)

Regardless, I think this statement is incorrect. The type of the first argument to a classmethod is assumed to be `type[Self]` if unannotated. And the type of `self` is assumed to be `Self`.

docs/spec/distributing.rst
```diff
@@ -10,72 +10,428 @@ Stub files
...
 
-  where ``__init__.pyi`` contains a line such as ``from . import ham``
-  or ``from .ham import Ham``, then ``ham`` is an exported attribute
-  of ``spam``.
+    def foo(): ...  # compatible
+    def bar(): pass  # behavior undefined
 
-* Stub files may be incomplete. To make type checkers aware of this, the file
-  can contain the following code::
+All variants of overloaded functions and methods must have an ``@overload``
```
erictraut, line 286: This is covered in the "Overloads" section of the typing spec already and shouldn't be included in the "stubs" section.

docs/spec/distributing.rst
```diff
@@ -10,72 +10,428 @@ Stub files
...
 
-    def __getattr__(name) -> Any: ...
+    @overload
+    def foo(x: str) -> str: ...
+    @overload
+    def foo(x: float) -> int: ...
 
-  Any identifier not defined in the stub is therefore assumed to be of type
-  ``Any``.
+The following (which would be used in the implementation) is wrong in stubs::
```
erictraut, line 294: I think this should be removed from the typing spec as well. This is more of a guide thing.

docs/spec/distributing.rst
```diff
@@ -10,72 +10,428 @@ Stub files
...
+    @overload
+    def foo(x: str) -> str: ...
+    @overload
+    def foo(x: float) -> int: ...
+    def foo(x: str | float) -> Any: ...
+
+Aliases and NewType
+"""""""""""""""""""
+
+Type checkers should accept module-level type aliases, optionally using
```
erictraut, line 305: This section is also redundant. Type aliases are already covered in the "Type Aliases" section of the spec. That section already describes their form and where they can be used.

docs/spec/distributing.rst
```diff
@@ -10,72 +10,428 @@ Stub files
...
+
+Decorators
+""""""""""
+
+Type checkers are expected to understand the effects of all decorators defined
+in the ``typing`` module, plus these additional ones:
+
+ * ``classmethod``
+ * ``staticmethod``
+ * ``property`` (including ``.setter``)
```
erictraut, line 346: And presumably ".deleter".

docs/spec/distributing.rst
```diff
@@ -10,72 +10,428 @@ Stub files
...
+ * ``abc.abstractmethod``
+ * ``dataclasses.dataclass``
+ * functions decorated with ``@typing.dataclass_transform``
```
erictraut, line 349: Plus classes decorated with `@typing.dataclass_transform`.

docs/spec/distributing.rst
```diff
@@ -10,72 +10,428 @@ Stub files
...
+
+Version and Platform Checks
+"""""""""""""""""""""""""""
+
+Stub files for libraries that support multiple Python versions can use version
+checks to supply version-specific type hints. Stubs for different Python
+versions should still conform to the most recent supported Python version's
+syntax, as explain in the :ref:`stub-file-syntax` section above.
+
+Version checks are if-statements that use ``sys.version_info`` to determine the
```
erictraut, line 372: Version checking is already covered in other parts of the typing spec and shouldn't be duplicated here.

docs/spec/distributing.rst
```diff
@@ -10,72 +10,428 @@ Stub files
...
+
+    if sys.platform == 'linux' and (sys.version_info < (3,) or sys,version_info >= (3, 7)): ...
+
+Enums
+"""""
+
+Enum classes are supported in stubs, regardless of the Python version targeted by
+the stubs.
+
+Enum members should be specified with an unannotated assignment, for example as
```
erictraut, line 432: This is already covered in the "Enums" section of the typing spec and should not be duplicated here.

docs/spec/distributing.rst
```diff
@@ -266,3 +622,23 @@ of that Python version. This can be queried e.g.
...
+
+By convention, certain import forms indicate to type checkers that an imported
+symbol is re-exported and should be considered part of the importing module's
+public interface. All other imported symbols are considered private by default.
+
+The following import forms re-export symbols:
+
+* ``import X as X`` (a redundant module alias): re-exports ``X``.
+* ``from Y import X as X`` (a redundant symbol alias): re-exports ``X``.
+* ``from Y import *``: re-exports all symbols in ``Y`` that do not begin with
```
erictraut, line 639: I think this is wrong. All symbols imported by a wildcard (as indicated by the `__all__` symbol) should be re-exported, including those that begin with an underscore. See [this documentation](https://github.com/python/typing/blob/main/docs/guides/libraries.rst#library-interface-public-and-private-symbols) for details.

docs/spec/distributing.rst
```diff
@@ -266,3 +622,23 @@ of that Python version. This can be queried e.g.
...
+  an underscore.
+* ``from . import bar`` in an ``__init__`` module: re-exports ``bar`` if it does
```
erictraut, line 641: I don't think these last two bullets should be included. They kind of make sense for stubs, but they don't make sense for "py.typed" libraries. I think the rules for stubs and "py.typed" libraries should be the same. Pyright has never implemented these two rules for either stubs or "py.typed" libraries, and they are not included in [this guide](https://github.com/python/typing/blob/main/docs/guides/libraries.rst#library-interface-public-and-private-symbols).

erictraut (changes requested):
I focused my review on the typing spec section, so my comments are limited to that portion of the PR.

I think it's important to break this effort into two separate PRs — one for the typing spec (which the typing council will need to approve) and one for the guides documentation (which requires less formality and scrutiny). These two sections are each very significant, they target different audiences, and they have different review/approval processes.

For the typing spec, I think we need to be careful to include only information that pertains to type checker behaviors. The current PR contains a bunch of guidance directed at stub authors. This should be moved to guides, not included in the typing spec.

The current PR also contains a bunch of information that is covered already in other parts of the typing spec. In many cases, this new section describes existing information in incomplete and/or contradictory terms. I think all of the redundant information needs to be stripped out. It's fine to include references to other parts of the typing spec, but it the new section should not attempt to repeat the information from other sections. This will cause confusion and lead to inconsistencies in the spec.

The source material for this PR was written before "py.typed" libraries became popular. There are many similarities between stub files and inlined type information found within a "py.typed" library. I think the typing spec needs to describe the behaviors of type checkers with regard to both stubs and "py.typed" libraries. Given the large overlap between the two, these should be covered in the same section, and we should strive to reduce any differences between the two cases. [This document](https://github.com/python/typing/blob/main/docs/guides/libraries.rst) (which is a peer to the ["Writing Stubs" document](https://github.com/python/typing/blob/main/docs/guides/writing_stubs.rst) used as the source material for this PR) is my attempt to document the intended behaviors for "py.typed" libraries. I think the information in both of these documents needs to be consolidated into the typing spec.

docs/spec/distributing.rst
JelleZijlstra: If there is no `__all__`, then `import *` does re-export all names that do not start with an underscore (by analogy with Python's runtime behavior).

srittau:
> I think it's important to break this effort into two separate PRs — one for the typing spec (which the typing council will need to approve) and one for the guides documentation (which requires less formality and scrutiny). These two sections are each very significant, they target different audiences, and they have different review/approval processes.

I agree. When we start by splitting out the non-spec part first, we can also more easily merge non-spec behavior from the spec part into the guides section.

docs/spec/distributing.rst
erictraut: Yes, I agree that's the correct behavior if there's no `__all__`. That matches pyright's behavior. Not sure about other type checkers.

rchen152:
Thanks for the quick and thorough reviews! I'll address these comments, then break out the non-spec changes into another PR and aim to get those merged first.

docs/spec/distributing.rst
rchen152: I thought that classes decorated with dataclass_transform were inherited from, not used as decorators?

erictraut: That's correct, but the `dataclass_transform` is used then as a decorator on the metaclass or base class. This class decorator must be supported by type checkers. The current wording above says "functions decorated with `@typing.dataclass_transform`".

docs/guides/writing_stubs.rst
```diff
@@ -106,3 +106,473 @@ particularly with false positives.
...
+--------------------------
+
+Not all modules should be included into stubs.
+
+It is recommended to exclude:
+
+1. Implementation details, with `multiprocessing/popen_spawn_win32.py <https://github.com/python/cpython/blob/main/Lib/multiprocessing/popen_spawn_win32.py>`_ as a notable example
+2. Modules that are not supposed to be imported, such as ``__main__.py``
+3. Protected modules that start with a single ``_`` char. However, when needed protected modules can still be added (see :ref:`undocumented-objects` section below)
+
```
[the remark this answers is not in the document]
rchen152: I went ahead and addressed these future notes, since most of the fixes were fairly straightforward, and I'll be splitting up this PR anyway (which should help with the concerns about it being large/unfocused).

rchen152:
The changes to the writing_stubs guide have been split off into https://github.com/python/typing/pull/1817. That PR is ready for review; this one is not.

For now, I have *not* reverted the writing_stubs changes in this PR, in case it's useful to look at the latest commits here to see what I changed in response to review comments.

docs/spec/distributing.rst
rchen152: This was just saying that no other statements need to be supported, with "more complex" being flavor text. It's redundant anyway, so I deleted it.

docs/guides/writing_stubs.rst
```diff
@@ -337,6 +337,45 @@ No::
...
+  @overload
+  def foo(x: str) -> str: ...
+  @overload
+  def foo(x: float) -> int: ...
+  def foo(x: str | float) -> Any: ...
+
+Decorators
+----------
+
+Include only those decorators whose effects type checkers understand, enumerated
```
TeamSpen210, line 365: Maybe it’d be better to describe these as decorators that have “special behaviour”, since type checkers do understand other decorators via typeshed? With the current list the distinction seems more that these are special-cased, and their results can’t be described by regular types. And you want to apply the decorator results where possible to avoid relying on checker-specific inference behaviour in your library.

rchen152: Thanks for the feedback!

> type checkers do understand other decorators via typeshed

This isn't true for all type checkers. Pytype, for example, ignores any decorators in stubs outside of this list. I tweaked this sentence to say "all of the major type checkers" rather than just "type checkers," since I think that makes it a little clearer that the concern here is differences in behavior among type checkers.

docs/spec/distributing.rst
```diff
@@ -10,72 +10,428 @@ Stub files
...
+
+Stub files distinguish between imports that are re-exported and those
+that are only used internally. See :ref:`import-conventions`.
+
+Type aliases can be used to re-export an import under a different name::
+
+    from foo import bar as _bar
+    new_bar = _bar  # "bar" gets re-exported with the name "new_bar"
+
+Stubs support customizing star import semantics by defining a module-level
```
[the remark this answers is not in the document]
rchen152: I've done this (by moving the entire "Library Interface" section over), but I do want to point out that, right now, pytype supports only the first three forms in that list in stubs, and I don't know if it'll be a problem for typeshed if the spec differ from what's supported in practice.

docs/spec/distributing.rst
rchen152: What value is added by supporting things like arithmetic operators? My preference would be to allow only literal ints, strings, etc., on the principle that stubs shouldn't contain any unnecessary implementation detail.

docs/spec/distributing.rst
rchen152: This section was using "unspecified" to mean basically the same thing as "undefined." I've switched to consistently using "undefined."

docs/spec/distributing.rst
rchen152: Like the case with the limited `__all__` forms, class-scoped type aliases were originally excluded because pytype doesn't support them in stubs, and one of the intended purposes of the Type Stubs doc was to document what could be safely included in stubs targeting multiple type checkers. I'm not sure whether or not that's still a good reason to disallow them, since it does feel rather arbitrary.

docs/spec/distributing.rst
rchen152: Removed this sentence, since it didn't have anything stub-specific to say.

docs/spec/distributing.rst
rchen152: That's covered by "all decorators defined in the `typing` module" (in the paragraph above this list).

docs/spec/distributing.rst
rchen152: Updated this bullet point to describe the behavior when `__all__` is present.

docs/spec/distributing.rst
rchen152: That guide says, "If [imported symbols] use the [...] `from . import A` forms, symbol `A` is not private unless the name begins with an underscore. If a file `__init__.py` uses form `from .A import X`, symbol `A` is treated likewise," which I interpreted to mean what I wrote in the last two bullets. Did I misunderstand something?

docs/spec/distributing.rst
rchen152: Thanks for this feedback! I've redrafted this section to focus on what type checkers should support and moved any stub author guidance that seemed useful to keep to the writing stubs guide.

rchen152:
> The source material for this PR was written before "py.typed" libraries became popular. There are many similarities between stub files and inlined type information found within a "py.typed" library. I think the typing spec needs to describe the behaviors of type checkers with regard to both stubs and "py.typed" libraries. Given the large overlap between the two, these should be covered in the same section, and we should strive to reduce any differences between the two cases. [This document](https://github.com/python/typing/blob/main/docs/guides/libraries.rst) (which is a peer to the ["Writing Stubs" document](https://github.com/python/typing/blob/main/docs/guides/writing_stubs.rst) used as the source material for this PR) is my attempt to document the intended behaviors for "py.typed" libraries. I think the information in both of these documents needs to be consolidated into the typing spec.

I've now moved the entire "Library Interface" section from the typing libraries guide into the spec. The rest of the guide is about using setuptools for distribution, type completeness, and best practices for authoring inline types; it wasn't clear to me if any of this should be migrated to the spec.

I feel like stub files should have their own section, as a separate file type with some differences from Python source files.

rchen152:
Ok, I believe this PR is ready for another round of review! I was going to update my post on discuss.python.org, but the site's not loading for me at the moment...

docs/spec/distributing.rst
```diff
@@ -10,72 +10,160 @@ Stub files
...
+Stub files use a subset of the constructs used in Python source files, as
+described in :ref:`stub-file-supported-constructs` below. Type checkers should
+parse a stub that uses only such constructs without error and not interpret any
+construct in a contradictory manner. However, type checkers are not required to
+implement checks for all of these constructs and can elect to ignore unsupported
+ones. Additionally, type checkers can support constructs not described here.
+
+Type checkers should have a configurable search path for stub files. If a stub
+file is found, the type checker should not read the corresponding "real" module.
+See :ref:`mro` for more information.
```
carljm, line 36: I'm not sure I agree with this first sentence. Type checkers should allow adding extra search paths, but I don't think type checkers should necessarily provide "a configurable search path" for stub files specifically.

But I'm not sure we should be getting into any of that here; it's specified more carefully below, let's just refer to that.

```suggestion
If a stub
file is found for a module, the type checker should not read the corresponding "real" module.
See :ref:`mro` for more information.
```

carljm, line 36: I guess it predates this PR, but I'm not sure why `mro` is the reference tag for "import resolution ordering".

docs/spec/distributing.rst
```diff
@@ -10,72 +10,160 @@ Stub files
...
+
+.. _stub-file-syntax:
+
+Syntax
+^^^^^^
+
+Stub files are syntactically valid Python files in the earliest Python version
+that is not yet end-of-life. They use a ``.pyi`` suffix. The Python syntax used
+by stub files is independent from the Python versions supported by the
+implementation, and from the Python version the type checker runs under (if
```
carljm, line 46: The use of the word "implementation" to mean "the real module which the type stub is providing types for" is IMO not obvious (at first I thought it meant "the type checker".) I think it's a useful term, but we should introduce its meaning explicitly.

docs/spec/distributing.rst
carljm: I think the value added is sometimes clarity for readers, e.g. `SECONDS_IN_DAY: Final = 60 * 60 * 24` vs `SECONDS_IN_DAY: Final = 86400`. You can add the clarity in a comment, but then it risks getting out of date with the code.

I don't think the value is large, but I suspect the cost is also not large, since I would expect a type checker to support this in non-stubs anyway. 

Though I note that the spec for Literal currently says nothing about support for arithmetic on integer literals, so perhaps we should avoid saying too much here, and defer to that spec, and it can be expanded in a separate PR with more clarity.

I would favor simply saying here that inference of Final literals in stubs should match inference of Final literals in non-stubs.

docs/spec/distributing.rst
carljm: I do think these last two bullets make sense for `py.typed` libraries. It's common for libraries to define classes and functions in sub-modules, and then import them in `__init__.py` for the sole purpose of exposing a flat namespace to users of the library.

carljm (commented):
Thanks for working on this!

docs/spec/distributing.rst
JelleZijlstra: I suppose it stands for "Module Resolution Order" https://peps.python.org/pep-0561/#type-checker-module-resolution-order.

docs/spec/distributing.rst
```diff
@@ -10,72 +10,160 @@ Stub files
...
+  authors, which is especially useful when adding type annotations to
+  existing packages.
+* They can act as documentation, succinctly explaining the external
+  API of a package, without including the implementation or private
+  members.
+
+Stub files use a subset of the constructs used in Python source files, as
+described in :ref:`stub-file-supported-constructs` below. Type checkers should
+parse a stub that uses only such constructs without error and not interpret any
+construct in a contradictory manner. However, type checkers are not required to
```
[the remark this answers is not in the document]
rchen152: Ah, contradictory to the spec. Updated to be clearer.

docs/spec/distributing.rst
rchen152: Good point, I added a definition for "implementation" the first time it's used in the intro.

docs/spec/distributing.rst
```diff
@@ -266,3 +355,64 @@ of that Python version. This can be queried e.g.
...
+By convention, certain import forms indicate to type checkers that an imported
+symbol is re-exported and should be considered part of the importing module's
+public interface. All other imported symbols are considered private by default.
+
+The following import forms re-export symbols:
+
+* ``import X as X`` (a redundant module alias): re-exports ``X``.
+* ``from Y import X as X`` (a redundant symbol alias): re-exports ``X``.
+* ``from Y import *``: if ``Y`` defines a module-level ``__all__`` list,
+  re-exports all names in ``__all__``; otherwise, re-exports  all symbols in
```
erictraut, line 413: I think "all symbols in ``Y`` that do not begin with an underscore" may be too broad. It should be "all public symbols in the module's global scope". (Notably, this excludes imported symbols unless they are intended to be re-exported.)

docs/spec/distributing.rst
erictraut: I've seen cases where a "py.typed" library includes a `__init__.py` that imports submodules with the intent of referencing specific symbols in those submodules but without the intent of re-exporting that submodule. If we always assume that library authors want these submodules re-exported from the top-level module, we're taking away the ability for them to keep them private. If they want to make them public, they can explicitly do so by using a redundant form (`from . import bar as bar`). Pyright has enforced this for the last three years, and library authors have adapted accordingly. I'm uncomfortable changing the rules now.

erictraut (commented):
This is looking good. Thanks again for doing this @rchen152. I left a few comments.

docs/spec/distributing.rst
rchen152: I removed the description of how the inference works and linked to the appropriate section of the Literal spec.

docs/spec/distributing.rst
```diff
@@ -10,72 +10,161 @@ Stub files
...
-* Standard library modules for which type hints have not yet been
-  written
+In locations where value expressions can appear, such as the right-hand side of
+assignment statements and function parameter defaults, type checkers should
+support the following expressions:
 
-* Modules that must be compatible with Python 2 and 3
+* The ellipsis literal, ``...``, which can stand in for any value
+* Any value that is a
+  :ref:`legal parameter for typing.Literal <literal-legal-parameters>`
```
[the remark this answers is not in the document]
rchen152: Added both floating point literals and complex literals to this list.

docs/spec/distributing.rst
```diff
@@ -10,72 +10,161 @@ Stub files
...
+
+If a stub file is found for a module, the type checker should not read the
+corresponding "real" module. See :ref:`mro` for more information.
+
+.. _stub-file-syntax:
+
+Syntax
+^^^^^^
+
+Stub files are syntactically valid Python files in the earliest Python version
```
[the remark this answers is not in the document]
rchen152: I kept the first few sentences that describe the syntax of a stub file, since I think that's useful for type checkers, and deleted the part that went into what language features stubs ought to use.

docs/spec/distributing.rst
rchen152: That sounds reasonable to me. I changed this bullet point from "method aliases" to just "aliases" and added an example of a type alias to the code block below to make it clear that type aliases are included.

docs/spec/distributing.rst
rchen152: Ah, that makes a lot of sense. I've removed the last two bullets.

docs/spec/distributing.rst
```diff
@@ -10,72 +10,161 @@ Stub files
...
+Type checkers should fully support these constructs:
+
+* All features from the ``typing`` module of the latest released Python version
+* Comments, including type declaration (``# type: X``) and error suppression
+  (``type: ignore``) comments
+* Import statements, including the standard :ref:`import-conventions` and cyclic
+  imports
+* Module-level type aliases (e.g., ``X: TypeAlias = int``,
+  ``Y: TypeAlias = dict[str, _V]``)
+* Regular aliases (e.g., ``function_alias = some_function``) at both the module-
```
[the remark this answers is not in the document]
rchen152: That was me trying to distinguish between type aliases and other aliases like function aliases, which is no longer necessary.

docs/spec/distributing.rst
```diff
@@ -10,72 +10,161 @@ Stub files
...
+use it to mark coroutines, even if the implementation still uses the
+``@coroutine`` decorator. On the other hand, the ``type`` soft keyword from
+:pep:`695`, introduced in Python 3.12, is not valid in stubs until Python 3.11
+reaches end-of-life in October 2027.
+
+Stubs are treated as if ``from __future__ import annotations`` is enabled. In
+particular, forward references do not need to be quoted, and syntax from newer
+versions than otherwise supported may be used in annotation contexts. For
+example, the pipe union syntax (``X | Y``) introduced in Python 3.10 may be used
+in annotation contexts even before Python 3.9 reaches end-of-life.
```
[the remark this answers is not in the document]
rchen152: I was thinking about the fact that something like `MODULE_ATTR = int | str` doesn't work in 3.9 and below, even when `from __future__ import annotations` is in use, but on second thought, that's not a *syntax* error, so pipe union syntax can indeed be used everywhere. I'll remove the references to "annotation contexts". (For the record, I meant places where annotation expressions can be used, like variable annotations or function parameter annotations.)

docs/spec/distributing.rst
```diff
@@ -10,72 +10,161 @@ Stub files
...
+
+For example, Python 3.7 added the ``async`` keyword (see :pep:`492`). Stubs may
+use it to mark coroutines, even if the implementation still uses the
+``@coroutine`` decorator. On the other hand, the ``type`` soft keyword from
+:pep:`695`, introduced in Python 3.12, is not valid in stubs until Python 3.11
+reaches end-of-life in October 2027.
+
+Stubs are treated as if ``from __future__ import annotations`` is enabled. In
+particular, forward references do not need to be quoted, and syntax from newer
+versions than otherwise supported may be used in annotation contexts. For
```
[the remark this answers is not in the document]
rchen152: Actually, I think this sentence does need to describe where syntax from newer versions can be used; otherwise, it just sounds like it's directly contradicting the earlier info about what syntax stubs use. I changed this to say "in annotation expressions" rather than "in annotation contexts." Hopefully that's clearer, since "annotation expression" is a term that is defined in the glossary.

docs/spec/distributing.rst
```diff
@@ -10,72 +10,154 @@ Stub files
...
+Syntax
+^^^^^^
+
+Stub files are syntactically valid Python files in the earliest Python version
+that is not yet end-of-life. They use a ``.pyi`` suffix. The Python syntax used
+by stub files is independent from the Python versions supported by the
+implementation, and from the Python version the type checker runs under (if
+any).
+
+Type checkers should treat stubs as if ``from __future__ import annotations`` is
```
JelleZijlstra, line 50: This may be useful as a mnemonic but it's not a particularly precise description. For example, forward references in type aliases are valid in stubs, but they would fail at runtime even with the future import enabled. Moreover, `from __future__ import annotations` is about to be deprecated.

I would suggest something like:

> Type checkers should support all type system features in stub files, including those introduced in newer versions. For example, the pipe union syntax (``X | Y``) introduced in Python 3.10 may be used even before Python 3.9 reaches end-of-life. Type checkers should also support forward references in all contexts, without the need to quote strings. For example, forward references may be used in annotations and in the right-hand side of assignments used as type aliases.

docs/spec/distributing.rst
```diff
@@ -10,72 +10,154 @@ Stub files
...
+.. _stub-file-supported-constructs:
+
+Supported Constructs
+^^^^^^^^^^^^^^^^^^^^
+
+Type checkers should fully support these constructs:
+
+* All features from the ``typing`` module of the latest released Python version
+* Comments, including type declaration (``# type: X``) and error suppression
+  (``type: ignore``) comments
```
JelleZijlstra, line 65:
```suggestion
  (``# type: ignore``) comments
```

docs/spec/distributing.rst
```diff
@@ -10,72 +10,154 @@ Stub files
...
+* Import statements, including the standard :ref:`import-conventions` and cyclic
+  imports
+* Aliases, including type aliases, at both the module and class level
```
JelleZijlstra, line 68: Should type aliases at the class level be included? I'm not sure they're well-specified and I don't think we use them in typeshed.

docs/spec/distributing.rst
```diff
@@ -10,72 +10,154 @@ Stub files
...
-  automatically become exported attributes of their parent module
-  when imported. For example, if the ``spam`` package has the
-  following directory structure::
+Class definition syntax follows general Python syntax, but type checkers
+are expected to understand only the following constructs in class bodies:
 
-      spam/
-          __init__.pyi
-          ham.pyi
+* The ellipsis literal ``...`` is used for empty class bodies. Using ``pass`` in
```
JelleZijlstra, line 116: That seems unnecessarily strict; it seems fine that `pass` is also allowed (even if it's not stylistically preferred).

docs/spec/distributing.rst
erictraut: Hmm, I like the previous phrasing better. The proposed change doesn't fully capture the intended behavior. It's not just that type checkers should *support* forward references. They should *assume forward references within all type expressions* in a stub file, as they do when `from __future__ import annotations` is used in a source file. The distinction is important because symbol resolution rules change when assuming forward references.

Perhaps we should say that "all type expressions in a stub file should be evaluated as though they are quoted even if they are not".

docs/spec/distributing.rst
erictraut: Type aliases within classes are used in type stubs today. Maybe not in typeshed, but in other libraries. I think it would be problematic not to include them.

docs/spec/distributing.rst
erictraut: I think it's OK to support `pass`, but we may need to revisit this question if/when we standardize the way to mark a protocol method as abstract versus implemented. In source files, type checkers use (non-specified) heuristics to determine whether the implementation is "trivial", but different rules are needed for stubs.

docs/spec/distributing.rst
rchen152: Let me think a bit about how to rephrase this. @JelleZijlstra I see what you mean about the current wording being incorrect, but supporting "all type system features" doesn't seem quite right either. For example, type checkers aren't expected to support the new `type` keyword (PEP 695) in stubs for a while yet, right, since it's new syntax in 3.12?

docs/spec/distributing.rst
rchen152: I believe they're not used in typeshed because pytype doesn't support them in stubs ^^; I don't feel strongly about this either way, but it seems better to allow them unless there's a compelling reason not to.

docs/spec/distributing.rst
rchen152: Again, I don't feel super strongly about this, but I slightly prefer to not allow `pass` because allowing it would mean that type checkers need to support two ways of expressing the same thing, for no clear benefit.

docs/spec/distributing.rst
erictraut: A conformant type checker should definitely support the `type` keyword regardless of whether it's used in a ".pyi" or a ".py" file.

As a stub author, you can decide whether or not to use this feature based on whether the most popular type checkers support this feature, but this is not a concern of the typing spec.

docs/spec/distributing.rst
erictraut: I don't think it's appropriate for the typing spec to say whether `pass` should be "disallowed" in a stub. After all, type checkers don't report errors in stubs that are imported by programs.

The typing spec should indicate whether a type checker should be required to handle `pass` in a stub. I'm ambivalent on this question. If we don't say that it's required, then the behavior will be undefined.

For what it's worth, I have seen `pass` used in some stubs. Not in typeshed, but in other stubs.

rchen152: I don't think anyone's advocating that we require `pass` to be an error. As I understand them, the options being considered are (1) specify that `pass` in class and function bodies means the same thing as `...` (the "allow" option),or (2) leave the behavior of `pass` undefined (the "disallow" option).

docs/spec/distributing.rst
rchen152: @erictraut What version of Python syntax do you think type checkers ought to support in stubs? Right now, this PR says that stubs use the syntax of the earliest Python version not yet end-of-life (currently 3.8). That's what was suggested [here](https://github.com/python/typing/issues/1605#issuecomment-1908644244), and I went with it because it seemed sensible and unambiguous. But if we stick with this, it would be contradictory to require type checkers to support newer syntax unless it's a case like the pipe union syntax, where `from __future__ import annotations` can be used to cheat, basically.

JelleZijlstra: The distinction I would make is between syntactic and non-syntactic features. For syntactic features (such as the `type` statement), stubs can only use features that are present in the oldest supported Python version. For non-syntactic type system features (such as the `|` operator, which does not change the parser), features from any Python version should be allowed.

erictraut: > stubs can only use features that are present in the oldest supported Python version

Why? I think this is good guidance for stub authors — and therefore should be added to the "guides". It's also understandable that typeshed contributors should be compelled to follow this rule. But it doesn't sound like something that belongs in the typing spec. The typing spec describes the type system and how type checkers should work. Conformant type checkers should support all of the latest syntax regardless of whether that's in stubs or source files.

JelleZijlstra: The reason to add this behavior to the spec is that it is part of the contract between stub authors and type checkers. A conformant type checker must accept `int | str` in a stub, even if it runs in Python 3.9 mode; a conformant type checker does not need to accept `type T = int` in a stub if it runs in Python 3.9 mode.

erictraut: > a conformant type checker does not need to accept type T = int in a stub if it runs in Python 3.9 mode.

I don't think that's documented anywhere in the typing spec currently. I don't remember seeing that, and I can't find any such statement in the spec.

This strikes me as an implementation limitation of mypy (because it chooses to use the parser built in to CPython). I think there are multiple compelling reasons why this implementation decision should be revisited by mypy maintainers, so I'd prefer not to codify this in the typing spec.

srittau: The fact of the matter is that mypy – despite its shortcomings – is probably the most (explicitly) used and arguably most important type checker.[^1] In many cases, it's not super important (although of course not desirable) when the current behavior of mypy (or any other type checker for that matter) diverges from the spec, since users will usually use only one type checker for a particular project and can work around the specific idiosyncrasies. But in the case of library interfaces (which type stubs are), conformance is key. This is why I don't think the spec can just ignore mypy in this case – at least not as long there is no change in sight.[^2]

That said, I suggest that the `type` statement and potentially other newer syntax constructs should be allowed in type stubs, but support should be made optional. (A "MAY" in [RFC 2119](https://datatracker.ietf.org/doc/html/rfc2119#section-5) parlance.) The guide can then recommend against using those constructs for the time being. The spec can periodically be updated to require support for certain syntax features as mypy (and other type checkers) catch up to support them unconditionally.

[^1]: pyright is such a fantastic type checker and I wish mypy would come close, especially when it comes to support for new features and response time for bug fixes.
[^2]: I also wish mypy would switch to an external ast parser and not rely on the ast module. Or at least significantly increase the minimum required Python version to run mypy.

rchen152: I took a stab at rewording this paragraph; PTAL.

I think it's important for the spec to explicitly state which version of Python syntax should be supported in stubs. Unlike for Python source files, the version isn't necessarily coupled to the runtime, so which one to use is potentially unclear. (Maybe this is already obvious to everyone, but it hadn't been stated yet, so I just wanted to make sure we're on the same page here.)

I agree with Sebastian that this version should be the earliest non-EOL'd version until the situation with mypy changes. One of the [purposes](https://typing.readthedocs.io/en/latest/spec/type-system.html#purpose) of the spec is to "provide library authors guarantees to rely on when working with multiple type checkers." The more the spec diverges from reality, the less useful it becomes for that purpose.

erictraut: I recognize and appreciate that mypy is important. For that reason, I think it's good for us to provide guidance to stub authors in the "guides" section that they should avoid the use of newer language features if they want to maximize compatibility.

I don't think it makes sense for the typing spec to say that conformant type checkers SHOULD support `type` statements in ".py" files, but MAY support `type` statements in stub files. Stub files shouldn't be special in this regard. Conformant type checkers should support all of the features in the latest published version of Python.

I'll note that mypy fails to conform with the typing spec [in a number of areas](https://htmlpreview.github.io/?https://github.com/python/typing/blob/main/conformance/results/results.html). While that's not ideal, it's understandable. We should make sure that the community understands where mypy is not conformant so they can work around those limitations. Let's make sure we document this, but IMO this documentation doesn't belong in the typing spec.

One option is to amend the typing spec with some variant of what Jelle suggests above: "support for newer typing features that rely on grammar changes is optional if the type checker is implemented in Python and is running on an older version of Python." This statement would apply to both ".py" and ".pyi" files. I personally don't think the spec should say that because it would be codifying a current limitation in mypy. As a TC member, I would vote against such a change, but maybe other TC members would be in favor of it.

I understand why mypy originally leveraged the CPython parser, but that implementation decision has not aged well. And it's an implementation decision that could easily be revisited — especially now that the CPython code base uses the PEG parser. Mypy maintainers could incorporate the latest parser form the CPython code base if this were a priority.

rchen152: Hmm, I think we're at a bit of an impasse here. I understand where you're coming from, but it feels unnecessarily confusing to me for the spec and guides to say different things on supported syntax in stubs.

Discussion on this PR seems to have largely died down, so maybe it's a good time to ask the rest of the typing council to weigh in. I'll make a note of this and other unresolved points when opening the typing council issue.

docs/spec/distributing.rst
JelleZijlstra: Thinking about this again, the current text is fine.

docs/spec/distributing.rst
```diff
@@ -10,72 +10,156 @@ Stub files
...
+by stub files is independent from the Python versions supported by the
+implementation, and from the Python version the type checker runs under (if
+any).
+
+Type checkers should evaluate all annotation expressions as if they are quoted.
+Consequently, forward references do not need to be quoted, and syntax from newer
+versions than otherwise supported may be used in annotations. For example, the
+pipe union syntax (``X | Y``) introduced in Python 3.10 may be used even before
+Python 3.9 reaches end-of-life. Outside of annotations, type checkers may but
+are not required to support type system features that use newer syntax.
```
JelleZijlstra, line 55: I think this text is still somewhat confusing, in particular because it uses "syntax" twice in different meanings. I would restrict "syntax" to what the parser looks at, so the ``|`` union operator is not "syntax". I tried to make the text more explicit about the requirement: stubs should be *parsable* in all Python versions, but once the code is parsed, it may use features that do not exist in the earliest supported Python version.

```suggestion
Stub files are syntactically valid Python files with a ``.pyi`` suffix. They
should be parseable (e.g., with :py:func:`ast.parse`) in all Python versions
supported by the implementation. For example, defining a type alias using the
``type`` keyword is only accepted by the Python parser in Python 3.12 and later,
so stubs supporting Python 3.11 or earlier versions should not use this syntax.
Type checkers may choose to support syntactic features from newer Python versions
in stub files, but stubs that rely on such features may not be portable to all
type checkers.

Type checkers should evaluate all :term:`annotation expressions <annotation expression>` as if they are quoted.
Consequently, forward references do not need to be quoted, and type system
features that do not depend on Python syntax changes are supported in stubs regardless
of the Python version supported. For example, the use of the ``|`` operator
to create unions (``X | Y``) was introduced in Python 3.10, but may be used
even in stubs that support Python 3.9 and older versions. 
```

This I believe accurately describes the current state of stubs in e.g. typeshed. Restricting stubs to what can be parsed by `ast.parse` allows type checkers that rely on `ast.parse` (mypy and pyanalyze) to work and more generally makes it possible for tools that are written in Python to analyze stub files without having to build their own parser. I think that's useful and desirable.

erictraut: I think this phrasing is better than previous drafts, but I'm still uneasy with it. I again question whether guidance to stub authors (what they should or should not put in the stubs) belongs in the typing spec. The typing spec is meant to specify the required and optional behavior for type checkers. From that perspective, the new sentence that begins "Type checkers may choose..." seems appropriate for the typing spec, but the sentence preceding it seems out of place. Is there a way it can be rephrased such that it indicates what is expected of type checkers?

In the phrase "in all Python versions supported by the implementation", what does "implementation" mean? Is this referring to the type checker implementation or the implementation of the library associated with the stub? If it's the latter, that would seem to imply that a library that supports Python 2.x wouldn't be able to use any Python 3.x syntax in its corresponding stub.

How about something like "A type checker implemented in Python may depend upon the Python parser (e.g., with :py:func:`ast.parse`) to parse source and stub files. If this type checker runs on an older version of Python, it will not be able to parse newer syntactic constructs. For example, a type alias defined with the `type`` keyword is accepted by the Python parser starting in Python 3.12, so a stub that uses this syntax may not be supported by a type checker running on Python 3.11 or older. For further guidance about authoring stubs, refer to <link to guide>".

JelleZijlstra: > The typing spec is meant to specify the required and optional behavior for type checkers. 

In PEP 729, we wrote that type checkers are one audience for the spec, but not the only one: https://peps.python.org/pep-0729/#specification-for-the-type-system. I see this section as specifying what is required from both type checkers and stub authors. On the one hand, we have to specify what type checkers need to support in order to understand a conformant stub; on the other hand, we have to specify what features stub authors can use when they write stubs in order to be conformant.

From that perspective, I actually feel that the "Type checkers may choose..." sentence is redundant. Type checkers may always choose to support more than what the spec requires, but the spec should be precise about the minimum required, so that stub authors can know what to rely on.

> what does "implementation" mean?

An earlier paragraph in this document defines "implementation" (and it means what you guessed). But you make a good point about Python 2; we shouldn't require Python 2 syntax. A good general solution would be to require that stubs are parseable only in Python version that are still supported by CPython (i.e., currently 3.8 through 3.13).

> A type checker implemented in Python may depend upon the Python parser (e.g., with :py:func:ast.parse) to parse source and stub files

This feels more like a motivation than a specification. The spec should primarily say what behavior is required, not why it is that way, though we can add explanatory text where it is helpful.

With that in mind here's another suggested text:

```suggestion
Stub files are syntactically valid Python files with a ``.pyi`` suffix. They
should be parseable (e.g., with :py:func:`ast.parse`) in all Python versions
that are supported by the implementation and that are still supported
by the CPython project. For example, defining a type alias using the
``type`` keyword is only accepted by the Python parser in Python 3.12 and later,
so stubs supporting Python 3.11 or earlier versions should not use this syntax.
This allows type checkers implemented in Python to parse stub files using
functionality from the standard library.
Type checkers may choose to support syntactic features from newer Python versions
in stub files, but stubs that rely on such features may not be portable to all
type checkers. Type checkers may also choose to support Python versions that
are no longer supported by CPython; if so, they cannot rely on standard library
functionality to parse stub files.

Type checkers should evaluate all :term:`annotation expressions <annotation expression>` as if they are quoted.
Consequently, forward references do not need to be quoted, and type system
features that do not depend on Python syntax changes are supported in stubs regardless
of the Python version supported. For example, the use of the ``|`` operator
to create unions (``X | Y``) was introduced in Python 3.10, but may be used
even in stubs that support Python 3.9 and older versions. 
```

erictraut: This phrasing still makes me uneasy, but I think its meaning is now clear and it represents a reasonable compromise from where we started. Thanks for listening to my concerns.

rchen152: @JelleZijlstra Thanks, I think this is a good compromise, and the wording is clearer than before.

docs/spec/distributing.rst
sobolevn: I also had this question while reading this section. Probably

```
* functions decorated with ``@typing.dataclass_transform``
``` 

can be deleted, since 

```
Type checkers are expected to understand the effects of all decorators defined
in the ``typing`` module, plus these additional ones:
```

covers it.

docs/spec/distributing.rst
```diff
@@ -266,3 +358,60 @@ of that Python version. This can be queried e.g.
...
 binary, in case it is not in the path.
+
+.. _library-interface:
+
+Library interface (public and private symbols)
+----------------------------------------------
+
+If a ``py.typed`` module is present, a type checker will treat all modules
+within that package (i.e. all files that end in ``.py`` or ``.pyi``) as
+importable unless the file name begins with an underscore. These modules
```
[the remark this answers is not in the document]
erictraut: Which "object" are you referring to here? This statement is talking about modules (files) in a package.

docs/spec/distributing.rst
```diff
@@ -266,3 +358,60 @@ of that Python version. This can be queried e.g.
...
+-  Local variables within a function (including nested functions) are
+   always considered private.
+
+The following idioms are supported for defining the values contained
+within ``__all__``. These restrictions allow type checkers to statically
+determine the value of ``__all__``.
+
+-  ``__all__ = ('a', b')``
+-  ``__all__ = ['a', b']``
+-  ``__all__ += ['a', b']``
```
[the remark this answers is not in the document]
erictraut: I don't think that's necessary. Type checkers generally don't type check stubs (i.e. they don't emit errors for stubs that your code imports). Of course, if you ask a type checker to type check a stub, it should perform all of its normal type checks, but that's true of all elements within the stub, not just `__all__ += ...` statements.

docs/spec/distributing.rst
```diff
@@ -266,3 +358,60 @@ of that Python version. This can be queried e.g.
...
+within that package (i.e. all files that end in ``.py`` or ``.pyi``) as
+importable unless the file name begins with an underscore. These modules
+comprise the supported interface for the library.
+
+Each module exposes a set of symbols. Some of these symbols are
+considered "private” — implementation details that are not part of the
+library’s interface. Type checkers can use the following rules
+to determine which symbols are visible outside of the package.
+
+-  Symbols whose names begin with an underscore (but are not dunder
```
sobolevn, line 377: What about objects `@typing.type_check_only`? Are they exposed?

erictraut: `typing.type_check_only` is orthogonal to whether a symbol should be considered public or private. It means that a type checker should report an error if that symbol is used in a runtime context. These symbols can still be used in type expressions if they are publicly exposed from a library or stub file.

docs/spec/distributing.rst
carljm: I think the reasoning behind the inclusion of this bullet, with its current wording, is that a function decorated with `typing.dataclass_transform` is itself a decorator (which is not itself defined in the `typing` module), whose function must be understood by type checkers.

A class decorated by `typing.dataclass_transform` is not then itself a decorator, so this case is fully covered by "all decorators defined in the `typing` module."

carljm (approved):
Looks good to me. Thanks for working on this!

carljm:
@erictraut You've signed off on this change over at https://github.com/python/typing-council/issues/34, but your review status on this PR is still "changes requested."

erictraut (approved):

erictraut:
@carljm, thanks for noting that. I just marked the PR as approved. 

@rchen152, thanks again for all your work on this update! All TC members have signed off on it, so I'll proceed with the merge. You can announce it on the discord channel if you'd like.

Issue #1758, opened 2024-06-02
pytype conformance tests crash if no `python3.11` executable is in `PATH`


AlexWaygood:
I attempted to run the conformance tests locally using Python 3.12, following the instructions given in https://github.com/python/typing/blob/main/conformance/README.md#running-the-conformance-test-tool. The mypy, pyright and pyre tests all ran, but the pytype tests crashed with the following traceback:

```pytb
Running tests for pytype
Traceback (most recent call last):
  File "/Users/alexw/dev/typing/conformance/.venv/lib/python3.12/site-packages/pytype/config.py", line 704, in _store_python_version
    self.output_options.python_exe = compiler.get_python_executable(version)
                                     ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/Users/alexw/dev/typing/conformance/.venv/lib/python3.12/site-packages/pytype/pyc/compiler.py", line 135, in get_python_executable
    raise PythonNotFoundError()
pytype.pyc.compiler.PythonNotFoundError

During handling of the above exception, another exception occurred:

Traceback (most recent call last):
  File "/Users/alexw/dev/typing/conformance/src/main.py", line 258, in <module>
    main()
  File "/Users/alexw/dev/typing/conformance/src/main.py", line 251, in main
    run_tests(root_dir, type_checker, test_cases, skip_timing=options.skip_timing)
  File "/Users/alexw/dev/typing/conformance/src/main.py", line 30, in run_tests
    tests_output = type_checker.run_tests([file.name for file in test_cases])
                   ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/Users/alexw/dev/typing/conformance/src/type_checker.py", line 287, in run_tests
    options = pytype_config.Options.create(python_version=(3, 11), quick=True)
              ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/Users/alexw/dev/typing/conformance/.venv/lib/python3.12/site-packages/pytype/config.py", line 107, in create
    return cls(options)
           ^^^^^^^^^^^^
  File "/Users/alexw/dev/typing/conformance/.venv/lib/python3.12/site-packages/pytype/config.py", line 88, in __init__
    Postprocessor(names, opt_map, options, self).process()
[...]

PR #1822, opened 2024-07-27
Add conformance tests for PEP 702's `@deprecated`

### LINKED ISSUE #1821: Conformance tests: PEP 702
PEP 702's `@deprecated` has been [a part of the specs](https://typing.readthedocs.io/en/latest/spec/directives.html#deprecated) for a while now, yet there aren't any conformance tests for it. Is it simply because no one have volunteered to do so (in which case I'll be interested in submitting a PR), or are such tests not stored in this repository?

InSyncWithFoo:
Fixes #1821.

I'm not sure how to make Pytype run properly (#1758), probably because I'm using [Rye](https://github.com/astral-sh/rye), so I'm skipping manual tests.


conformance/tests/directives_deprecated_pep702_example.py
```diff
@@ -0,0 +1,44 @@
+"""
```
erictraut, line 1: I'm not sure it makes sense for this feature to have three separate test files in the conformance suite. It's a pretty straightforward feature, so it probably makes sense to combine them into a single test "directives_deprecated.py" that appears in the conformance summary as a single row. Just trying to keep the test granularity consistent between features.

In any case, the test name should not include the PEP number. The PEP is the original spec, but it is superseded by the typing spec. The fact that the feature was originally part of a PEP is a historical footnote only.

conformance/tests/directives_deprecated_protocol.py
```diff
@@ -0,0 +1,34 @@
...
+
+    @deprecated("Deprecated")
+    def foo(self) -> None: ...
+
+    def bar(self) -> None: ...
+
+
+class Fooer(Fooable):
+
+    def foo(self) -> None:  # E?: Implementation of deprecated method foo
```
JelleZijlstra, line 26: I don't think this should be a warning at all, unless it is marked with `@override`.

The scenario envisioned in the quoted text is actually a little different and more convoluted, something like this:

```
class Fooable(Protocol):
    def foo(self) -> None: ...

class Concrete:
    @deprecated("Deprecated")
    def foo(self) -> None: ...

def take_fooable(f: Fooable) -> None: ...

def caller(c: Concrete) -> None:
    take_fooable(c)  # Concrete is a Fooable, but only because of a deprecated method
```

erictraut (changes requested):
Thanks for doing this!

This looks pretty good, but I think the three test files should be combined into one for the reasons I mention in the comment.

You should also run the test suite and manually score the four type checkers, update the corresponding toml files, and do a final run to generate the [conformance summary report](https://htmlpreview.github.io/?https://github.com/python/typing/blob/main/conformance/results/results.html).

conformance/results/pyre/aliases_explicit.toml
```diff
@@ -9,26 +9,6 @@ Incorrectly rejects import alias of `TypeAlias` when used to define type alias.
 Does not report invalid specialization of already-specialized generic type alias.
 """
 output = """
-aliases_explicit.py:23:0 Incompatible variable type [9]: GoodTypeAlias9 is declared to have type `TA` but is used as type `Type[typing.Callable[..., Variable[$synthetic_attribute_resolution_variable]]]`.
```
erictraut, line 12: Hmm, it looks like pyre isn't running properly on your system. It should produce errors in these cases. You'll need to determine why it doesn't work on your system.

conformance/results/pyright/directives_deprecated.toml
```diff
@@ -0,0 +1,32 @@
+conformant = "Partial"
+notes = """
+Does not report error for deprecated magic methods
```
erictraut, line 3: Nit: for consistency, please add a period at the end of these sentences. These appear in the hover text within the summary report.

conformance/results/pyright/directives_deprecated.toml
```diff
@@ -0,0 +1,32 @@
...
+Outputs an extra diagnostic when the imported symbol is used
```
erictraut, line 4: I don't think this violates the spec. This is a use of a deprecated symbol.

conformance/results/pytype/directives_deprecated.toml
```diff
@@ -0,0 +1,26 @@
+conformant = "Unsupported"
+notes = """
+Does not support @deprecated
```
erictraut, line 3: Nit: Please include period at end of this sentence.

conformance/results/mypy/directives_deprecated.toml
```diff
@@ -0,0 +1,20 @@
+conformant = "Unsupported"
+notes = """
+Does not support @deprecated
```
erictraut, line 3: Nit: Please include period at end of this sentence. This is displayed in the hover text within the conformance summary report.

erictraut (changes requested):

conformance/tests/_directives_deprecated_library.py
```diff
@@ -0,0 +1,44 @@
+"""
+Support module for directive_deprecated_pep702_example.
```
erictraut, line 2: This header comment is no longer correct.

conformance/tests/directives_deprecated.py
```diff
@@ -0,0 +1,108 @@
...
+# > For deprecated overloads, this includes all calls that resolve to the deprecated overload.
+
+library.foo(1)  # E: Use of deprecated overload for foo
+library.foo("x")  # no error
+
+
+ham = Ham()  # no error (already reported above)
+
+
+# > Any syntax that indirectly triggers a call to the function.
```
erictraut, line 36: Another useful test is to verify that a `__call__` marked as deprecated (e.g. within a callback protocol) triggers a deprecation check if the object is called.

conformance/tests/directives_deprecated.py
```diff
@@ -0,0 +1,108 @@
+"""
+Tests the warnings.deprecated function.
+"""
+
+# pyright: reportDeprecated=true
+
+# Specification: https://typing.readthedocs.io/en/latest/spec/directives.html#deprecated
+# See also https://peps.python.org/pep-0702/
+
+# > Type checkers should produce a diagnostic whenever they encounter a usage of an object
```
erictraut, line 10: There's an open question in the spec about whether decorators applied on top of a `@deprecated` decorator should retain the deprecation message — and under what circumstance. 

```python
def other_decorator[**P, R](c: Callable[P, R]) -> Callable[P, R]: ...

@other_decorator
@deprecated("???")
def func() -> None: ...
```

In pyright's implementation, I chose to record the deprecated message as part of the function's signature, and this signature is then bound to a ParamSpec. This retains the deprecated message in cases where the deprecated function is passed through other decorators. The spec isn't really clear about this behavior, so maybe it's best to leave out of the conformance suite for now. Maybe @JelleZijlstra has thoughts about this.

erictraut (commented):
I'm not sure what to do about the pyre tests. I don't think we can commit the PR as is. Options:
1. You find a way to run the pyre tests on your machine.
2. Someone else runs the tests and updates the PR on your branch before you commit it.
3. You revert all of the pyre test results, we merge your PR, and then I immediately follow it up with another PR that addresses the pyre results.

Option 1 is cleanest if you're able to do it. If that's not feasible, then option 3 might work the best.

conformance/tests/directives_deprecated.py
JelleZijlstra: I feel the behavior users would expect is that the deprecation gets preserved.

JelleZijlstra:
I'm running the script locally now and will update pyre.

conformance/results/pyre/directives_reveal_type.toml
```diff
@@ -1,5 +1,6 @@
 conformant = "Partial"
-notes = """"
+notes = """
+"
```
JelleZijlstra, line 3: There was an extra `"` at the beginning of the line, somehow my rerun separated it out.
```suggestion
```

erictraut (approved):
Thanks again for doing this! The latest draft looks good to me.

PR #1841, opened 2024-08-15
Specify that `type[]` distributes over unions

InSyncWithFoo:
I think this much can be inferred, since the specs already said that:

* `type[A | B]` is supported, and
* `B` the class is a valid value for `type[A | B]`

### DIFF
--- a/docs/spec/special-types.rst
+++ b/docs/spec/special-types.rst
@@ -176,7 +176,7 @@ Note that it is legal to use a union of classes as the parameter for
       ...
 
 ``type[]`` distributes over unions:
-``type[A | B]`` is the same as ``type[A] | type[B]``.
+``type[A | B]`` is consistent with ``type[A] | type[B]``.
 
 However, the actual argument passed in at runtime must still be a
 concrete class object, e.g. in the above example::
--- a/docs/spec/special-types.rst
+++ b/docs/spec/special-types.rst
@@ -176,7 +176,7 @@ Note that it is legal to use a union of classes as the parameter for
       ...
 
 ``type[]`` distributes over unions:
-``type[A | B]`` is equivalent to ``type[A] | type[B]``.
+``type[A | B]`` is :term:`equivalent` to ``type[A] | type[B]``.
 
 However, the actual argument passed in at runtime must still be a
 concrete class object, e.g. in the above example::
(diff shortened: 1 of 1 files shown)


erictraut:
Since this is a substantive change to the spec, it should go through formal TC review. Please follow the process documented [here](https://github.com/python/typing-council#decisions).

docs/spec/special-types.rst
```diff
@@ -175,7 +175,10 @@ Note that it is legal to use a union of classes as the parameter for
       user = new_user(user_class)
       ...
 
-However the actual argument passed in at runtime must still be a
+``type[]`` distributes over unions:
+``type[A | B]`` is consistent with ``type[A] | type[B]``.
```
JelleZijlstra, line 179: I think the right term to use is "equivalent" (https://typing.readthedocs.io/en/latest/spec/glossary.html#term-equivalent).

erictraut: If I'm understanding the terms correctly, `equivalent` applies only to fully-static types. I think we want to accommodate types that include gradual forms here, no? After all, `type[int] | type[list[Any]]` describes the same set of values as `type[int | list[Any]]`.

JelleZijlstra: No, equivalent also applies to types that are not fully static; see the linked definition. Fully static types are equivalent if they cover the same set of values; types that are not fully static are equivalent if the set of types they can materialize to is the same.

erictraut: Ah yes, I was thrown off by [this table](https://typing.readthedocs.io/en/latest/spec/concepts.html#summary-of-type-relations). I should have read the paragraph below the table.

docs/spec/special-types.rst
```diff
@@ -175,7 +175,10 @@ Note that it is legal to use a union of classes as the parameter for
       user = new_user(user_class)
       ...
 
-However the actual argument passed in at runtime must still be a
+``type[]`` distributes over unions:
+``type[A | B]`` is equivalent to ``type[A] | type[B]``.
```
JelleZijlstra, line 179:
```suggestion
``type[A | B]`` is :term:`equivalent` to ``type[A] | type[B]``.
```

JelleZijlstra (approved):

carljm (approved):

hauntsaninja (approved):
Thank you!

PR #1855, opened 2024-09-16
Change int to Any in unpacking unbounded tuple types

denisborisov:
It looks like there should be `Any` instead of `int`, because the documentation says the following:
"""
We can also pass a `*tuple[int, ...]` wherever a `*Ts` is expected. This is useful when we have particularly dynamic code and cannot state the precise number of dimensions **or the precise types** for each of the dimensions.
"""

### DIFF
--- a/docs/spec/generics.rst
+++ b/docs/spec/generics.rst
@@ -1262,7 +1262,7 @@ unnecessary ``TypeVarTuple``:
     process_batch_channels(z)  # Error: Expected Channels.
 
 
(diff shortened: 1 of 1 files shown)


srittau (approved):

srittau:
Thanks, this seems to be a [PEP 646](remnant from https://peps.python.org/pep-0646/#unpacking-unbounded-tuple-types), where `Tuple[int, ...]` as used in the section as an example.

JelleZijlstra (approved):

PR #1860, opened 2024-09-27
Permit inferring Self for unannotated self

hauntsaninja:
https://discuss.python.org/t/permit-inferring-self-for-unannotated-self/65190
https://typing--1860.org.readthedocs.build/en/1860/spec/annotations.html

### DIFF
--- a/docs/spec/annotations.rst
+++ b/docs/spec/annotations.rst
@@ -355,9 +355,7 @@ In most cases the first argument of instance and class methods
 If the argument is not annotated, then for instance methods it is
 assumed to have the type of the containing class or :ref:`Self
 <self>`, and for class methods the type object type corresponding to
-the containing class object or ``type[Self]``. In a class method, the
-precise type of the first argument cannot be represented using the
-available type notation.
+the containing class object or ``type[Self]``.
 
 In addition, the first argument in an instance method can be annotated
 with a type variable. In this case the return type may use the same
--- a/docs/spec/annotations.rst
+++ b/docs/spec/annotations.rst
@@ -349,16 +349,15 @@ types cannot be specified::
 Annotating instance and class methods
 -------------------------------------
 
-In most cases the first argument of class and instance methods
+In most cases the first argument of instance and class methods
 (conventionally named ``self`` or ``cls``) does not need to be annotated.
 
 If the argument is not annotated, then for instance methods it is
 assumed to have the type of the containing class or :ref:`Self
 <self>`, and for class methods the type object type corresponding to
-the containing class object or ``type[Self]``. For example, in class
-``A`` the first argument of an instance method has the implicit type
-``A``. In a class method, the precise type of the first argument
-cannot be represented using the available type notation.
+the containing class object or ``type[Self]``. In a class method, the
+precise type of the first argument cannot be represented using the
+available type notation.
 
 In addition, the first argument in an instance method can be annotated
 with a type variable. In this case the return type may use the same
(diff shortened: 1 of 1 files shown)


docs/spec/annotations.rst
```diff
@@ -354,9 +350,16 @@ Annotating instance and class methods
...
-type corresponding to the containing class object for class methods.
+(conventionally named ``self`` or ``cls``) does not need to be annotated.
+
+If the argument is not annotated, then for instance methods it is
+assumed to have the type of the containing class or :ref:`Self
+<self>`, and for class methods the type object type corresponding to
+the containing class object or ``type[Self]``. For example, in class
+``A`` the first argument of an instance method has the implicit type
+``A``. In a class method, the precise type of the first argument
+cannot be represented using the available type notation.
```
stroxler, line 361: I'm not 100% sure what this last sentence means.

I think `type[Self]` is a fairly precise type - it's a type variable bound above by `type[<current-class>]` but can be narrowed on callsites where we know more.

stroxler (commented):
Overall looks good to me but I was confused by some of the wording

docs/spec/annotations.rst
hauntsaninja: Yeah, that sentence is from PEP 484, it confused me as well. I thought it was maybe talking about the usual thing where args to `cls(...)` are hopeless, but I don't really know

JelleZijlstra: Yes, that sentence makes no sense to me either, I think we should remove it.

erictraut: Except for the last sentence, I think it looks reasonable. 

I do think the spec should eventually be updated to mandate that `self` has the implied type `Self` and `cls` has the implied type `type[Self]`. Many important use cases involving the `Self` type break if this isn't the case. In retrospect, PEP 673 (which introduced the `Self` type) should have specified this IMO. But if you'd prefer to hold off on that change, then your proposed wording seems fine.

docs/spec/annotations.rst
```diff
@@ -353,10 +349,14 @@ types cannot be specified::
...
-does not need to be annotated, and it is assumed to have the
-type of the containing class for instance methods, and a type object
-type corresponding to the containing class object for class methods.
+In most cases the first argument of instance and class methods
+(conventionally named ``self`` or ``cls``) does not need to be annotated.
+
+If the argument is not annotated, then for instance methods it is
+assumed to have the type of the containing class or :ref:`Self
+<self>`, and for class methods the type object type corresponding to
+the containing class object or ``type[Self]``.
```
carljm, line 358: I don't think it is accurate to make an explicit equivalence here between the type `Self` and "the type of the containing class". This is not actually what the specification for `Self` says: `Self` is not "the type of the containing class," it is "a typevar with a bound of the containing class". This distinction is important for how the `Self` type is handled when used in other arguments, and in inheritance cases. 

I suggest we simply don't attempt a mini-definition of the meaning of `Self` here, and allow the `Self` spec to handle this in depth.

```suggestion
assumed to have the type :ref:`Self <self>`, and for class methods ``type[Self]``.
```

carljm (commented):
Thanks for clarifying this!

docs/spec/annotations.rst
carljm: Oh! I see how that could be the interpretation of the "or" here. If that's the intention, then I suggest we make it a bit more explicit. (I'll make a separate suggested edit, since it would involve one more line than I included in this one.)

carljm, line 358:
```suggestion
If the argument is not annotated, then for instance methods it may be
inferred to have either the type of the containing class, or the type :ref:`Self
<self>`. For class methods it may be inferred to have either the type object
type corresponding to the containing class object, or ``type[Self]``.
```

carljm: Personally I think inferring `Self` is the better option here, and we could also consider just specifying that, but I see that may be a bigger change than was intended in this PR.

PR #1863, opened 2024-10-10
Updated conformance tests and results for latest versions of type che…

erictraut:
…ckers.

One test case in the `generics_syntax_infer_variance` was commented out temporarily until we decide how to handle variance inference for dataclasses in Python 3.13.

Mypy version was updated from 1.11.1 to 1.11.2. Minor error message change but no other substantive changes to the conformance results.

Pyright version was updated from 1.1.374 to 1.1.384. Many error messages changed slightly. Tests also identified three conformance regressions.

Pytype version was updated from 2024.04.11 to 2024.09.13. Error message format changed significantly, requiring updates to the test logic. No other substantive changes to conformance results.

Changed test infrastructure to make sure old versions of type checkers are uninstalled before installing new to guarantee correct version and increase determinism of timings.


hauntsaninja (commented):
On macOS there's an issue where a fresh install of (mypyc-compiiled) mypy will trigger a malware scanner and so the first time will often be really slow. In mypy_primer I work around this by running `mypy --version` once before recording any times.

We should maybe strip ANSI codes from the pytype output

conformance/results/pytype/aliases_implicit.toml
```diff
@@ -43,8 +335,8 @@ Line 117: Expected 1 errors
...
+Line 54: Unexpected errors ["aliases_implicit.py:54:9: \\x1b[1m\\x1b[31merror\\x1b[39m\\x1b[0m: in <module>: Invalid type annotation '[str, str]'  [invalid-annotation]", "aliases_implicit.py:54:9: \\x1b[1m\\x1b[31merror\\x1b[39m\\x1b[0m: in <module>: Invalid type annotation 'Callable[Concatenate[int, P], R][[str, str], None]'  [invalid-annotation]"]
```
JelleZijlstra, line 338: Does this coloring show up correctly in the output tables we generate? It might be better to configure pytype to not use color, if possible.

JelleZijlstra (approved):

conformance/results/pytype/aliases_implicit.toml
erictraut: No, this doesn't show up in the output report. 

Unfortunately, I don't see a way to disable color. If there is, it's not documented.

PR #1864, opened 2024-10-11
Fixed a bug in the protocols_generic.py test related to protocol matc…

erictraut:
…hing when method-scoped type variables were used. Both pyright and mypy were previously reported as failing this test, but they were actually working correctly. This was a bug in the original test case (which I wrote).

Updated to the latest version of pytest (2024.10.11). It updates many error messages but has no substantive changes to the conformance results.


JelleZijlstra (approved):
Agree this case should be an error.

PR #1868, opened 2024-10-23
Suggest `pyupgrade` or `ruff` for modernizing superseded typing features

Zac-HD:
Given the existence of automatic upgrade tools, I'd rather not do this by hand; and I think it's worth giving the tip that they exist for readers who don't know the ecosystem as well as many maintainers might.


hauntsaninja (approved):
Thanks!

PR #1871, opened 2024-10-24
Add missing return type to generic_syntax_infer_variance.py

migeed-z:
Since the spec does not specify that inferring the return type is required, we add a missing return annotation to this test case. 

The reason for this is that if an inference algorithm depends on the return annotation, it will fail and be incorrectly non-conformant.


conformance/tests/generics_syntax_infer_variance.py
```diff
@@ -2,27 +2,30 @@
...
 
 
 T = TypeVar("T", infer_variance=True)
 K = TypeVar("K", infer_variance=True)
 V = TypeVar("V", infer_variance=True)
 
-S1 = TypeVar("S1", covariant=True, infer_variance=True)  # E: cannot use covariant with infer_variance
+S1 = TypeVar(
+    "S1", covariant=True, infer_variance=True
+)  # E: cannot use covariant with infer_variance
```
erictraut, line 18: It looks like you reformatted this line and the one below. This changes the test because the comment is now on the wrong line.

conformance/tests/generics_syntax_infer_variance.py
```diff
@@ -2,27 +2,30 @@
 Tests the handling of "infer_variance" parameter for TypeVar.
 """
 
+from dataclasses import dataclass
```
erictraut, line 5: Please move this import back to its previous location.

conformance/tests/generics_syntax_infer_variance.py
```diff
@@ -2,27 +2,30 @@
...
-S2 = TypeVar("S2", contravariant=True, infer_variance=True)  # E: cannot use contravariant with infer_variance
+S2 = TypeVar(
+    "S2", contravariant=True, infer_variance=True
+)  # E: cannot use contravariant with infer_variance
 
 
 class ShouldBeCovariant1(Generic[T]):
-    def __getitem__(self, index: int) -> T:
-        ...
+    def __getitem__(self, index: int) -> T: ...
```
erictraut, line 26: Please don't needlessly reformat the file because it impacts all of the conformance results.

conformance/results/pyright/generics_syntax_infer_variance.toml
```diff
@@ -1,62 +1,66 @@
...
   "ShouldBeContravariant1[int]" is not assignable to "ShouldBeContravariant1[float]"
     Type parameter "T@ShouldBeContravariant1" is contravariant, but "int" is not a supertype of "float"
       "float" is not assignable to "int" (reportAssignmentType)
 """
-conformance_automated = "Pass"
+conformance_automated = "Fail"
 errors_diff = """
+Line 18: Expected 1 errors
+Line 22: Expected 1 errors
+Line 17: Unexpected errors ['generics_syntax_infer_variance.py:17:42 - error: TypeVar cannot be both covariant and contravariant (reportGeneralTypeIssues)']
```
erictraut, line 64: These unexpected errors are caused by the reformatting in the test.

erictraut (changes requested):

conformance/results/pyright/generics_syntax_infer_variance.toml
migeed-z: Thank you for the feedback! I fixed the formatting issues.

erictraut (approved):
Thanks for making the changes. LGTM!

PR #1877, opened 2024-10-29
Mention use of cast in enum type stub example

hauntsaninja:
In case people gloss over the text, it's good for the example to recommend the thing with the most static information

### DIFF
--- a/docs/spec/enums.rst
+++ b/docs/spec/enums.rst
@@ -124,7 +124,7 @@ statically in cases where dynamic values are used.
         species: str  # Non-member attribute
 
         CAT = 1  # Member attribute with known value and type
-        DOG = cast(str, ...)  # Member attribute with unknown value and known type
+        DOG = cast(int, ...)  # Member attribute with unknown value and known type
         BIRD = ...  # Member attribute with unknown value and type
 
 * Members defined within an enum class should not include explicit type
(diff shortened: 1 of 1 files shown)


docs/spec/enums.rst
```diff
@@ -123,8 +123,9 @@ statically in cases where dynamic values are used.
         genus: str  # Non-member attribute
         species: str  # Non-member attribute
 
-        CAT = ...  # Member attribute
-        DOG = ...  # Member attribute
+        CAT = 1  # Member attribute with known value and type
+        DOG = cast(str, ...)  # Member attribute with unknown value and known type
```
carljm, line 127: As a reader, I think I would be confused by the use of `str` here, given that this appears to be a reimplementation of the above `Pet` enum. Why not just use `int` and keep things consistent?

hauntsaninja, line 127:
```suggestion
        DOG = cast(int, ...)  # Member attribute with unknown value and known type
```

hauntsaninja: Changed it! I was anchored to some of the conversations we had around using `_value_` as a way to spell this (which would work for homogenous types, but not heterogeneous types)

carljm (approved):
Looks fine to me! This is a pretty minor change, but I guess technically any spec change is supposed to go to the whole typing council? Might be good to get at least one more opinion?

erictraut:
Looks good to me too.

JelleZijlstra:
I think this is fine to merge without the full approval procedure because it doesn't change any normative part of the spec, it just clarifies an example.

PR #1879, opened 2024-10-31
Clarify some behavior around user-defined generic classes

rchen152:
* `Protocol[T]` behaves like `Generic[T]` when defining a generic class.
* It's an error to include both `Generic[T]` and `Protocol[T]`.
* When `Generic[T]` is omitted, type variables are taken in order of first appearance in generic base classes.

Discussion: https://discuss.python.org/t/clarifying-the-rules-for-subclassing-generic-classes/69698.

### DIFF
--- a/docs/spec/generics.rst
+++ b/docs/spec/generics.rst
@@ -93,8 +93,9 @@ This is equivalent to omitting the generic notation and just saying
 User-defined generic types
 --------------------------
 
-You can include a ``Generic`` base class to define a user-defined class
-as generic.  Example::
+You can define a user-defined class as generic by including a ``Generic``
+base class, either explicitly or implicitly via the ``Protocol[T]``
+shorthand for :ref:`generic protocols<generic-protocols>`. Example::
 
   from typing import TypeVar, Generic
   from logging import Logger
@@ -144,7 +145,6 @@ A generic type can have any number of type variables, and type variables
 may be constrained. This is valid::
 
   from typing import TypeVar, Generic
-  ...
 
   T = TypeVar('T')
   S = TypeVar('S')
@@ -156,29 +156,16 @@ Each type variable argument to ``Generic`` must be distinct. This is
 thus invalid::
 
   from typing import TypeVar, Generic
-  ...
 
   T = TypeVar('T')
 
   class Pair(Generic[T, T]):   # INVALID
       ...
 
-The ``Generic[T]`` base class is redundant in simple cases where you
-subclass some other generic class and specify type variables for its
-parameters::
-
-  from typing import TypeVar
-  from collections.abc import Iterator
-
-  T = TypeVar('T')
-
-  class MyIter(Iterator[T]):
-      ...
-
-That class definition is equivalent to::
-
-  class MyIter(Iterator[T], Generic[T]):
-      ...
+When no ``Generic[T]`` or ``Protocol[T]`` base class is present, a defined
+class is generic if you subclass one or more other generic classes and
+specify type variables for their parameters. See :ref:`generic-base-classes`
+for details.
 
 You can use multiple inheritance with ``Generic``::
 
@@ -402,6 +389,7 @@ instead is preferred. (First, creating the subscripted class,
 e.g. ``Node[int]``, has a runtime cost. Second, using a type alias
 is more readable.)
 
+.. _`generic-base-classes`:
 
 Arbitrary generic types as base classes
 ---------------------------------------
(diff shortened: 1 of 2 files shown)


docs/spec/generics.rst
```diff
@@ -93,8 +93,9 @@ This is equivalent to omitting the generic notation and just saying
 User-defined generic types
 --------------------------
 
-You can include a ``Generic`` base class to define a user-defined class
-as generic.  Example::
+You can define a user-defined class as generic by including a ``Generic``
+base class, either explicitly or implicitly via the ``Protocol[T]``
```
sobolevn, line 97: Maybe we can change the `Protocol[T]` part a bit here? Why `T`?

JelleZijlstra: This sentence also sounds like it predates PEP 695, so it may be worth rephrasing a bit more.

docs/spec/protocol.rst
```diff
@@ -271,7 +272,8 @@ non-protocol generic types::
           ...
 
 ``Protocol[T, S, ...]`` is allowed as a shorthand for
-``Protocol, Generic[T, S, ...]``.
+``Protocol, Generic[T, S, ...]``. It is an error to include both
+``Protocol[T, S, ...]`` and ``Generic[T, S, ...]`` in a class definition.
```
JelleZijlstra, line 276: This should also be an error: `class C[T](Protocol[T]):`. The PEP 695 syntax creates an implicit inheritance from Generic, but I think it's better to be explicit.

rchen152: Added

docs/spec/generics.rst
rchen152: Thanks, I rephrased it. PTAL :)

docs/spec/generics.rst
```diff
@@ -458,8 +450,43 @@ Also consider the following example::
...
+      ...
+  class Child(Parent1[T1, T3], Parent2[T2, T3]):
+      ...
+
+That ``Child`` definition is equivalent to::
+
+  class Child(Parent1[T1, T3], Parent2[T2, T3], Generic[T1, T3, T2]):
+      ...
 
+Type checkers may warn when the type variable order is inconsistent::
```
JelleZijlstra, line 476: Should we mandate an error here instead? This code doesn't make sense, because a `Child[int, str]` is both a `Parent[int, str]` and a `Parent[str, int]` due to the double inheritance. So a type checker should error here, because it can't build up this class in a correct way.

rchen152: The reason I didn't mandate an error is that mypy currently doesn't error on this. But I'm happy to change it if we think that would be better.

docs/spec/generics.rst
```diff
@@ -93,8 +93,13 @@ This is equivalent to omitting the generic notation and just saying
...
 
-You can include a ``Generic`` base class to define a user-defined class
-as generic.  Example::
+You can define a user-defined class as generic in three ways: by including a
+``Generic`` base class, by using the  new generic class syntax in Python 3.12
+and higher, or by including a ``Protocol`` base class parameterized with type
+variables. The third approach also marks the class as a protocol - see
+:ref:`generic protocols<generic-protocols>` for more information.
+
+Example using ``Generic``::
```
erictraut, line 102: I think it would be good to specify the following rules as well. These are enforced by the runtime, but a type checker should detect and report these error conditions.

If `Generic` is included as a base class, all type parameters must be specified as type arguments to `Generic`.

```python
class Bad1(Iterable[T], Generic): ...  # Error: T does not appear in Generic type arg list
```

If `Protocol` is used as a base class, type parameters can optionally be specified as type arguments to `Protocol`, or `Protocol` can be used without type arguments.

```python
class Bad2(Iterable[T], Protocol[T]): ...  # OK
class Good2(Iterable[T], Protocol): ...  # OK
```

If type parameters are specified using either `Generic` or `Protocol` type arguments, all type parameters for the class must appear within the type argument list. A type checker should report an error if a type variable appears elsewhere in the base class list and that type variable is not included as a type argument for `Generic` or `Protocol`.

```python
class Bad3(Iterable[T], Generic[S]): ...  # Error: T does not appear in Generic type arg list
class Bad4(Iterable[T], Protocol[S]): ...  # Error: T does not appear in Protocol type arg list
```

docs/spec/generics.rst
```diff
@@ -118,14 +123,14 @@ as generic.  Example::
...
-Or, in Python 3.12 and higher, by using the new syntax for generic
-classes::
+Or, using the new generic class syntax::
 
   class LoggedVar[T]:
       # methods as in previous example
 
 This implicitly adds ``Generic[T]`` as a base class and type checkers
-should treat the two largely equivalently (except for variance, see below).
+should treat the two definitions of ``LoggedVar`` largely equivalently (except
```
erictraut, line 132: Nit: There should be a comma between "as a base class" and "and type checkers".

docs/spec/generics.rst
```diff
@@ -458,8 +450,43 @@ Also consider the following example::
   class MyDict(Mapping[str, T]):
       ...
 
-In this case MyDict has a single parameter, T.
+In this case ``MyDict`` has a single parameter, ``T``.
```
erictraut, line 453: For clarity, this should probably say "type parameter" rather than just "parameter".

docs/spec/generics.rst
rchen152: Thanks, these are all great suggestions! I've included them, rephrased slightly to try to fit the flow of the text better. I also noted that all arguments to `Generic` or `Protocol` need to be type variables, since I didn't see that mentioned anywhere.

docs/spec/generics.rst
carljm: I don't know if there is a rationale behind mypy's behavior, or how hard it would be to change, but it seems to me that this should be mandated as an error; I don't see any way to make sense of this definition.

> This code doesn't make sense, because a `Child[int, str]` is both a `Parent[int, str]` and a `Parent[str, int]` due to the double inheritance.

To make sure I'm understanding: you meant `Grandparent` here? There is no double inheritance from `Parent`.

erictraut (approved):

docs/spec/generics.rst
rchen152: Ok, I updated this section to mandate an error.

JelleZijlstra (approved):

docs/spec/generics.rst
```diff
@@ -93,8 +93,13 @@ This is equivalent to omitting the generic notation and just saying
 User-defined generic types
 --------------------------
 
-You can include a ``Generic`` base class to define a user-defined class
-as generic.  Example::
+You can define a user-defined class as generic in three ways: by including a
```
JukkaL, line 96: Should the case where `Generic` is optional be explicitly mentioned here? I mean this:
```py
from typing import TypeVar

T = TypeVar("T")

class A(list[T]): pass
```
Alternatively, this could rewritten so that it's clear that these three aren't the only ones.

JukkaL (approved):
Looks good to me, but one thing seemed a little unclear.

rchen152:
All TC members have either directly approved this PR or signed off at https://github.com/python/typing-council/issues/38.

Issue #39, opened 2015-01-15
TypeVar example in PEP is confusing


JukkaL:
[I think Guido pointed this out elsewhere, but maybe this should be addressed separately here so that it won't be forgotten.]

The following type variable constraint example from the PEP is confusing:

    from typing import Iterable
    X = TypeVar('X')
    Y = TypeVar('Y', Iterable[X])
    def filter(rule: Callable[[X], bool], input: Y) -> Y:
    ...

I think this could be written equivalently without having a constraint:

    from typing import Iterable
    X = TypeVar('X')
    def filter(rule: Callable[[X], bool], input: Iterable[X]) -> Iterable[X]:
    ...

Also, if this is similar to the `values` keyword argument to mypy's `typevar`, having a type variable with a single value would be equivalent to having a type alias (since subtypes are not valid), so the only interesting use cases would have 2 or more constraints.

There is actually no non-trivial example of using constraints in the PEP. I suggest adding an `AnyStr` example, as this seems like the most common use case for constraints (at least of the sort supported by mypy right now). Maybe the discussion should mention how this is different from bounded polymorphism/quantification as in Java, for example, as this is probably the most non-mainstream feature introduced in the PEP.

brakhane:
Does that mean that there is no way to specify that two non union types must/will be the same? In the above example, I would have expected that the first filter function is guaranteed to return a file object if called with a file object, but the second definition could return a sequence instead.

That would be indeed unexpected and should be documented. It also seems inconsistent:

    X = TypeVar("X")
    def copy(original: X) -> X

    Y  = TypeVar("Y", Iterable[Y])
    def copy_it(original: Y) -> Y

Both declarations look identical, but while the first one will guarantee that isinstance(original, copy) is true, the second one doesn't; copy((1,2,3,4,5)) could return a list, for example.

gvanrossum:
Yes, generic types and type variables need to be explained better. I think you did get the distinction between constrained and unconstrained type variables right (though your words are a bit confusing).

The best example of constrained type variables is AnyStr. See the explanation of this in http://mypy.readthedocs.org/en/latest/generics.html .

I would also like to point out that requiring a function to return an object of the exact same type as an argument is usually unreasonable, since being able to extract the type (e.g. `x.__class__`) is not sufficient to be able to create an instance. (How would you know the constructor signature?) There is some discussion of this in https://github.com/JukkaL/mypy/issues/539 (search for "Eugene").

JukkaL:
[...]

PR #1890, opened 2024-12-01
Specs clarification: Forbid `ClassVar` and `Final` within `TypedDict` body

InSyncWithFoo:
* Discussion: <i>[What does it mean for a `TypedDict` to have a `ClassVar` field?](https://discuss.python.org/t/72880)</i>
* Typing council issue: [#39: <i>Typing spec update: `ClassVar` and `Final` within `TypedDict`</i>](https://github.com/python/typing-council/issues/39)

At runtime, both `ClassVar` and `Final` are not recognized as valid qualifiers for `TypedDict` items:

```pycon
>>> from typing import TypedDict, ClassVar, Final
>>> class D(TypedDict):
...     a: ClassVar[int]
...
[truncated]
TypeError: typing.ClassVar[int] is not valid as type argument
>>> class D(TypedDict):
...     a: Final[int]
...
[truncated]
TypeError: typing.Final[int] is not valid as type argument
```

### DIFF
--- a/docs/spec/class-compat.rst
+++ b/docs/spec/class-compat.rst
@@ -89,6 +89,10 @@ annotated in ``__init__`` or other methods, rather than in the class::
       def __init__(self, content):
           self.content: T = content
 
+``ClassVar`` cannot be used as a qualifier for a :ref:`TypedDict <typeddict>`
+item or a :ref:`NamedTuple <namedtuple>`field. Such usages are also hard
+errors at runtime.
+
 .. _`override`:
 
 ``@override``
--- a/docs/spec/class-compat.rst
+++ b/docs/spec/class-compat.rst
@@ -90,7 +90,7 @@ annotated in ``__init__`` or other methods, rather than in the class::
           self.content: T = content
 
 ``ClassVar`` cannot be used as a qualifier for a :ref:`TypedDict <typeddict>`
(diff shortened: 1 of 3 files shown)


JelleZijlstra:
I'd prefer to approach this by being precise about where `ClassVar` and `Final` *are* allowed, as opposed to saying something specific to TypedDicts. For example, are they allowed in NamedTuple bodies?

docs/spec/class-compat.rst
```diff
@@ -89,6 +89,10 @@ annotated in ``__init__`` or other methods, rather than in the class::
       def __init__(self, content):
           self.content: T = content
 
+``ClassVar`` cannot be used as a qualifier for a :ref:`TypedDict <typeddict>`
+item or a :ref:`NamedTuple <namedtuple>` field. Such usages are also hard
```
erictraut, line 93: I don't think the term "hard error" is used anywhere else in the spec. I recommend changing to "Such usage also generates an error at runtime".

(Same comment for `Final` below.)

erictraut (approved):

carljm (approved):

JukkaL (approved):

rchen152 (approved):

JelleZijlstra (approved):
I feel this could still be phrased in a more general way, but we can change that later.

PR #1891, opened 2024-12-02
Remove paragraph mentioning minor edits through the mailing list

Viicos:


JelleZijlstra:
Yes, let's remove the whole paragraph. The statement in the paragraph is true for the whole spec: minor changes can be done through Typing Council action.

Viicos:
Done :+1:

JelleZijlstra (approved):
Thanks. I don't think this needs to go through the formal Typing Council decision process, but cc @carljm @JukkaL @rchen152 if you disagree.

carljm (approved):

PR #1892, opened 2024-12-03
Update conformance tests for recent generics spec change

rchen152:
Updates conformance tests for https://github.com/python/typing/pull/1879. I only added tests for the changes to generics.rst because the changes to protocol.rst are already reflected in existing tests.


erictraut (approved):
Thanks for doing this. The change LGTM.

JelleZijlstra (approved):

PR #1915, opened 2025-01-30
Allowed bare `typing.Final` in an annotation expression

superbobry:
The current grammar only allows `typing.Final[...]`, but not `typing.Final`.


JelleZijlstra (approved):

erictraut:
Yes, good catch.

hauntsaninja (approved):

PR #1916, opened 2025-01-31
Add conformance test for `@final` combined with `@property`

samwgoldman:
The language of the typing spec indicates that the `@final` decorator may be used with properties:

> The method decorator version may be used with all of instance methods, class methods, static methods, and properties.

I noticed this case while working on Pyre. I observed that Mypy and Pyright also differ in behavior, so it seemed like a good conformance test to add.

See discussion here:
https://discuss.python.org/t/proposal-allow-chaining-final-decorator-when-previous-decorators-return-a-non-function/78918

I also added an example to the spec itself. I think it's a practical example, but also happy to remove it as well, since the existing content makes it pretty clear that the example should be allowed.


hauntsaninja (approved):
Looks good, although I think conformance in pyright/qualifiers_final_decorator.toml should be updated to match the new conformance_automated value

conformance/results/pyright/qualifiers_final_decorator.toml
```diff
@@ -1,4 +1,4 @@
-conformant = "Pass"
+conformant = "Fail"
```
erictraut, line 1: This should say "Partial", not "Fail".

erictraut:
Thanks for reporting this and adding the conformance test. FYI, I've filed the following bug in the pyright issue tracker: https://github.com/microsoft/pyright/issues/9795.

samwgoldman:
Thanks for the guidance!

hauntsaninja:
I guess always a risk changing a pyright conformant value, because odds are 50/50 that Eric will have fixed it before the PR is merged :-)

docs/spec/qualifiers.rst
```diff
@@ -66,6 +66,22 @@ implementation (or on the first overload, for stubs)::
        def method(self, x=None):
            ...
 
+The ``@final`` decorator can be combined with previous decorators, like
```
JelleZijlstra, line 69: Somebody on Discuss brought up that perhaps you should put `@property` above `@final`. Do we need to specify that explicitly?

For `classmethod` + `final`, currently both [mypy](https://mypy-play.net/?mypy=latest&python=3.12&gist=b97323c411bc7615c2f3fd2307ee2c21) and [pyright](https://pyright-play.net/?code=GYJw9gtgBALgngBwJYDsDmUkQWEMrCoCGANgFBkDGJRAzrVAIIBcZU7UAAoSqWx52p1aEAKYwAFmAAm-dtNHACxEgH1CIWjAAU1WgEooAWgB8mFDGZQQ4gK4gUUAIxyorwTXpjJM9zz4cUApK-mq0opRgKNK6JAbGZqiW1nYOUABMrlSeDABC2oz6rIEewt5SsoHByrxqGlqx8abmyTYw9o4AzBQlQl7iFa7VoarhkdGNhs1JVm0dUAAsQA) allow putting the decorators in any order.

My preference would be to make the spec say explicitly that either order is allowed and add conformance tests for it.

erictraut: What is the correct behavior if a property getter is marked final but a setter or deleter is not (or vice versa)? I ask because the different ordering would seem to imply different behaviors. Does the `final` apply to the `property` object or the method that becomes the `fget` callable?

samwgoldman: I think it's a good point that the spec text here it too specific ("previous" implies an order, should read "other").

On the other hand, I think it gets a bit too precise to enumerate orders. In Pyre (and I suspect Pyright too) the non-conformance arises from attaching the decorator metadata directly to a function type, so a chained decorator that returns a non-function runs into problems.

In terms of typing, the `@property` and `@final` signatures compose in either order, so I think it's clear that the definition is legal in either order.

samwgoldman: > What is the correct behavior if a property getter is marked final but a setter or deleter is not (or vice versa)? I ask because the different ordering would seem to imply different behaviors. Does the `final` apply to the `property` object or the method that is wrapped by the property?

This is a great question. I don't feel strongly, but one option:

1. The `final`-ness applies to the slot on the class, not the type itself (property descriptor or otherwise)
2. If the final property is a descriptor, then the setter should not be called on that type (either through `p: Final = Descriptor()` or `@final`-decorated descriptor)

PR #1918, opened 2025-02-03
Add additional Unpack/kwargs conformance test cases

yangdanny97:
These two examples from the typing spec don't appear to have corresponding cases in the conformance suite:

EXAMPLE 1

> The reverse situation where the destination callable contains **kwargs: Unpack[TypedDict] and the source callable doesn’t contain **kwargs should be disallowed. This is because, we cannot be sure that additional keyword arguments are not being passed in when an instance of a subclass had been assigned to a variable with a base class type and then unpacked in the destination callable invocation

https://typing.readthedocs.io/en/latest/spec/callables.html#source-contains-kwargs-and-destination-doesn-t

EXAMPLE 2

> Therefore, kwargs hinted with an unpacked TypedDict can only be passed to another function if the function to which unpacked kwargs are being passed to has **kwargs in its signature as well, because then additional keywords would not cause errors at runtime during function invocation. Otherwise, the type checker should generate an error.

https://typing.readthedocs.io/en/latest/spec/callables.html#passing-kwargs-inside-a-function-to-another-function

This PR adds 2 new test cases and updates the generated conformance output.

Pyre's output is super fishy here, it emits an error message on line 138, but it's not what I would expect. If we agree that these tests should be added, I'll dig into Pyre's behavior some more.

Example 2 I have some questions about. At a glance it seems reasonable, but I thought it was really weird to specify this requirement just for unpacked kwargs, but not typed dicts in other argument positions (which should still be susceptible to these issues due to structural typing). For example:

```
def func7(*, v1: int, v3: str, v2: str = "") -> None: ...

def func9(x: TD2, **kwargs: Unpack[TD2]) -> None:
    func7(**kwargs)  # not OK
    func7(**x)  # OK according to the spec, but not safe
```


conformance/tests/callables_kwargs.py
```diff
@@ -121,3 +121,19 @@ def func5(v1: int, **kwargs: Unpack[TD2]) -> None:  # E: parameter v1 overlaps w
 
 def func6(**kwargs: Unpack[T]) -> None:  # E: unpacked value must be a TypedDict, not a TypeVar bound to TypedDict.
     ...
+
+
+def func7(*, v1: int, v3: str, v2: str = "") -> None:
```
erictraut, line 126: Please add the quoted text from the spec for each of the test cases. I see that you added it for the second test case but not the first. In both cases, use the `# >` quoting syntax so it's consistent with all of the other tests and the [README guidance](https://github.com/python/typing/blob/main/conformance/README.md#notes-about-tests).

conformance/results/mypy/callables_kwargs.toml
```diff
@@ -1,4 +1,4 @@
-conformant = "Pass"
+conformant = "Partial"
```
erictraut, line 1: Please add a multiline "notes" entry that summarizes the test failures. See other test result files for examples. This should be done for the pyright results as well.

erictraut (commented):
Thanks for the contribution! Overall, it looks good. Just a few minor changes needed for consistency with existing tests and test results.

erictraut (approved):
LGTM

erictraut:
@yangdanny97, this PR looks good to me. If you think it's complete, I can merge it.

erictraut:
@yangdanny97, looking at this closer, I agree with you that Example 2 seems inconsistent. It doesn't make any sense that there should be a difference between the two cases in your code snippet above. I'm not sure what to do about this. Implementing this behavior inconsistently will require some ugly special-casing within pyright's logic — and presumably in the other type checkers as well.

Rather than add Example 2 to the conformance tests, perhaps we should amend the typing spec to remove this inconsistency. Thoughts? If that sounds good to you, please update the PR to remove Example 2, and then either you or I can start a discussion in the [typing forum](https://discuss.python.org/c/typing/32) to propose changing the spec.

yangdanny97:
I had a look at Pyre's implementation. I think the error emitted for the second example is a different bug, and Pyre doesn't implement the desired behavior of banning unpacked kwargs in this situation.

I'll update the results to mark Pyre as partial here, and then we should be good to go.

yangdanny97:
> @yangdanny97, looking at this closer, I agree with you that Example 2 seems inconsistent. It doesn't make any sense that there should be a difference between the two cases in your code snippet above. I'm not sure what to do about this. Implementing this behavior inconsistently will require some ugly special-casing within pyright's logic — and presumably in the other type checkers as well.
> 
> Rather than add Example 2 to the conformance tests, perhaps we should amend the typing spec to remove this inconsistency. Thoughts? If that sounds good to you, please update the PR to remove Example 2, and then either you or I can start a discussion in the [typing forum](https://discuss.python.org/c/typing/32) to propose changing the spec.

Sure, I can update the PR to remove the second example until we sort out the spec. Ideally we have this check apply to all typed dict spreads, or none of them.

yangdanny97:
Posted https://discuss.python.org/t/typing-spec-inconsistency-for-unpacking-typed-dict-kwargs/79640

PR #1926, opened 2025-02-17
Updated conformance tests for pyright 1.1.394.

erictraut:
This version of pyright fixes a conformance issue detected by the `callables_kwargs` test. Also updated the `annotations_generatosr` and `annotations_coroutines` test cases to accommodate a more precise (narrower) type inferred for `async def` functions. Pyright now infers `types.CoroutineType` rather than `typing.Coroutine`. Both are correct and should be accepted by the test.

### DIFF
--- a/conformance/tests/annotations_coroutines.py
+++ b/conformance/tests/annotations_coroutines.py
@@ -16,7 +16,10 @@ async def func1(ignored: int, /) -> str:
     return "spam"
 
 
-assert_type(func1, Callable[[int], Coroutine[Any, Any, str]])
+# Don't use assert_type here because some type checkers infer
+# the narrower type types.CoroutineType rather than typing.Coroutine
+# in this case.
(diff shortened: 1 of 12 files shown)


conformance/tests/annotations_coroutines.py
```diff
@@ -16,7 +16,10 @@ async def func1(ignored: int, /) -> str:
     return "spam"
 
 
-assert_type(func1, Callable[[int], Coroutine[Any, Any, str]])
+# Don't use assert_type here because some type checkers infer
+# the narrower type types.CoroutineType rather than typing.Coroutine
```
JelleZijlstra, line 20: This can wait but I'd favor making the conformance tests test for CoroutineType here in the future. It's the precise type at runtime, and ideally it's what type checkers should infer.

JelleZijlstra (approved):

PR #1937, opened 2025-03-01
Added "Importing Final Variables" section

erictraut:
For details, refer to [this discussion](https://discuss.python.org/t/imported-final-variable/82429).

### DIFF
--- a/docs/spec/qualifiers.rst
+++ b/docs/spec/qualifiers.rst
@@ -207,6 +207,34 @@ following should be allowed::
 item or a :ref:`NamedTuple <namedtuple>` field. Such usage also generates
 an error at runtime.
 
+
+Importing ``Final`` Variables
+^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
+
+If a module declares a ``Final`` variable and another module imports that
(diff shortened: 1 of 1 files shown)


carljm (commented):
This looks good to me. Should we also add the examples here as conformance tests?

erictraut:
I was planning to add the conformance tests as a separate PR if/when this is approved. I'll wait the required week for additional community input before I formally submit it for consideration by the full TC.

samwgoldman:
Are the examples given in this PR intended to be exhaustive? If not, it’s not clear to me what kinds of code would erase the final qualifier.  What about `PI = lib.PI`?

erictraut:
> What about `PI = lib.PI`?

That doesn't involve an "import of a Final variable by name or wildcard", so it doesn't match the situation described in the proposed text. This is simply a usage of a final — in the RHS of an assignment statement. 

If you think that's not clear from the current wording, let me know if you can think of a better way to phrase it.

samwgoldman:
I think it would help to specify an `import` statement, to clarify that statement refers to syntax, not a conceptual import.

erictraut:
Change has been approved by full TC, so merging.

PR #1940, opened 2025-03-05
Update URLs to typing.python.org

srittau:
@EricTraut The conformance tests also reference typing.readthedocs.io (a lot). I've left those alone for now, since I'm not whether just replacing the links there would work.


carljm (approved):
I think the links in the conformance suite test files could be updated, and then re-running the conformance suite would update the toml files.

JelleZijlstra:
Thank you!

PR #1942, opened 2025-03-10
Implemented conformance tests for imported Final variables

erictraut:


conformance/results/pytype/qualifiers_final_annotation.toml
```diff
@@ -123,5 +134,9 @@ Line 147: Expected 1 errors
 Line 149: Expected 1 errors
 Line 152: Expected 1 errors
 Line 155: Expected 1 errors
+Line 166: Expected 1 errors
+Line 170: Expected 1 errors
 Line 93: Unexpected errors ['qualifiers_final_annotation.py:93:1: \\x1b[1m\\x1b[31merror\\x1b[39m\\x1b[0m: in <module>: Class ClassCChild overrides final class attribute BORDER_WIDTH, defined in base class ClassC [final-error]']
+Line 164: Unexpected errors ["qualifiers_final_annotation.py:164:1: \\x1b[1m\\x1b[31merror\\x1b[39m\\x1b[0m: in <module>: Can't find module '_qualifiers_final_annotation_1'. [import-error]"]
+Line 168: Unexpected errors ["qualifiers_final_annotation.py:168:1: \\x1b[1m\\x1b[31merror\\x1b[39m\\x1b[0m: in <module>: Can't find module '_qualifiers_final_annotation_2'. [import-error]"]
```
carljm, line 141: It seems pytype can't find the helper modules at all? This suggests the root problem here is something in the conformance tests' path setup for pytype.

carljm (approved):
Thank you!

conformance/results/pytype/qualifiers_final_annotation.toml
erictraut: We may need help from pytype maintainers here.

So far, I haven't seen any interest expressed by the pytype maintainers to be conformant with the typing spec, so maybe we should drop pytype from the test entirely? Maintaining support for pytype has been a bit of a burden (much more so than the other three type checkers), and I haven't seen any sign that anyone (including the pytype team) care about the results of these tests.

carljm: I think there will be pytype maintainers at the PyCon typing summit, so I can discuss this general topic with them there. I don't think figuring out this issue needs to block this PR.

JelleZijlstra (approved):

PR #1957, opened 2025-04-01
Added a few more overload evaluation tests.

erictraut:
These uncovered a few more cases where mypy and pyright do not conform to the standard.

### DIFF
--- a/conformance/results/mypy/overloads_evaluation.toml
+++ b/conformance/results/mypy/overloads_evaluation.toml
@@ -4,7 +4,7 @@ Does not expand boolean arguments to Literal[True] and Literal[False].
 Does not expand enum arguments to literal variants.
 Does not expand tuple arguments to possible combinations.
 Does not evaluate Any in some cases where overload is ambiguous.
-Evaluates Any is some cases where overload is not ambiguous.
+Evaluates Any in some cases where overload is not ambiguous.
 """
 conformance_automated = "Fail"
 errors_diff = """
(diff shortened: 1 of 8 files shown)


conformance/tests/overloads_evaluation.py
```diff
@@ -176,49 +211,126 @@ def check_expand_tuple(v: int | str) -> None:
...
+
+
+def example6(a: float, b: T) -> T: ...
+
+
+def check_example6(a: list, b: Any) -> None:
+    m: list[int] = []
+
+    v1 = example6(1, a)
+    assert_type(v1, float)
```
carljm, line 298: This test case becomes a bit subtle, so maybe some comments here would be useful to the reader? IIUC, in this case step 5 eliminates the second overload, since all possible materializations of `list[Any]` are assignable to `Any`.

conformance/tests/overloads_evaluation.py
```diff
@@ -176,49 +211,126 @@ def check_expand_tuple(v: int | str) -> None:
...
+
+    v2 = example6(1, b)
+    assert_type(v2, float)
```
carljm, line 301: And here similarly all possible materializations of `Any` are assignable to `Any`, so all subsequent overloads are eliminated in step 5.

conformance/tests/overloads_evaluation.py
```diff
@@ -176,49 +211,126 @@ def check_expand_tuple(v: int | str) -> None:
...
+
+    v3 = example6(1, m)
+    assert_type(v3, float)
```
carljm, line 304: I guess here step 5 still applies even if all arguments are fully static types? So we eliminate the second overload because all possible materialization of the arguments will match the first overload?

conformance/tests/overloads_evaluation.py
```diff
@@ -176,49 +211,126 @@ def check_expand_tuple(v: int | str) -> None:
...
+
+    v4 = example6(1.0, "")
+    assert_type(v4, str)
```
carljm, line 307: It seems like an unwarranted assumption here that this must be `str` and not `Literal[""]` -- is that literal-widening behavior of typevar constraint solving already mandated elsewhere in the spec?

If we want to avoid that issue here (which seems wise to me if it is not already clearly mandated), we could use an argument to `check_example6` typed as `str`, instead of using `""`.

conformance/results/mypy/overloads_evaluation.toml
```diff
@@ -3,41 +3,51 @@ notes = """
 Does not expand boolean arguments to Literal[True] and Literal[False].
 Does not expand enum arguments to literal variants.
 Does not expand tuple arguments to possible combinations.
+Does not evaluate Any in some cases where overload is ambiguous.
+Evaluates Any is some cases where overload is not ambiguous.
```
carljm, line 7:
```suggestion
Evaluates Any in some cases where overload is not ambiguous.
```

carljm (commented):
Thank you!

PR #1958, opened 2025-04-01
Allow highlighting decorators instead of function signatures

### LINKED ISSUE #1955: Conformance tests: classes_override.py should allow highlighting the decorator
While working on support for this test in PyCharm, I noticed that it requires highlighting the function signature instead of the decorator (which is how PyCharm currently handles it). 
For example:
```python
    @staticmethod
    @override
    def static_method1() -> int:  # E: no matching signature in ancestor
        return 1

    @classmethod
    @override
    def class_method1(cls) -> int:  # E: no matching signature in ancestor
        return 1
```
This would be easy to change on our end, but I believe allowing the option to highlight the decorator would be preferable for several reasons:

1. Highlighting the function signature may overlap with other checks, such as PyCharm inspections or other linters.
2. If the decorator is removed, the error no longer applies, so highlighting the decorator itself makes sense.
3. Other conformance tests already allow highlighting decorators. For example:
  * qualifiers_final_decorator.py
```python
@final  # E[func]: not allowed on non-method function.
def func1() -> int:  # E[func]
    return 0
```
  * overloads_definitions.py

```python
@overload  # E[func2]
def func2(x: int) -> int:  # E[func2]: no implementation
    ...
```

If this change makes sense, I’d be happy to submit a PR to update the test.

alvov26:
Fixes #1955

### DIFF
--- a/conformance/tests/classes_override.py
+++ b/conformance/tests/classes_override.py
@@ -49,8 +49,8 @@ class ChildA(ParentA):
     def method2(self, x: int | str) -> int | str:  # OK
         return 0
 
-    @override
-    def method3(self) -> int:  # E: no matching signature in ancestor
+    @override  # E[method3]
+    def method3(self) -> int:  # E[method3]: no matching signature in ancestor
         return 1
 
     @overload  # E[method4]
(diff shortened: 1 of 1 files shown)


JelleZijlstra (approved):

alvov26:
Oh, no, that is not what I wanted to do, sorry.

JelleZijlstra:
You're doing well! I didn't merge immediately in case someone else had comments. I'll give it another day or so then merge.

PR #1959, opened 2025-04-05
Assign all TypeVars to variables

yangdanny97:
There were a few places in the conformance tests that had calls to `TypeVar` without assigning them to variables. This PR changes them to be assigned to variables, for consistency.

Based on the results, it appears that some existing typecheckers do not check TypeVars that are not assigned to variables, so the typechecker logic did not get exercised previously.


erictraut:
Thanks for the fix. Looks good to me.

PR #1960, opened 2025-04-05
Delete "Passing kwargs inside a function to another function section"

yangdanny97:
The context is in example 2 from https://github.com/python/typing/pull/1918

It doesn't make sense to enforce this requirement for just `**kwargs` but not any other arguments annotated with a TypedDict, and additionally, no typecheckers currently implement this.

It seems like https://discuss.python.org/t/typing-spec-inconsistency-for-unpacking-typed-dict-kwargs/79640 didn't get any responses. I'll bump the discussion again but for now I'll make a PR here so I don't forget about it.


JelleZijlstra:
Sorry we let this sit. We've since made some similar conformance suite changes in #2272 and #2301, and we've accepted PEP 728. In that light, it might be time to revise the spec to work consistent with PEP 728's new mechanisms: closed TypedDicts can't have extra keys; open ones can.

yangdanny97:
I'll close this one and make a new PR when I have time

PR #1965, opened 2025-04-10
Update docs on equivalence between gradual types 

MatthewMckee4:
Update docs to include that the relation on "equivalent" gradual types is an equivalence relation.

Or i guess more explicitly state that it is an equivalence relation.


carljm:
Closing and reopening the PR to see if that nudges CLA bot.

carljm:
Looks like that worked!

carljm (approved):
This looks good to me. I don't think it's a substantive change, so I don't think it requires a full typing council approval process, but I'll wait a day or two before merging in case anyone else wants to weigh in.

JelleZijlstra (approved):

PR #1971, opened 2025-04-13
Clarify that a TypeVarTuple default always has to be unpacked

rchen152:
The spec says that:
1. The default of a TypeVarTuple can be another TypeVarTuple
2. TypeVarTuples are always unpacked.

Together, these suggest that the default of a TypeVarTuple has to be an *unpacked* TypeVarTuple, and indeed [pyright](https://pyright-play.net/?pyrightVersion=1.1.386&code=GYJw9gtgBALgngBwJYDsDmUkQWEMoAqiApgGoCGIBArggDbEA0UAqiguQMYDWAUAQGcAjFAC8hEhSq0GACgDkgofICU-AQCYxEhGUo16xBYI3zmAE2LBy1OjFFK1ggMzaiuqQbmKBzs1EtrW3s2Dh4AbSUAXRUgA) and [mypy](https://mypy-play.net/?mypy=latest&python=3.12&gist=b6b01c533c7930e1e520765b773d1fe1) both agree. This PR makes that clear and adds a corresponding conformance test.

I don't think this needs a discussion + Typing Council approval, since it's just a clarification of an existing rule, but happy to go through the full process if anyone disagrees.


erictraut:
I agree this is just a clarification. The new wording is clearer.

rchen152:
Oops, I forgot about this PR for a bit XD I'll regenerate the conformance results and merge.

Issue #1973, opened 2025-04-17
Error in `TypedDict` assignability rule


Viicos:
The spec [says](https://typing.python.org/en/latest/spec/typeddict.html#assignability):

> First, any TypedDict type is [assignable](https://typing.python.org/en/latest/spec/glossary.html#term-assignable) to `Mapping[str, object]`.
> 
> Second, a TypedDict type `B` is [assignable](https://typing.python.org/en/latest/spec/glossary.html#term-assignable) to a TypedDict `A` if and only if both of these conditions are satisfied:
> 
> - For each key in `A`, `B` has the corresponding key and the corresponding value type in `B` is [consistent](https://typing.python.org/en/latest/spec/glossary.html#term-consistent) with the value type in `A`.
> 
> - For each required key in `B`, the corresponding key is required in `A`. For each non-required key in `B`, the corresponding key is not required in `A`.

However, the following seems to be accepted by type checkers:

```python
from typing import TypedDict


class A(TypedDict):
    a: int

class B(TypedDict):
    a: int
    b: str

def func(a: A): pass

b: B = ...
func(b)  # OK
```

However, following the second point, any required key in `B` should be required in `A`. Or, b is required in `B` but not required (because not defined) in `A`.

@JelleZijlstra suggested that the second bullet point should be updated to:

> For each required key in `A`, the corresponding key is required in `B`. For each non-required key in `A`, the corresponding key is not required in `B`.

Should we also say something about extra keys are allowed to be present in `B`?

erictraut:
Thanks for pointing this out.

This was due to a clerical error. At some point, someone did a pass through the typing spec and changed all of the assignability rules to assign `B` to `A` rather than the other way around. The symbol names shouldn't matter in assignability rules, but it's useful to be consistent throughout the document, so the update was welcomed. It appears that the first bullet was updated in this pass but the second was not. You can see this was the case when you compare the current typing spec with the language in the [original PEP](https://peps.python.org/pep-0589/#type-consistency). So I agree with Jelle that the second bullet point should be updated such that `A` and `B` are swapped.

I don't think there's a need to say anything about extra keys. That's implied by the current rules, and adding additional (unnecessary) rules could make the spec less clear.

Viicos:
> I don't think there's a need to say anything about extra keys. That's implied by the current rules, and adding additional (unnecessary) rules could make the spec less clear.

PEP 728 will probably update it anyway. I pushed a fix in https://github.com/python/typing/pull/1974.

[...]

PR #1974, opened 2025-04-17
Fix assignability rule for typed dictionaries

Viicos:
Fixes https://github.com/python/typing/issues/1973.


JelleZijlstra (approved):
Looks good. I don't think this needs the full Typing Council approval process because it fixes what is essentially a typo.

PR #1977, opened 2025-04-19
Conformance - document Python 3.11 requirement for pytype tests

Andrej730:
Not sure if it's still required-required by pytype - I haven't investigated, but wanted to document as it is the current conformance requirement.

Source:
https://github.com/python/typing/blob/0b9fed83ee2ac10aeac187d4182ae949f9fc25f0/conformance/src/type_checker.py#L313-L318


Andrej730:
Closing after https://github.com/python/typing/pull/2010

PR #1978, opened 2025-04-19
typeddicts - update conformance tests according to the spec

Andrej730:
Added missing tests to typeddicts_operations.py based on the current state of https://typing.python.org/en/latest/spec/typeddict.html#supported-and-unsupported-operations

Stumbled upon type checkers having a different behaviour for `typed_dict[str]`, investigated the spec and the spec is clear about type checkers generally rejecting operations with arbitrary strings - so wanted to reflect the spec in conformance tests.

Added tests for:
1. Disallowed `popitem`.
2. Not requires keys - spec allows both behaviours.
3. Extra keys in `get` and with `in` - spec is not certain about this, so I guess it allows both behaviours.
4. Operations with variable keys - disallowed descructive and non-destructive operations except `.get` and `in`.

Added baseline output commit just for a reference - to confirm that there no other typecheckers version changes at play in the updated output. I can remove it before merge to avoid additional diffs.

### DIFF
--- a/conformance/tests/typeddicts_operations.py
+++ b/conformance/tests/typeddicts_operations.py
@@ -54,6 +54,8 @@ movie.get("name")
 # It's not clear from the spec what type this should be.
 movie.get("other")  # E?
 
+# It's not clear from the spec whether it's allowed.
+reveal_type("other" in movie)  # E?
 
 movie.clear()  # E: clear not allowed
 movie.popitem()  # E: popitem not allowed
@@ -73,6 +75,12 @@ movie_optional: MovieOptional = {}
 # even if the key 'x' is not required.
 movie_optional["name"]  # E?
 assert_type(movie_optional.get("name"), str | None)
+
+# It's not clear from the spec what type this should be.
+reveal_type(movie_optional.get("other"))  # E?
+
+# It's not clear from the spec whether it's allowed.
+reveal_type("other" in movie_optional)  # E?
 
 movie_optional.clear()  # E: clear not allowed
 movie_optional.popitem()  # E: popitem not allowed
--- a/conformance/tests/typeddicts_operations.py
+++ b/conformance/tests/typeddicts_operations.py
@@ -69,6 +69,9 @@ class MovieOptional(TypedDict, total=False):
 
 movie_optional: MovieOptional = {}
 
+# > Type checkers may allow reading an item using d['x']
+# even if the key 'x' is not required.
+movie_optional["name"]  # E?
 assert_type(movie_optional.get("name"), str | None)
 
 movie_optional.clear()  # E: clear not allowed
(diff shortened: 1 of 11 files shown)


conformance/tests/typeddicts_operations.py
```diff
@@ -32,19 +32,33 @@ class Movie(TypedDict):
 movie = {"name": "", "year": 1900, "other": 2}  # E: extra key
 
 
-def func1(variable_key: str):
+def func1(variable_key: str, existing_movie: Movie):
     # > A key that is not a literal should generally be rejected.
     movie: Movie = {variable_key: "", "year": 1900}  # E: variable key
 
+    # Destructive operations.
+    existing_movie[variable_key] = 1982  # E: variable key
```
erictraut, line 40: I'd like to get clarification on the wording in the spec before we add these tests. I've started a discussion [here](https://discuss.python.org/t/indexing-typeddict-with-a-non-literal-variable/88929). Feel free to add your viewpoint to the thread.

conformance/tests/typeddicts_operations.py
```diff
@@ -32,19 +32,33 @@ class Movie(TypedDict):
...
+
 
 # It's not clear from the spec what type this should be.
 movie.get("name")
 
 # It's not clear from the spec what type this should be.
 movie.get("other")  # E?
 
+# It's not clear from the spec whether it's allowed.
+reveal_type("other" in movie)  # E?
```
erictraut, line 58: In general, we should not add tests to the conformance suite in areas where the spec is not clear. If you think that clarity is needed (and is beneficial), then please start a discussion about amending the spec.

erictraut (changes requested):

conformance/tests/typeddicts_operations.py
Andrej730: I think I've overlooked the spec a bit - I've made assumption that `in` with unknown key is ambigious similar to `movie.get("other")  # E?` but now I've found that spec actually states that all operations with unknown key should be rejected:

> The use of a key that is not known to exist should be reported as an error, even if this wouldn’t necessarily generate a runtime type error.

So I assume it's the same situation as with arbitrary strings - spec prohibits it, but in reality 
- setitem, getitem, del - rejected by everyone (mypy, Pyre, pyright)
- `.pop("other")` - rejected by Pyre and mypy
- `.get("other")` - only rejected by Pyre
- `"other" in movie` is not rejected by any type checker

Can you please take a look if I'm not missing something? And I'll start a similar discussion.

Andrej730: I think there is a value in clarifying the spec. Currently it says that use of unknown keys should be reported as an error
but the spec is interpreted by each type checker in a bit different way. I've started a discussion to gather opinions from the community - https://discuss.python.org/t/typeddict-operations-with-unknown-literal-keys/89653

Andrej730:
As this PR got outdated and there's no resolution, I'll split it into two issues - #2053 and #2054

PR #1979, opened 2025-04-19
Clarify namedtuple member rules

yangdanny97:
This PR clarifies the typing spec around named tuple fields and adds a few conformance test cases:
- fields are not being allowed to begin with underscore (and its interaction with `rename`)
- which members are counted/not counted as fields

Analogous restrictions are outlined in the [spec for enums](https://typing.python.org/en/latest/spec/enums.html#defining-members), but they are not present in the spec for named tuples.

### DIFF
--- a/conformance/results/mypy/namedtuples_define_class.toml
+++ b/conformance/results/mypy/namedtuples_define_class.toml
@@ -11,11 +11,19 @@ namedtuples_define_class.py:46: error: Argument 2 to "Point" has incompatible ty
 namedtuples_define_class.py:47: error: Argument "units" to "Point" has incompatible type "int"; expected "str"  [arg-type]
 namedtuples_define_class.py:48: error: Too many arguments for "Point"  [call-arg]
 namedtuples_define_class.py:49: error: Unexpected keyword argument "other" for "Point"  [call-arg]
-namedtuples_define_class.py:59: error: Non-default NamedTuple fields cannot follow default fields  [misc]
-namedtuples_define_class.py:98: error: Argument 2 to "Property" has incompatible type "float"; expected "str"  [arg-type]
-namedtuples_define_class.py:105: error: NamedTuple should be a single base  [misc]
+namedtuples_define_class.py:59: error: Invalid statement in NamedTuple definition; expected "field_name: field_type [= default]"  [misc]
+namedtuples_define_class.py:65: error: Missing positional argument "units" in call to "Point2"  [call-arg]
+namedtuples_define_class.py:67: error: Too many values to unpack (2 expected, 3 provided)  [misc]
+namedtuples_define_class.py:76: error: NamedTuple field name cannot start with an underscore: _y  [misc]
+namedtuples_define_class.py:86: error: Non-default NamedTuple fields cannot follow default fields  [misc]
+namedtuples_define_class.py:125: error: Argument 2 to "Property" has incompatible type "float"; expected "str"  [arg-type]
+namedtuples_define_class.py:132: error: NamedTuple should be a single base  [misc]
 """
 conformance_automated = "Fail"
 errors_diff = """
-Line 79: Expected 1 errors
+Line 69: Expected 1 errors
+Line 106: Expected 1 errors
+Line 59: Unexpected errors ['namedtuples_define_class.py:59: error: Invalid statement in NamedTuple definition; expected "field_name: field_type [= default]"  [misc]']
+Line 65: Unexpected errors ['namedtuples_define_class.py:65: error: Missing positional argument "units" in call to "Point2"  [call-arg]']
+Line 67: Unexpected errors ['namedtuples_define_class.py:67: error: Too many values to unpack (2 expected, 3 provided)  [misc]']
 """
--- a/conformance/results/mypy/namedtuples_define_functional.toml
+++ b/conformance/results/mypy/namedtuples_define_functional.toml
@@ -11,8 +11,15 @@ namedtuples_define_functional.py:43: error: Argument "x" to "Point6" has incompa
 namedtuples_define_functional.py:52: error: "namedtuple()" has duplicate field name "a"  [misc]
 namedtuples_define_functional.py:53: error: "namedtuple()" field name "def" is a keyword  [misc]
 namedtuples_define_functional.py:54: error: "namedtuple()" field name "def" is a keyword  [misc]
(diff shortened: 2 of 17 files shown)


erictraut:
Yes, please start a thread in the discussion forum. I don't think these changes will be controversial, but we should let the community weigh in.

yangdanny97:
Posted https://discuss.python.org/t/adding-namedtuple-member-naming-rules-to-the-typing-spec/88882

WSH032:
Hi, I'm not sure if this is the right place for this discussion—if not, please let me know (and sorry!).

I originally raised [this issue in pyright](https://github.com/microsoft/pyright/issues/10461#issuecomment-2888714348), but I think here might be a more appropriate place to discuss it.

Suppose I do have members that start with `"_"`:

```py
NT5 = namedtuple("NT5", ["abc", "_1"], rename=True)  # OK
NT5(abc="", _1=1)  # OK
```

How can I provide type annotations for them in a `.pyi` file?

```py
class NT5(NamedTuple):
    abc: str
    _1: int  # <- HOW ?
```

- Can we allow an exemption for field names matching the pattern `_{index}` (e.g., `_1`)?
- Or, could the spec recommend that even if a type checker rejects a field name like `_1`, it should still provide type analysis for that field:

    ```py
    from collections import namedtuple

    from typing_extensions import TYPE_CHECKING, NamedTuple, assert_type

    if not TYPE_CHECKING:
        NT5 = namedtuple("NT5", ["abc", "_1"], rename=True)  # OK
    else:

        class NT5(NamedTuple):
            abc: str
            _1: int  # type: ignore
            #        #  ^ suppress false positive


    nt5 = NT5(abc="", _1=1)  # should be ok

    assert_type(nt5[0], str)
    assert_type(nt5[1], int)  # should be ok

    assert_type(nt5.abc, str)
    assert_type(nt5._1, int)  # should be ok
    ```

    Currently, mypy and pyright [behave differently on this point](https://github.com/microsoft/pyright/issues/10461#issuecomment-2888729606).

JelleZijlstra:
I'm with Eric on the linked issue that this doesn't make sense to support. If you use different types on both arms of `if TYPE_CHECKING`, you'll see things like this happen.

> Can we allow an exemption for field names matching the pattern `_{index}` (e.g., `_1`)?

No, unless the runtime has that exemption. We should match the runtime behavior.

> could the spec recommend that even if a type checker rejects a field name like `_1`, it should still provide type analysis for that field

Type checkers could do that but I don't think it should be the spec's business to decide how type checkers should proceed once they've detected a type error.

JelleZijlstra:
And to give a constructive suggestion for your use case: Maybe just don't use namedtuple? If you require names that start with an underscore, it's not the right tool for the job.

WSH032:
Fair enough! Thanks for your quick response! 👍

> No, unless the runtime has that exemption

It would be nice if `typing.NamedTuple` also supported a `rename=True` option at runtime (I think that's outside the scope of this issue discussion 😂).

> And to give a constructive suggestion for your use case: Maybe just don't use namedtuple? If you require names that start with an underscore, it's not the right tool for the job.

I can't control the code on the other end; I can only provide type annotations for it (like in a `.pyi` file).

The workaround I've come up with for now is:

```py
class NT5(tuple[str, int]):
    abc: str
    _1: int

    __match_args__ = ("abc", "_1")

    def __new__(cls, abc: str, _1: int) -> Self: ...
```

But this isn't as straightforward as `typing.NamedTuple`.

Anyway, thanks for your help! 🙏

JelleZijlstra (approved):

conformance/results/mypy/namedtuples_define_functional.toml
```diff
@@ -11,8 +11,15 @@ namedtuples_define_functional.py:43: error: Argument "x" to "Point6" has incompa
 namedtuples_define_functional.py:52: error: "namedtuple()" has duplicate field name "a"  [misc]
```
UnboundVariable, line 11: The toml results files should be updated so the "conformant" field value is changed from "Pass" to "Partial", and a "notes" field should be added in each case specifying what failed for that type checker. These notes appear in the [conformance results report](https://htmlpreview.github.io/?https://github.com/python/typing/blob/main/conformance/results/results.html) when you mouse over a box that says "Partial" so viewers can easily tell which tests are failing for that type checker.

erictraut:
The change looks good to me, but you'll need to update the branch, resolve merge conflicts, and re-run the tests with the latest versions of the type checkers.

Thanks for the contribution!

yangdanny97:
Ok, I've rebased and updated the results.

conformance/results/mypy/namedtuples_define_class.toml
```diff
@@ -11,11 +11,19 @@ namedtuples_define_class.py:46: error: Argument 2 to "Point" has incompatible ty
...
+namedtuples_define_class.py:59: error: Invalid statement in NamedTuple definition; expected "field_name: field_type [= default]"  [misc]
+namedtuples_define_class.py:65: error: Missing positional argument "units" in call to "Point2"  [call-arg]
+namedtuples_define_class.py:67: error: Too many values to unpack (2 expected, 3 provided)  [misc]
+namedtuples_define_class.py:76: error: NamedTuple field name cannot start with an underscore: _y  [misc]
+namedtuples_define_class.py:86: error: Non-default NamedTuple fields cannot follow default fields  [misc]
+namedtuples_define_class.py:125: error: Argument 2 to "Property" has incompatible type "float"; expected "str"  [arg-type]
+namedtuples_define_class.py:132: error: NamedTuple should be a single base  [misc]
 """
-conformance_automated = "Fail"
+conformance_automated = "Partial"
```
erictraut: The conformance_automated field should be "Fail", since the automated tests indicate a failure. There should be a separate field "conformant" at the top of the ".toml" file. This is the manual scoring field — which is the value that is officially reported in the conformance test results. It should have the the value "Partial" in this case, and there should be a "notes" field that explains why this was scored as "Partial".

Same goes for the other results below.

erictraut (changes requested):

conformance/results/mypy/namedtuples_define_class.toml
yangdanny97: Oops, fixed

erictraut:
The PR looks good to me, but we need to wait for @carljm and @JukkaL to review and sign off on the change [here](https://github.com/python/typing-council/issues/47).

carljm (approved):

PR #1981, opened 2025-04-22
Make optional type narrowing actually optional in tuples_type_compat.py

missingdays:
This pull request marks optional type narrowing in tuples_type_compat.py with `E?` to signify its optionality. Otherwise, it is treated as obligatory.

See the comment `NOTE: This type narrowing functionality is optional, not mandated.` at the beginning of each function.

### DIFF
--- a/conformance/tests/tuples_type_compat.py
+++ b/conformance/tests/tuples_type_compat.py
@@ -71,15 +71,15 @@ def func4(
 def func5(val: tuple[int] | tuple[str, str] | tuple[int, *tuple[str, ...], int]):
     if len(val) == 1:
         # Type can be narrowed to tuple[int].
-        assert_type(val, tuple[int])  # tuple[int]
+        assert_type(val, tuple[int])  # E?: tuple[int]
 
     if len(val) == 2:
         # Type can be narrowed to tuple[str, str] | tuple[int, int].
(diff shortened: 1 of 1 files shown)


erictraut:
I agree these should be marked with "E?" rather than "E" because they are optional in the spec. Thanks for the fix.

erictraut (approved):

conformance/tests/tuples_type_compat.py
```diff
@@ -71,15 +71,15 @@ def func4(
 def func5(val: tuple[int] | tuple[str, str] | tuple[int, *tuple[str, ...], int]):
     if len(val) == 1:
         # Type can be narrowed to tuple[int].
-        assert_type(val, tuple[int])  # tuple[int]
+        assert_type(val, tuple[int])  # E?: tuple[int]
```
JelleZijlstra, line 74: This makes the test not very useful (without manual review) since type checkers will now pass no matter what type they infer for `val`. One solution could be to use two types and asserts that exactly one of them errors, by using an error group and asserting for both the narrowed and non-narrowed type. However, we don't currently check in that case that there is exactly one error. I can send a PR doing that.

missingdays: Sounds reasonable to me.

PR #1982, opened 2025-04-22
Conformance: Make "# E[tag]" exactly once, add support for at least once

JelleZijlstra:
Followup from https://github.com/python/typing/pull/1981#discussion_r2054352100.

The documentation said `# E[tag]` meant the error must appear on exactly one of the tagged lines, but the implementation enforced at least one error instead. I think "exactly one" is usually the better choice, so I made it do that. However, in a few cases we do want "at least one". For example, with two mutually recursive type aliases, mypy errors on both and pyright only on the first. Both of those behaviors seem sensible, so I added a new syntax `# E[tag+]` that means an error must appear on at least one of the tagged lines.

Also added a section to the README explaining all the syntax.


conformance/tests/overloads_definitions.py
```diff
@@ -70,28 +70,28 @@ def not_abstract(self, x: str) -> str: ...
 class C:
     @overload  # E[func5]
     @staticmethod
-    def func5(x: int) -> int:  # E[func5]
+    def func5(x: int, /) -> int:  # E[func5]
```
JelleZijlstra, line 73: I tried to get rid of various unrelated errors detected by some type checkers so that this test case focuses on only the issue it's trying to test for.

erictraut (approved):
Good idea!

PR #1985, opened 2025-04-24
Add missing import in enums_members.py

PetrGolubev:


JelleZijlstra:
This isn't strictly needed, as type checkers can assume `reveal_type()` is always imported. However, we don't currently specify this (https://typing.python.org/en/latest/spec/directives.html#reveal-type).

PetrGolubev:
My guess was that it's more reliable to have an actual import here since the spec doesn't explicitly say that it's optional.
And a type checker may choose to report an error as it causes an exception in runtime. Other tests that use `reveal_type` also import it.

PR #1987, opened 2025-04-30
Replace typing_extensions -> typing in examples

### LINKED ISSUE #1986: Examples shouldn't use `typing_extensions` imports for stable `typing` features
# Description

There are a few code examples in the docs which import things from `typing_extensions` that are by now available in Python's `typing` standard library module:

<details>
<summary>List</summary>

https://github.com/python/typing/blob/3149e8ce32ff5bb3c1afd01467d2d7b1f5b64c95/docs/guides/unreachable.rst?plain=1#L57

https://github.com/python/typing/blob/3149e8ce32ff5bb3c1afd01467d2d7b1f5b64c95/docs/reference/generics.rst?plain=1#L655

... and 4 more occurrences in `reference/generics.rst`.

https://github.com/python/typing/blob/3149e8ce32ff5bb3c1afd01467d2d7b1f5b64c95/docs/reference/protocols.rst?plain=1#L75

... and 4 more occurrences in `reference/protocols.rst`.

https://github.com/python/typing/blob/3149e8ce32ff5bb3c1afd01467d2d7b1f5b64c95/docs/reference/quality.rst?plain=1#L43

</details>

It probably makes sense to use `typing` imports for all of these instead, possibly with the `typing_extensions` alternative mentioned.

# How to fix?

What is the policy for these? Should they just be replaced with bare `typing` imports or something more like
```python
from typing import Protocol # or from typing_extensions for Python < 3.8
```
?
I haven't seen the latter in any of the other imports, which just import from `typing` without comment (EDIT: at least not in the code itself, but usually it is mentioned as an alternative for older versions somewhere around the code), but who knows...

smheidrich:
Fixes #1986.

### DIFF
--- a/docs/guides/unreachable.rst
+++ b/docs/guides/unreachable.rst
@@ -54,7 +54,7 @@ As an example, consider this simple calculator:
 .. code:: python
 
    import enum
-   from typing_extensions import Never
+   from typing import Never
 
    def assert_never(arg: Never) -> Never:
        raise AssertionError("Expected code to be unreachable")
--- a/docs/reference/generics.rst
+++ b/docs/reference/generics.rst
@@ -651,8 +651,7 @@ for a more faithful type annotation:
 
 .. code-block:: python
 
-   from typing import Callable, TypeVar
-   from typing_extensions import ParamSpec
+   from typing import Callable, ParamSpec, TypeVar
 
    P = ParamSpec('P')
    T = TypeVar('T')
@@ -668,8 +667,7 @@ alter the signature of the input function:
 
 .. code-block:: python
 
-   from typing import Callable, TypeVar
-   from typing_extensions import ParamSpec
+   from typing import Callable, ParamSpec, TypeVar
 
    P = ParamSpec('P')
    T = TypeVar('T')
@@ -692,8 +690,7 @@ Or insert an argument:
 
 .. code-block:: python
 
-    from typing import Callable, TypeVar
-    from typing_extensions import Concatenate, ParamSpec
+    from typing import Callable, Concatenate, ParamSpec, TypeVar
 
     P = ParamSpec('P')
     T = TypeVar('T')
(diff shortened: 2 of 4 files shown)


docs/guides/unreachable.rst
```diff
@@ -54,7 +54,7 @@ As an example, consider this simple calculator:
 .. code:: python
 
    import enum
-   from typing_extensions import Never
+   from typing import Never
```
smheidrich, line 57: In general, are these "simple" replacements fine or should it be something more like 

```python
from typing import Never  # or from typing_extensions for Python < 3.11
```

?

Haven't seen any of the latter for other imports, but who knows...

JelleZijlstra: Let's discuss on the issue for now

docs/reference/generics.rst
```diff
@@ -663,13 +662,17 @@ for a more faithful type annotation:
            return func(*args, **kwds)
        return wrapper
 
+.. note::
+
+   To use this feature on Python versions earlier than 3.10, you will need to
+   import ``ParamSpec`` from ``typing_extensions``.
```
srittau:
```suggestion
   import ``ParamSpec`` and ``Concatenate`` from ``typing_extensions``.
```

docs/guides/unreachable.rst
srittau, line 57: Never needs an infobox as well. (Added in Python 3.11.)

srittau (changes requested):
Thanks, and sorry for the delay. I think your hybrid approach is the best of both worlds. A few remarks below.

docs/guides/unreachable.rst
smheidrich: I think I didn't add it for that one because it's already explained in the prose at the start of the page; but I've added an additional infobox below the code snippet now for consistency (and perhaps easier discoverability for people who only skim the page) :+1:

PR #1993, opened 2025-05-06
Explain that `GenericAlias` instances are not considered to be a class

AbduazizZiyodov:
I added new explanations to the part of the document that relates to that issue (after explanation of "type erasure").

Related issue: python/cpython#132467

### DIFF
--- a/docs/spec/generics.rst
+++ b/docs/spec/generics.rst
@@ -397,8 +397,8 @@ the runtime class of the objects created by instantiating them doesn't
 record the distinction. This behavior is called "type erasure"; it is
 common practice in languages with generics (e.g. Java, TypeScript).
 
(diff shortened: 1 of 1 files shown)


JelleZijlstra:
I don't think this belongs in the typing spec since it's about runtime behavior.

AbduazizZiyodov:
@JelleZijlstra @erictraut 

Got you, thanks. 

This part of the document(from typing docs) was _"quoted"_ in the issue, I though that I should updated that part. Now I need to find out where I should document that behaviour in python docs. 

If you have any suggestions, let me know, appreciated )

JelleZijlstra:
This could be added to https://docs.python.org/3.14/library/stdtypes.html#generic-alias-type

PR #2000, opened 2025-05-13
Add some exceptions when commenting Any

srittau:


docs/guides/writing_stubs.rst
```diff
@@ -772,18 +772,26 @@ all type checkers::
...
-When using ``Any``, document the reason for using it in a comment. Ideally,
-document what types could be used.
+:class:`object` instead.
+
+When using :ref:`Any`, document the reason for using it in a comment, unless the
+reason is obvious. Ideally, document what types could be used. Obvious
+reasons can include:
+
+* Using :ref:`Any` as a type argument for a generic to say "any object of this
+  type is allowed", e.g. ``type[Any]``.
```
JelleZijlstra, line 790: This should be `type[object]`.

JelleZijlstra: Yes, that's a better example. In general you need this for invariant generics; for covariant ones `object` is fine.

JelleZijlstra (approved):

PR #2008, opened 2025-05-20
Modified `generics_type_erasure.py` test to not require an error when…

erictraut:
… accessing a generic attribute on a class when a subtype could provide specialization for that generic attribute.

Pyright recently made a modification to eliminate a false positive error in this case. See https://github.com/microsoft/pyright/discussions/10303 for details. I don't think the conformance test should mandate an error in this case. The spec certainly doesn't say anything about it, and the error is potentially a false positive.


erictraut:
@JelleZijlstra, I think you wrote this test originally. Could you review the proposed change and let me know what you think?

JelleZijlstra:
And in fact the example comes from PEP 484: https://peps.python.org/pep-0484/#instantiating-generic-classes-and-type-erasure

hauntsaninja:
Yup, I added the test because of the relevant line in PEP 484 / the spec.

This PR makes sense given that none of pytype/pyre/mypy error here. I am sympathetic to the use case in the pyright discussion, since I can't think of a great alternative for them. If we want to try and close down the hole introduced, we could consider erroring on assignment to such variables.

JelleZijlstra:
I feel given that we're changing the output of something that has been explicitly specified to raise an error, this is a substantive change and should go through Council approval.

jorenham (approved):

carljm (approved):

davidhalter (approved):

PR #2013, opened 2025-05-24
spec: various minor changes

JelleZijlstra:
- Replace an incorrect use of "subtype" with "assignable"
- Remove discussion of str/bytes Literal types that seems geared towards
  Python 2. Instead, say straightforwardly that strings (str objects) and
  bytes objects are supported.
- Rewrite an example that relies on the int/float/complex special case.
  The example was correct but I think it's confusing to readers if we have
  examples unnecessarily rely on the special case. Instead, use example
  types with more straightforward behavior.

### DIFF
--- a/docs/spec/literal.rst
+++ b/docs/spec/literal.rst
@@ -98,7 +98,7 @@ expressions, and nothing else.
 Legal parameters for ``Literal`` at type check time
 """""""""""""""""""""""""""""""""""""""""""""""""""
 
-``Literal`` may be parameterized with literal ints, byte and unicode strings,
+``Literal`` may be parameterized with literal ints, strings, `bytes` objects,
 bools, Enum values and ``None``. So for example, all of
 the following would be legal::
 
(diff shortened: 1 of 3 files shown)


docs/spec/literal.rst
```diff
@@ -98,16 +98,16 @@ expressions, and nothing else.
...
-``Literal`` may be parameterized with literal ints, byte and unicode strings,
+``Literal`` may be parameterized with literal ints, strings, `bytes` objects,
 bools, Enum values and ``None``. So for example, all of
 the following would be legal::
 
    Literal[26]
    Literal[0x1A]  # Exactly equivalent to Literal[26]
    Literal[-4]
    Literal["hello world"]
+   Literal[u"hello world"]  # Exactly equivalent to Literal["hello world"]
```
erictraut, line 109: Recommend deleting "Exactly" here. Including it might be confusing. (I'm not sure how equivalency could be inexact. :) )

docs/spec/literal.rst
```diff
@@ -98,16 +98,16 @@ expressions, and nothing else.
 Legal parameters for ``Literal`` at type check time
 """""""""""""""""""""""""""""""""""""""""""""""""""
 
-``Literal`` may be parameterized with literal ints, byte and unicode strings,
+``Literal`` may be parameterized with literal ints, strings, `bytes` objects,
```
erictraut, line 101: For consistency, maybe use the class names for `int`, `str`, `bytes`, and `bool`. With the proposed change, only `bytes` uses the actual class name.

Also, I think this requires double backticks.

docs/spec/literal.rst
JelleZijlstra: I did this for consistency with a few lines up, but I'll remove it there too.

JukkaL (approved):
LGTM

PR #2015, opened 2025-05-28
(🎁) conformance: `NewType` isn't a class

KotlinIsland:
i can't update the results on my machine, sorry!

### DIFF
--- a/conformance/tests/aliases_newtype.py
+++ b/conformance/tests/aliases_newtype.py
@@ -14,6 +14,9 @@ u2: UserId = UserId(42)  # OK
 
 assert_type(UserId(5) + 1, int)
 
+# > NewType('Derived', Base) returns a dummy function
+_: type = UserId  # E: functions are not instances of `type`
+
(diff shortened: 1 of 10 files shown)


conformance/tests/aliases_newtype.py
```diff
@@ -14,6 +14,8 @@
 
 assert_type(UserId(5) + 1, int)
 
+UserId.mro()  # E: NewType is not an instance of `type`
```
erictraut, line 17: By my reading of the spec, this shouldn't result in a type error. Can you point to the text in the spec that would indicate that it should? All conformance tests should be motivated by the language in the typing spec. If you think that there's a hole in the typing spec, please work through the process to change the spec.

KotlinIsland: great, this has been updated and pushed

erictraut (approved):

JelleZijlstra (approved):

PR #2016, opened 2025-05-28
(🎁) conformance: check for variance on constrained type variables

KotlinIsland:
i can't update the results on my machine, sorry!


JelleZijlstra:
Note that variance is meaningful on constrained type variables if one constraint is assignable to the other. (Given a `class X[T: (int, bool)]`, `X[bool]` is assignable to `X[int]` only if `T` is covariant.)

PR #2017, opened 2025-05-28
conformance: dataclasses_inheritance.py - add missing # E

KalininDL:
It seems that # E is missing in the snippet bellow
```
@dataclass
class DC5:
    # This should generate an error because a default value of
    # type list, dict, or set generate a runtime error.
    x: list[int] = []
```

Could you please take a look?

### DIFF
--- a/conformance/results/mypy/dataclasses_inheritance.toml
+++ b/conformance/results/mypy/dataclasses_inheritance.toml
@@ -1,8 +1,12 @@
-conformant = "Pass"
+conformant = "Partial"
+notes = """
+Does not report when the default value of dataclass field is mutable (is of type list, dict, or set).
+"""
 output = """
 dataclasses_inheritance.py:60: error: Cannot override instance variable (previously declared on base class "DC6") with class variable  [misc]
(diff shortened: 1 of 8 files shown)


conformance/results/mypy/dataclasses_inheritance.toml
```diff
@@ -1,8 +1,9 @@
-conformant = "Pass"
+conformant = "Partial"
```
erictraut, line 1: Please add a "notes" section here with a one-line summary of the failure(s). These notes appear when hovering over the results in the [conformance report](https://htmlpreview.github.io/?https://github.com/python/typing/blob/main/conformance/results/results.html).

conformance/results/pyright/dataclasses_inheritance.toml
```diff
@@ -1,8 +1,9 @@
-conformant = "Pass"
+conformant = "Partial"
```
erictraut, line 1: Same here — add a "notes" section and rerun the tests to produce an updated summary report.

conformance/results/mypy/dataclasses_inheritance.toml
KalininDL: Aha, now I got how it works, thanks!

PR #2035, opened 2025-06-29
update results for new pyright version

JelleZijlstra:
Two more tests became conformant; a minor error message change in another.

Also update to mypy 1.16.1 but it doesn't have any relevant changes.


erictraut:
Thanks, I forgot to update this after publishing the latest pyright version.

PR #2045, opened 2025-07-16
Table of Content improvements

srittau:
* Remove extra top levels from main ToC.
* Don't include sub-chapters of the specification in the main ToC.
* Reduce the ToC level by one on the guides index.

### DIFF
--- a/docs/index.rst
+++ b/docs/index.rst
@@ -5,18 +5,33 @@ Static Typing with Python
 Guides
 ======
 
-.. toctree::
-   :maxdepth: 2
+..
+   Keep in sync with docs/guides/index.rst.
 
-   guides/index
+.. toctree::
+   :maxdepth: 1
+
+   guides/libraries
+   guides/writing_stubs
+   guides/modernizing
+   guides/unreachable
+   guides/type_narrowing
+   guides/typing_anti_pitch
 
 Reference
 =========
 
-.. toctree::
-   :maxdepth: 2
+..
+   Keep in sync with docs/reference/index.rst.
 
-   reference/index
+.. toctree::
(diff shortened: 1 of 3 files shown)


docs/index.rst
```diff
@@ -27,7 +42,7 @@ Specification
 =============
 
 .. toctree::
-   :maxdepth: 2
+   :maxdepth: 1
```
JelleZijlstra: I actually find it nicer to keep multiple levels of TOC here so you don't have to click through.

srittau:
Rationale:

* This was always bothering me as redundant and ugly looking. Disadvantage: We now have a duplication in the ToC sources.
* ~~The spec is a very specialized document, not suitable for the vast majority of Python users. I feel one extra click for those specialists is worth the clearer ToC for ordinary users. *Edit:* Especially since the spec has many chapters that will push more relevant info to the bottom of the page.~~
* On the guides index, directly linking to sub-chapters of specific guides didn't seem worth it to me. Users will usually want to read a complete guide.

docs/index.rst
srittau: See below for the rationale.

srittau: @JelleZijlstra Do you agree with the rationale below? Otherwise I could revert this, I'm not too attached to it.

JelleZijlstra: I'd still strongly prefer to keep another level of TOC. The spec is relatively specialized but it isn't (shouldn't be, at least!) completely inaccessible to most users, and I think it's nice to be able to go directly to a page you're interested in.

JelleZijlstra (approved):

PR #2048, opened 2025-07-18
Edit error marker comments on some tests

yangdanny97:
See inline comments for why


conformance/tests/enums_behaviors.py
```diff
@@ -23,8 +23,13 @@ class Color(Enum):
 # > constructor. Instead, the call performs a value-based lookup of an
 # > enum member.
 
-assert_type(Color["RED"], Color)  # 'Literal[Color.RED]' is also acceptable
```
yangdanny97, line 26: felt odd to assert_type here and say that another type is also acceptable. I tried to put both versions down, so that type checkers have discretion on which one to implement w/o the test failing.

I assume that if a checker emits an error on both lines the test will fail - could the maintainers please confirm this assumption?

conformance/tests/qualifiers_final_decorator.py
```diff
@@ -65,7 +65,7 @@ def method3() -> None:  # E[method3]
         pass
 
     @overload  # E[method4]
-    def method4(self, x: int) -> int:
+    def method4(self, x: int) -> int:  # E[method4]
```
yangdanny97, line 68: Pyrefly emits diagnostics on the line of the `def`, not the line with the `@overload`

This is similar to what I did in https://github.com/python/typing/pull/1980

conformance/tests/enums_behaviors.py
JelleZijlstra: That's right; this is documented in the "Test case syntax" section in conformance/README.

JelleZijlstra (approved):

rchen152 (approved):

PR #2052, opened 2025-07-21
Typing concepts: add that object | Any and Never | Any can be simplified

JelleZijlstra:
See https://discuss.python.org/t/interactions-with-never-leading-to-undesirable-assignability/99445/25

I feel this is a straightforward oversight so I don't think this needs to go through the full approval process.

### DIFF
--- a/docs/spec/concepts.rst
+++ b/docs/spec/concepts.rst
@@ -298,9 +298,9 @@ visualize this analogy in the following table:
    * - ``B`` is :term:`equivalent` to ``A``
      - ``B`` is :term:`consistent` with ``A``
 
-We can also define an **equivalence** relation on gradual types: the gradual 
-types ``A`` and ``B`` are equivalent (that is, the same gradual type, not 
-merely consistent with one another) if and only if all materializations of 
+We can also define an **equivalence** relation on gradual types: the gradual
+types ``A`` and ``B`` are equivalent (that is, the same gradual type, not
+merely consistent with one another) if and only if all materializations of
 ``A`` are also materializations of ``B``, and all materializations of ``B``
 are also materializations of ``A``.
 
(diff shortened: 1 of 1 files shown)


docs/spec/concepts.rst
```diff
@@ -368,10 +368,16 @@ can likewise be materialized to ``T1 | T2``. Thus, the gradual types ``S1`` and
 
 If ``B`` is a subtype of ``A``, ``B | A`` is equivalent to ``A``.
 
-This rule applies only to subtypes, not assignable-to. The union ``T | Any`` is
+This rule applies only to subtypes, not assignable-to. For any type ``T``
+other than the top and bottom types (``object`` and ``Never``), the union ``T | Any`` is
```
hauntsaninja: nit:
```suggestion
(other than the top and bottom types ``object`` and ``Never``), the union ``T | Any`` is
```

i put the whole thing in the parentheses bc it's more of a technicality, so want to minimise the amount it distracts from the main sentence

JelleZijlstra: I think your suggestion duplicates the first line:
```suggestion
(other than the top and bottom types ``object`` and ``Never``), the union ``T | Any`` is
```

docs/spec/concepts.rst
```diff
@@ -368,10 +368,16 @@ can likewise be materialized to ``T1 | T2``. Thus, the gradual types ``S1`` and
...
+This rule applies only to subtypes, not assignable-to. For any type ``T``
+(other than the top and bottom types ``object`` and ``Never``), the union ``T | Any`` is
 not reducible to a simpler form. It represents an unknown static type with
 lower bound ``T``. That is, it represents an unknown set of objects which may
 be as large as ``object``, or as small as ``T``, but no smaller.
+The exceptions are ``object`` and ``Never``. The union ``object | Any`` is equivalent to
+``object``, because ``object`` is a type containing all values and therefore the ``Any``
+cannot add any values. Similarly, ``Never | Any`` is equivalent to ``Any``, because
+``Never`` is a type containing no values, so that including it in a union cannot add any
+values to the type.
```
sharkdp, line 380: There is another exception explained in the paragraph below (for equivalent gradual types). To be honest, I'm not sure if it really helps with clarity to add more exceptions here. It feels to me like the original intent of this paragraph was maybe just to explain that `T | Any` is not **generally** reducible to a simpler form (`Any`)?

Also, the existing sentence "It represents an unknown static type with lower bound `T`" is only true if `T` is fully-static.

Maybe both of these problems could be solved by going with something like:

> This rule applies only to subtypes, not assignable-to. The union `T | Any` is not generally reducible to a simpler form. For example, `int | Any` represents an unknown static type with lower bound `int`. That is, it represents an unknown set of objects which may be as large as `object`, or as small as `int`, but no smaller.
>
> Equivalent gradual types can, however, be simplified from unions […]

jorenham, line 380: `Never | Any ≅ Any` can be generalized into `Never | T ≅ T` (for all types `T`), but I'm not sure if that axiom is already in the spec.

PR #2061, opened 2025-08-01
Small improvements to the conformance tests

rchen152:
A few tweaks to the conformance tests that I'm hoping are uncontroversial. Happy to split up this PR if any of these seem like they need more discussion.

* When a generator is annotated incorrectly, pyright reports errors on both the return type and the `yield`, while mypy reports an error on just the return type. IMO both of these are reasonable choices, so I suggest we mark the error on the `yield` as optional.
* Most of the errors in class_override.py are tagged as allowed to occur on either the `@override` line or the `def` line. One error was missing a tag on the `def` line, which I've added.
* The `dataclass_transform` decorator should not be applied to more than one of an overloaded function's overloads and implementation (https://typing.python.org/en/latest/spec/dataclasses.html#specification). dataclasses_transform_func.py applied `dataclass_transform` to two overloads of the same function, so I removed the decorator from the second overload.


JelleZijlstra (approved):

erictraut:
LGTM. Thanks @rchen152!

PR #2062, opened 2025-08-02
Update unreachable.rst: mention match exhaustiveness codes from checkers type …

wyattscarpenter:
This will be useful for anyone actually trying to practically check and existing code base for exhaustive matches who is consulting this page.

### DIFF
--- a/docs/guides/unreachable.rst
+++ b/docs/guides/unreachable.rst
@@ -86,6 +86,8 @@ and is also present in ``typing_extensions`` starting at version 4.1.
 However, it is also possible to define a similar function in your own
 code, for example if you want to customize the runtime error message.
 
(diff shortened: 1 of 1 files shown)


docs/guides/unreachable.rst
```diff
@@ -86,6 +86,8 @@ and is also present in ``typing_extensions`` starting at version 4.1.
 However, it is also possible to define a similar function in your own
 code, for example if you want to customize the runtime error message.
 
+This use of match statements for comprehensive matching is common enough that some typecheckers have direct support for checking all match statements for exhaustiveness, regardless of whether or not they have an assert_never-style function in their default arm. For instance, in pyright this can be enabled by enabling the diagnostic code reportMatchNotExhaustive (on by default in strict mode), and in mypy it can be enabled by enabling the error code exhaustive-match.
```
srittau: Please add a few line breaks for easier diffs in the future. Also, a few keywords should be marked with ``:


```suggestion
This use of match statements for comprehensive matching is common
enough that some typecheckers have direct support for checking all match
statements for exhaustiveness, regardless of whether or not they have an
``assert_never``-style function in their default arm. For instance, in pyright
this can be enabled by enabling the diagnostic code ``reportMatchNotExhaustive``
(on by default in strict mode), and in mypy it can be enabled by enabling
the error code ``exhaustive-match``.
```

wyattscarpenter: Thanks!

srittau (approved):

PR #2067, opened 2025-08-11
Add the Zuban Type Checker

davidhalter:
I have added the type checker that I have recently written. The results should be mostly the same, since it's mostly using the same tests.

I have only changed a few things:

- I changed a few tests, mostly uncontroversial line change. I'm happy to revert these changes, but I feel like they all make sense and should not need much discussion.
- I added a way to fill in `conformance = "Pass"` for unknown conformance if the automated tests pass. This was useful while adding the type checker and will probably be useful for other type checkers as well.
- The seconds counter is one digit more precise if it's faster than a second.

I would be happy if this was merged soon, because the merge conflicts can be kind of annoying for this one.

### DIFF
--- a/conformance/src/type_checker.py
+++ b/conformance/src/type_checker.py
@@ -291,11 +291,6 @@ class ZubanLSTypeChecker(MypyTypeChecker):
                 [sys.executable, "-m", "pip", "install", "zuban"],
                 check=True,
             )
-
-            # Run "mypy --version" to ensure that it's installed and to work
-            # around timing issues caused by malware scanners on some systems.
-            self.get_version()
-
             return True
         except CalledProcessError:
             print("Unable to install zuban")
--- a/conformance/tests/qualifiers_annotated.py
+++ b/conformance/tests/qualifiers_annotated.py
@@ -115,7 +115,4 @@ class ClassC(TypedDict):
 
 TA1: TypeAlias = Annotated[int | str, ""]
 TA2 = Annotated[Literal[1, 2], ""]
-
-T = TypeVar("T")  # E?: T is already defined
-
 TA3 = Annotated[T, ""]
--- a/conformance/results/mypy/qualifiers_annotated.toml
+++ b/conformance/results/mypy/qualifiers_annotated.toml
@@ -28,8 +28,6 @@ qualifiers_annotated.py:91: error: "<typing special form>" not callable  [operat
 qualifiers_annotated.py:92: error: "<typing special form>" not callable  [operator]
 qualifiers_annotated.py:98: error: Invalid type: ClassVar nested inside other type  [valid-type]
 qualifiers_annotated.py:100: error: Final can be only used as an outermost qualifier in a variable annotation  [valid-type]
-qualifiers_annotated.py:119: error: Cannot redefine "T" as a type variable  [misc]
-qualifiers_annotated.py:119: error: Invalid assignment target  [misc]
 """
 conformance_automated = "Fail"
 errors_diff = """
--- a/conformance/results/results.html
+++ b/conformance/results/results.html
@@ -256,7 +256,7 @@
 <th class="column col2 partially-conformant">Partial</th>
 <th class="column col2 conformant">Pass</th>
 <th class="column col2 not-conformant"><div class="hover-text">Unsupported<span class="tooltip-text" id="bottom"><p>Does not support generic defaults.</p></span></div></th>
-<th class="column col2 partially-conformant">Partial</th>
+<th class="column col2 partially-conformant"><div class="hover-text">Partial<span class="tooltip-text" id="bottom"><p>Does not reject TypeVars with defaults after a TypeVarTuple</p><p>Type parameter defaults are not bound by attribute access</p><p>ParamSpec after TypeVarTuple is not always handled correctly</p></span></div></th>
 </tr>
 <tr><th class="column col1">&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;generics_defaults_referential</th>
 <th class="column col2 partially-conformant">Partial</th>
--- a/conformance/results/zuban/generics_defaults.toml
+++ b/conformance/results/zuban/generics_defaults.toml
@@ -1,4 +1,9 @@
 conformant = "Partial"
+notes = """
+Does not reject TypeVars with defaults after a TypeVarTuple
+Type parameter defaults are not bound by attribute access
+ParamSpec after TypeVarTuple is not always handled correctly
+"""
 conformance_automated = "Fail"
 errors_diff = """
 Line 141: Expected 1 errors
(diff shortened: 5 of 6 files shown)


davidhalter:
I just signed https://www.python.org/psf/contrib/contrib-form/, I hope that will fix CI eventually.

conformance/tests/typeddicts_alt_syntax.py
```diff
@@ -42,4 +42,4 @@
 
 movie2: Movie2
 movie2 = {"name": "Blade Runner", "year": 1982}  # E?: May generate errors if keyword-argument syntax is unsupported
-movie2 = {"name": "Blade Runner", "year": ""}  # E: Incorrect type for year
```
davidhalter, line 45: It feels like this error is also optional, because as it says above, the keyword argument syntax is not mandatory.

conformance/tests/qualifiers_annotated.py
```diff
@@ -116,6 +116,6 @@ class ClassC(TypedDict):
 TA1: TypeAlias = Annotated[int | str, ""]
 TA2 = Annotated[Literal[1, 2], ""]
 
-T = TypeVar("T")
+T = TypeVar("T")  # E?: T is already defined
```
JelleZijlstra: I think we should just remove this line, it looks like an oversight that the TypeVar was redefined.

Type checkers probably should error on this line, but it's not the subject of this test case so better to leave it out.

conformance/tests/typeddicts_alt_syntax.py
JelleZijlstra: Yes, that makes sense to me. If type checkers aren't expected to support this syntax for creating a TypedDict, they also don't need to understand it.

JelleZijlstra:
Thanks for adding this! I looked over the test file changes and they mostly look good, as noted above. I didn't review all the results for Zuban.

One concern: I know you're using a partly paid model, where some users have to pay for usage of the type checker. How does that system interact with usage of Zuban in this repo? For example, could we run into issues if the conformance suite grows past some line count, or if we tried to run all type checkers in CI?

conformance/src/type_checker.py
```diff
@@ -273,9 +273,78 @@ def parse_errors(self, output: Sequence[str]) -> dict[int, list[str]]:
...
+
+            # Install the latest version.
+            run(
+                [sys.executable, "-m", "pip", "install", "zuban"],
+                check=True,
+            )
+
+            # Run "mypy --version" to ensure that it's installed and to work
+            # around timing issues caused by malware scanners on some systems.
+            self.get_version()
```
erictraut: Is this needed for zuban? This was a hack that was added for mypy specifically (see comment). If it is needed, please update this comment. If it's not needed, it can be deleted.

erictraut:
Thanks for adding this.

I didn't do a line-by-line review, but I did skim through it. I have a few questions:
1. I noticed that many of the "Notes" fields in the Zuban results are identical to the "Notes" field for mypy. Is this correct? Unless Zuban is based on mypy directly (and is therefore bug-for-bug compatible), it seems unlikely that the conformance failure modes would be identical between the two type checkers.
2. I noticed that some of the tests that were scored as "Partial" are missing a "Notes" field. For consistency, please provide a "Notes" field for all test cases that are not scored as "Pass". These notes appear in the conformance summary report when you mouse over a box that says "Partial".

conformance/tests/qualifiers_annotated.py
davidhalter: I removed it.

davidhalter:
> One concern: I know you're using a partly paid model, where some users have to pay for usage of the type checker. How does that system interact with usage of Zuban in this repo? For example, could we run into issues if the conformance suite grows past some line count, or if we tried to run all type checkers in CI?

No, I'm most likely going to open source it anyway in the near future and at the moment there the line count limits are disabled.

>  I noticed that many of the "Notes" fields in the Zuban results are identical to the "Notes" field for mypy. Is this correct? Unless Zuban is based on mypy directly (and is therefore bug-for-bug compatible), it seems unlikely that the conformance failure modes would be identical between the two type checkers.

It is based on Mypy's test suite, so if you look at the automatic test results and open them side by side you quite often see the exact same results. If the results are not the same the issues are sometimes still the same even if there is an additional error, it still stems from the same fact (e.g. some sort of thing is not supported, because it's not covered by Mypy's test suite).

I also specifically fixed the issues that would differ from Mypy's results to be more similar to Mypy, so I wouldn't have to write my own Notes. I'm lazy in a weird way :)

> I noticed that some of the tests that were scored as "Partial" are missing a "Notes" field. For consistency, please provide a "Notes" field for all test cases that are not scored as "Pass". These notes appear in the conformance summary report when you mouse over a box that says "Partial".

I will add notes.

srittau:
I'm concerned about adding support for a type checker with a proprietary license. I'm volunteering my (and my company's) time for work on typing, and the contract has always been that the software we're directly supporting is licensed freely. For example, while we're directly supporting pyright, which is MIT licensed, we're not directly supporting the proprietary pylance language server. (Of course, we're indirectly supporting proprietary software and improving typing as a whole and fixing bugs reported by developers and users of proprietary software, but that benefits all typing users.)

Frankly speaking, if I'm supposed to support proprietary software, me or my company expects to be paid for it. So, I'm -1 on merging this (and any similar) PRs.

conformance/src/type_checker.py
davidhalter: Good catch, I removed it. I guess I didn't proof-read my code well enough.

davidhalter:
@erictraut @JelleZijlstra  I think I fixed the remaining issues.

@srittau Did you read the part about open sourcing?

srittau:
If Zuban would go Open Source, my concerns would obviously be irrelevant.

davidhalter:
Does anybody know why the CLA part of CI is still failing? I signed it yesterday.

JelleZijlstra:
I hoped that closing and reopening would fix the CLA issue but no dice. @ambv could you help take a look?

> No, I'm most likely going to open source it anyway in the near future and at the moment there the line count limits are disabled.

That's good to know!

@srittau I broadly agree with this concern (and I probably wouldn't support adding Zuban to typeshed CI while it's still proprietary), but I think it's not too bad for the conformance suite, where we don't have to do that much more than run the type checker and record the results.

brianschubert:
> I just signed https://www.python.org/psf/contrib/contrib-form/, I hope that will fix CI eventually.

It looks like the CLA form you signed requires manual processing, which probably hasn't gone through yet. The other way to sign the CLA is by clicking the button in the bot comment / visiting https://cla.python.org/, which is a fully automated process

srittau:
> @srittau I broadly agree with this concern (and I probably wouldn't support adding Zuban to typeshed CI while it's still proprietary), but I think it's not too bad for the conformance suite, where we don't have to do that much more than run the type checker and record the results.

Fair enough, especially since I personally stay away from the conformance tests.

davidhalter:
@brianschubert Thanks that went through.
@srittau @JelleZijlstra I wouldn't even have tried to add Zuban to typeshed, because I know how much work that would be for you guys and appreciate what you are working on. I think here it's a bit different, but I doubt that matters in a few weeks.

PR #2072, opened 2025-08-17
spec: Rewrite TypedDict spec

JelleZijlstra:
This is an edit of the TypedDict spec for clarity and flow. My goal was to
unify the pieces of the spec that derive from the various PEPs into a coherent
whole. I removed excessive examples and motivations: the spec should specify,
not justify. The length of the spec chapter is reduced by more than half.

The general approach I took is to first define the kinds of TypedDicts that
can exist, then explain the syntax for defining TypedDicts, then discuss
other aspects of TypedDict types.

I introduce some new terminology around PEP 728 to make it easier to talk
about the different kinds of TypedDict. TypedDicts are defined to have a
property called openness, which can have three states:
- Open: all TypedDicts prior to PEP 728
- Closed: no extra keys are allowed (closed=True)
- With extra items: extra_items=... from PEP 728

I retained existing text where it made sense but also wrote some from
scratch.

### DIFF
--- a/docs/spec/typeddict.rst
+++ b/docs/spec/typeddict.rst
@@ -112,7 +112,8 @@ in the class definition:
 * ``closed``: a boolean literal (``True`` or ``False``) indicating whether
   the TypedDict is :term:`closed` (``True``) or :term:`open` (``False``).
   The latter is the default, except when inheriting from another TypedDict that
-  is not open (see :ref:`typeddict-inheritance`).
+  is not open (see :ref:`typeddict-inheritance`), or when the ``extra_items``
+  argument is also used.
   As with ``total``, the value must be exactly ``True`` or ``False``. It is an error
   to use this argument together with ``extra_items=``.
 * ``extra_items``: indicates that the TypedDict has :term:`extra items`. The argument
@@ -492,7 +493,7 @@ Multiple inheritance does not allow conflicting types for the same item::
 Subtyping and assignability
 ---------------------------
 
-Because TypedDict types are :term:`structural` types, a TypedDict ``T1`` is assignable to another
+Because TypedDict types are :term:`structural` types, a TypedDict ``T1`` is :term:`assignable` to another
 TypedDict type ``T2`` if the two are structurally compatible, meaning that all operations that
 are allowed on ``T2`` are also allowed on ``T1``. For similar reasons, TypedDict types are
 generally not assignable to any specialization of ``dict`` or ``Mapping``, other than ``Mapping[str, object]``,
@@ -501,7 +502,10 @@ to these types.
 
 The rest of this section discusses the :term:`subtyping <subtype>` rules for TypedDict in more detail.
 As with any type, the rules for :term:`assignability <assignable>` can be derived from the subtyping
-rules using the :term:`materialization <materialize>` procedure.
+rules using the :term:`materialization <materialize>` procedure. Generally, this means that where
+":term:`equivalent`" is mentioned below, the :term:`consistency <consistent>` relation can be used instead
+when implementing assignability, and where ":term:`subtyping <subtype>`" between elements of a
+TypedDict is mentioned, assignability can be used instead when implementing assignability between TypedDicts.
 
 Subtyping between TypedDict types
 ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
@@ -518,6 +522,8 @@ The conditions are as follows:
 
     - It must also be required in ``B``.
     - If it is read-only in ``A``, the item type in ``B`` must be a subtype of the item type in ``A``.
+      (For :term:`assignability <assignable>` between two TypedDicts, the first item must instead
+      be assignable to the second.)
 
     - If it is mutable in ``A``, it must also be mutable in ``B``, and the item type in ``B`` must be
       :term:`equivalent` to the item type in ``A``. (It follows that for assignability, the two item types
@@ -536,14 +542,14 @@ The conditions are as follows:
     - If it is mutable in ``A``:
 
       - If ``B`` has an item with the same key, it must also be mutable, and its item type must be
-        :term:`equivalent` to the item type in ``A``. (As before, it follows that for assignability, the two item types
-        must be :term:`consistent`.)
+        :term:`equivalent` to the item type in ``A``.
 
       - Else:
 
         - If ``B`` is closed, the check fails.
         - If ``B`` has extra items, the extra items type must not be read-only and must
           be :term:`equivalent` to the item type in ``A``.
+
 - If ``A`` is closed, ``B`` must also be closed, and it must not contain any items that are not present in ``A``.
 - If ``A`` has read-only extra items, ``B`` must either be closed or also have extra items, and the extra items type in ``B``
   must be a subtype of the extra items type in ``A``. Additionally, for any items in ``B`` that are not present in ``A``,


JelleZijlstra:
Posted about this on Discuss: https://discuss.python.org/t/revision-of-the-typeddict-spec/102675

docs/spec/glossary.rst
```diff
@@ -84,6 +99,12 @@ This section defines a few terms that may be used elsewhere in the specification
       runtime code using :pep:`526` and
       :pep:`3107` syntax (the filename ends in ``.py``).
 
+   item
+      In the context of a :ref:`TypedDict <typeddict>`, an item is a key/value
+      pair defined in the TypedDict definition. Each item has a name (the key)
+      and a type (the value). Items may be :term:`required` or
```
jorenham: maybe

```suggestion
      and a type (of the value). Items may be :term:`required` or
```

docs/spec/glossary.rst
```diff
@@ -52,6 +59,14 @@ This section defines a few terms that may be used elsewhere in the specification
       also materializations of ``B``, and all materializations of ``B`` are
       also materializations of ``A``.
 
+   extra items
+      A :ref:`TypedDict <typeddict>` type with extra items may contain arbitrary
+      additional :term:`items <item>` beyond those specified in the TypedDict definition, but those
+      items must be of the type specified by the TypedDict definition.
```
jorenham: perhaps

```suggestion
      items must be of the type specified by that definition.
```

docs/spec/glossary.rst
```diff
@@ -110,12 +131,41 @@ This section defines a few terms that may be used elsewhere in the specification
...
+
    package
       A directory or directories that namespace Python modules.
       (Note the distinction between packages and :term:`distributions <distribution>`.
       While most distributions are named after the one package they install, some
       distributions install multiple packages.)
 
+   read-only
+      A read-only :term:`item` in a :ref:`TypedDict <typeddict>` may not be modified.
+      Attempts to assign to or delete that item
```
jorenham: it could be a bit easier to read this way, as it requires less "mental RAM"

```suggestion
      Attempts to delete or assign to that item
```

jorenham (commented):
I picked some nits; hope you don't mind.

docs/spec/callables.rst
```diff
@@ -180,24 +180,61 @@ generated. For example::
...
-        title: str
-        year: NotRequired[int]
+Required and non-required items
+^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
 
+Items in a TypedDict may be either :term:`required` or :term:`non-required`.
 When using a ``TypedDict`` to type ``**kwargs`` all of the required and
 non-required keys should correspond to required and non-required function
-keyword parameters. Therefore, if a required key is not supported by the
+keyword parameters. Therefore, if a required key is not provided by the
```
PIG208, line 189: Perhaps we can use the updated terminology, referring to "items" instead of "keys" consistently.

docs/spec/typeddict.rst
```diff
@@ -128,95 +244,232 @@ key, and the ``'name'`` key is missing::
...
+
+However, subclasses may also explicitly use the ``closed`` and ``extra_items`` arguments
+to change the openness of the TypedDict, but in some cases this yields a type checker error.
+If the base class is open, all possible states are allowed in the subclass: it may remain open,
+it may be closed (with ``closed=True``), or it may have extra items (with ``extra_items=...``).
+If the base class is closed, any child classes must also be closed.
+If the base class has extra items, but they are not read-only, the child class must also allow
+the same extra items. If the base class has read-only extra items, the child class may be closed,
+or it may redeclare its extra items with a type that is :term:`assignable` to the base class type.
+Child classes may also have mutable extra items if the base class has read-only extra items.
```
PIG208: Organizing this section into bullets might help with readability, since it exhausts all possible openness a base TypedDict class can have.

docs/spec/glossary.rst
```diff
@@ -27,6 +27,13 @@ This section defines a few terms that may be used elsewhere in the specification
       ``B``, respectively, such that ``B'`` is a subtype of ``A'``. See
       :ref:`type-system-concepts`.
 
+   closed
+      A :ref:`TypedDict <typeddict>` type is closed if it may not contain any
+      additional :term:`items <item>` beyond those specified in the TypedDict definition.
+      A closed TypedDict can be created using the ``closed=True`` argument to
+      :py:func:`typing.TypedDict`.
+      Compare :term:`extra items` and :term:`open`.
```
PIG208, line 35: It might help to mention that TypedDict with `extra_items=Never` is considered `closed` too.

docs/spec/typeddict.rst
```diff
@@ -227,195 +480,181 @@ Additional notes on TypedDict class inheritance:
...
+A TypedDict type is a subtype of ``dict[str, VT]`` if the following conditions are met:
 
-      c: C = {'x': 0, 'y': 'foo'}
-      g(c)
-      c['y'] + 'bar'  # Runtime error: int + str
+- The TypedDict type has mutable :term:`extra items` of a type that is :term:`equivalent` to ``VT``.
+- All items on the TypedDict satisfy the following conditions:
+  - The value type of the item is :term:`equivalent` to ``VT``.
+  - The item is not read-only.
+  - The item is not required.
```
PIG208, line 626: Looks like this sublist doesn't render correctly:

> - All items on the TypedDict satisfy the following conditions: - The value type of the item is [equivalent](https://typing--2072.org.readthedocs.build/en/2072/spec/glossary.html#term-equivalent) to VT. - The item is not read-only. - The item is not required.

docs/spec/typeddict.rst
```diff
@@ -227,195 +480,181 @@ Additional notes on TypedDict class inheritance:
...
-      def g(a: A) -> None:
-          f(a)  # Type check error: 'A' not assignable to Dict[str, int]
+    not_required_num_dict: IntDictWithNum = {"num": 1, "bar": 2}
+    regular_dict: dict[str, int] = not_required_num_dict  # OK
+    f(not_required_num_dict)  # OK
 
-      b: B = {'x': 0, 'y': 'foo'}
-      g(b)
-      b['y'] + 'bar'  # Runtime error: int + str
+In this case, methods that are previously unavailable on a TypedDict are allowed,
```
PIG208: nit: "previously" made sense when it meant "prior to this PEP", but here these methods do not become available on the TypedDict. In the context of the spec, we can replace "previously" with "generally".

PIG208 (commented):
Went through the changes and left some comments. Looks good to me overall!

docs/spec/glossary.rst
JelleZijlstra: I think the description as a "key-value pair" is actually not great. An item isn't really a key-value pair, it's a key-type pair plus two bits for the requiredness and mutability. Let me rewrite this.

docs/spec/callables.rst
JelleZijlstra: I went over all usages of "key" and changed some to item, but in some cases I think it's clearer to say "key".

docs/spec/typeddict.rst
JelleZijlstra: Reworded with "otherwise"

JelleZijlstra:
Thanks for the feedback! I pushed some changes.

docs/spec/glossary.rst
```diff
@@ -52,6 +59,14 @@ This section defines a few terms that may be used elsewhere in the specification
       also materializations of ``B``, and all materializations of ``B`` are
       also materializations of ``A``.
 
+   extra items
+      A :ref:`TypedDict <typeddict>` type with extra items may contain arbitrary
+      additional :term:`items <item>` beyond those specified in the TypedDict definition, but those
+      items must be of the type specified by that definition.
```
carljm: Two wording clarity quibbles here:

1. I understand an "item" to mean a key-value pair, in which case an "item" does not have a type, its value has a type.
2. "The type specified by that definition" I think is supposed to mean "the type specified by the `extra_items` argument in that definition", but it doesn't say so clearly.

docs/spec/typeddict.rst
```diff
@@ -1,47 +1,62 @@
...
-particular, :term:`assignability <assignable>` should be :term:`structural`: a
-more specific TypedDict type can be assignable to a more general TypedDict
-type, without any inheritance relationship between them.
+(Originally specified in :pep:`589`, with later additions: ``Required``
+and ``NotRequired`` in :pep:`655`, use with ``Unpack`` in :pep:`692`,
+``ReadOnly`` in :pep:`705`, and ``closed=True`` and ``extra_items=`` in :pep:`728`.)
+
+A TypedDict type represents ``dict`` objects that contain only keys of
+type ``str``. There are restrictions on which string keys are valid, and
+which values can be associated with each key. Values that are members of a
```
carljm: I feel the word "members" is too overloaded/unclear, it often means something like "attribute". I think "inhabit" is an unambiguous and technically correct term here. (It should also be added to the glossary.)

```suggestion
which values can be associated with each key. Values that inhabit a
```

docs/spec/typeddict.rst
```diff
@@ -52,41 +67,142 @@ A TypedDict type can be defined using the class definition syntax with
...
+* ``extra_items``: indicates that the TypedDict has :term:`extra items`. The argument
+  must be a :term:`annotation expression` specifying the type of the extra items.
+  The :term:`type qualifier` ``ReadOnly[]`` may be used to indicate that the extra items are
+  :term:`read-only`. Other type qualifiers are not allowed. If the extra items type
+  is ``Never``, no extra items are allowed, so this is equivalent to ``closed=True``.
+
+The body of the class definition defines the :term:`items <item>` of the TypedDict type.
+It may also contain a docstring or ``pass`` statements (primarily to allow the creation of
+an empty TypedDict). No other statements are allowed, and type checkers should report an
+error if any are present. Type comments are not supported for creating TypedDict items.
```
carljm, line 128: Is this restriction clearly stated in the prior text? I would assume so, since the intent of this PR is just to reorganize, not make new normative statements, but I can't quickly find it by searching. (I also assume its a pre-existing restriction since mypy and pyright both emit a diagnostic about it.)

Anyway, it may be orthogonal to this PR if this was a pre-existing restriction, but I think this is overly and unnecessarily restrictive. I think the following typed-dict should be accepted:

```py
from typing import TypedDict
import sys


class TD(TypedDict):
    x: int
    if sys.version_info >= (3, 12):
        y: str

def f(x: TD):
    reveal_type(x["y"])
```

FWIW, today both pyright and mypy error on the conditional statement in the body. Mypy then considers the field `y` to never exist. Pyright still respects the existence and type of the field (depending on the Python version). Pyrefly and ty both also respect the field as conditionally-defined, and don't emit any error -- which I guess is not spec-compliant.

docs/spec/typeddict.rst
```diff
@@ -128,95 +244,237 @@ key, and the ``'name'`` key is missing::
...
-The TypedDict ``XYZ`` has three items: ``x`` (type ``int``), ``y``
-(type ``str``), and ``z`` (type ``bool``).
+   class X(TypedDict):
+       x: str
+       y: ReadOnly[int]
 
-A TypedDict cannot inherit from both a TypedDict type and a
-non-TypedDict base class other than ``Generic``.
+   class Y(X):
+       x: int  # Type check error: cannot overwrite TypedDict field "x"
```
carljm, line 378: I think it would help to (also?) show an example item that is overridden in a covariant way (subclass type assignable to superclass type) but without superclass type being read-only, to clarify that that is still not allowed. The current example might allow the reader to think the override was only disallowed because `int` is not assignable to `str`.

carljm (approved):
Looks good to me, and much clearer! Thank you.

docs/spec/typeddict.rst
JelleZijlstra: Good point. I agree that it makes sense to support such statically known conditions. I got this from the existing text: https://typing.python.org/en/latest/spec/typeddict.html#class-based-syntax "The class body should only contain lines with item definitions of the form key: value_type, optionally preceded by a docstring".

I'd support a separate spec change to allow your example.

docs/spec/glossary.rst
JelleZijlstra: Reworded as you suggest.

docs/spec/typeddict.rst
JelleZijlstra: Done, I'll also add it to the glossary and link the one place in concepts.rst that uses the word "inhabit".

docs/spec/typeddict.rst
JelleZijlstra: Done. I am avoiding the term "covariant" here though; the whole chapter in fact does not mention variance.

Thanks for the review!

docs/spec/typeddict.rst
carljm: https://discuss.python.org/t/allowing-typeddict-items-to-be-defined-conditionally-via-statically-known-conditions/102973

rchen152 (approved):
Awesome, this is much better organized than before!

docs/spec/typeddict.rst
```diff
@@ -227,195 +487,182 @@ Additional notes on TypedDict class inheritance:
...
-This section provides an alternative syntax that can be backported to
-older Python versions such as 3.5 and 2.7 that don't support the
-variable definition syntax introduced in :pep:`526`.  It
-resembles the traditional syntax for defining named tuples::
+    - It must also be required in ``B``.
+    - If it is read-only in ``A``, the item type in ``B`` must be a subtype of the item type in ``A``.
 
-    Movie = TypedDict('Movie', {'name': str, 'year': int})
+    - If it is mutable in ``A``, it must also be mutable in ``B``, and the item type in ``B`` must be
+      :term:`equivalent` to the item type in ``A``. (It follows that for assignability, the two item types
```
erictraut, line 529: I don't think equivalence is correct here. I think it should say "consistent with". 

```python
class Foo(TypedDict):
    x: list[int]

class Bar(Foo):
    x: list[Any] # OK
```

We should be very suspect if the word "equivalent" shows up anywhere in sublcassing or assignability rules in the spec.

docs/spec/typeddict.rst
```diff
@@ -227,195 +487,182 @@ Additional notes on TypedDict class inheritance:
...
-refers to a dictionary object does not need to be supported, to simplify
-implementation.
+        - If ``B`` is closed, the check succeeds.
+        - If ``B`` has extra items, the extra items type must be a subtype of the item type in ``A``.
 
+    - If it is mutable in ``A``:
 
-.. _typeddict-assignability:
+      - If ``B`` has an item with the same key, it must also be mutable, and its item type must be
+        :term:`equivalent` to the item type in ``A``. (As before, it follows that for assignability, the two item types
```
erictraut: I think this should say "consistent with" rather than "equivalent to".

docs/spec/typeddict.rst
```diff
@@ -227,195 +487,182 @@ Additional notes on TypedDict class inheritance:
...
+        must be :term:`consistent`.)
 
-Assignability
-^^^^^^^^^^^^^
+      - Else:
 
-First, any TypedDict type is :term:`assignable` to ``Mapping[str, object]``.
+        - If ``B`` is closed, the check fails.
+        - If ``B`` has extra items, the extra items type must not be read-only and must
+          be :term:`equivalent` to the item type in ``A``.
```
erictraut, line 551: Consistent with

docs/spec/typeddict.rst
```diff
@@ -227,195 +487,182 @@ Additional notes on TypedDict class inheritance:
...
+- If ``A`` is closed, ``B`` must also be closed, and it must not contain any items that are not present in ``A``.
+- If ``A`` has read-only extra items, ``B`` must either be closed or also have extra items, and the extra items type in ``B``
+  must be a subtype of the extra items type in ``A``. Additionally, for any items in ``B`` that are not present in ``A``,
+  the item type must be a subtype of the extra items type in ``A``.
+- If ``A`` has mutable extra items, ``B`` must also have mutable extra items, and the extra items type in ``B``
+  must be :term:`equivalent` to the extra items type in ``A``. Additionally, for any items in ``B`` that are not present in ``A``,
```
erictraut, line 558: consistent with

docs/spec/typeddict.rst
```diff
@@ -227,195 +487,182 @@ Additional notes on TypedDict class inheritance:
...
+  the item type must be :term:`equivalent` to the extra items type in ``A``.
```
erictraut, line 559: consistent with

docs/spec/typeddict.rst
```diff
@@ -227,195 +487,182 @@ Additional notes on TypedDict class inheritance:
...
-      def f(a: A) -> None:
-          a['y'] = 1
+However, a TypedDict with :term:`extra items` may be a subtype of ``dict[str, VT]``,
+provided certain conditions are met, because it introduces sufficient restrictions
+for this subtyping relation to be safe.
+A TypedDict type is a subtype of ``dict[str, VT]`` if the following conditions are met:
 
-      def g(b: B) -> None:
-          f(b)  # Type check error: 'B' not assignable to 'A'
+- The TypedDict type has mutable :term:`extra items` of a type that is :term:`equivalent` to ``VT``.
```
erictraut, line 621: Consistent with

docs/spec/typeddict.rst
```diff
@@ -227,195 +487,182 @@ Additional notes on TypedDict class inheritance:
...
+- All items on the TypedDict satisfy the following conditions:
 
-      c: C = {'x': 0, 'y': 'foo'}
-      g(c)
-      c['y'] + 'bar'  # Runtime error: int + str
+  - The value type of the item is :term:`equivalent` to ``VT``.
```
erictraut, line 624: Consistent with

docs/spec/typeddict.rst
JelleZijlstra: This rule list is specifically about subtyping, not assignability. There is a note that this implies that for assignability, you can read "consistency" wherever "equivalence" shows up. Perhaps it would be clearer to describe the assignability procedure, but subtyping is more fundamental.

erictraut: Yeah, I think the current wording is confusing and not as useful to users of the type system or for type checker authors. I see your point about subtyping being more fundamental, but assignability rules are more important for understanding typing and for properly implementing a type checker. I doubt if any of the major type checkers have code that tests for subtyping. Pyright does not. It's all about assignability.

The note helps, but it doesn't cover all aspects of the description. For example, the note doesn't apply to the statement "If it is read-only in A, the item type in B must be a subtype of the item type in A". Without additional notes, the assignability rules are arguably ambiguous.

JelleZijlstra: That makes sense and I'm happy to recast this in terms of assignability (and consistency), but I'd like to hear what others think (maybe @carljm?).

carljm: > assignability rules are more important for understanding typing

I'm not sure I agree with this. For a surface-level understanding, maybe. It's very hard to understand the logic behind the actual assignability rules without understanding their relationship to subtyping via materialization.

> and for properly implementing a type checker

That really depends on how the type checker chooses to implement the assignability rules.

> I doubt if any of the major type checkers have code that tests for subtyping.

Ty certainly does, and according to a conversation I had with Jukka at PyCon, mypy does as well. Subtyping is the relevant relation for simplifying a union (or in the case of ty, intersection) type. Since ty uses intersections for narrowing, their correct simplification is important.

I think it is important that the spec be clear about the rules for both subtyping and assignability. In general I think the best way to do that is to present the rules for subtyping, and derive the rules for assignability from the definition of materialization. I don't think an accurate definition of subtyping can be clearly derived from a definition of assignability.

> the note doesn't apply to the statement "If it is read-only in A, the item type in B must be a subtype of the item type in A".

I think the note already does correctly apply here. If you use the subtyping relation as given for two TypedDicts, and apply materialization, you will get a correct assignability relation.

I think it might be helpful to expand the note about materialization to say explicitly here what is already said in the core concepts section: that this means that where equivalence is mentioned for subtyping, it should be consistency instead for assignability, and where subtyping is mentioned, it should be assignability instead.

erictraut: OK, I buy that argument.

Thanks for the additional clarity, Jelle — and a big thanks for revamping this entire chapter. It's a big improvement!

I'll sign off on the change.

PR #2073, opened 2025-08-20
Make some minor fixes and improvements to the typedddicts_extra_items test

rchen152:
I'm in the process of implementing PEP 728 in Pyrefly and noticed a few things that could be improved in the corresponding conformance test.

* The tests for setting `closed=False` when inheriting from a TypedDict with `closed=True` or `extra_items` were also adding an illegal extra item in the subclass. Pyright was correctly flagging the extra item but incorrectly not flagging `closed=False`, meaning it missed the expected error but generated an unexpected one on the same line. I changed the subclass bodies to `pass` so that the automated conformance checker can detect the missing errors.
* Most of the errors on specific TypedDict items are marked as allowing to occur on either the `class` line or the `key: type` line. I found two that were marked as expected on only the `class` line, so I changed them to be allowed on either line.
* I didn't see a test for `closed=True` being forbidden when subclassing a TypedDict with non-read-only extra items, so I added it. Pyright already passes it.
* I also added a test verifying that a type checker flags the case of an item having a type that is incompatible with read-only extra items. Pyright passes this as well.
* The update script emitted a warning about `zuban` not being installed, so I added `zuban` to requirements.txt.


conformance/tests/typeddicts_extra_items.py
```diff
@@ -178,15 +181,18 @@ class Child(Parent, extra_items=int): # E: Cannot change 'extra_items' type unle
...
 
-class MovieNotRequiredYear(MovieBase2):  # E: 'int | None' is not consistent with 'int'
-    year: NotRequired[int]
+class MovieNotRequiredYear(MovieBase2):  # E[MovieNotRequiredYear]: 'int | None' is not consistent with 'int'
+    year: NotRequired[int]  # E[MovieNotRequiredYear]
 
 class MovieWithYear(MovieBase):  # OK
     year: NotRequired[int | None]
 
+class MovieWithDirector(MovieBase):  # E[MovieWithDirector]: 'str' is not assignable to 'int | None'
```
JelleZijlstra: Maybe use MovieBase2? MovieBase was defined a long way up (line 33) and has ReadOnly extra items, which adds an additional complication. The test immediately above this one probably should also have used MovieBase2.

rchen152: I intentionally used `MovieBase` in order to test `ReadOnly` extra items; the non-`ReadOnly` variant is already covered. Do you think it would be better to define a `MovieBase3` instead of reusing `MovieBase`?

JelleZijlstra: I think so, it's hard to follow the tests when we reuse something that is defined so far up.

We could also use something other than movies in the test (heresy?).

rchen152: Indeed heresy ;) Done.

PR #2081, opened 2025-09-01
Fixed pyrefly results output

erictraut:
This PR adds code to change pyrefly diagnostics to be recorded in the results as relative paths rather than absolute. 

It also fixes the summary report so it looks better with five columns.

It doesn't change any of the conformance results.


erictraut:
cc @yangdanny97, @rchen152, @JelleZijlstra for visibility

PR #2089, opened 2025-09-28
Updated the typing spec to use modern (PEP 695) generics syntax

erictraut:
This PR attempts to update all places in the spec where the old-style `TypeVar`, `ParamSpec` and `TypeVarTuple` constructors are used in code samples, with the exception of samples that are intended to explain the older syntax. This should improve readability.

Since it doesn't make any substantive changes to the spec, I don't think this requires approval from the full typing council, but it would be good to have a thorough review.

### DIFF
--- a/docs/spec/annotations.rst
+++ b/docs/spec/annotations.rst
@@ -367,7 +367,7 @@ with a type variable. In this case the return type may use the same
 type variable, thus making that method a generic function. For example::
 
   class Copyable:
-      def copy[T: 'Copyable'](self: T) -> T:
+      def copy[T: Copyable](self: T) -> T:
           # return a copy of self
 
   class C(Copyable): ...
--- a/docs/spec/annotations.rst
+++ b/docs/spec/annotations.rst
@@ -379,7 +379,7 @@ of the first argument::
 
   class C:
       @classmethod
-      def factory[T: 'C'](cls: type[T]) -> T:
+      def factory[T: C](cls: type[T]) -> T:
           # make a new instance of cls
 
   class D(C): ...
--- a/docs/spec/annotations.rst
+++ b/docs/spec/annotations.rst
@@ -366,9 +366,8 @@ In addition, the first argument in an instance method can be annotated
 with a type variable. In this case the return type may use the same
 type variable, thus making that method a generic function. For example::
 
-  T = TypeVar('T', bound='Copyable')
   class Copyable:
-      def copy(self: T) -> T:
+      def copy[T: 'Copyable'](self: T) -> T:
           # return a copy of self
 
   class C(Copyable): ...
@@ -378,10 +377,9 @@ type variable, thus making that method a generic function. For example::
 The same applies to class methods using ``type[]`` in an annotation
 of the first argument::
 
-  T = TypeVar('T', bound='C')
   class C:
       @classmethod
-      def factory(cls: type[T]) -> T:
+      def factory[T: 'C'](cls: type[T]) -> T:
           # make a new instance of cls
 
   class D(C): ...
--- a/docs/spec/dataclasses.rst
+++ b/docs/spec/dataclasses.rst
@@ -146,7 +146,11 @@ customization of default behaviors:
 
 .. code-block:: python
 
-  def dataclass_transform[T](
+  class _IdentityCallable(Protocol):
+      def __call__[T](self, arg: T, /) -> T:
+          ...
+
+  def dataclass_transform(
       *,
       eq_default: bool = True,
       order_default: bool = False,
@@ -154,7 +158,7 @@ customization of default behaviors:
       frozen_default: bool = False,
       field_specifiers: tuple[type | Callable[..., Any], ...] = (),
       **kwargs: Any,
-  ) -> Callable[[T], T]: ...
+  ) -> _IdentityCallable: ...
 
 * ``eq_default`` indicates whether the ``eq`` parameter is assumed to
   be True or False if it is omitted by the caller. If not specified,
--- a/docs/spec/dataclasses.rst
+++ b/docs/spec/dataclasses.rst
@@ -72,12 +72,10 @@ Decorator function example
 
 .. code-block:: python
 
-  _T = TypeVar("_T")
-
   # The ``create_model`` decorator is defined by a library.
   # This could be in a type stub or inline.
   @typing.dataclass_transform()
-  def create_model(cls: Type[_T]) -> Type[_T]:
+  def create_model[T](cls: type[T]) -> type[T]:
       cls.__init__ = ...
       cls.__eq__ = ...
       cls.__ne__ = ...
@@ -148,9 +146,7 @@ customization of default behaviors:
 
 .. code-block:: python
 
-  _T = TypeVar("_T")
-
-  def dataclass_transform(
+  def dataclass_transform[T](
       *,
       eq_default: bool = True,
       order_default: bool = False,
@@ -158,7 +154,7 @@ customization of default behaviors:
       frozen_default: bool = False,
       field_specifiers: tuple[type | Callable[..., Any], ...] = (),
       **kwargs: Any,
-  ) -> Callable[[_T], _T]: ...
+  ) -> Callable[[T], T]: ...
 
 * ``eq_default`` indicates whether the ``eq`` parameter is assumed to
   be True or False if it is omitted by the caller. If not specified,
--- a/docs/spec/generics.rst
+++ b/docs/spec/generics.rst
@@ -210,22 +210,15 @@ defined class is generic if you subclass one or more other generic classes and
 specify type variables for their parameters. See :ref:`generic-base-classes`
 for details.
 
-You can use multiple inheritance with ``Generic``::
+You can use multiple inheritance with generic classes::
 
-  from typing import TypeVar, Generic
-  from collections.abc import Sized, Iterable, Container
+  from collections.abc import Container, Iterable, Sized
 
-  T = TypeVar('T')
-
-  class LinkedList(Sized, Generic[T]):
+  class LinkedList[T](Sized):
       ...
 
-  K = TypeVar('K')
-  V = TypeVar('V')
-
-  class MyMapping(Iterable[tuple[K, V]],
-                  Container[tuple[K, V]],
-                  Generic[K, V]):
+  class MyMapping[K, V](Iterable[tuple[K, V]],
+                        Container[tuple[K, V]]):
       ...
 
 Subclassing a generic class without specifying type parameters assumes
@@ -245,8 +238,9 @@ Generic metaclasses are not supported.
 Scoping rules for type variables
 --------------------------------
 
-Type variables follow normal name resolution rules.
-However, there are some special cases in the static typechecking context:
+When using the generic class syntax introduced in Python 3.12, the location of
+its declaration defines its scope. When using the older syntax, the scoping
+rules are more subtle and complex.
 
 * A type variable used in a generic function could be inferred to represent
   different types in the same code block. Example::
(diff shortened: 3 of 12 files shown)


docs/spec/generics.rst
```diff
@@ -344,13 +338,9 @@ Instantiating generic classes and type erasure
 ----------------------------------------------
 
 User-defined generic classes can be instantiated. Suppose we write
-a ``Node`` class inheriting from ``Generic[T]``::
+a ``Node`` class using the new generic class syntax::
```
hauntsaninja: nit:
```suggestion
a ``Node`` class::
```

docs/spec/annotations.rst
```diff
@@ -366,9 +366,8 @@ In addition, the first argument in an instance method can be annotated
 with a type variable. In this case the return type may use the same
 type variable, thus making that method a generic function. For example::
 
-  T = TypeVar('T', bound='Copyable')
   class Copyable:
-      def copy(self: T) -> T:
+      def copy[T: 'Copyable'](self: T) -> T:
```
AlexWaygood: Forward references are natively supported in annotation scopes, so stringifying the bound here is unnecessary


```suggestion
      def copy[T: Copyable](self: T) -> T:
```

docs/spec/annotations.rst
```diff
@@ -378,10 +377,9 @@ type variable, thus making that method a generic function. For example::
 The same applies to class methods using ``type[]`` in an annotation
 of the first argument::
 
-  T = TypeVar('T', bound='C')
   class C:
       @classmethod
-      def factory(cls: type[T]) -> T:
+      def factory[T: 'C'](cls: type[T]) -> T:
```
AlexWaygood:
```suggestion
      def factory[T: C](cls: type[T]) -> T:
```

docs/spec/dataclasses.rst
```diff
@@ -148,17 +146,15 @@ customization of default behaviors:
 
 .. code-block:: python
 
-  _T = TypeVar("_T")
-
-  def dataclass_transform(
+  def dataclass_transform[T](
```
AlexWaygood: This implies that the TypeVar is scoped to the function, but it is actually scoped to the function's return annotation. To be fully explicit and unambiguous here, you could consider using a callback protocol for the return annotation, which is what we do in the annotations for the runtime implementation of `dataclass_transform` at https://github.com/python/cpython/blob/3779f2b95ebf68e2d5f8ea27c7a7baa0879f5496/Lib/typing.py#L3597

docs/spec/generics.rst
```diff
@@ -210,22 +210,15 @@ defined class is generic if you subclass one or more other generic classes and
 specify type variables for their parameters. See :ref:`generic-base-classes`
 for details.
 
-You can use multiple inheritance with ``Generic``::
+You can use multiple inheritance with generic classes::
```
AlexWaygood: This subsection about multiple inheritance appears to be specifically about the rules at runtime for using multiple inheritance with the class `typing.Generic`. I'm not totally sure this subsection belongs in the typing spec at all, but I think if it is to be kept here then it defeats the point for this subsection to use PEP-695 syntax. I'd revert the changes made to this subsection

docs/spec/generics.rst
```diff
@@ -245,8 +238,9 @@ Generic metaclasses are not supported.
 Scoping rules for type variables
 --------------------------------
 
-Type variables follow normal name resolution rules.
-However, there are some special cases in the static typechecking context:
+When using the generic class syntax introduced in Python 3.12, the location of
+its declaration defines its scope. When using the older syntax, the scoping
+rules are more subtle and complex.
```
AlexWaygood:
```suggestion
rules are more subtle and complex:
```

docs/spec/generics.rst
```diff
@@ -665,8 +651,18 @@ ParamSpec
...
+(for a function, class, or type alias).
+
+.. code-block::
+
+   def decorator[**P](func: Callable[P, int]) -> Callable[P, str]: ...
+
+   class CallbackWrapper[T, **P]:
+       callback: Callable[P, T]
+
+Prior to 3.12, the ``ParamSpec`` constructor can be used::
```
AlexWaygood: You don't want to use `::` here because you have `.. code-block::` immediately below

```suggestion
Prior to 3.12, the ``ParamSpec`` constructor can be used:
```

docs/spec/generics.rst
```diff
@@ -702,62 +698,54 @@ parameter specification variable (``Callable[Concatenate[int, P], int]``\ ).
                       parameter_specification_variable
                    "]"
 
-where ``parameter_specification_variable`` is a ``typing.ParamSpec`` variable,
-declared in the manner as defined above, and ``concatenate`` is
+where ``parameter_specification_variable`` is introduced either inline (using
+``**``) or via ``typing.ParamSpec`` as shown above, and ``concatenate`` is
```
AlexWaygood:
```suggestion
``**`` in a type parameter list) or via ``typing.ParamSpec`` as shown above, and ``concatenate`` is
```

docs/spec/generics.rst
```diff
@@ -820,9 +806,7 @@ evaluated in the same way.
 
 .. code-block::
 
-   U = TypeVar("U")
-
-   class Y(Generic[U, P]):
+   class Y[U, **P]:
      f: Callable[P, str]
```
AlexWaygood, line 817:
```suggestion
       f: Callable[P, str]
```

docs/spec/generics.rst
```diff
@@ -820,9 +806,7 @@ evaluated in the same way.
...
      prop: U
```
AlexWaygood, line 818:
```suggestion
       prop: U
```

docs/spec/generics.rst
```diff
@@ -1101,27 +1085,27 @@ In the same way that a normal type variable is a stand-in for a single
 type such as ``int``, a type variable *tuple* is a stand-in for a *tuple* type such as
 ``tuple[int, str]``.
 
-Type variable tuples are created and used with:
+In Python 3.12 and newer, type variable tuples can be introduced inline by prefixing
+their name with ``*`` inside a generic parameter list.
```
AlexWaygood: This is the term that users will find in the CPython docs:

```suggestion
their name with ``*`` inside a type parameter list.
```

docs/spec/overload.rst
```diff
@@ -40,28 +40,28 @@ Another example where ``@overload`` comes in handy is the type of the
...
-
   @overload
-  def map(func: Callable[[T1], S], iter1: Iterable[T1]) -> Iterator[S]: ...
+  def map[T1, S](func: Callable[[T1], S], iter1: Iterable[T1]) -> Iterator[S]: ...
   @overload
-  def map(func: Callable[[T1, T2], S],
-          iter1: Iterable[T1], iter2: Iterable[T2]) -> Iterator[S]: ...
+  def map[T1, T2, S](
+      func: Callable[[T1, T2], S],
+      iter1: Iterable[T1], iter2: Iterable[T2],
```
AlexWaygood:
```suggestion
      iter1: Iterable[T1],
      iter2: Iterable[T2],
```

docs/spec/overload.rst
```diff
@@ -91,11 +91,7 @@ A constrained ``TypeVar`` type can sometimes be used instead of
 using the ``@overload`` decorator. For example, the definitions
 of ``concat1`` and ``concat2`` in this stub file are equivalent::
 
-  from typing import TypeVar
-
-  AnyStr = TypeVar('AnyStr', str, bytes)
-
-  def concat1(x: AnyStr, y: AnyStr) -> AnyStr: ...
+  def concat1[AnyStr: (str, bytes)](x: AnyStr, y: AnyStr) -> AnyStr: ...
```
AlexWaygood: I'd prefer to avoid using the name `AnyStr` for the type variable. It doesn't mean "any string" and has nothing to do with `Any` either; we deprecated the symbol at runtime for these reasons.

docs/spec/protocol.rst
```diff
@@ -266,25 +266,25 @@ Generic protocols are important. For example, ``SupportsAbs``, ``Iterable``
...
-``Protocol[T, S, ...]`` is allowed as a shorthand for
-``Protocol, Generic[T, S, ...]``. It is an error to combine
-``Protocol[T, S, ...]`` with ``Generic[T, S, ...]``, or with the new syntax for
-generic classes in Python 3.12 and above::
+The older syntax ``Protocol[T, S, ...]`` remains available as a shorthand for
+``Protocol, Generic[T, S, ...]``. It is an error to combine the shorthand with
+``Generic[T, S, ...]`` or to mix it with the new ``class Iterable[T]`` form::
 
-  class Iterable(Protocol[T], Generic[T]):   # INVALID
+  class Iterable[T](Protocol, Generic[T]):   # INVALID
```
AlexWaygood: I agree that this modified example demonstrates something a type checker should reject, but I don't think it demonstrates the same thing that it used to (and it seems to me that it makes the example immediately below it redundant). I'd revert this change

erictraut:
Thanks for the review @AlexWaygood. I think I've addressed your feedback.

I'm going to leave this unmerged a bit longer to give others a chance to review if they'd like.

docs/spec/generics.rst
```diff
@@ -665,8 +658,18 @@ ParamSpec
 Declaration
 """"""""""""
 
-A parameter specification variable is defined in a similar manner to how a
-normal type variable is defined with ``typing.TypeVar``.
+In Python 3.12 and newer, a parameter specification variable can be introduced
+inline by prefixing its name with ``**`` inside a generic parameter list
```
JelleZijlstra:
```suggestion
inline by prefixing its name with ``**`` inside a type parameter list
```

docs/spec/generics.rst
```diff
@@ -702,62 +705,54 @@ parameter specification variable (``Callable[Concatenate[int, P], int]``\ ).
...
 .. code-block::
 
-   T = TypeVar("T")
-   P_2 = ParamSpec("P_2")
-
-   class X(Generic[T, P]):
-     f: Callable[P, int]
-     x: T
+   from collections.abc import Callable
+   from typing import Concatenate, ParamSpec
```
JelleZijlstra:
```suggestion
   from typing import Concatenate
```
Import looks unused

docs/spec/generics.rst
```diff
@@ -1924,11 +1899,9 @@ types or an unpacked, in-scope ``TypeVarTuple`` (see `Scoping Rules`_).
 
 ::
 
-   DefaultTs = TypeVarTuple("DefaultTs", default=Unpack[tuple[str, int]])
+   class Foo[*Ts = *tuple[str, int]]: ...
 
-   class Foo(Generic[*DefaultTs]): ...
-
-   reveal_type(Foo)               # type is type[Foo[DefaultTs = *tuple[str, int]]]
+   reveal_type(Foo)               # type is type[Foo[Ts = *tuple[str, int]]]
```
JelleZijlstra:
```suggestion
   reveal_type(Foo)               # type is type[Foo[str, int]]
```
The previous example here isn't valid syntax. Though this whole example is a bit questionable, but the output I suggested matches what pyright says.

docs/spec/literal.rst
```diff
@@ -733,18 +733,16 @@ Conditional statements and expressions work as expected:
         return result  # OK
 
 
-Interaction with TypeVars and Generics
-""""""""""""""""""""""""""""""""""""""
+Interaction with Type Variables and Generics
```
JelleZijlstra, line 736: Out of scope for this PR but I feel this whole section is basically redundant and could be deleted. LiteralString is a type, so it can be used in any place that an arbitrary type can be used. Some of the Literal-related text above is similarly redundant.

docs/spec/protocol.rst
```diff
@@ -266,25 +266,26 @@ Generic protocols are important. For example, ``SupportsAbs``, ``Iterable``
...
 
   class Iterable[T](Protocol[T]):   # INVALID
       ...
 
-User-defined generic protocols support explicitly declared variance.
-Type checkers will warn if the inferred variance is different from
-the declared variance. Examples::
+When using the generics syntax introduced in Python 3.12, the variance of
+type variables is inferred. When using the pre-3.12 generics syntax, variance
+must be specified. Type checkers will warn if the declared variance does not
```
JelleZijlstra:
```suggestion
must be specified (unless ``infer_variance=True`` is used). Type checkers will warn if the declared variance does not
```

JelleZijlstra (approved):
Thank you!

PR #2093, opened 2025-10-08
Remove tests with invalid syntax (Python 3.14)

sharkdp:
The use of named expressions and await expressions results in syntax errors on Python 3.14 (see below). This PR therefore suggests to remove two tests from `conformance/tests/qualifiers_annotated.py` that were making use of this now-invalid syntax.

```
▶ uvx python@3.14               
Python 3.14.0 (main, Oct  7 2025, 15:35:21) [Clang 20.1.4 ] on linux
>>> from typing import Annotated, Literal
>>> Good9: Annotated[Literal[3, 4, 5, None], x := 3]
  File "<python-input-1>", line 1
    Good9: Annotated[Literal[3, 4, 5, None], x := 3]
                                             ^^^^^^
SyntaxError: named expression cannot be used within an annotation
>>> async def func3() -> None:
...     Good10: Annotated[str, await func3()]
...     
  File "<python-input-2>", line 2
    Good10: Annotated[str, await func3()]
                           ^^^^^^^^^^^^^
SyntaxError: await expression cannot be used within an annotation
>>> l
  File "<python-input-3>", line 1, in <module>
    l
NameError: name 'l' is not defined
```

### DIFF
--- a/conformance/src/type_checker.py
+++ b/conformance/src/type_checker.py
@@ -365,7 +365,7 @@ class PyreflyTypeChecker(TypeChecker):
 
     def run_tests(self, test_files: Sequence[str]) -> dict[str, str]:
         proc = run(
-            ["pyrefly", "check", "--output-format", "min-text", "--no-summary"],
+            ["pyrefly", "check", "--output-format", "min-text", "--summary=none"],
             stdout=PIPE,
             text=True,
             encoding="utf-8",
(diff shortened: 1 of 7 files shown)


conformance/src/type_checker.py
```diff
@@ -365,7 +365,7 @@ def get_version(self) -> str:
 
     def run_tests(self, test_files: Sequence[str]) -> dict[str, str]:
         proc = run(
-            ["pyrefly", "check", "--output-format", "min-text", "--no-summary"],
+            ["pyrefly", "check", "--output-format", "min-text", "--summary=none"],
```
sharkdp, line 368: This is a drive-by fix. The latest version of pyrefly does not seem to support `--no-summary` anymore. Let me know if this should be removed or pulled out into a second PR.

PR #2096, opened 2025-10-10
Make deferred-annotation errors optional (depends on version)

sharkdp:
The forward references in `conformance/tests/annotations_forward_refs.py` are no longer errors on Python 3.14, so it seems reasonable to me to mark them as `E?` instead of `E`, because whether or not this is an error depends on how the type checkers are configured (what Python version are they targeting?).

I saw that a similar pattern was used in `typeddicts_alt_syntax.py` to mark version-depend errors with `E?`, but I'm obviously open to other suggestions (e.g. removing these test assertions alltogether, if possible).

### DIFF
--- a/conformance/tests/annotations_forward_refs.py
+++ b/conformance/tests/annotations_forward_refs.py
@@ -19,8 +19,8 @@ def func1(
     assert_type(p4, list[ClassA | int])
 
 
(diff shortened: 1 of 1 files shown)


JelleZijlstra:
Thanks, this seems fine. Maybe in the long term we should have some way to configure the Python version in the conformance test suite.

PR #2102, opened 2025-10-15
Add ty and Zuban to the list of type checkers that exist

AlexWaygood:
Unfortunately, this leads to 3/4 type checkers in this list advertising themselves as "fast", which feels a bit silly... ideally I'd put more information here about how ty tries to distinguish itself from the other two fast type checkers, but it feels unfair to have a more verbose description for ty than is afforded to the other three type checkers here :-)

We could also just get rid of the descriptions for all four type checkers; I think it would probably be fine to say that readers of these docs have to go elsewhere to try to decide which tool is right for them

### DIFF
--- a/docs/index.rst
+++ b/docs/index.rst
@@ -79,14 +79,11 @@ Typing-related Tools
 Type Checkers
 -------------
 
-* `mypy <http://mypy-lang.org/>`_, the reference implementation for type
-  checkers.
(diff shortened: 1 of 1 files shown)


hauntsaninja (approved):
Yeah, the descriptions just seem not helpful. I would remove them entirely. We should also mention zuban

PR #2105, opened 2025-10-20
Document the trouble with lambdas in the guides

### LINKED ISSUE #2020: Indicate in the docs that static typing doesn't work with lambda functions
I sometimes get asked why a `lambda` function can't be type-checked.  And I believe the answer is that the arguments and result of a `lambda` can't be inferred by a static type checker.

It might be worth adding something at https://typing.python.org/en/latest/spec/callables.html#callables that tells people that `lambda` functions can't be type checked (or if they can, what are the restrictions), and that the workaround is to create a typed function `def` that does what the lambda does.

wyattscarpenter:
Closes #2020


docs/guides/lambda.rst
```diff
@@ -0,0 +1,65 @@
...
+
+``def f(x: object) -> object: return x``
+
+The second workaround is equivalent: assigning the lambda to a variable, and annotating
+the type of that variable with a Callable.
+
+``f: Callable[[object], object] = lambda x: x``
+
+Type comments on function definitions do not actually work on lambda, nor do
+normal type comments help (although you can use a type comment on an assignment
```
JelleZijlstra: Might be better to not mention type comments? They're effectively deprecated, and in the context we're discussing I don't think they add anything over variable annotations.

docs/guides/lambda.rst
```diff
@@ -0,0 +1,65 @@
...
+
+3. Many lambdas get deduced as ``Any``, which might suppress the reporting of other
+type errors.
+
+4. However, many lambdas get deduced fine, and for those it's not a problem.
+
+5. If you want to annotate the type of lambdas, you can bind them and annotate them
+there.
+
+6. Most type checkers have a setting that will warn you if anything gets deduced as
```
JelleZijlstra, line 87: Could be useful to mention what those rules are for at least some type checkers (minimally mypy and pyright), so the guide is more practically useful.

docs/guides/lambda.rst
wyattscarpenter: They are a random intrusion of esoterica here, but they *are* a potential answer to "where would the type annotations go?", so I think it makes sense to mention that they don't work.

JelleZijlstra: I feel many readers probably won't even know what type comments are, so discussing them is more likely to confuse than to enlighten.

wyattscarpenter: Incorporated; thanks!

docs/guides/lambda.rst
wyattscarpenter: Incorporated; thanks!

PR #2106, opened 2025-10-21
Update Modernizing document for Python 3.14

srittau:
Document that the future import isn't necessary anymore.


srittau:
@JelleZijlstra Maybe you'd like to review this, considering that PEP 749 was yours?

PR #2114, opened 2025-11-17
Best Practices: Split an example

srittau:
Split the text and examples for `Any` vs `object` into one for arguments and one for callback return types. This separated these two related, but separate suggestions. It should make the examples a bit clearer and less convoluted.


srittau:
https://typing--2114.org.readthedocs.build/en/2114/reference/best_practices.html#using-any-and-object

PR #2116, opened 2025-11-22
add ci/cd for conformance testing and option to skip install type checkers

Cloud0310:
As the both pyrefly and zuban is under active development, the results of conformance test are often stale, so I added actions for auto update weekly.
Due to the action doesn't need to clean install cache every time, added cmd option to skip installation check for type checkers.


Cloud0310:
I guess maybe we can deploy the test result to github pages? Current solution is kinda bad. Would love to help.

erictraut:
The conformance tests require manual "grading" and the addition of notes for each failure. These parts of the tests cannot be automated. This manual process is made easier is to by using the automated tests and diffs. This CI script will make it more difficult (almost impossible) to do this manual process. For that reason, I'm not in favor of adding this.

PR #2137, opened 2025-12-17
feat: add ty

SimonSchick:
See https://docs.astral.sh/ty/

I'm uncertain what steps need to be taken to mark tests as unsupported rather than `Unknown` in bulk.
Should I mark all failing tests as `unsupported` for the time being?


conformance/results/pyrefly/version.toml
```diff
@@ -1 +1 @@
-version = "pyrefly 0.45.2"
+version = "pyrefly 0.46.0"
```
SimonSchick, line 1: This auto-updated, please let me know if this should be reverted.

conformance/src/type_checker.py
```diff
@@ -200,6 +200,79 @@ def parse_errors(self, output: Sequence[str]) -> dict[int, list[str]]:
             line_to_errors.setdefault(int(lineno), []).append(line)
         return line_to_errors
 
+class TyTypeChecker(TypeChecker):
```
SimonSchick, line 203: This is entirely based on the `Pyright` checker.

conformance/src/type_checker.py
```diff
@@ -200,6 +200,79 @@ def parse_errors(self, output: Sequence[str]) -> dict[int, list[str]]:
...
+        a = set[str]()
+        for line in output:
+            if not line or line[0].isspace():
+                continue
+            # Ignore indented notes
+            assert line.count(":") >= 3, f"Failed to parse line: {line!r}"
+            _, lineno, kind, _ = line.split(":", maxsplit=3)
+            kind = kind.split()[-1]
+            a.add(kind)
+            if kind not in ("major", "minor"):
```
SimonSchick, line 265: Not sure if we want to re-map these to `error` and `warning` but this seems to work fine-ish?

conformance/src/type_checker.py
```diff
@@ -200,6 +200,79 @@ def parse_errors(self, output: Sequence[str]) -> dict[int, list[str]]:
...
+
+    def get_version(self) -> str:
+        proc = run(
+            [sys.executable, "-m", "ty", "--version"], stdout=PIPE, text=True
+        )
+        return proc.stdout.strip()
+
+    def run_tests(self, test_files: Sequence[str]) -> dict[str, str]:
+        command = [sys.executable, "-m", "ty", "check", ".", "--output-format", "gitlab"]
+        proc = run(command, stdout=PIPE, text=True, encoding="utf-8")
```
SimonSchick, line 235: `gitlab` seems to be the only json output mode.

conformance/src/type_checker.py
SimonSchick: I am also open to move these checkers into separate files, it's getting a lil crowded.

hauntsaninja:
Thanks, but folk from ty don't yet want to add themselves to the conformance suite

erictraut:
@SimonSchick, are you working with the Astral team on this? Ty is still in prerelease, so I've been waiting for them to give the green light before adding them to the conformance tests. If you have their approval, then great. Otherwise, I think we should hold off on this.

SimonSchick:
I do not but you are welcome to continue this PR at a later time.

srittau:
Thanks for your contribution. Per the discussion, I'm closing this for now, until the ty developers agree to add it to the conformance tests.

PR #2138, opened 2025-12-18
Use `pyproject.toml` instead of `setup.py`

### LINKED ISSUE #2121: Migrate typing documentation to `pyproject.toml`
# Documentation

The official typing documentation recommends using a `setup.py` without `pyproject.toml` (https://typing.python.org/en/latest/guides/libraries.html#companion-type-stub-package, `Doc/library/typing.rst`). This relies on pre-PEP 621 fallback behavior, while we should teach users to follow PEP 621 and use `pyproject.toml`. Can we migrate this documentation to use `pyproject.toml`, including `[project]` and `build-system]` as well as `[tool.*]` for configuration? 

The documentation at from packaging.python.org has a tab selection for different build backends, with a default of hatchling instead of setuptools (https://packaging.python.org/en/latest/tutorials/packaging-projects/#choosing-a-build-backend). We found that this is easier to work with than setuptools. A possible alternative would be having the same tabs on typing.python.org (I don't know if that's supported in the renderer), or to move the packaging part of the documentation to packaging.python.org.

konstin:
The main motivation for this change is to migrate from the non-standard `setup.py` to the standard PEP 517 `setup.py`. The secondary motivation is to use a build backend that's easier to use and harder to get wrong than setuptools.

Modern build backends and much easier to use than setuptools, and will generally include `py.typed` and `.pyi` files automatically. This means we can remove the setuptools-specific documentation. We use `uv_build` for the stubs package example since it works out of the box with it.

packaging.python.org has a tab system to show different build backends (https://packaging.python.org/en/latest/guides/writing-pyproject-toml/). Afaik typing.python.org doesn't, so we only show one build backend.

Fixes #2121

### DIFF
--- a/docs/guides/libraries.rst
+++ b/docs/guides/libraries.rst
@@ -88,10 +88,7 @@ A typical directory structure would look like:
       stuff.py
       py.typed
 
-It's important to ensure that the ``py.typed`` marker file is included in the
-distributed package. Modern build backends such as ``hatchling`` include it
-by default.
-
+Note the py.typed should be located inside the package, along with ``__init__.py``.
 
 Type stub files included in the package
 ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^


docs/guides/libraries.rst
```diff
@@ -150,41 +136,28 @@ For example:
...
-      version="0.1",
-      package_data={"my_great_package-stubs": ["__init__.pyi", "stuff.pyi"]},
-      packages=["my_great_package-stubs"]
-   )
+   [build-system]
+   requires = ["hatchling"]
+   build-backend = "hatchling.build"
 
+   [tool.hatch.build.targets.wheel]
+   packages = ["src/my_great_package-stubs"]
```
konstin: uv-build supports those without extra configuration, but hatchling is wider used.

docs/guides/libraries.rst
```diff
@@ -82,25 +82,25 @@ A typical directory structure would look like:
...
 
-   setup.py
+   pyproject.toml
    my_great_package/
       __init__.py
       stuff.py
       py.typed
 
 It's important to ensure that the ``py.typed`` marker file is included in the
-distributed package. If using ``setuptools``, this can be achieved like so:
```
konstin: This note is afaik only relevant for setuptools anymore. Since the other build backends I'm aware support this in the default configuration, I've shortened it to this hint, the specifics fit better in the setuptools documentation.

srittau: I would rather make a general statement, instead of focusing on hatchling. Something like "Most modern build tools like ``hatchling`` include it by default."

hauntsaninja: `setuptools>=69` includes this by default too, so I think we can just remove this statement / I think we can just not mention build backends in this guide

konstin: I shrunk the note to mentioning it needs to be included and all modern package managers do it by default, i.e. don't put it in your gitignore where hatchling skips it.

konstin:
With the latest iteration, only the stubs package has a `pyproject.toml`, for all other cases the defaults from you package manager / following https://packaging.python.org/en/latest/guides/writing-pyproject-toml/ does the right thing.

konstin:
> For the stubs-only package, I'd vote we change the example to uv-build because that's a better experience for these users. If hatchling later adds the same feature, then we can edit to mention that.

Do you still want to show the `pyproject.toml` contents? To get stubs with uv-build, the basic getting-started instructions work:

```toml
[project]
name = "my-great-package-stubs"
version = "0.1.0"
requires-python = ">=3.14"

[build-system]
requires = ["uv_build>=0.9.18,<0.10.0"]
build-backend = "uv_build"
```

hauntsaninja:
I'm okay with that for the "Companion type stub package" example. Let's change requires python to 3.10 though. If someone blindly copies and pastes, they will notice faster if python version is too low rather than too high

PR #2144, opened 2026-01-04
Change the typing spec around string references

davidhalter:
I added this after the discussion here: https://discuss.python.org/t/annotation-string-references-in-class-scope-in-conformance-tests/105439

I'm not 100% sure about the wording, but I hope the direction is fine. I would like to gather some feedback before presenting this to the typing council.

Please also merge https://github.com/python/typing/pull/2139 before this pull request. Otherwise it will be very hard to update Zuban's conformance test results in this pull request.


davidhalter:
@JelleZijlstra Could you please pre-review this? What do you think about this spec change?

docs/spec/annotations.rst
```diff
@@ -243,8 +243,7 @@ The string literal should contain a valid Python expression (i.e.,
 ``compile(lit, '', 'eval')`` should be a valid code object) and it
 should evaluate without errors once the module has been fully loaded.
 The local and global namespace in which it is evaluated should be the
-same namespaces in which default arguments to the same function would
-be evaluated.
+same namespaces in which normal non-string types would be evaluated.
```
JelleZijlstra: The "local and global namespaces" wording is imprecise because the names may come from another namespace, like the type parameter scope. Similarly, the wording about "once the module has been fully loaded" on line 244 is questionable, because similar reasoning applies to a class scope. And regarding the "evaluate without errors" part, mypy and pyright both accept `def f(x: "int | str") -> None: pass` in 3.9 mode, but that annotation wouldn't evaluate without errors. I also don't particularly like the motivating example, since in Python 3.14 you don't actually need to quote that annotation.

So in summary, I think we need to rewrite a bit more. I would drop the example (this is a spec, not a user guide). The spec text could be something like "The string literal should contain a syntactically valid Python expression (i.e., `compile(lit, '', 'eval')` should succeed) that evaluates to a valid :term:`annotation expression`. Names within the expression are looked up in the same way as they would be looked up at runtime in Python 3.14 and higher if the annotation was not enclosed in a string literal. Thus, name lookup follows general rules (e.g., the current function, class, or module scope first, and the builtin scope last), but names defined later within the same scope can be used in an earlier annotation."

davidhalter: I pushed my latest changes. This incorporates your changes. I'm kind of happy with how you worded it, so I didn't really modify anything. Please let me know if further changes are needed.

conformance/tests/annotations_forward_refs.py
```diff
@@ -79,21 +78,21 @@ class ClassD:
...
 
+    x: "int" = 0  # # E: Refers to a local int as well
+
     def __init__(self) -> None:
         self.ClassC = ClassC()
 
 
-assert_type(ClassD.str, str)
-assert_type(ClassD.x, int)
+assert_type(ClassD.str, Any)
```
erictraut: These lines should be removed. The spec does not (nor should) specify what types should be evaluated by type checkers in error cases. Since these are now error cases, the conformance test should not enforce any particular behavior here.

erictraut (dismissed):

conformance/tests/annotations_forward_refs.py
```diff
@@ -79,21 +78,21 @@ class ClassD:
...
+    str: "str" = ""  # E: circular reference
 
     def int(self) -> None:  # OK
         ...
 
-    x: "int" = 0  # OK
-
     y: int = 0  # E: Refers to local int, which isn't a legal type expression
 
+    x: "int" = 0  # # E: Refers to a local int as well
```
carljm: It seems like this example would be even more illuminating if placed before `def int`

davidhalter: Do you think I should move both x and y? In 3.14 it should not matter where the annotations are defined, but type checkers might still work a bit different.

conformance/tests/annotations_forward_refs.py
davidhalter: That makes sense. I didn't want to change too much to the original test. Those statements have been removed.

conformance/tests/annotations_forward_refs.py
carljm: If we're aiming for this test to really clarify the behavior, I think having a version of both x and y, both before and after `def int`, would be useful. Their behavior should not be the same. The stringified version should refer to `def int` regardless of location; the non stringified version will refer to the global `int` if before `def int`

davidhalter: This is not how I understand Python 3.14 works. Maybe @JelleZijlstra can clarify. If I understand it correctly, both before and after would refer to the `def int`.

carljm: The 3.14 behavior would be that before or after `def int`, the annotation `int` refers to `def int`. So that should be the behavior of the stringified `"int"` annotation.

The behavior of the unstringified `int` annotation should depend on the Python version. Checking under 3.14 it should behave the same as the stringified annotation. Checking under an earlier Python, it should resolve as I described above.

I don't interpret this PR to be specifying that non-stringified annotations in Python versions earlier than 3.14 should behave as if under 3.14.

So you're right -- my comment above was making an unwarranted assumption that the conformance tests specify behavior on an older version of Python. But there's no reason that should be assumed. I think this highlights an existing problem, which is that it's not clear (or documented anywhere, as far as I can find) what Python version the conformance tests should assume.

If the conformance tests should always be assumed to check under the most recent Python, then stringified and non-stringified annotations should behave the same.

davidhalter: Agreed. Now I'm just not sure what to do. :smile:  What would you suggest?

carljm: Probably for purposes of this PR the best option is to avoid checking behavior that should depend on the Python version? So that implies avoiding an unstringified `int` annotation before `def int`. Which is fine -- there already isn't one. So just ignore that part of my suggestion :) 

I also think the conformance tests should ideally be clearer about Python version, and probably include a facility for explicitly checking some files under a specified version. But that's a bigger project, definitely out of scope for this PR.

davidhalter:
I think I have integrated all the changes. Is it time to open an issue on the Typing Council’s issue tracker asking for a decision?

conformance/tests/annotations_forward_refs.py
carljm: I still think having `z: "int"` before `def int`, as well as the `x: "int"` after (my original suggestion in this thread) is useful, and doesn't depend on Python version.

It's the expanded idea of also having `foo: int` before `def int` that would depend on Python version, and thus isn't worth doing here.

docs/spec/annotations.rst
```diff
@@ -222,32 +222,14 @@ String annotations
...
-same namespaces in which default arguments to the same function would
-be evaluated.
-
-Moreover, the expression should be parseable as a valid type hint, i.e.,
-it is constrained by the rules from :ref:`the expression grammar <expression-grammar>`.
+The string literal should contain a syntactically valid Python expression
+(i.e., ``compile(lit, '', 'eval')`` should succeed) that evaluates to a valid
+:term:`annotation expression`. Names within the expression are looked up in the
+same way as they would be looked up at runtime in Python 3.14 and higher if the
+annotation was not enclosed in a string literal. Thus, name lookup follows
```
JelleZijlstra: Are we in agreement that this is how lookup should happen regardless of the Python version? So even on earlier versions, stringified annotations should use 3.14 semantics for name resolution. That was my intent when writing this, but not clear if we agree that's how it should be done.

If so, this should make it clearer:
```suggestion
:term:`annotation expression`. Regardless of the Python version used, names within the expression are looked up in the
same way as they would be looked up at runtime in Python 3.14 and higher if the
annotation was not enclosed in a string literal. Thus, name lookup follows
```

davidhalter: I committed your change, because I believe it makes sense to specify that explicitly. If this does not meet the consensus, I will revert.

carljm: This was my understanding of the intent of the change.

conformance/tests/annotations_forward_refs.py
carljm: @davidhalter I think my last comment in this thread is not addressed? I do still think there's a clarifying addition that can be made to these tests that doesn't depend on Python version: add an attribute with annotation `"int"`, before `def int(...)`.

docs/spec/annotations.rst
```diff
@@ -222,32 +222,14 @@ String annotations
...
-should evaluate without errors once the module has been fully loaded.
-The local and global namespace in which it is evaluated should be the
-same namespaces in which default arguments to the same function would
-be evaluated.
-
-Moreover, the expression should be parseable as a valid type hint, i.e.,
-it is constrained by the rules from :ref:`the expression grammar <expression-grammar>`.
+The string literal should contain a syntactically valid Python expression
+(i.e., ``compile(lit, '', 'eval')`` should succeed) that evaluates to a valid
+:term:`annotation expression`. Regardless of the Python version used, names within the expression are looked up in the
```
carljm: Nit: word-wrap this line consistently with the rest of the text? it's too long

carljm: I would say just "that is a valid annotation expression". An annotation expression itself is a syntactic form, not a runtime object, so nothing can "evaluate to" an annotation expression.

carljm (approved):
One wording nit, one formatting nit, and one conformance suite nit :) But overall this looks great to me.

docs/spec/annotations.rst
davidhalter: I will apply both of your suggestions if there are no other opinions. Since you wrote "one conformance suite nit", did you just forget to add that?

carljm: It was a comment on a pre-existing thread: https://github.com/python/typing/pull/2144#discussion_r2739017520

rchen152 (approved):
Looks good to me - much more consistent and clearly specified than before

carljm (approved):
This looks good to me. Thanks @davidhalter for getting this clarified.

davidhalter:
I have integrated all of Carl's suggestions. I will update the conformance tests as soon as the typing council approves this change. If I update it now we probably just run into merge conflicts, since especially pyrefly changes a lot.

@carljm Please let me know if you think something needs more work.

JelleZijlstra (approved):

zzzeek:
Hi, can someone explain the intent of this change to me?

Given, under python 3.14:

```
Python 3.14.0 (main, Oct 20 2025, 16:44:45) [GCC 14.3.1 20250808 (Red Hat 14.3.1-3)] on linux
>>> class A: pass
... 
>>> class X:
...     def A(self) -> "A": pass
...     
>>> class Y:
...     def A(self) -> A: pass
...     
>>> typing.get_type_hints(X.A)
{'return': <class '__main__.A'>}
>>> typing.get_type_hints(Y.A)
{'return': <function Y.A at 0x7f2052bc3c10>}
>>> 

```

does this change propose that it would be impossible for `get_type_hints(X.A)` to return `class A` under any circumstances, even with the quotes?

JelleZijlstra:
Yes

zzzeek:
are you going to change the behavior of `get_type_hints()` ?    is this a 3.15 change?  is there a pep?  it should be apparent that this is an enormous backwards-incompatible change I hope?


edit: the pep is pep-749

zzzeek:
also is this change intended to take place regardless of whether a file is in pep-563 mode or pep-649 mode?

zzzeek:
existing libraries which use this pattern will have to use:

```
>>> _a_cls = A
>>> class Z:
...     def A(self) -> _a_cls: pass
...     
>>> typing.get_type_hints(Z.A)
{'return': <class '__main__.A'>}
```

CaselIT:
this would be on top of this breaking change in 3.13 -> 3.14
```
Python 3.13.9 | packaged by Anaconda, Inc. | (main, Oct 21 2025, 19:09:58) [MSC v.1929 64 bit (AMD64)] on win32
>>> class A: pass
...
>>> class F:
...     def A(self)->A: pass
...
>>> import typing
>>> typing.get_type_hints(F.A)
{'return': <class '__main__.A'>}
```
```
Python 3.14.3 | packaged by Anaconda, Inc. | (main, Feb 24 2026, 22:45:56) [MSC v.1942 64 bit (AMD64)] on win32
>>> class A: pass
...
>>> class F:
...     def A(self) -> A: pass
...
>>> import typing
>>> typing.get_type_hints(F.A)
{'return': <function F.A at 0x0000025520D22610>}
```

zzzeek:
I went to where I knew I'd find examples of this, in `datetime.pyi`, and indeed they are using an alias to work around the ambiguity:  https://github.com/python/typeshed/blob/main/stdlib/datetime.pyi#L296

jorenham (approved):
When reading this at first I was a bit worried that this would cause issues for NumPy, where builtin shadowing (e.g. `numpy.bool`) and class shadowing (`numpy.dtype` / `numpy.ndarray.dtype`) are pretty common things. However, this was recently fixed by @JelleZijlstra in https://github.com/numpy/numpy/pull/31951, so that's no longer an issue.

Anyway, this change reduces ambiguity while improving precision in a directly testable way. The main selling point of string annotations are forward references, so it seems natural for it to shadow *"as above, so below"*. The other explicit "no special casing" vibes I'm getting from this I also really like.

carljm:
@CaselIT 

> this would be on top of this breaking change in 3.13 -> 3.14

Not "on top of" - it's the same change. What this PR does is reflect that very same change in runtime semantics of annotations in the specified type checker behavior, so that the two are consistent.

@zzzeek 

> also is this change intended to take place regardless of whether a file is in [pep-563](https://peps.python.org/563) mode or [pep-649](https://peps.python.org/649) mode?

I think this is a really good question, and personally I would be open to specifying a more gradual transition path here, even though it would mean type checkers have to maintain more different behaviors at once.

Specifically, we could say that stringified annotations (whether explicitly strings, or implicitly stringified via `from __future__ import annotations`) would retain semantics compatible with the runtime behavior of `typing.get_type_hints` on strings (avoiding the pressure to change that runtime behavior in a backwards-incompatible way), while specifying that all annotations in Python 3.14+ use deferred-evaluation semantics with normal Python scoping rules, just like 3.14+ does at runtime.

rchen152:
I was struggling to wrap my head around the matrix of runtime behaviors involved here, so I put together a little table for myself. Copying it below in case it helps anyone else (all code snippets are in the body of a class `A`):

Code | `get_type_hints` in 3.12.10 | `get_type_hints` in 3.14.2
-- | -- | --
`str: str` | `{'str': <class 'str'>}` | `{'str': <class 'str'>}`
`str: "str"` | `{'str': <class 'str'>}` | `{'str': <class 'str'>}`
`x: int; def int(self): ...` | `{'x': <class 'int'>}` | `{'x': <function A.int>}`
`x: "int"; def int(self): ...` | `{'x': <function A.int>}` | `{'x': <function A.int>}`
`def int(self): ...; y: int` | `{'y': <function A.int>}` | `{'y': <function A.int>}`
`def int(self): ...; y: "int"` | `{'y': <function A.int>}` | `{'y': <function A.int>}`
`def int(self) -> int: ...` | `{'return': <class 'int'>}` | `{'return': <function A.int>}`
`def int(self) -> "int": ...` | `{'return': <class 'int'>}` | `{'return': <class 'int'>}`

@carljm If I understand correctly, you're proposing that all stringified annotations be resolved exactly as they are at runtime, while non-string annotations are resolved as they would be in 3.14? If so, then at least for the above examples, that seems to simplify down to "match 3.14 runtime behavior exactly, even where it's conceptually inconsistent," which I can get behind, even though it's a little unsatisfying.

CaselIT:
That table is really useful.
The last row highlights something that's kinda inconsistent in 3.14. I would be nicer if it did resolve to the function even there. At least for consistency sake

JelleZijlstra:
I'd rather not rely on the behavior of `get_type_hints()` too much, it's deliberately buggy for backwards compatibility.

AlexWaygood:
I fully agree, `get_type_hints` definitely shouldn't be used as any kind of a reference here -- it has very questionable behaviour even on the latest Python version

rchen152:
Is there a better reference than `typing.get_type_hints` we can use then? My (possibly mistaken) understanding was that `get_type_hints` is the closest thing we have to an "official" runtime interpretation of type hints and therefore the thing to look at if we're concerned about compatibility with runtime behavior and previous versions.

To be clear, I think that if we were starting from a clean slate, a simple and consistent interpretation of type annotations as proposed here would be ideal. But we're not, so I'm trying to understand what a compromise that still moves us in the right direction would look like.

jorenham:
> I'd rather not rely on the behavior of `get_type_hints()` too much, it's deliberately buggy for backwards compatibility.

Hmm this is also news to me. 

Then in that case I agree that we shouldn't build upon unstable grounds. But at the same time I also think it's a good idea to minimize the differences between the static- and runtime-typing worlds.
Changing perspective, maybe there's some way that we could "fix" `get_type_hints` (or perhaps introduce an alternative) so that it'll match the behavior of whatever we decide here in this PR (i.e. after this is merged)?

AlexWaygood:
The non-buggy modern alternative is `annotationlib.get_annotations`, which is already backported in `typing_extensions`. I don't think it's workable to change `get_type_hints`; its questionable behaviour is too deeply embedded

jorenham:
> The non-buggy modern alternative is `annotationlib.get_annotations`,

Ah perfect; that sounds like a good alternative to `typing.get_type_hints` that we could reference here instead.

JelleZijlstra:
Generally the reference should be what happens if you don't stringify the annotation in 3.14+. Annotations are now always lazily evaluated.

rchen152:
Perfect, thanks. I retried using `annotationlib.get_annotations` in 3.14 and `typing_extensions.get_annotations` in 3.12. [1] The matrix came out the same as it did for `typing.get_type_hints`.

If the intended reference is the non-stringified 3.14 results, then this PR matches the intended runtime behavior in 3.14 onward, which is great.

`def int(self) -> int: ...` still gives a different result from `def int(self) -> "int": ...` with `get_annotations(eval_str=True)`; is that intentional? That particular example seems to be the exact case in which the backwards compatibility implications of this change are the most concerning.

[1] using this code, if anyone wants to check my homework ;) I added `from __future__ import annotations` and `eval_str=True` to see how stringified annotations evaluate.
```
from annotationlib import get_annotations

class A:
  str: str

class B:
  x: int
  def int(self): ...
  y: int

class C:
  def int(self) -> int: ...

print(get_annotations(A))
print(get_annotations(B))
print(get_annotations(C.int))
```

carljm:
@rchen152 Yes, I think that's a good summary of my suggestion above: to match the runtime behavior of 3.14 exactly, assuming that `typing.get_type_hints` or `annotationlib.get_annotations(eval_str=True)` is used to access the annotations and resolve stringified ones. Let's call this proposal "compatible".

The alternative proposal on the table we could call "simple" -- that proposal is that _all_ annotations (including stringified ones) should be resolved the way they would resolve at runtime if they were non-stringified annotations on 3.14+.

To the extent that stringified annotations (mostly) eventually go away, the two proposals will converge, since they differ only in the handling of stringified annotations.

The advantage of "simple" is that there is a single consistent model of name resolution for type checkers to implement. The disadvantages are:

* To be compliant, type checkers have to outright break backwards compatibility for existing code using stringified annotations in these edge cases, as mentioned by some users above.
* Type checkers will resolve stringified annotations in a way that does not match the semantics used by the only tools the stdlib provides to resolve them at runtime. (Thus we can't really claim there is a "single consistent model" for users to understand.)

The "compat" model preserves backwards-compatibility and full consistency with runtime behavior, allowing users to migrate to the new name-resolution semantics on their own schedule (by moving from stringified annotations to non-stringified PEP 647/747 in 3.14+). The cost is more complexity in type checker implementations.

It increasingly seems to me that "compat" is the choice that better serves users of the type system.

(There is a third possible choice, which is to fully embrace the backwards-incompatible change by also changing the runtime behavior of `typing.get_type_hints`, but I don't see any appetite for that.)

davidhalter:
What's the consensus here between [Carl's](https://github.com/python/typing/pull/2144#issuecomment-5242395896) simple and compat proposal? I'm personally for "simple", because I don't like the complexity of `compat`. I'm aware of the runtime issues, but I think most typing users do care about static analysis and not runtime. It seems unexpected that changing `Foo[int]` to `"Foo[int]"` would change its meaning.

I would like to move forward with this, because currently it feels like there are 6 competing ways between type checkers how we see forward references and I'm happy to move on with the majority of the council here.

carljm:
I am happy to move forward with either, if authors of other type checkers are OK with a backwards-incompatible change to their handling of stringified annotation forward references in the ambiguous cases. ty already implements "simple", so that's less work for me :)

rchen152:
I prefer "simple", although I'd also be okay with "compat". Given the concerns Carl raised, I was going to try implementing it in Pyrefly to make sure I didn't run into any showstopping issues before offering an opinion, but if it's already how ty works, I'm much less worried on that front.

(I'll still implement it in Pyrefly as soon as I can and report back if I run into unanticipated problems, but I don't think we need to wait on that.)

davidhalter:
I ran the tests with all the type checkers and added notes.

davidhalter:
@carljm , I think you're the only one who has not ticked off https://github.com/python/typing-council/issues/51.

I guess that there won't be more feedback here, so I feel like if you're ok with this change, we should be able to merge.

conformance/tests/annotations_forward_refs.py
```diff
@@ -79,23 +78,21 @@ class ClassD:
 
     ClassF: "ClassF"  # E: circular reference
 
-    str: "str" = ""  # OK
+    str: "str" = ""  # E: circular reference
+
+    z: "int" = 0  # E: Refers to the local int function
```
carljm: Could we add positive assertions where both the shadowed name and the shadowing name are valid types? The changed shadowing cases all require errors, so they don't establish that we select the correct type when resolution should succeed. For example:

```python
class Token: ...

class C:
    def method(self, x: "Token") -> "Token":
        return x

    class Token: ...

assert_type(C().method(C.Token()), C.Token)
```

With the quotes removed, CPython 3.14.4 resolves both annotations to `C.Token`, as the proposed rule requires. Mypy 2.1.0, Pyright 1.1.410, and Zuban 0.9.2 instead select the global `Token` for the quoted example. Pyright and Zuban also select the global when `C.Token` precedes the method. Zuban currently passes this PR's test, so this is a concrete gap in what that pass establishes.

The equivalent case with a nested function and a later `class Token` in its enclosing function also selects the wrong global in all three checkers. Could we cover both class and enclosing-function scopes, with bindings before and after the annotation, and assert the resulting types? A method whose own name shadows its annotation, such as `def int(self) -> "int"`, would also be useful: that should resolve to the method and be rejected.

conformance/tests/annotations_forward_refs.py
```diff
@@ -79,23 +78,21 @@ class ClassD:
 
     ClassF: "ClassF"  # E: circular reference
 
-    str: "str" = ""  # OK
+    str: "str" = ""  # E: circular reference
```
carljm: Could we pair this with an annotation-only declaration and assert its type?

```python
class C:
    str: "str"       # OK: resolves to builtin str

class D:
    str: "str" = ""  # Error: assigned value shadows str

def check(c: C) -> None:
    assert_type(c.str, str)
```

An annotation alone creates no runtime class attribute. CPython 3.14.4 resolves the unquoted first example to builtin `str`, but Zuban 0.9.2 rejects the quoted version as an undefined name even though it passes our updated test. The existing `ClassC: "ClassC"` covers module fallback, but doesn't catch this builtin fallback failure.

There is also a subtler three-scope case that I think deserves coverage:

```python
class Token: ...

def outer():
    class Token: ...

    class C:
        value: "Token"
        Token: "type[str]"

    return C
```

In CPython 3.14.4's unquoted equivalent, the annotation for `value` in `C` resolves to the module-level `Token`, skipping `outer.Token`. Removing the annotation-only declaration of `C.Token` makes it resolve to `outer.Token` instead. I verified the same distinction for a method parameter annotation. Could we make those fallback rules explicit in the tests as well?

docs/spec/annotations.rst
```diff
@@ -223,32 +223,14 @@ String annotations
...
-
-Moreover, the expression should be parseable as a valid type hint, i.e.,
-it is constrained by the rules from :ref:`the expression grammar <expression-grammar>`.
+The string literal should contain a syntactically valid Python expression
+(i.e., ``compile(lit, '', 'eval')`` should succeed) that is a valid
+:term:`annotation expression`. Regardless of the Python version used, names
+within the expression are looked up in the same way as they would be looked up
+at runtime in Python 3.14 and higher if the annotation was not enclosed in a
+string literal. Thus, name lookup follows general rules (e.g., the current
+function, class, or module scope first, and the builtin scope last), but names
```
carljm: Could we include quoted annotations that distinguish a type-parameter scope from a same-named global?

```python
class T: ...

class C[T]:
    def identity(self, value: "T") -> "T":
        return value

assert_type(C[int]().identity(1), int)
```

Zuban 0.9.2 resolves these annotations to the global class `T`, despite passing this PR's forward-reference test. CPython 3.14.4's unquoted equivalent resolves to the type parameter, and ty 0.0.75 passes the assertion. Our existing generic-scoping tests don't exercise this quoted shadowing case. Could we cover both class and method type parameters, so a checker can't pass while using a separate string resolver that loses the type-parameter scope?

docs/spec/annotations.rst
```diff
@@ -223,32 +223,14 @@ String annotations
...
-same namespaces in which default arguments to the same function would
-be evaluated.
-
-Moreover, the expression should be parseable as a valid type hint, i.e.,
-it is constrained by the rules from :ref:`the expression grammar <expression-grammar>`.
+The string literal should contain a syntactically valid Python expression
+(i.e., ``compile(lit, '', 'eval')`` should succeed) that is a valid
+:term:`annotation expression`. Regardless of the Python version used, names
+within the expression are looked up in the same way as they would be looked up
+at runtime in Python 3.14 and higher if the annotation was not enclosed in a
```
carljm: How should this rule apply to local-variable annotations, which have no runtime evaluation to use as a reference?

```python
class T: ...

def f() -> None:
    x: "T"
    class T: ...
    x = T()
```

[PEP 649 explicitly excludes local-variable annotations from its runtime machinery](https://peps.python.org/pep-0649/#annotations-on-local-variables-inside-functions). I would expect `x` to use the function-local `T`, including its later declaration, but the reference to runtime lookup doesn't directly specify that behavior. Mypy 2.1.0, Pyright 1.1.410, and Zuban 0.9.2 choose global `T` here, while ty 0.0.75 chooses local `T`.

Could we state the static lookup rule for these annotations explicitly and add a test? Contrasting a function's signature, a local-variable annotation in its body, and a nested function's signature would help pin down which function scope is meant.

docs/spec/annotations.rst
```diff
@@ -223,32 +223,14 @@ String annotations
...
+string literal. Thus, name lookup follows general rules (e.g., the current
```
carljm: Could we also add controls for where visibility of the enclosing class stops? For example, given both global `Token` and `Outer.Token`, an annotation in `Outer.Inner.method` should skip `Outer` and resolve global `Token`; so should an annotation on a function nested inside `Outer.method`. Without the global definition, those names should be unresolved.

I verified these boundaries with unquoted annotations on CPython 3.14.4. They would complement the positive method-signature cases and catch an implementation that searches every enclosing class indiscriminately. This could be follow-up coverage if we want to keep this change smaller.

conformance/tests/annotations_forward_refs.py
```diff
@@ -57,10 +57,9 @@ def invalid_annotations(
     pass
 
 
-# > It should evaluate without errors once the module has been fully loaded.
-# > The local and global namespace in which it is evaluated should be the same
-# > namespaces in which default arguments to the same function would be evaluated.
-
+# > Names within the expression are looked up in the same way as they would be
+# > looked up at runtime in Python 3.14 and higher if the annotation was not
+# > enclosed in a string literal.
```
carljm: Would it make sense to add a control distinguishing lexical class names from inherited attributes?

```python
class Token: ...

class Base:
    class Token: ...

class Derived(Base):
    value: "Token"
```

I would expect `Derived.value` to have the module-level `Token` type: looking up a name in the annotation's class namespace is different from looking up an attribute through the class's MRO. An `assert_type` check here, and the equivalent method-signature case, would guard that boundary. This could also be follow-up coverage.

conformance/tests/annotations_forward_refs.py
```diff
@@ -79,23 +78,21 @@ class ClassD:
...
+    z: "int" = 0  # E: Refers to the local int function
 
     def int(self) -> None:  # OK
         ...
 
-    x: "int" = 0  # OK
+    y: int = 0  # E: Refers to the local int function, which isn't a legal type expression
 
-    y: int = 0  # E: Refers to local int, which isn't a legal type expression
+    x: "int" = 0  # E: Refers to a local int function
```
carljm: Could we exercise this same shadowing with both whole and partial quoting, such as `value: "list[int]"` and `value: list["int"]`, before and after `def int`? Both forms should resolve the quoted `int` to the class method and be rejected.

The existing `func1` covers both forms only for an unambiguous module name. It doesn't establish that parsing a nested forward reference preserves the same scope and lookup timing as parsing a whole quoted annotation. This seems useful as additional coverage, even if we add it separately.

docs/spec/annotations.rst
```diff
@@ -223,32 +223,14 @@ String annotations
...
-should evaluate without errors once the module has been fully loaded.
-The local and global namespace in which it is evaluated should be the
-same namespaces in which default arguments to the same function would
-be evaluated.
-
-Moreover, the expression should be parseable as a valid type hint, i.e.,
-it is constrained by the rules from :ref:`the expression grammar <expression-grammar>`.
+The string literal should contain a syntactically valid Python expression
+(i.e., ``compile(lit, '', 'eval')`` should succeed) that is a valid
+:term:`annotation expression`. Regardless of the Python version used, names
```
carljm: Could we explicitly cover `from __future__ import annotations` as well? The discussion's "simple" model includes implicitly stringified annotations, while this paragraph describes an expression enclosed in a string literal.

A companion test with the future import could repeat the before/after class-shadowing cases using bare annotations and compare them with explicitly quoted annotations. It would help make clear whether both are intended to use the same lookup rules on older target versions as well as 3.14, and catch a checker that updates its explicit-string resolver but leaves its future-annotations path unchanged. This could be follow-up coverage, but I think the intended semantics should be clear.

conformance/results/mypy/annotations_forward_refs.toml
```diff
@@ -2,7 +2,7 @@ conformant = "Partial"
 notes = """
 Does not report error for a forward reference that is not enclosed in quotes.
 Does not report error for use of quoted type with `|` operator (runtime error).
-Incorrectly generates error for quoted type defined in class scope.
+Resolves forward references in type annotations at the point of definition instead of end-of-scope
```
carljm: Could we qualify this? Mypy's behavior depends on the annotation context, rather than simply resolving all forward references at the point of definition. For example, with mypy 2.1.0:

```python
x: "int" = 0                 # Accepted: builtin int

def f(x: "int") -> None: ... # Error: resolves to the function below

def int() -> None: ...
```

So this behavior also occurs for module-level variable annotations, while a module-level function signature resolves the later global binding. Mypy also resolves later class/function-local definitions when there isn't an already-visible competing name.

Maybe something like "In some annotation contexts, resolves quoted names to builtin or outer-scope bindings even when later definitions shadow them" would be more accurate? The same blanket sentence appears in the other checker results, so I think those should describe their observed discrepancies individually as well.

conformance/results/ty/annotations_forward_refs.toml
```diff
@@ -1,12 +1,10 @@
 conformance_automated = "Fail"
 conformant = "Partial"
 notes = """
-Resolves references in type annotations as referring to end-of-scope types (<https://discuss.python.org/t/annotation-string-references-in-class-scope-in-conformance-tests/105439>, <https://github.com/python/typing/pull/2144>)
+Fails to reject `str: "str"` forward references
```
carljm: Could we include the initializer and class context in this description? `str: "str"` without an assignment is valid: an annotation alone creates no class attribute, so lookup falls back to builtin `str`. The failing test is specifically `str: "str" = ""`, where the assigned value shadows the builtin.

Maybe "Fails to reject the initialized class attribute `str: \"str\" = \"\"`, whose value shadows the builtin type" would capture the actual discrepancy? ty 0.0.75 accepts both forms, but only the initialized form should be rejected under the proposed rule.

conformance/results/pyright/annotations_forward_refs.toml
```diff
@@ -1,4 +1,7 @@
-conformant = "Pass"
+conformant = "Partial"
+notes = """
+Resolves forward references in type annotations at the point of definition instead of end-of-scope
```
carljm: Could we describe the class-scope lookup discrepancy directly here? "At the point of definition" doesn't explain the recorded output: Pyright also accepts `x: "int"` on line 90, after `def int`, while rejecting the unquoted `y: int` on line 88.

A focused probe with Pyright 1.1.410 likewise selects a global class `Token` for a quoted class-body annotation both before and after a nested `class Token`. With no same-named global, it can resolve a later nested class, so this isn't a general inability to resolve forward references either.

Maybe "Resolves quoted class-body annotations to builtin or outer-scope bindings even when class attributes shadow them" would describe the observed behavior more accurately?

conformance/results/pyrefly/annotations_forward_refs.toml
```diff
@@ -1,12 +1,11 @@
 conformant = "Partial"
 notes = """
-Types in quotes incorrectly refer to shadowing class member.
-Does not reject some type forms that require quotes.
+Resolves forward references in type annotations at the point of definition instead of end-of-scope
```
carljm: Could we narrow this to the shadowing cases? Pyrefly does resolve references to later definitions when there is no already-visible competing binding; for example, a nested function's quoted signature can refer to a later class in its enclosing function.

The missing errors here are specifically `str: "str" = ""` and `z: "int"` before the class's `int` method; the quoted reference after that method is correctly rejected. I confirmed these distinctions with the pinned Pyrefly 1.3.0.dev1.

Maybe "Does not report errors for quoted class-body annotations shadowed by the annotated assignment or a later class attribute" would describe those discrepancies without suggesting that forward references in general use definition-time lookup?

conformance/results/pycroscope/annotations_forward_refs.toml
```diff
@@ -1,19 +1,20 @@
 conformant = "Partial"
 notes = """
 Fails to reject `"x" | int` annotations that fail at runtime.
+Resolves forward references in type annotations at the point of definition instead of end-of-scope
```
carljm: Could we make this specific to the class-scope shadowing failures? Pycroscope 0.5.0 does use later bindings in some contexts: in an importable module, both `value: "int" = 0` and a function parameter annotated with `"int"` report an invalid annotation when followed by `def int() -> None: ...`. An ordinary quoted reference to a later global class also succeeds.

Perhaps "Does not account for class-scope shadowing in some quoted annotations" would describe the missing errors on `ClassD.str` and `ClassD.z` more precisely? The separate note about rejecting valid quoted annotations matches this fixture's output, but that behavior depends on the module failing to import, so it doesn't establish a general definition-time lookup rule.

Could we also regenerate the notes in `results.html` after adjusting the checker-specific descriptions?

conformance/tests/annotations_forward_refs.py
```diff
@@ -57,10 +57,9 @@ def invalid_annotations(
     pass
 
 
-# > It should evaluate without errors once the module has been fully loaded.
-# > The local and global namespace in which it is evaluated should be the same
-# > namespaces in which default arguments to the same function would be evaluated.
-
+# > Names within the expression are looked up in the same way as they would be
```
carljm: Could we add coverage along these lines? The patch checks successful namespace selection as well as invalid type expressions: module and class shadowing, function scopes, annotation-only fallback, nested-class and inheritance boundaries, type parameters, partial quoting, and future annotations.

It also makes two assumptions explicit in the spec: local-variable annotations use the function body's lexical scope, including later bindings; implicitly stringified annotations use the same lookup rule as explicit strings.

I checked 57 corresponding lookups on CPython 3.14.4 with unquoted annotations and no future import, and ran all six checker versions pinned by this PR. The added tests expose further disagreements; ty passes the new scope, function, and generic fixtures, with its existing self-referential initialized-attribute issue also appearing in the future-import fixture. Result snapshots can be regenerated once we agree on these expectations.

<details>
<summary>Proposed test and spec diff against 927aa611</summary>

```diff
diff --git a/conformance/tests/annotations_forward_refs.py b/conformance/tests/annotations_forward_refs.py
index cdc92a2f4d..a54d7a0138 100644
--- a/conformance/tests/annotations_forward_refs.py
+++ b/conformance/tests/annotations_forward_refs.py
@@ -7,7 +7,7 @@ Tests the handling of forward references in type annotations.
 
 
 import types
-from typing import assert_type, Any
+from typing import assert_type
 
 
 def func1(
@@ -76,9 +76,9 @@ class ClassC:
 class ClassD:
     ClassC: "ClassC"  # OK
 
-    ClassF: "ClassF"  # E: circular reference
+    ClassF: "ClassF"  # E: No bound name ClassF is available
 
-    str: "str" = ""  # E: circular reference
+    str: "str" = ""  # E: Refers to the assigned string, which is not a type
 
     z: "int" = 0  # E: Refers to the local int function
 
diff --git a/docs/spec/annotations.rst b/docs/spec/annotations.rst
index 7ff8e0f70f..197b86d40d 100644
--- a/docs/spec/annotations.rst
+++ b/docs/spec/annotations.rst
@@ -227,11 +227,18 @@ The string literal should contain a syntactically valid Python expression
 (i.e., ``compile(lit, '', 'eval')`` should succeed) that is a valid
 :term:`annotation expression`. Regardless of the Python version used, names
 within the expression are looked up in the same way as they would be looked up
-at runtime in Python 3.14 and higher if the annotation was not enclosed in a
-string literal. Thus, name lookup follows general rules (e.g., the current
+at runtime in Python 3.14 and higher without ``from __future__ import annotations``,
+if the annotation was not enclosed in a string literal.
+Thus, name lookup follows general rules (e.g., the current
 function, class, or module scope first, and the builtin scope last), but names
 defined later within the same scope can be used in an earlier annotation.
 
+This lookup rule also applies to annotations implicitly stringified by
+``from __future__ import annotations``. Local-variable annotations are not
+evaluated at runtime; their names are resolved statically using the normal
+lexical scope rules for the function body, including bindings that follow
+the annotation.
+
 If a triple quote is used, the string should be parsed as though it is
 implicitly surrounded by parentheses. This allows newline characters to be
 used within the string literal::
diff --git a/conformance/tests/annotations_forward_refs_functions.py b/conformance/tests/annotations_forward_refs_functions.py
new file mode 100644
index 0000000000..1dc7475991
--- /dev/null
+++ b/conformance/tests/annotations_forward_refs_functions.py
@@ -0,0 +1,112 @@
+"""
+Tests forward-reference lookup in function signatures and local annotations.
+"""
+
+from typing import assert_type
+
+
+class Token:
+    ...
+
+
+# A function's parameter names and body-local names are not in scope in its
+# signature. Both annotations below refer to the module-level Token.
+def parameter_scope(Token: int, value: "Token") -> "Token":
+    return value
+
+
+def body_scope(value: "Token") -> "Token":
+    class Token:
+        ...
+
+    return value
+
+
+assert_type(parameter_scope(0, Token()), Token)
+assert_type(body_scope(Token()), Token)
+
+
+# A nested function's signature does see its enclosing function's locals,
+# including definitions after the signature. The enclosing Token wins over
+# the same-named module-level type in either order.
+def enclosing_before() -> None:
+    class Token:
+        ...
+
+    def identity(value: "Token") -> "Token":
+        assert_type(value, Token)
+        return value
+
+    assert_type(identity(Token()), Token)
+
+
+def enclosing_after() -> None:
+    def identity(value: "Token") -> "Token":
+        assert_type(value, Token)
+        return value
+
+    class Token:
+        ...
+
+    assert_type(identity(Token()), Token)
+
+
+# Forward lookup also works when there is no competing module-level binding.
+def enclosing_without_global() -> None:
+    def identity(value: "LocalToken") -> "LocalToken":
+        assert_type(value, LocalToken)
+        return value
+
+    class LocalToken:
+        ...
+
+    assert_type(identity(LocalToken()), LocalToken)
+
+
+# Lookup stops at the nearest enclosing function that defines the name.
+def nearest_enclosing() -> None:
+    class Token:
+        ...
+
+    def middle() -> None:
+        def identity(value: "Token") -> "Token":
+            assert_type(value, Token)
+            return value
+
+        class Token:
+            ...
+
+        assert_type(identity(Token()), Token)
+
+    middle()
+
+
+# Local-variable annotations use the function's own scope, unlike its
+# signature. They are not evaluated at runtime; these tests specify the
+# corresponding static lookup rule, including later local definitions.
+def local_before() -> None:
+    class Token:
+        ...
+
+    value: "Token" = Token()
+    assert_type(value, Token)
+
+
+def local_after() -> None:
+    value: "Token"
+
+    class Token:
+        ...
+
+    value = Token()
+    assert_type(value, Token)
+
+
+def local_without_global() -> None:
+    value: "LocalValue"
+
+    class LocalValue:
+        ...
+
+    value = LocalValue()
+    assert_type(value, LocalValue)
diff --git a/conformance/tests/annotations_forward_refs_future.py b/conformance/tests/annotations_forward_refs_future.py
new file mode 100644
index 0000000000..6688cd174e
--- /dev/null
+++ b/conformance/tests/annotations_forward_refs_future.py
@@ -0,0 +1,98 @@
+"""
+Tests forward-reference lookup with ``from __future__ import annotations``.
+"""
+
+from __future__ import annotations
+
+from typing import assert_type
+
+
+# Implicitly stringified annotations and explicit strings use the same lookup
+# rules as unquoted, deferred annotations in Python 3.14.
+
+class Token:
+    ...
+
+
+class ClassScope:
+    def bare_before(self, value: Token) -> Token:
+        return value
+
+    def quoted_before(self, value: "Token") -> "Token":
+        return value
+
+    class Token:
+        ...
+
+    def bare_after(self, value: Token) -> Token:
+        return value
+
+    def quoted_after(self, value: "Token") -> "Token":
+        return value
+
+
+# All four signatures refer to the nested class, including those that precede it.
+class_scope = ClassScope()
+class_token = ClassScope.Token()
+assert_type(class_scope.bare_before(class_token), ClassScope.Token)
+assert_type(class_scope.quoted_before(class_token), ClassScope.Token)
+assert_type(class_scope.bare_after(class_token), ClassScope.Token)
+assert_type(class_scope.quoted_after(class_token), ClassScope.Token)
+
+
+# An annotation without an initializer creates no class attribute that could
+# shadow the builtin. Adding an initializer changes the lookup result.
+class AnnotationOnly:
+    str: str
+    bytes: "bytes"
+
+
+def check_annotation_only(value: AnnotationOnly) -> None:
+    assert_type(value.str, str)
+    assert_type(value.bytes, bytes)
+
+
+class Initialized:
+    str: str = ""  # E: The assigned class attribute is not a valid type
+    bytes: "bytes" = b""  # E: The assigned class attribute is not a valid type
+
+
+# Non-type class members shadow builtins before and after their definitions,
+# including within the annotation of the member itself.
+class InvalidClassScope:
+    bare_before: int  # E: Refers to the class-local method
+    quoted_before: "int"  # E: Refers to the class-local method
+
+    def int(self) -> int:  # E: Refers to the method itself
+        ...
+
+    bare_after: int  # E: Refers to the class-local method
+    quoted_after: "int"  # E: Refers to the class-local method
+
+    def str(self) -> "str":  # E: Refers to the method itself
+        ...
+
+
+# Enclosing function locals also take precedence over globals, even when the
+# local class is defined after the nested function's annotations.
+def enclosing() -> None:
+    def bare_before(value: Token) -> Token:
+        return value
+
+    def quoted_before(value: "Token") -> "Token":
+        return value
+
+    class Token:
+        ...
+
+    def bare_after(value: Token) -> Token:
+        return value
+
+    def quoted_after(value: "Token") -> "Token":
+        return value
+
+    token = Token()
+    assert_type(bare_before(token), Token)
+    assert_type(quoted_before(token), Token)
+    assert_type(bare_after(token), Token)
+    assert_type(quoted_after(token), Token)
diff --git a/conformance/tests/annotations_forward_refs_generics.py b/conformance/tests/annotations_forward_refs_generics.py
new file mode 100644
index 0000000000..d06da53b92
--- /dev/null
+++ b/conformance/tests/annotations_forward_refs_generics.py
@@ -0,0 +1,86 @@
+"""
+Tests that quoted annotations respect type parameter scopes.
+"""
+
+from typing import assert_type
+
+
+# A same-named global type makes choosing the wrong scope observable even when
+# both possible resolutions would be valid type expressions.
+class T:
+    pass
+
+
+class U:
+    pass
+
+
+def identity[T](value: "T") -> "T":
+    return value
+
+
+def check_identity(integer: int, string: str) -> None:
+    assert_type(identity(integer), int)
+    assert_type(identity(string), str)
+
+
+class Box[T]:
+    value: "T"
+
+    def __init__(self, value: "T") -> None:
+        self.value = value
+
+    def identity(self, value: "T") -> "T":
+        return value
+
+    def unquoted(self, value: T) -> T:
+        return value
+
+    # The method's own type parameter takes precedence over the global U;
+    # references to T still select the enclosing class's type parameter.
+    def pair[U](self, value: "U") -> "tuple[T, U]":
+        return self.value, value
+
+    # Whole and partial quoting must use the same type parameter scopes.
+    def lists[U](self, first: "list[T]", second: list["U"]) -> "tuple[list[T], list[U]]":
+        return first, second
+
+    # A method's type parameter does not leak into a sibling method.
+    def global_type(self, value: "U") -> "U":
+        return value
+
+
+def check_box(integer: int, string: str, integers: list[int], strings: list[str]) -> None:
+    box = Box[int](integer)
+    assert_type(box.value, int)
+    assert_type(box.identity(integer), int)
+    assert_type(box.unquoted(integer), int)
+    assert_type(box.pair(string), tuple[int, str])
+    assert_type(box.lists(integers, strings), tuple[list[int], list[str]])
+    assert_type(box.global_type(U()), U)
+
+
+# A nested function can capture the enclosing function's type parameter.
+def outer[T](value: T, string: str) -> "T":
+    def inner(item: "T") -> "T":
+        return item
+
+    # Its own type parameter and the enclosing one remain distinct.
+    def pair[U](first: "T", second: "U") -> "tuple[T, U]":
+        return first, second
+
+    assert_type(pair(value, string), tuple[T, str])
+    return inner(value)
+
+
+def check_outer(integer: int, string: str) -> None:
+    assert_type(outer(integer, string), int)
+    assert_type(outer(string, string), str)
+
+
+# Generic declarations do not change resolution in the surrounding module.
+def global_type(value: "T") -> "T":
+    return value
+
+
+assert_type(global_type(T()), T)
diff --git a/conformance/tests/annotations_forward_refs_scopes.py b/conformance/tests/annotations_forward_refs_scopes.py
new file mode 100644
index 0000000000..0482829e27
--- /dev/null
+++ b/conformance/tests/annotations_forward_refs_scopes.py
@@ -0,0 +1,192 @@
+"""
+Tests namespace selection and shadowing in quoted annotations.
+"""
+
+import builtins
+from typing import assert_type
+
+
+# A later module binding shadows a builtin in both variable and function annotations.
+module_value: "bytes"
+
+
+def module_before(value: "bytes") -> "bytes":
+    return value
+
+
+class bytes:
+    pass
+
+
+module_value = bytes()
+assert_type(module_value, bytes)
+assert_type(module_before(bytes()), bytes)
+
+
+class Token:
+    pass
+
+
+GlobalToken = Token
+
+
+# Class annotations and method signatures see the class's final namespace.
+class ClassScope:
+    before: "Token"
+
+    def method_before(self, value: "Token") -> "Token":
+        return value
+
+    class Token:
+        pass
+
+    after: "Token"
+
+    def method_after(self, value: "Token") -> "Token":
+        return value
+
+    def parameter_name(self, Token: "Token") -> "Token":
+        return Token
+
+
+def check_class_scope(value: ClassScope) -> None:
+    assert_type(value.before, ClassScope.Token)
+    assert_type(value.after, ClassScope.Token)
+    assert_type(value.method_before(ClassScope.Token()), ClassScope.Token)
+    assert_type(value.method_after(ClassScope.Token()), ClassScope.Token)
+    assert_type(value.parameter_name(ClassScope.Token()), ClassScope.Token)
+
+
+# Forward references also work when no outer binding competes with the local one.
+class NoOuterBinding:
+    value: "Local"
+
+    class Local:
+        pass
+
+
+def check_no_outer_binding(value: NoOuterBinding) -> None:
+    assert_type(value.value, NoOuterBinding.Local)
+
+
+# An annotation-only class name falls back to the global or builtin namespace.
+class AnnotationOnly:
+    str: "str"
+    Token: "Token"
+
+
+def check_annotation_only(value: AnnotationOnly) -> None:
+    assert_type(value.str, builtins.str)
+    assert_type(value.Token, GlobalToken)
+
+
+# A later assignment creates a class binding even when separated from its annotation.
+class SplitAssignment:
+    str: "str"  # E: Refers to the assigned string, which is not a type
+    str = ""
+
+
+# A method's own binding is visible to its quoted signature.
+class OwnMethodName:
+    def int(self) -> "int":  # E: Refers to this method, which is not a type
+        raise NotImplementedError
+
+
+# Whole and partial quotation preserve lookup of the quoted name.
+class PartialQuotes:
+    whole_before: "list[int]"  # E: Refers to the class's int method
+    partial_before: list["int"]  # E: Refers to the class's int method
+
+    def int(self) -> None:
+        pass
+
+    whole_after: "list[int]"  # E: Refers to the class's int method
+    partial_after: list["int"]  # E: Refers to the class's int method
+
+
+# Only the immediately enclosing class namespace is visible to an annotation.
+class Outer:
+    class Token:
+        pass
+
+    class Inner:
+        value: "Token"
+
+        def method(self, value: "Token") -> "Token":
+            return value
+
+    class InnerWithBinding:
+        value: "Token"
+
+        class Token:
+            pass
+
+    def method(self) -> None:
+        def nested(value: "Token") -> "Token":
+            return value
+
+        assert_type(nested(GlobalToken()), GlobalToken)
+
+
+def check_nested_class(value: Outer.Inner) -> None:
+    assert_type(value.value, GlobalToken)
+    assert_type(value.method(GlobalToken()), GlobalToken)
+
+
+def check_nested_class_binding(value: Outer.InnerWithBinding) -> None:
+    assert_type(value.value, Outer.InnerWithBinding.Token)
+
+
+class OuterOnly:
+    class Hidden:
+        pass
+
+    class Inner:
+        value: "Hidden"  # E: The outer class namespace is not visible here
+
+    def method(self) -> None:
+        def nested(value: "Hidden") -> None:  # E: The class namespace is not visible here
+            pass
+
+
+# Inherited attributes do not participate in lexical name lookup.
+class Base:
+    class Token:
+        pass
+
+
+class Derived(Base):
+    value: "Token"
+
+    def method(self, value: "Token") -> "Token":
+        return value
+
+
+def check_inherited(value: Derived) -> None:
+    assert_type(value.value, GlobalToken)
+    assert_type(value.method(GlobalToken()), GlobalToken)
+
+
+# An annotation-only class declaration changes closure capture even without a value.
+def class_fallback() -> None:
+    class Token:
+        pass
+
+    class Captured:
+        value: "Token"
+
+        def method(self, value: "Token") -> "Token":
+            return value
+
+    class Declared:
+        value: "Token"
+        Token: "type[builtins.str]"
+
+        def method(self, value: "Token") -> "Token":
+            return value
+
+    def check(captured: Captured, declared: Declared) -> None:
+        assert_type(captured.value, Token)
+        assert_type(captured.method(Token()), Token)
+        assert_type(declared.value, GlobalToken)
+        assert_type(declared.method(GlobalToken()), GlobalToken)
```

</details>

docs/spec/annotations.rst
```diff
@@ -223,32 +223,14 @@ String annotations
...
+within the expression are looked up in the same way as they would be looked up
+at runtime in Python 3.14 and higher if the annotation was not enclosed in a
+string literal. Thus, name lookup follows general rules (e.g., the current
+function, class, or module scope first, and the builtin scope last), but names
+defined later within the same scope can be used in an earlier annotation.
```
carljm, line 233: I think we should explicitly specify that the same rules apply to annotations implicitly stringified by `from __future__ import annotations`. And this should be tested in the conformance suite, too.

conformance/tests/annotations_forward_refs.py
carljm, line 60: I think another case we should probably test is type parameter scopes, e.g.

```py
class T: ...

class C[T]:
    def identity(self, value: "T") -> "T":
        return value

assert_type(C[int]().identity(1), int)
```

The quoted `"T"` annotations here should refer to the class typevar, not the global class `T`.

conformance/tests/annotations_forward_refs.py
carljm:
```suggestion
    z: "int" = 0  # E: Refers to the local int function, which isn't a valid type
```

conformance/tests/annotations_forward_refs.py
```diff
@@ -79,23 +78,21 @@ class ClassD:
...
-    str: "str" = ""  # OK
+    str: "str" = ""  # E: circular reference
+
+    z: "int" = 0  # E: Refers to the local int function
 
     def int(self) -> None:  # OK
         ...
 
-    x: "int" = 0  # OK
+    y: int = 0  # E: Refers to the local int function, which isn't a legal type expression
```
carljm:
```suggestion
    y: int = 0  # E: Refers to the local int function, which isn't a valid type
```

conformance/tests/annotations_forward_refs.py
carljm:
```suggestion
    x: "int" = 0  # E: Refers to the local int function, which isn't a valid type
```

conformance/tests/annotations_forward_refs.py
carljm: I think we should also have at least one example of a _valid_ forward reference -- all the below examples now error.

conformance/results/mypy/annotations_forward_refs.toml
carljm: Several type checkers have this note, but I don't think it's really accurate as a blanket statement for any of them; each type checker has subtleties about how specific cases are handled. I think ideally these notes would be more specific to the precise test case(s) actually failed by the checker.

carljm (commented):
Did another review pass on this.

conformance/tests/annotations_forward_refs.py
davidhalter: Added this.

conformance/tests/annotations_forward_refs.py
davidhalter: I added a few. The only questionable one might be Line 83, because that's a legitimate failure now pre-Python3.14. I can also make it a `# E?  This will be a mandatory error in the future, once pre-3.14 Python is end-of-life`. Let me know what you think.

conformance/results/mypy/annotations_forward_refs.toml
davidhalter: I have tried to improve the notes slightly. To me it feels like

```
Resolves forward references in type annotations in classes at the point of definition instead of end-of-scope
```

Represents Mypy, Pyroscope and Pyrefly pretty well. They all have the same issues.

docs/spec/annotations.rst
davidhalter: I added it to the spec. Please review. I will copy the `tests/annotations_forward_refs.py` to `tests/annotations_forward_refs_future.py` and add notes once we agree on the rest in this PR.

docs/spec/annotations.rst
```diff
@@ -308,6 +290,14 @@ _module_._class_ name::
     from models.a import A
     from models.b import B
 
+from __future__ import annotations
+----------------------------------
+
+The presence of the import `from __future__ import annotations` must not
+influence type checking.  Annotations must be resolved in the exact same way as
+if the import was not present.
```
carljm, line 298: Sorry, I should have been clearer in my last comment. I do not believe that this is the equivalence we should specify. Part of the point of using `from __future__ import annotations` is to enable forward references in type annotations, since the annotations are stringified at runtime and won't error. I think the equivalence that we should specify (and test) is that `from __future__ import annotations` causes all annotations to behave the same way as if they were explicitly stringified, using end-of-scope resolution.

For a pre-3.14 target, this should work with `from __future__ import annotations` but fail without it:

```python
class C:
    x: B
    
    class B: ...
```

conformance/tests/annotations_forward_refs.py
```diff
@@ -75,25 +74,52 @@ class ClassC:
...
 
 class ClassD:
-    ClassC: "ClassC"  # OK
+    # OK
+    simple_attr: ClassA
+    ClassB: ClassB
+    ClassC: "ClassC"
+    bytes_direct: bytes
+    bytes: "bytes"
+    inner1: ClassInner
```
carljm: I agree with your suggestion to use `# E?` here. This unquoted reference raises `NameError` on Python 3.12, which the suite currently targets, and works on 3.14. So a type checker erroring here under 3.12 is behaving correctly, IMO. It's the same version-dependent situation as the existing cases at lines 22, 23, and 65.

conformance/results/mypy/annotations_forward_refs.toml
```diff
@@ -2,7 +2,14 @@ conformant = "Partial"
 notes = """
 Does not report error for a forward reference that is not enclosed in quotes.
 Does not report error for use of quoted type with `|` operator (runtime error).
-Incorrectly generates error for quoted type defined in class scope.
+Resolves forward references in type annotations in classes at the point of definition instead of end-of-scope
```
carljm: I think this still overgeneralizes slightly: mypy accepts the new `inner2: "ClassInner"` before `ClassInner` is defined, so it does resolve some class forward references to later definitions.

I think we could literally just add the word "some" and avoid inaccuracy here, without needing to precisely describe the failed cases.

```suggestion
Resolves some forward references in type annotations in classes at the point of definition instead of end-of-scope
```

conformance/tests/annotations_forward_refs.py
```diff
@@ -75,25 +74,52 @@ class ClassC:
...
+    inner2: "ClassInner"
+
+    class ClassInner:
+        ...
+
+    inner_after1: ClassInner
+    inner_after2: "ClassInner"
```
carljm, line 90: I would add `assert_type` checks for these valid attributes, especially `bytes` and the nested-class references. At the moment, a checker could silently resolve them to `Any` and still pass.

For example:

```python
def check_valid_attributes(d: ClassD) -> None:
    assert_type(d.bytes, bytes)
    assert_type(d.inner2, ClassD.ClassInner)
    assert_type(d.inner_after2, ClassD.ClassInner)
```

davidhalter: Good point! Changed.

docs/spec/annotations.rst
davidhalter: So this would only change the resolution for pre-3.14 targets? I assume that would just mean that the tests marked with `# E?: Runtime error prior to 3.14: requires quotes` must not be errors I assume?

Sorry for asking for clarification again. My draft for the spec would be:

```
from __future__ import annotations
----------------------------------

If `from __future__ import annotations` is present in a file, annotations are always resolved at the end of the scope, as they are in Python 3.14 and later.
```

What do you think?

carljm: > So this would only change the resolution for pre-3.14 targets?

Yes, that's right, and I think your draft text looks reasonable. I'd slightly prefer referring directly to the string-annotation rules we just specified:

"For purposes of type checking, annotations in a file containing from `__future__ import annotations` must use the same name-resolution rules as explicitly stringified annotations, regardless of the target Python version."

This makes the equivalence explicit and avoids implying that Python 3.14 literally evaluates annotations at the end of the scope.

> I assume that would just mean that the tests marked with `# E?: Runtime error prior to 3.14: requires quotes` must not be errors I assume?

Yes, in the `from __future__ import annotations` copy of those tests, those should become required successes.

davidhalter: I did pretty much all of that. Note that I did not copy all of the tests from `annotations_forward_refs.py`, but only the relevant tests.

conformance/tests/annotations_future_import.py
```diff
@@ -0,0 +1,45 @@
...
+forward1: ClassA
+forward2: list[ClassA]
+forward3: "ClassA"
+
+
+class ClassA:
+    ...
+
+
+class ClassB:
```
carljm, line 23: Could we also include the shadowing cases (`ClassD.z`, `ClassD.x`, and `ClassD.str`) in this test? The current cases establish that forward references work, but don't check that a class binding takes precedence over an outer name.

For example, pyright passes this file but also reports no errors with this added:

```python
class Shadowing:
    before: int = 0  # E: resolves to Shadowing.int

    def int(self) -> None:
        pass

    after: int = 0  # E: resolves to Shadowing.int
    str: str = ""  # E: resolves to the class attribute
```

Under the future import, the specified rules require errors for all three annotations.

conformance/results/pyrefly/annotations_future_import.toml
```diff
@@ -0,0 +1,17 @@
+conformant = "Partial"
+notes = """
+Names in annotations that refer to definitions after them are not resolved in the precence of `from __future__ import annotations`
```
carljm: Could we narrow this to “Forward references to nested classes are not resolved”? The recorded failures are for `ClassInner`; the module-level `ClassA` and `ClassC` forward references succeed, so the current wording overstates the limitation. I think the same wording change applies to pycroscope's corresponding note.

docs/spec/annotations.rst
carljm: Looks good. I do think there would be a few more tests that would be relevant to copy; commented separately.

carljm (approved):
Couple remaining nits, but looks great, thank you!

conformance/results/pyrefly/annotations_future_import.toml
davidhalter: Makes sense!

davidhalter:
Ok now it should be ready. At least now this file shows how different the type checkers think about forward references: None of the type checkers agree on the errors :smile: 

There are probably a few new notes that might not be great. Will merge after a last OK from Carl.

Thanks for all the feedback!

conformance/results/pycroscope/annotations_future_import.toml
```diff
@@ -0,0 +1,26 @@
+conformant = "Partial"
+notes = """
+Names in annotations that refer to definitions after them are not resolved in the precence of `from __future__ import annotations`
```
carljm: Could we apply the same wording correction here as in the Pyrefly notes: “Forward references to nested classes are not resolved”? The module-level `ClassA` and `ClassC` forward references succeed, so this still overstates the limitation. (Also, "precence" is currently mis-spelled.)

conformance/results/pyrefly/annotations_future_import.toml
```diff
@@ -0,0 +1,21 @@
+conformant = "Partial"
+notes = """
+Forward references to nested classes are not resolved
```
carljm, line 3: Could we also add a short note covering the newly exposed shadowing failures? The results now record missing errors for `str: str = ""` and `z: int = 0` before `def int` (lines 37 and 39), but the notes only describe nested-class failures. Something like “Resolves some class annotations to outer names instead of shadowing class bindings” would cover those cases. Zuban's notes have the same omission.

carljm (approved):

davidhalter:
I think I addressed everything. Will merge tomorrow if there's nothing else.

carljm (approved):
Pushed a few minor nits, looks good to me as-is. Will let you merge it.

davidhalter:
Thanks all! This was quite a bit of work and I'm happy it's finally landing.

carljm:
Thanks @davidhalter for all your work on this!

PR #2146, opened 2026-01-06
Clarify that `tuple` should not be prohibited as an argument to `type`

### LINKED ISSUE #2145: Is `tuple` allowed as an argument to `type`?
Conformance test:
https://github.com/python/typing/blob/211b0cb5470e856f6093f7c3d688295f72f78bb6/conformance/tests/specialtypes_type.py#L69

Spec:
https://github.com/python/typing/blob/211b0cb5470e856f6093f7c3d688295f72f78bb6/docs/spec/special-types.rst?plain=1#L191-L192

The conformance test seems to be in contradiction with the spec

fangyi-zhou:
Closes #2145

This spec change reconciles the typing spec and the conformance testsuite. `tuple` is an object of the `type` class, so it should be accepted as a value of `type`.

### DIFF
--- a/docs/spec/special-types.rst
+++ b/docs/spec/special-types.rst
@@ -188,9 +188,8 @@ concrete class object, e.g. in the above example::
 ``type[T]`` where ``T`` is a type variable is allowed when annotating the
 first argument of a class method (see the relevant section).
 
-Any other special constructs like ``Callable`` are not allowed
-as an argument to ``type``,
-but ``tuple`` is allowed as an argument to ``type``.
+Any other :term:`special forms <special form>` like ``Callable`` are not
+allowed as an argument to ``type``.
 
 There are some concerns with this feature: for example when
 ``new_user()`` calls ``user_class()`` this implies that all subclasses


erictraut (commented):
I don't think the new wording should mention `tuple` at all. It's not a special form. It's simply a class.

The problem here is the phrase `special constructs`, which is not defined anywhere in the spec and is therefore ambiguous. I think we should change this to `special form`, which is defined in the glossary.

encukou:
Maybe add an extra sentence now?
```
Tuples, for example ``type[A, B]``, are also not allowed.
```

erictraut:
> Maybe add an extra sentence now?

No, this is already covered elsewhere in the spec. We try to avoid redundancies because they create confusion and potential inconsistencies.

fangyi-zhou:
> > Maybe add an extra sentence now?
> 
> No, this is already covered elsewhere in the spec. We try to avoid redundancies because they create confusion and potential inconsistencies.

@erictraut Do you mean this is covered in the grammar in https://typing.python.org/en/latest/spec/annotations.html#type-and-annotation-expressions?

erictraut:
> Do you mean this is covered in the grammar

Yes

rchen152 (approved):
I agree with Jelle and Eric's points at https://discuss.python.org/t/spec-change-clarify-that-tuple-should-not-be-prohibited-as-an-argument-to-type/105590 that this section is in need of a broader fix, but this change as-is is already an improvement (removes the obviously incorrect statement that `type[tuple]` is not allowed and the ambiguous term "special constructs"), so I'm in favor of accepting it.

fangyi-zhou:
Are we happy to merge the current change as is? Or we would want to take this opportunity to improve upon it more?

carljm:
I am happy to merge this as-is and consider broader improvements separately, but I would like more than two typing council members to approve before I merge it.

erictraut:
I'm good with this change.

PR #2148, opened 2026-01-08
Change the spec for class access of Self variables

davidhalter:
I have tried to create a first draft for disallowing access to variables through the class object containing Self, as discussed [here](https://discuss.python.org/t/disallow-access-to-class-variables-with-self/105434).

The general reception of this proposed change was pretty positive, but I guess we are all not sure how this rule should be worded.

The by far biggest issue is that I write:

> Accessing a variable through the class object that has a type annotation where
the annotation **contains an occurrence of Self** is not allowed, because Self can
be different depending on the class in the inheritance hierarchy:             

where **"contains an occurrence of Self"** is according to Eric a bit of a new concept. I guess I understand that this is not well defined in the glossary or anywhere else and can of course be used for other things as well. I have tried to skim the Spec and found the following usages in the spec:

- glossary.rst: A type is "fully static" if it does not **contain any gradual form**.
- glossary.rst: All types in the Python type system are "gradual". A gradual type may be a fully static type, or it may be Any, or a type that **contains ``Any``** or another gradual form.
- generics.rst: We reject type aliases **containing ``Self``**. Supporting ``Self``

There's also two times where "in" is used:

- generics.rst: Note that we reject ``Self`` **in** staticmethods. ``Self`` does not add much
- generics.rst: Likewise, we reject ``Self`` **in** metaclasses. ``Self`` consistently refers to the

I have not found other language about this. I have tried to search for "depends on", "references", "mentions", "introduces" and none of them are used.

However I feel like this concept is often talked in almost implicit ways about TypeVars/Self. For example in the chapter "Type Parameters as Parameters to Generics" (talks about `T1 = TypeVar("T2", default=list[T1])`):

> Type parameters are valid as parameters to generics inside of a ``default``

I feel like there's a couple of other places where different words are used, but it's not like this concept is new.

Please let me know what you all think. I'm happy to improve the wording.

### DIFF
--- a/conformance/tests/generics_self_advanced.py
+++ b/conformance/tests/generics_self_advanced.py
@@ -43,5 +43,5 @@ class ChildB(ParentB):
         cls.a # E: Access to generic instance variables via class is ambiguous
         assert_type(cls.method1(), Self)
 
-ClassB.a # E: Ambigous access
-ClassB.a = ClassB.method1() # E: Ambigous write
+ChildB.a # E: Ambigous access
+ParentB.a = [ChildB.method1()] # E: Ambigous write


conformance/tests/generics_self_advanced.py
```diff
@@ -40,6 +40,8 @@ def method2(self) -> None:
     @classmethod
     def method3(cls) -> None:
         assert_type(cls, type[Self])
-        assert_type(cls.a, list[Self])
-        assert_type(cls.a[0], Self)
+        cls.a # E: Access to generic instance variables via class is ambiguous
         assert_type(cls.method1(), Self)
+
+ClassB.a # E: Ambigous access
+ClassB.a = ClassB.method1() # E: Ambigous write
```
carljm: Are these supposed to be `ChildB`?

carljm (commented):
I commented more generally at https://discuss.python.org/t/disallow-access-to-class-variables-with-self/105434/8

conformance/tests/generics_self_advanced.py
davidhalter: You are right, this was a typo. My bad, thanks for looking into it!

carljm:
I'm afraid that I don't think this change makes sense as currently worded, and I don't think we should add this to the spec. As shown in my [Discuss comment](https://discuss.python.org/t/disallow-access-to-class-variables-with-self/105434/8), the soundness issues mentioned as rationale here are unrelated to access via the class. Any inheritance of any mutable attribute typed with `Self` is unsound even if only ever accessed via instances. In the code sample added to the spec here, the actual unsound line of code is `class D(C): pass`, not `C.others` or `D.others`.

It would be controversial to require soundness here, because people use mutable `Self` attributes and inherit classes containing them. But I don't see the rationale for specifically erroring on access via the class object.

(Edit to clarify: by "as currently worded" I mean "if it is worded as a restriction on access via class objects", I don't mean more minor wording changes around how to specify that.)

PR #2153, opened 2026-01-19
Allow type checkers to ignore specific error codes

davidhalter:
Changes the error code from `additional_stuff` to `assignment` (as Mypy uses it).

This was the [proposal](https://discuss.python.org/t/allow-type-ignore-mypy-style-error-codes-in-conformance-tests/105433/5) Rebecca:

> For # type: ignore[additional_stuff], I can see the argument for allowing an error about additional_stuff being an invalid code, but it seems tricky to update that line to allow both mypy’s and pyright’s behavior, since they’re so different. (Mypy also reports an assignment error, since that error code isn’t suppressed.) Maybe we could change it to:
>
> `z: int = ""  # type: ignore[assignment]`
>
> so that the expected behavior in all type checkers is that there’s no error?


JelleZijlstra:
This will need changes in the spec text as well as the conformance suite, and it will need to be approved by the Typing Council. (Personally, I'm supportive of doing something like this, but haven't thought too much about the exact proposal yet.)

davidhalter:
@JelleZijlstra What do you have in mind? My problem is that the spec is currently not specific about this at all (see [this discussion](https://discuss.python.org/t/allow-type-ignore-mypy-style-error-codes-in-conformance-tests/105433)) and I haven't gotten much feedback on where people would want to take this in the same thread.

Would you want to add something to the spec like this:

> Any text following `# type: ignore` is ignored, provided there is at least one whitespace character after `# type: ignore`.
>
> The only exception is the form `# type: ignore[assignment]`, where the bracketed portion is used by mypy to restrict the ignored error codes. Other type checkers may ignore the bracketed codes and treat this form as equivalent to `# type: ignore`.

This is of course just my initial idea, any guidance is appreciated. The current spec doesn't really talk about this, so this should be more of an addition.

JelleZijlstra:
Something like that, yes. If the spec doesn't currently discuss this, that means we should add it.

The spec text shouldn't call out any specific type checker (like mypy). Instead, we should generically say that type checkers may use syntax like `[code]` to restrict the type ignore.

davidhalter:
I have pushed a new update to the spec. It's just a small addition and I hope it reflects the general sentiment of the discussion in the typing discuss. As always I'm happy to integrate feedback.

davidhalter:
@JelleZijlstra Would you mind looking at this spec change? I'd prefer at least a single opinion before I bring this to the typing council.

docs/spec/directives.rst
```diff
@@ -73,6 +73,13 @@ other comments and linting markers:
 
   # type: ignore # <comment or other marker>
 
+Any text following ``# type: ignore`` is ignored, provided there is at least one
+whitespace character after ``# type: ignore``. The form of
+``# type: ignore[error_code1, error_code2]`` can be used to filter error codes
```
JelleZijlstra: I'm not sure I follow this line. The spec should primarily say how type checkers should behave, not what it's good for. Can you make sure the spec text makes it very precise exactly what behavior is expected?

JelleZijlstra: I would expect the behavior to be something like this:
- `# type: ignore` must ignore all errors
- `# type: ignore[code]` may ignore errors, depending on the type checker.

That is, pyright's behavior of ignoring the code and always ignoring errors remains conformant, but mypy's behavior of ignoring the error only when the code matches newly becomes conformant.

We'll also need to be clear about what the following do:

```
s: str = 1  # type: ignored
s: str = 1  # type: ignore, because I feel like it
s: str = 1  # type: ignore because why not
```

davidhalter: @JelleZijlstra Thanks for the feedback

I have tried to integrate as much as possible. Please let me know what you think.

>     # type: ignore must ignore all errors
>     # type: ignore[code] may ignore errors, depending on the type checker.
> 
> That is, pyright's behavior of ignoring the code and always ignoring errors remains conformant, but mypy's behavior of ignoring the error only when the code matches newly becomes conformant.

Just to make sure we're on the same page: I 100% agree and don't want to introduce anything else. I'm probably just struggling how to word this exactly.

docs/spec/directives.rst
```diff
@@ -73,6 +73,25 @@ other comments and linting markers:
 
   # type: ignore # <comment or other marker>
 
+Text following ``# type: ignore`` must be ignored, provided there is at
+least one whitespace character after it::
+
+    # Not valid because of the "d" after ignore; does not supress the type error
+    s: str = 1  # type: ignored
+    # Not valid because of the comma; does not supress the type error
```
JelleZijlstra: What's the motivation again for this part? I'm not sure what we get out of separating the cases with and without the comma, or even why we allow random text after `ignore` at all.

docs/spec/directives.rst
```diff
@@ -73,6 +73,25 @@ other comments and linting markers:
 
   # type: ignore # <comment or other marker>
 
+Text following ``# type: ignore`` must be ignored, provided there is at
```
JelleZijlstra: "Ignored" is a bit confusing since there's multiple levels of ignoring going on. I would word it as saying certain patterns are equivalent to plain `# type: ignore`.

docs/spec/directives.rst
davidhalter: The motiviation is that some people mentioned [on discuss](https://discuss.python.org/t/allow-type-ignore-mypy-style-error-codes-in-conformance-tests/105433/5) that allowing text after `# type: ignore` has a clear intent and we should therefore allow it.

I'm personally in the camp of being more strict here, `# type: ignore foo` would therefore add an error that this is an invalid type: ignore. But I feel like I'm in the minority.

> I'm not sure what we get out of separating the cases with and without the comma, or even why we allow random text after ignore at all.

I'm also open to just dropping that particular case, not mentioning the comma case and leave type checkers room to deal with that in different ways. I don't think it's that important (and probably really rare in actual usage).

[Rebecca mentioned](https://discuss.python.org/t/allow-type-ignore-mypy-style-error-codes-in-conformance-tests/105433/10):

> We could require whitespace after the ignore. So `# type: ignored` would not be a valid ignore, but `# type: ignore whatever` would be.
> (I checked what pyrefly does, and we require either whitespace or a “word boundary”

I'm fine with having it either way, but I think defining what a `word boundary` is in the spec is a bit difficult. I also think that allowing stuff like `# type: ignore.foo` is a bit weird.

docs/spec/directives.rst
davidhalter: Makes sense! Rephrased it.

docs/spec/directives.rst
JelleZijlstra: I don't feel too strongly about the exact behavior but I'd prefer we have a precise spec here to remove ambiguity.

davidhalter: Whitespace is pretty strong, no? It seems like you want the comma to be added. How about we treat whitespace the same as `;`, and `,`? I feel like these are probably most likely to be a useful boundary. Everything else to me feels unnecessary. Do you have another idea how we could phrase it? I also don't have strong feelings.

I'm also questioning whether I should even try to add this to the spec. I am mostly trying to get the part about type ignore filters (`type: ignore[assignment]`) in here. This just felt like a natural addition which was under-specified. Would you recommend just removing that part focusing on the type ignore filters?

carljm: I think deciding how to specify handling of arbitrary text after `# type: ignore` is quite a challenging project, given the diversity of type checker behaviors. IMO it really requires some research/experimentation with the ecosystem corpus to make an informed decision.

I would support not attempting to specify that in this PR, and limiting this PR to just the code/filter stuff below (which I support as written.)

davidhalter:
I removed the parts where I tried to specify how whitespace after `# type: ignore` works. Now it's just about error codes.

docs/spec/directives.rst
```diff
@@ -73,6 +73,15 @@ other comments and linting markers:
...
   # type: ignore # <comment or other marker>
 
+The form of ``# type: ignore[...]`` may be used to filter errors depending on
+the type checker:
+
+- In ``# type: ignore[my_code1]``, a type checker may ignore the error code
+  ``my_code1`` and choose to treat this form as equivalent to ``# type: ignore``.
+- Alternatively in ``# type: ignore[my_code1]`` a type checker may suppress the
+  error only if the error cause matches the error code ``my_code1``.
+- ``# type: ignore`` must always suppress all errors.
```
carljm:
```suggestion
The form ``# type: ignore[...]`` may be used to suppress only type errors with a
given error code, though support for this is optional and may vary by type checker:

- Given ``# type: ignore[my_code1]``, a type checker may ignore the error code
  ``my_code1`` and choose to treat this form as equivalent to ``# type: ignore``.
- Alternatively, given ``# type: ignore[my_code1]`` a type checker may suppress the
  error only if the error cause matches the error code ``my_code1``.
- A bare ``# type: ignore`` must always suppress all type errors.
```

conformance/tests/directives_type_ignore.py
```diff
@@ -11,7 +11,7 @@
 y: int = ""  # type: ignore - additional stuff
 
 # The following type violation should be suppressed.
```
carljm: Given the wording in the spec here, a type checker should be permitted to _not_ suppress this error, if its error code for `z: int = ""` is not `assignment`. (This change will make ty pass this test.)

```suggestion
# The following type violation may be suppressed, depending whether the
# type checker uses the error code `assignment` for invalid assignments,
# and/or how it treats unknown error codes.
```

conformance/tests/directives_type_ignore.py
```diff
@@ -11,7 +11,7 @@
...
-z: int = ""  # type: ignore[additional_stuff]
+z: int = ""  # type: ignore[assignment]
```
carljm:
```suggestion
z: int = ""  # type: ignore[assignment]  # E?
```

carljm: tbh I think it would be more "fair" to _not_ change this line to use the type code that matches mypy/zuban. There's no specified requirement that all type checkers must use the type code `assignment` for this situation. So I think all type checkers that do strict handling of ignores with error codes should equally rely on `# E?` on this line.

davidhalter: This is definitely fair. I'm indifferent here and will wait if there are any other opinions here.

docs/spec/directives.rst
```diff
@@ -73,6 +73,15 @@ other comments and linting markers:
 
   # type: ignore # <comment or other marker>
 
+The form ``# type: ignore[...]`` may be used to suppress only type errors with a
+given error code, though support for this is optional and may vary by type checker:
```
JukkaL, line 77: What about also documenting that multiple error codes can be given (comma-separated), and unrecognized error codes should be ignored, if a type checker supports ignoring errors based on specific error code, as @ilevkivskyi suggested [here](https://discuss.python.org/t/allow-type-ignore-mypy-style-error-codes-in-conformance-tests/105433/11).

conformance/tests/directives_type_ignore.py
AlexWaygood: I obviously don't have a vote here, but I agree with @carljm :)

ty has the same behaviour as zuban and mypy here but has different error codes. I would expect this change to result in ty being marked fully conformant with this file, just like mypy and zuban, but it is not because you're using a mypy/zuban-specific error code

JelleZijlstra: Yes, I think it might be best to actually use a completely made up error code.

`# type: ignore[an-empty-str-is-not-an-int]`, for the type checkers with really specific error codes.

docs/spec/directives.rst
davidhalter: Are there any other opinions on this? I'm pretty open to changes here.

I personally think that the comma separated form doesn't need to be specified explicitly.

> unrecognized error codes should be ignored, if a type checker supports ignoring errors based on specific error code

Does this mean that Mypy would need to ignore the error code "undefined"? Currently it adds errors at least when using `--warn-unused-ignores`: https://mypy-play.net/?mypy=latest&python=3.12&flags=warn-unused-ignores&gist=27ebc9511f193a9f348b0235412cbcc9

conformance/tests/directives_type_ignore.py
davidhalter: Oh nice, I didn't even realize that Ty was also affected by this. It feels like the change makes even more sense then. I changed it towards the random error code Jelle proposed.

AlexWaygood: yes, we changed our behaviour recently because we came to the conclusion that mypy's/zuban's behaviour was much more user-friendly for folks trying to run multiple type checkers on their codebase (not unusual for library authors)

conformance/tests/directives_type_ignore.py
```diff
@@ -10,8 +10,10 @@
 # The following type violation should be suppressed.
 y: int = ""  # type: ignore - additional stuff
 
-# The following type violation should be suppressed.
-z: int = ""  # type: ignore[additional_stuff]
+# The following type violation may be suppressed, depending whether the
+# type checker uses the error code `assignment` for invalid assignments,
+# and/or how it treats unknown error codes.
```
carljm: I know I'm editing myself here, but this comment needs another update now that the next line is using a made-up error code that no type checker would ever use:
```suggestion
# The following type violation may be suppressed, if the type checker
# actually uses the code `an-empty-str-is-not-an-int` (unlikely!), or if
# it treats unknown codes as blanket ignores.
```

docs/spec/directives.rst
carljm: I personally think that the approach outlined by @ilevkivskyi is the right one for multi-type-checker interoperability (much better than requiring totally separate ignore comments for each type checker), and I would be happy to fully specify it. But that would be a specification change that puts pyright and pyrefly out of conformance. If we did that, then I think the fulI behavior, including multiple comma separated codes, should be specified. I guess maybe we need feedback at least from pyrefly here: is the "always blanket ignore" behavior pyrefly's UX preference, or just something that was done to satisfy the previous conformance suite requirements?

But I think as long as the specification here is looser, as currently written (to accommodate both the pyright/pyrefly behavior and the mypy/zuban/ty behavior), it's not really important to document the multiple-error-codes handling, since effectively all this spec is now saying is that type checkers can do whatever they want with the information inside the brackets.

(Either way, I don't think the spec should _require_that type checkers be silent about unknown codes; it should be OK to warn users about them, as mypy does with `--warn-unused-ignores`. This is better behavior for users who aren't using multiple type checkers. Ideally users in multi-type-checker scenarios can turn off that warning if they don't want it.)

I'm happy with the current state of this PR as a first step, at least.

carljm (approved):

JelleZijlstra (approved):

conformance/tests/directives_type_ignore.py
davidhalter: Yeah, that's really my fault. Sorry for that!

davidhalter:
> I personally think that the approach outlined by @ilevkivskyi is the right one for multi-type-checker interoperability (much better than requiring totally separate ignore comments for each type checker), and I would be happy to fully specify it. But that would be a specification change that puts pyright and pyrefly out of conformance. 

I'd also be happy to add that to the spec. I think if we want to do that we should ask for feedback on Discuss again, because people like @jorenham (scipy-stubs maintainer) might want to weigh in there for these multi-typechecker scenarios.

jorenham:
> I'd also be happy to add that to the spec. I think if we want to do that we should ask for feedback on Discuss again, because people like @jorenham (scipy-stubs maintainer) might want to weigh in there for these multi-typechecker scenarios.

I don't have any particular preference to be honest. Anything that'll allow me to simplify things like 

```py
    # NOTE: These raise a `ValueError` if passed anything other than `ClusterNode`
    @override
    # pyrefly: ignore [bad-override]
    def __eq__(self, node: ClusterNode, /) -> bool: ...  # type: ignore[override]  # pyright: ignore[reportIncompatibleMethodOverride]  # ty: ignore[invalid-method-override]
```

([src](https://github.com/scipy/scipy-stubs/blob/73bcb363d74a55c5af3bbf200067d40ba20e14de/scipy-stubs/cluster/hierarchy.pyi#L88-L91)), and I'll be happy.

docs/spec/directives.rst
rchen152: I'm fine with the current wording in this PR.

As for @ilevkivskyi's suggestion - I personally like it, but I'd like to run this by some other pyrefly devs who have thought a lot more than I have about configuration and adoption, to make sure this isn't incompatible with our tooling in some way that I'm missing.

rchen152: > As for @ilevkivskyi's suggestion - I personally like it, but I'd like to run this by some other pyrefly devs [...]

We discussed this and flagged a couple questions/concerns:
* Are there any plans to standardize what lines errors are reported on? Without standardization there, this proposed change seems less useful, since multiple `# type: ignore`s on different lines would still be needed to satisfy multiple type checkers.
* Pyrefly has a `--remove-unused-ignores` flag to remove `# pyrefly: ignore` directives that are no longer used (useful for automated pyrefly version upgrades). This would be trickier in a world in which every type checker uses `# type: ignore`, since error codes could be used by other checkers rather than unused.

AlexWaygood:
> I don't have any particular preference to be honest. Anything that'll allow me to simplify things like
> 
> ```python
>     # NOTE: These raise a `ValueError` if passed anything other than `ClusterNode`
>     @override
>     # pyrefly: ignore [bad-override]
>     def __eq__(self, node: ClusterNode, /) -> bool: ...  # type: ignore[override]  # pyright: ignore[reportIncompatibleMethodOverride]  # ty: ignore[invalid-method-override]
> ```

FWIW since we already changed ty's behaviour here, I believe you can already simplify it slightly to

```py
    # NOTE: These raise a `ValueError` if passed anything other than `ClusterNode`
    @override
    # pyrefly: ignore [bad-override]
    def __eq__(self, node: ClusterNode, /) -> bool: ...  # type: ignore[override, ty:invalid-method-override]  # pyright: ignore[reportIncompatibleMethodOverride]
```

davidhalter:
I'm not really sure how you handle situations like this, do I merge now that everyone has ticked off the box? I understand that the discussion around standardizing `# type ignore[<content>]` is still going on (where `<content>` could be defined as a comma separated list with namespaces and such). I'm supportive of that change, but I would prefer to merge this one first.

docs/spec/directives.rst
carljm: We should probably move this to a Discuss thread if someone wants to champion actually specifying the multiple-codes-in-type-ignore approach. I think the issues raised by the Pyrefly team are definitely real ones.

Regarding the first, I would just say that it's already a minority of type errors where checkers vary on the reported line, so the proposal is still a strict improvement for the ~90+% where type checkers naturally agree (because there's no real ambiguity). So some effort to standardize more of course wouldn't hurt, but this doesn't seem like a blocker.

Regarding the second, what ty does is expect all our codes in `# type: ignore[...]` to be prefixed with `ty:`, so we can still flag/remove all unused definitely-for-ty ignores, which is on par with what we can do with `# ty: ignore`. Whether we should ever remove (or flag as unused) ignores which might apply to other type checkers, is I think a UX question for each type checker. Might be useful for single-type-checker users, but definitely not desirable for multi-type-checker users, so at best it would need to be configurable.

carljm:
@davidhalter I think once you have five checkmarks on the typing-council issue and at least one approving review on the PR, you are good to merge.

jorenham:
> FWIW since we already changed ty's behaviour here, I believe you can already simplify it slightly to
> 
> ```python
>     # NOTE: These raise a `ValueError` if passed anything other than `ClusterNode`
>     @override
>     # pyrefly: ignore [bad-override]
>     def __eq__(self, node: ClusterNode, /) -> bool: ...  # type: ignore[override, ty:invalid-method-override]  # pyright: ignore[reportIncompatibleMethodOverride]
> ```

This unfortunately will cause mypy to complain:

```
Unused "type: ignore[ty:invalid-method-override]" comment
```

ilevkivskyi:
> This unfortunately will cause mypy to complain

This is a bug (or at least something that we want to change) in mypy.

davidhalter:
It seems like merging this PR (even though it was green) broke the CI pipeline on main.

@JelleZijlstra Would you have time to look at this. It feels like a weird issue with Pyright when they do new releases:

https://github.com/python/typing/actions/runs/24907018988/job/72938873294

I can try to fix this, but I don't have that much time until Monday. And you seem to have implemented these checks in the CI. (Not blaming you here at all, thanks for your work!)

JelleZijlstra:
That's a fun one! We probably should have a post-processing step to remove such errors.

PR #2155, opened 2026-01-20
Add disclaimer to conformance results

hauntsaninja:
I've now seen a few different times people misunderstand exactly what spec conformance implies, so here's a PR adding a small disclaimer

### DIFF
--- a/conformance/results/results.html
+++ b/conformance/results/results.html
@@ -166,8 +166,8 @@
             <h3>Python Type System Conformance Test Results</h3>
             <p class="disclaimer">
                 While spec conformance is important for the ecosystem, we don't recommend using it
-                as the primary basis for choosing a type checker. It is not (and is not meant to be)
-                representative of the things users typically care about.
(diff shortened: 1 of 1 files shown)


conformance/src/results_template.html
```diff
@@ -148,13 +148,27 @@
...
     </style>
 </head>
 
 <body>
     <div class="content_container">
         <header>
             <h3>Python Type System Conformance Test Results</h3>
+            <p class="disclaimer">
+                While spec conformance is important for the ecosystem, we don't recommend using it
+                as the primary basis for choosing a type checker. It is not (and is not meant to be)
```
JelleZijlstra: This second sentence feels a bit too strong for me. Sure, a lot of the conformance suite is about esoteric topics that don't affect users much, and there are many important aspects of a practical type checker that aren't reflected in the conformance suite, but I do believe that spec conformant behavior should be user-friendly behavior.

hauntsaninja: Okay, tried a reword

PR #2160, opened 2026-01-27
Update conformance test results.

rchen152:
Updates results for new versions of pyrefly, pyright, zuban.


conformance/results/pyright/tuples_type_compat.toml
```diff
@@ -1,4 +1,7 @@
-conformant = "Pass"
+conformant = "Partial"
+notes = """
```
JelleZijlstra, line 2: @erictraut fyi, recent pyright appears to regress here

rchen152: I opened a pyright issue with a bit more detail: https://github.com/microsoft/pyright/issues/11249

Issue #2162, opened 2026-01-28
Callable allows variadic args


JelleZijlstra:
The [spec](https://typing.python.org/en/latest/spec/callables.html#callable) claims "The Callable form provides no way to specify keyword-only parameters, variadic parameters, or default argument values."

In fact, PEP-646 syntax allows spelling a Callable type with variadic positional arguments, as specified [here](https://typing.python.org/en/latest/spec/generics.html#type-variable-tuples-with-callable). We should edit the spec to clarify this.

JelleZijlstra:
Fixed by #2164

PR #2164, opened 2026-01-28
Clarify that variadic positional parameters can be expressed with Callable

Daraan:
As @JelleZijlstra noted [here](https://github.com/python/peps/pull/4764#discussion_r2734695090) the [current notion](https://typing.python.org/en/latest/spec/callables.html#callable) of

> The Callable form provides no way to specify [...] variadic parameters [...]

is not correct as we can express variadic positional parameters with an unpacked tuple, e.g. `Callable[[*tuple[int, ...]], str]`.
This PR adds clarification for this case.

I've removed the notion about variadics completely in that paragraph, an alternative would be 

> no way to specify keyword-only parameters, variadic keyword-only parameters, or ..." 

which feels redundant. Feel free to share other suggestions

Resolves: #2162

### DIFF
--- a/docs/spec/callables.rst
+++ b/docs/spec/callables.rst
@@ -536,10 +536,11 @@ to be :term:`consistent` with any input parameters::
         f4: CallbackWithStr[...] = cb  # Error
 
 
-Variadic positional parameter types can be defined with ``...`` in combination
-with an unpacked tuple of the form ``tuple[int, ...]``. The first argument
-specifies the type of the variadic parameters. For example, the following
-defines a callable that accepts any number of integer arguments::
+Variadic positional parameter types can be defined with with an unpacked tuple
+of the form ``*tuple[int, ...]`` in the parameter list. The first argument
+specifies the type of the variadic parameters.
+For example, the following defines a callable that accepts any number of integer
+arguments::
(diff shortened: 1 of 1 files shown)


docs/spec/callables.rst
```diff
@@ -535,6 +535,14 @@ to be :term:`consistent` with any input parameters::
         f3: CallbackWithInt[...] = cb  # OK
         f4: CallbackWithStr[...] = cb  # Error
 
+
+Variadic positional parameter types can be defined with ``...`` in combination
```
carljm: Is this supposed to mention `*` rather than `...`? Doesn't make sense to say "`...` in combination with an unpacked tuple of the form `tuple[int, ...]`"

docs/spec/callables.rst
```diff
@@ -535,6 +535,15 @@ to be :term:`consistent` with any input parameters::
         f3: CallbackWithInt[...] = cb  # OK
         f4: CallbackWithStr[...] = cb  # Error
 
+
+Variadic positional parameter types can be defined with with an unpacked tuple
```
erictraut, line 539: Typo: "with with" should be "with"

PR #2173, opened 2026-02-12
Update specification for directives for sys.implementation and sys.platform checks.

Josverl:
This PR adds additional detail to the specification for Version and Platform checking.
Specificially it aims to add support for typechers to add support for :
 - checks on `sys.implementation.name`
 - membership checks (in tuple) 
 - negative membership checks ( not in tuple) 

References :
 - https://discuss.python.org/t/proposal-to-improve-support-for-other-python-platforms-in-the-typing-specification/91877/1
 - https://github.com/python/typing/discussions/1953

### DIFF
--- a/docs/spec/directives.rst
+++ b/docs/spec/directives.rst
@@ -200,7 +200,7 @@ Type checkers should support the following comparison patterns:
 
     Common values: ``"linux"``, ``"darwin"``, ``"win32"``, ``"emscripten"``, ``"wasi"``
 
-The membership checks ``in``  and ``not in``, only support simple containment testing to a tuple of literal strings.
+The membership checks ``in`` and ``not in`` only support simple containment testing with a tuple of literal strings.
 
 .. code-block:: python
    :caption: Example `sys.platform`
--- a/docs/spec/directives.rst
+++ b/docs/spec/directives.rst
@@ -177,7 +177,7 @@ Type checkers should support the following comparison patterns:
     * ``sys.version_info < <2-tuple>``
 
 Comparisons checks are only supported against the first two elements of the version tuple. 
-Use of named attributes is not supported.
+Use of named attributes is not mandated.
 
 .. code-block:: python
    :caption: Example `sys.version_info`
@@ -244,7 +244,7 @@ Type checkers should support the following comparison patterns:
     * ``sys.implementation.version < <2-tuple>``
 
 Comparisons checks are only supported against the first two elements of the implementation version tuple.
-Use of named attributes is not supported.
+Use of named attributes is not mandated.
 
 .. code-block:: python
    :caption: Example `sys.implementation.version`
@@ -266,7 +266,7 @@ No support for complex expressions
 ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
 
 Type checkers are only required to support the above patterns, and are not required to evaluate complex expressions involving these variables.
-For example, the pattern ``sys.platform == "linux"`` is supported but other syntax variants such as ``platform == "linux"`` and ``"win" not in sys.platform`` are not supported.
+For example, the pattern ``sys.platform == "linux"`` is supported but other syntax variants such as ``platform == "linux"`` and ``"win" not in sys.platform`` are not mandated.
 
 Therefore checkers are **not required** to understand obfuscations such as:
 
--- a/docs/spec/directives.rst
+++ b/docs/spec/directives.rst
@@ -154,7 +154,7 @@ left undefined by the typing spec at this time.
 Version and platform checking
 -----------------------------
 
-Type checkers should support narrowing based on:
+Type checkers should understand code paths as definitely reachable or not reachable due to comparison tests against these symbols:
     * ``sys.version_info``
     * ``sys.platform``
     * ``sys.implementation.version``
@@ -176,8 +176,9 @@ Type checkers should support the following comparison patterns:
     * ``sys.version_info >= <2-tuple>``
     * ``sys.version_info < <2-tuple>``
 
-Comparisons checks are only supported against the first two elements of the version tuple. 
-Use of named attributes is not mandated.
+Comparison checks are only supported against the first two elements of the version tuple.
+It should be noted that type checkers may choose to also support the 3-tuple ``sys.version_info >= <3-tuple>``.
+Type checkers are not expected to support comparisons with named attributes of `sys.version_info`.
 
 .. code-block:: python
    :caption: Example `sys.version_info`
@@ -186,8 +187,10 @@ Use of named attributes is not mandated.
    import sys
    if sys.version_info >= (3, 12):
        # Python 3.12+
+   elif sys.version_info >= (3, 11):
+       # Python 3.11 
    else:
-       # Python 3.11 and lower
+       # Python 3.10 and lower
 
 sys.platform checks
 ^^^^^^^^^^^^^^^^^^^
@@ -195,12 +198,13 @@ sys.platform checks
 Type checkers should support the following comparison patterns:
     * ``sys.platform == <string literal>``
     * ``sys.platform != <string literal>``
-    * ``sys.platform in <tuple of string literals>``
-    * ``sys.platform not in <tuple of string literals>``
+    * ``sys.platform.startswith(<string literal>)``
+    * ``sys.platform in <set of string literals>``
+    * ``sys.platform not in <set of string literals>``
 
     Common values: ``"linux"``, ``"darwin"``, ``"win32"``, ``"emscripten"``, ``"wasi"``
 
-The membership checks ``in`` and ``not in`` only support simple containment testing with a tuple of literal strings.
+The membership checks ``in`` and ``not in`` only support simple containment testing with a set of literal strings.
 
 .. code-block:: python
    :caption: Example `sys.platform`
@@ -220,10 +224,12 @@ sys.implementation.name checks
 Type checkers should support comparison patterns:
     * ``sys.implementation.name == <string literal>``
     * ``sys.implementation.name != <string literal>``
-    * ``sys.implementation.name in <tuple of string literals>``
-    * ``sys.implementation.name not in <tuple of string literals>``
+    * ``sys.implementation.name in <set of string literals>``
+    * ``sys.implementation.name not in <set of string literals>``
 
+    Default value: ``"cpython"``, unless configured otherwise.
     Common values: ``"cpython"``, ``"pypy"``, ``"micropython"``, ``"graalpy"``, ``"jython"``, ``"ironpython"``
+
 
 .. code-block:: python
    :caption: Example `sys.implementation.name`
@@ -239,12 +245,16 @@ Type checkers should support comparison patterns:
 sys.implementation.version checks
 ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
 
+``sys.implementation.version`` is a tuple, in the same format as sys.version_info. However it represents the version of the Python implementation 
+rather than the version of the Python language. This has a distinct meaning from the specific version of the Python language to which the currently 
+running interpreter conforms. For CPython this is the same as `sys.version_info`.
+
 Type checkers should support the following comparison patterns:
     * ``sys.implementation.version >= <2-tuple>``
     * ``sys.implementation.version < <2-tuple>``
 
-Comparisons checks are only supported against the first two elements of the implementation version tuple.
-Use of named attributes is not mandated.
+Comparison checks are only supported against the first two elements of the implementation version tuple.
+Type checkers are not required to support comparisons against named attributes of `sys.implementation.version`.  
 
 .. code-block:: python
    :caption: Example `sys.implementation.version`
@@ -256,39 +266,45 @@ Use of named attributes is not mandated.
    if sys.implementation.name == "micropython" and sys.implementation.version >= (1, 24):
        # MicroPython version 1.24 and above
 
-.. note::
-
-    ``sys.implementation.version`` is a tuple, in the same format as sys.version_info. However it represents the version of the Python implementation rather than the version of the Python language. 
-    This has a distinct meaning from the specific version of the Python language to which the currently running interpreter conforms. For CPython this is the same as `sys.version_info`.
-
 
 No support for complex expressions
 ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
 
-Type checkers are only required to support the above patterns, and are not required to evaluate complex expressions involving these variables.
-For example, the pattern ``sys.platform == "linux"`` is supported but other syntax variants such as ``platform == "linux"`` and ``"win" not in sys.platform`` are not mandated.
+Type checkers are required to support the above patterns, and are not required to evaluate other comparisons or other syntax variants.
 
 Therefore checkers are **not required** to understand obfuscations such as:
 
 .. code-block:: python
     :caption: Examples of unsupported or overly complex version/platform checks
-    :emphasize-lines: 3,5,7
+    :emphasize-lines: 4,6,8
 
     import sys
     from sys import platform
+
     if "".join(reversed(sys.platform)) == "xunil": 
-        # Linux specific code
+        # Typecheckers will not be required to understand this obfuscated check
     if platform == "linux":
-        # Linux specific code
+        # Typecheckers will not be required to understand this import alias for sys.platform
     if "win" not in sys.platform:
-        # Non-Windows specific code
+        # Typecheckers will not be required to understand this reversed membership check
 
 
 Configuration
 ^^^^^^^^^^^^^
 
-Type checkers should provide configuration or CLI options to specify target sys.version, sys.platform, sys.implementation.name and sys.implementation.version. 
-The exact mechanism for this is implementation-defined by the type checker.
+Type checkers must provide configuration or CLI options to specify target ``sys.version``, ``sys.platform``, ``sys.implementation.name`` and ``sys.implementation.version``.
+
+================================  ==========================  ==============  ===========================================================================
+  Symbol                            Suggested Format            Example         Suggested Default
+================================  ==========================  ==============  ===========================================================================
+``sys.version``                   string ``"major.minor"``    ``"3.11"``      The version of the Python interpreter used to run the type checker.
+``sys.platform``                  lowercase string            ``"linux"``     The platform of the Python interpreter used to run the type checker.
+``sys.implementation.name``       lowercase string            ``"cpython"``   ``"cpython"`` unless configured otherwise.
+``sys.implementation.version``    string ``"major.minor"``    ``"3.14"``      The value used for ``sys.version`` unless configured otherwise.
+================================  ==========================  ==============  ===========================================================================
+
+The configuration options should allow users to specify the target values for these symbols, so that type checkers can evaluate the version and platform checks correctly.
+The exact mechanism and name for these configuration options is implementation-specific, and defined by each type checker.
 
 .. _`deprecated`:
 
--- a/docs/spec/directives.rst
+++ b/docs/spec/directives.rst
@@ -154,23 +154,141 @@ left undefined by the typing spec at this time.
 Version and platform checking
 -----------------------------
 
-Type checkers are expected to understand simple version and platform
-checks, e.g.::
+Type checkers should support narrowing based on:
+    * ``sys.version_info``
+    * ``sys.platform``
+    * ``sys.implementation.version``
+    * ``sys.implementation.name``
 
-  import sys
+Type checkers should support combining these checks with:
+    * A ``not`` unary operator
+    * An ``and`` or ``or`` binary operator
 
-  if sys.version_info >= (3, 12):
-      # Python 3.12+
-  else:
-      # Python 3.11 and lower
+Type checkers are only required to support the fully-qualified form (e.g., ``sys.platform``).
+Support for aliases or import variants (e.g., ``from sys import platform``) is not required, though type checkers may choose to support them.
 
-  if sys.platform == 'win32':
-      # Windows specific definitions
-  else:
-      # Posix specific definitions
+The comparison patterns for these variables are described in more detail in the following paragraphs.
 
-Don't expect a checker to understand obfuscations like
-``"".join(reversed(sys.platform)) == "xunil"``.
+sys.version_info checks
+^^^^^^^^^^^^^^^^^^^^^^^^
+
+Type checkers should support the following comparison patterns:
+    * ``sys.version_info >= <2-tuple>``
+    * ``sys.version_info < <2-tuple>``
+
+Comparisons checks are only supported against the first two elements of the version tuple. 
+Use of named attributes is not supported.
+
+.. code-block:: python
+   :caption: Example `sys.version_info`
+   :emphasize-lines: 2
+
+   import sys
+   if sys.version_info >= (3, 12):
+       # Python 3.12+
+   else:
+       # Python 3.11 and lower
+
+sys.platform checks
+^^^^^^^^^^^^^^^^^^^
+
+Type checkers should support the following comparison patterns:
+    * ``sys.platform == <string literal>``
+    * ``sys.platform != <string literal>``
+    * ``sys.platform in <tuple of string literals>``
+    * ``sys.platform not in <tuple of string literals>``
+
+    Common values: ``"linux"``, ``"darwin"``, ``"win32"``, ``"emscripten"``, ``"wasi"``
+
+The membership checks ``in``  and ``not in``, only support simple containment testing to a tuple of literal strings.
+
+.. code-block:: python
+   :caption: Example `sys.platform`
+   :emphasize-lines: 2,4
+
+   import sys
+   if sys.platform == 'win32':
+       # Windows specific definitions
+   if sys.platform in ("linux", "darwin"):
+       # Platform-specific stubs for Linux and macOS
+       ...
+
+
+sys.implementation.name checks
+^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
+
+Type checkers should support comparison patterns:
+    * ``sys.implementation.name == <string literal>``
+    * ``sys.implementation.name != <string literal>``
+    * ``sys.implementation.name in <tuple of string literals>``
+    * ``sys.implementation.name not in <tuple of string literals>``
+
+    Common values: ``"cpython"``, ``"pypy"``, ``"micropython"``, ``"graalpy"``, ``"jython"``, ``"ironpython"``
+
+.. code-block:: python
+   :caption: Example `sys.implementation.name`
+   :emphasize-lines: 2,4
+
+   import sys
+   if sys.implementation.name == "cpython":
+       # CPython-specific stub
+   if sys.implementation.name == "micropython":
+       # MicroPython-specific stub
+
+
+sys.implementation.version checks
+^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
+
+Type checkers should support the following comparison patterns:
+    * ``sys.implementation.version >= <2-tuple>``
+    * ``sys.implementation.version < <2-tuple>``
+
+Comparisons checks are only supported against the first two elements of the implementation version tuple.
+Use of named attributes is not supported.
+
+.. code-block:: python
+   :caption: Example `sys.implementation.version`
+   :emphasize-lines: 2,4
+
+   import sys
+   if sys.implementation.name == "pypy" and sys.implementation.version >= (7, 3):
+       # PyPy version 7.3 and above
+   if sys.implementation.name == "micropython" and sys.implementation.version >= (1, 24):
+       # MicroPython version 1.24 and above
+
+.. note::
+
+    ``sys.implementation.version`` is a tuple, in the same format as sys.version_info. However it represents the version of the Python implementation rather than the version of the Python language. 
+    This has a distinct meaning from the specific version of the Python language to which the currently running interpreter conforms. For CPython this is the same as `sys.version_info`.
+
+
+No support for complex expressions
+^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
+
+Type checkers are only required to support the above patterns, and are not required to evaluate complex expressions involving these variables.
+For example, the pattern ``sys.platform == "linux"`` is supported but other syntax variants such as ``platform == "linux"`` and ``"win" not in sys.platform`` are not supported.
+
+Therefore checkers are **not required** to understand obfuscations such as:
+
+.. code-block:: python
+    :caption: Examples of unsupported or overly complex version/platform checks
+    :emphasize-lines: 3,5,7
+
+    import sys
+    from sys import platform
+    if "".join(reversed(sys.platform)) == "xunil": 
+        # Linux specific code
+    if platform == "linux":
+        # Linux specific code
+    if "win" not in sys.platform:
+        # Non-Windows specific code
+
+
+Configuration
+^^^^^^^^^^^^^
+
+Type checkers should provide configuration or CLI options to specify target sys.version, sys.platform, sys.implementation.name and sys.implementation.version. 
+The exact mechanism for this is implementation-defined by the type checker.
 
 .. _`deprecated`:
 
--- a/docs/spec/directives.rst
+++ b/docs/spec/directives.rst
@@ -177,7 +177,7 @@ Type checkers should support the following comparison patterns:
     * ``sys.version_info < <2-tuple>``
 
 Comparison checks are only supported against the first two elements of the version tuple.
-It should be noted that type checkers may choose to also support the 3-tuple ``sys.version_info >= <3-tuple>``.
+Type checkers may choose to also support the 3-tuple ``sys.version_info >= <3-tuple>``.
 Type checkers are not expected to support comparisons with named attributes of `sys.version_info`.
 
 .. code-block:: python
@@ -199,6 +199,9 @@ Type checkers should support the following comparison patterns:
     * ``sys.platform == <string literal>``
     * ``sys.platform != <string literal>``
     * ``sys.platform.startswith(<string literal>)``
+    * ``sys.platform in <tuple of string literals>``
+    * ``sys.platform not in <tuple of string literals>``
+Type checkers may also support the following comparison patterns:
     * ``sys.platform in <set of string literals>``
     * ``sys.platform not in <set of string literals>``
 
@@ -247,7 +250,7 @@ sys.implementation.version checks
 
 ``sys.implementation.version`` is a tuple, in the same format as sys.version_info. However it represents the version of the Python implementation 
 rather than the version of the Python language. This has a distinct meaning from the specific version of the Python language to which the currently 
-running interpreter conforms. For CPython this is the same as `sys.version_info`.
+running interpreter conforms. For CPython (``sys.implementation.name == "cpython"``) this is the same as `sys.version_info`.
 
 Type checkers should support the following comparison patterns:
     * ``sys.implementation.version >= <2-tuple>``
@@ -292,7 +295,7 @@ Therefore checkers are **not required** to understand obfuscations such as:
 Configuration
 ^^^^^^^^^^^^^
 
-Type checkers must provide configuration or CLI options to specify target ``sys.version``, ``sys.platform``, ``sys.implementation.name`` and ``sys.implementation.version``.
+Type checkers must be able to retrieve the information from the python implementation's runtime environment, or provide configuration or CLI options to specify target ``sys.version``, ``sys.platform``, ``sys.implementation.name`` and ``sys.implementation.version``.
 
 ================================  ==========================  ==============  ===========================================================================
   Symbol                            Suggested Format            Example         Suggested Default


docs/spec/directives.rst
```diff
@@ -155,12 +160,65 @@ checks, e.g.::
   else:
       # Python 3.11 and lower
 
+**Supported patterns:**
+
+* Equality: ``sys.version_info == (3, 10)``
```
carljm: This example doesn't make much sense, since it would never be true at runtime on any actual Python version.

A realistic example that might actually return `True` would need to look like `sys.version_info == (3, 13, 1, "final", 0)` -- but how useful is that in practice?

I would suggest not specifying that equality checks should be supported at all; I don't think they are usable for any realistic scenario.

docs/spec/directives.rst
```diff
@@ -155,12 +160,65 @@ checks, e.g.::
...
+* Inequality: ``sys.version_info != (3, 9)``
```
carljm: I think inequality checks should also not be supported, for the same reason discussed above for equality checks.

docs/spec/directives.rst
```diff
@@ -155,12 +160,65 @@ checks, e.g.::
...
+* Comparison: ``sys.version_info >= (3, 10)``, ``sys.version_info < (3, 12)``
+* Tuple slicing: ``sys.version_info[:2] >= (3, 10)``
+* Element access: ``sys.version_info[0] >= 3``
```
carljm: I think for these cases (and even for the basic comparison case) it's also important to specify which elements of the tuple are supported. For example, are type checkers expected to support `sys.version_info[2] == 1` (where `sys.version_info[2]` is the micro version). What about the fourth and fifth elements ("releaselevel" and "serial")?

I think we should be explicit that type checkers should only be expected to have special handling for the first two tuple elements (major and minor version); anything more than that adds a lot of complexity to type checker configuration for little to no gain.

carljm: What about the named attributes? Should type checkers support e.g. `sys.version_info.major >= 3`?

(I don't think this is useful enough to require, but maybe we should be explicit that support for it is not expected.)

docs/spec/directives.rst
```diff
@@ -155,12 +160,65 @@ checks, e.g.::
...
+* Inequality: ``sys.implementation.name != "cpython"``
+* Membership: ``sys.implementation.name in ("pypy", "graalpy")``
+* Negative membership: ``sys.implementation.name not in ("cpython", "pypy")``
+
+Common values: ``"cpython"``, ``"pypy"``, ``"micropython"``, ``"graalpy"``, ``"jython"``, ``"ironpython"``
+
+Configuration
+^^^^^^^^^^^^^
+
+Type checkers should provide configuration to specify target version, platform, and implementation. The exact mechanism is implementation-defined.
```
carljm: Target version to what granularity?

docs/spec/directives.rst
```diff
@@ -155,12 +160,65 @@ checks, e.g.::
...
+      ...
+
+**Supported patterns:**
+
+* Equality: ``sys.platform == "linux"``
+* Inequality: ``sys.platform != "win32"``
+* Membership: ``sys.platform in ("linux", "darwin")``
+* Negative membership: ``sys.platform not in ("win32", "cygwin")``
+
+sys.implementation.name checks
```
carljm, line 241: I think adding this has does have significant ecosystem costs (as outlined by @erictraut in https://discuss.python.org/t/proposal-to-improve-support-for-other-python-platforms-in-the-typing-specification/91877/3 ).

In the end I think it is probably worth it, because without it, type-checking for alternative implementations of Python seems quite difficult. The cost to type checkers themselves seems relatively low: one new config option, defaulting to `"cpython"`.

carljm (changes requested):

docs/spec/directives.rst
Josverl: I agree that the specific version should be adjusted, but I don't think that testing for equality should be dismissed, or truncated to an arbitrary number of nodes.

docs/spec/directives.rst
Josverl: Thanks, I think that is a useful restriction. 
Patch level should mean no API level changes, thus no changes in typing.

Josverl: Thanks, I think that is a useful restriction. 
Patch level should mean no API level changes, thus no changes in typing.

Josverl: From a MicroPyton perspective I concur, as it does not support named attributes for this, but I don't necessarily want to force that on other implementations.

docs/spec/directives.rst
carljm: I don't understand what the possible use case is here. Can you elaborate? 

To be clear, at runtime equality checks with `sys.version_info` will never return `True` unless you provide all five elements of the `sys.version_info` tuple. Every other equality check will always return `False`. So this is only usable if you check all five elements, and making _that_ useful requires that type checkers allow specifying the Python version down to "releaselevel" and "serial" level in their configurations.

From a type checker implementer perspective, I feel pretty strongly that we would never implement support for this, because it is a bunch of work and it is not useful in any realistic case. I see this as a blocker for this proposal.

docs/spec/directives.rst
Josverl: The ecosystem is larger than just one implementation. 
But I know I can't properly assess the effort needed to update and maintain the tooling I hope to influence.
So I'll leave the weighing up to those who can.

Josverl: The ecosystem is larger than just one implementation. 
But I know I can't properly assess the effort needed to update and maintain the tooling I hope to influence.
So I'll leave the weighing up to those who can.

docs/spec/directives.rst
Josverl: Fair point, I think; as you suggested; Major.Minor

docs/spec/directives.rst
Josverl: Thank you for the clear feedback,
Do I understand correctly that Major.Minor would  be acceptable for comparison?

Josverl: Thank you for the clear feedback,
Do I understand correctly that Major.Minor would  be acceptable for comparison?

Josverl: Thank you for the clear feedback,
Do I understand correctly that Major.Minor would  be acceptable for comparison?

carljm: Yes, I think that all `sys.version_info` checks should only handle major/minor. That implies that support for inequality/equality checks against all of `sys.version_info` must be removed from the proposal, because there is no way to perform a useful equality/inequality check against all of `sys.version_info` (as shown in this example) that only considers major/minor.

I'm fine with supporting `sys.version_info[:2] == (3, 14)`! So if we are explicit that equality/inequality checks are supported against only the first two elements, but not against the entire tuple, that resolves my objection.

docs/spec/directives.rst
```diff
@@ -145,8 +145,13 @@ left undefined by the typing spec at this time.
 Version and platform checking
 -----------------------------
 
-Type checkers are expected to understand simple version and platform
-checks, e.g.::
+
+Type checkers should support narrowing based on ``sys.version_info``, ``sys.platform``, and ``sys.implementation.name`` checks.
```
erictraut: I think it's important for the language to be very specific about the *exact* syntactical forms that are supported. Type checkers need to evaluate these expressions early in the analysis process, prior to type evaluation. That means they need to do exact tree matching of the AST. That means it's important to reduce the variations that are supported. For example, `sys.version_info <= (3, 10)` is fine, but `version_info <= (3, 10)` is not, nor is `(3, 10) > version_info`.

docs/spec/directives.rst
erictraut: I agree with Carl. Equality and inequality checks don't make sense. In think the only operators that make sense are `<` and `>=`.

docs/spec/directives.rst
```diff
@@ -155,12 +160,65 @@ checks, e.g.::
...
-Don't expect a checker to understand obfuscations like
+  if sys.platform in ("linux", "darwin"):
+      # Platform-specific stubs for Linux and macOS
+      ...
+
+**Supported patterns:**
+
+* Equality: ``sys.platform == "linux"``
+* Inequality: ``sys.platform != "win32"``
+* Membership: ``sys.platform in ("linux", "darwin")``
```
erictraut: I'm not convinced that containment should be supported here. There's already a way to express this in a way that all tools today support. I understand the argument that this is less verbose, but it's not that common for checks to include more than one platform, so I don't think there's a compelling argument to force all tools to support this additional form.

docs/spec/directives.rst
```diff
@@ -155,12 +160,65 @@ checks, e.g.::
...
-      # Posix specific definitions
 
-Don't expect a checker to understand obfuscations like
+  if sys.platform in ("linux", "darwin"):
+      # Platform-specific stubs for Linux and macOS
+      ...
+
+**Supported patterns:**
+
+* Equality: ``sys.platform == "linux"``
```
erictraut: Here also, I think it would be useful for the spec to be specific that `sys.platform == "linux"` is supported but other variants like `platform == "linux"` and `"linux" == sys.platform` are not.

docs/spec/directives.rst
```diff
@@ -155,12 +160,65 @@ checks, e.g.::
...
+  if sys.implementation.name == "cpython":
+      # CPython-specific stub
+  if sys.implementation.name == "micropython":
+      # MicroPython-specific stub
+
+**Supported patterns:**
+
+* Equality: ``sys.implementation.name == "cpython"``
+* Inequality: ``sys.implementation.name != "cpython"``
+* Membership: ``sys.implementation.name in ("pypy", "graalpy")``
```
erictraut: Here again, I don't think there's good justification for supporting a containment operator.

erictraut (changes requested):

docs/spec/directives.rst
Josverl: If all sys.version_info comparisons are only supported on the first two elements of the version tuple, the I think that therer is no need to support the explicit form `sys.version_info[:2]` either. 
I have removed that option.

docs/spec/directives.rst
Josverl: I see your point; removed

docs/spec/directives.rst
Josverl: adjusted accordingly.

docs/spec/directives.rst
Josverl: makes sense.
I added a note to explicitly clarify that.

docs/spec/directives.rst
JelleZijlstra: For deciding on issues like this it would be helpful to have a little summary of what type checkers currently support (like what I did in https://discuss.python.org/t/spec-change-clarify-that-tuple-should-not-be-prohibited-as-an-argument-to-type/105590/7). I'll gather a few variants and summarize it.

I think part of this change will be codifying what is already supported universally, and another part will be adding support for more features. The first group should be uncontroversial, and in the second group we should only force work on type checker authors if there's a clear use case.

Josverl: If possible this would really help with creating readable and maintainable type stubs for MicroPython.
even with the proposed sys.implementation.name check we still see significant API differences due the underlying MCU vendor SDKs being significantly different.
Though MicroPython tries to abstract much of these differences away that is not entirly possible as the underlying SDK or hardware is simply different. This is a common source of errors when code is ported from one MCU architecture to another. 
For instance Timers are available on all platforms, but with many different default values. 

While the below would work 
```py
# machine.pyi 
class Timer():
    """"Timer object"""

    if sys.implementation.name == "micropython" and (sys.platform == "esp32" or sys.platform ==  "mimxrt" or  sys.platform ==  "rp2" or sys.platform ==  "samd" or sys.platform ==  "stm32" or sys.platform ==  "alif" or sys.platform ==  "webassembly"): 
        @overload
        def __init__(
            self,
            id: int,
            /,
            *,
            mode: int = PERIODIC,
            period: int | None = None,
            callback: Callable[[Timer], None] | None = None,
            hard: bool | None = None,
        ):...
    elif sys.implementation.name == "micropython" and (sys.platform == "esp8266" or sys.platform ==  "unix" or  sys.platform ==  "windows"  or sys.platform ==  "zephyr"): 
        @overload
        def __init__(
            self,
            id: int = -1,
            /,
            *,
            mode: int = PERIODIC,
            period: int | None = None,
            callback: Callable[[Timer], None] | None = None,
        ):...
```
the below is much simpler to understand and maintain.

```py
class Timer():
    """"Timer object"""

    if sys.implementation.name == "micropython" and (sys.platform in ("esp32", "mimxrt", "rp2", "samd", "stm32", "alif", "webassembly")): 
        @overload
        def __init__(
            self,
            id: int,
            /,
            *,
            mode: int = PERIODIC,
            period: int | None = None,
            callback: Callable[[Timer], None] | None = None,
            hard: bool | None = None,
        ):...
    elif sys.implementation.name == "micropython" and (sys.platform in ("esp8266", "unix", "windows", "zephyr")): 
        @overload
        def __init__(
            self,
            id: int = -1,
            /,
            *,
            mode: int = PERIODIC,
            period: int | None = None,
            callback: Callable[[Timer], None] | None = None,
        ):...
```

I think it would be reasonable to explicitly restrict this to "a tuple of literal strings",assuming that simplies the implementation.
Negative membership option could also be omitted , I think that would still be sufficient.

docs/spec/directives.rst
Josverl: I addedd this for consistency, but I'm OK to remove ``sys.implementation.name in ("pypy", "graalpy")`` from the spec to simplify the implementation.

docs/spec/directives.rst
AlexWaygood: > Type checkers need to evaluate these expressions early in the analysis process, prior to type evaluation. That means they need to do exact tree matching of the AST. That means it's important to reduce the variations that are supported. For example, `sys.version_info <= (3, 10)` is fine, but `version_info <= (3, 10)` is not, nor is `(3, 10) > version_info`.

(None of this is true for how ty implements `sys.version_info` or `sys.platform` branches, FWIW)

carljm: I don't think that changes anything about what should be supported in the spec though, since it's true for all other type checkers.

docs/spec/directives.rst
```diff
@@ -145,8 +145,27 @@ left undefined by the typing spec at this time.
...
 
-Type checkers are expected to understand simple version and platform
-checks, e.g.::
+Type checkers should support narrowing based on:
+    * ``sys.version_info``
+    * ``sys.platform``
+    * ``sys.implementation.version``
+    * ``sys.implementation.name``
+
+The `<comparison>` patterns for these variables are described in more detail in the following paragraphs.
```
JelleZijlstra: I would add that the spec only requires support for exactly `sys.platform`, `sys.version_info` etc., with no aliases or `from` imports. Type checkers are free to support more but these are the only required ones.

docs/spec/directives.rst
```diff
@@ -155,13 +174,117 @@ checks, e.g.::
...
+Type checkers should support ``sys.platform`` comparisons:
+
+    Supported patterns:
+        * ``sys.platform <comparison> <string literal>``
+        * ``sys.platform in <tuple of string literals>``
+
+    `<comparison>` can be one of the following:
+        * Equality: ``==``
+        * Inequality: ``!=``
+        * Membership: ``in``
```
JelleZijlstra: This should be removed, since `comparison` only applies to the variant with a single string. We shouldn't support `sys.platform in "linuxdarwin"`.

docs/spec/directives.rst
```diff
@@ -155,13 +174,117 @@ checks, e.g.::
...
+
+
+sys.platform checks
+^^^^^^^^^^^^^^^^^^^
+
+Type checkers should support ``sys.platform`` comparisons:
+
+    Supported patterns:
+        * ``sys.platform <comparison> <string literal>``
+        * ``sys.platform in <tuple of string literals>``
```
JelleZijlstra: Or `not in`?

docs/spec/directives.rst
```diff
@@ -155,13 +174,117 @@ checks, e.g.::
...
+Type checkers should support ``sys.implementation.name`` comparisons:
+
+Supported patterns:
+    * ``sys.implementation.name <comparison> <string literal>``
+    * ``sys.implementation.name in <tuple of string literals>``
+
+    `<comparison>` can be one of the following:
+        * Equality: ``==``
+        * Inequality: ``!=``
+        * Membership: ``in``
```
JelleZijlstra: Same notes as for sys.platform

docs/spec/directives.rst
Josverl: added explanantion.

docs/spec/directives.rst
Josverl: the `not` / `and` / `or` are described further down to avoid neededing to repeat then for every possible clause, although `!=` _is_ repeated.
I could move that section to the top if that is clearer.

--------------
Multiple comparisons can be combined with:
    * A ``not`` unary operator
    * An ``and`` or ``or`` binary operator
--------------

docs/spec/directives.rst
Josverl: I now see this could be misread as allowing `in <string>` while I intended only `in <tuple of string literals>` 

I will remove this line, and keep the ``sys.platform in <tuple of string literals>``

docs/spec/directives.rst
Josverl: ditto

docs/spec/directives.rst
JelleZijlstra: That's different, it would allow `not (sys.platform in ("a", "b"))`, not `sys.platform not in ("a", "b")`.

Josverl: I see your point, then I think it is clearer to just enumerate all option directly, 

    Supported patterns:
        * ``sys.platform == <string literal>``
        * ``sys.platform != <string literal>``
        * ``sys.platform in <tuple of string literals>``
        * ``sys.platform not in <tuple of string literals>``

Josverl:
I have updated the text to better clarify the exact comparisons for each of the supported attributes,
 excluded the use of named attributes, 
 and did some more word-smithing to reduce unneeded verbosity.
I have also added highlighting to the examples, as a raiding aide.

docs/spec/directives.rst
Josverl: Added "Use of named attributes is not supported." to the spec.

docs/spec/directives.rst
Josverl: I have updated the spec based on this conversations and on Jelle's scan of implemented features/type checker.

docs/spec/directives.rst
```diff
@@ -145,23 +145,141 @@ left undefined by the typing spec at this time.
...
-``"".join(reversed(sys.platform)) == "xunil"``.
+sys.version_info checks
+^^^^^^^^^^^^^^^^^^^^^^^^
+
+Type checkers should support the following comparison patterns:
+    * ``sys.version_info >= <2-tuple>``
+    * ``sys.version_info < <2-tuple>``
+
+Comparisons checks are only supported against the first two elements of the version tuple. 
+Use of named attributes is not supported.
```
AlexWaygood:
```suggestion
Use of named attributes is not mandated.
```

Josverl: IIRC @carljm requested that addition. 
Agree that `not mandated` is clearer for the metadata.

< Same on line 238>

docs/spec/directives.rst
```diff
@@ -145,23 +145,141 @@ left undefined by the typing spec at this time.
...
+
+    ``sys.implementation.version`` is a tuple, in the same format as sys.version_info. However it represents the version of the Python implementation rather than the version of the Python language. 
+    This has a distinct meaning from the specific version of the Python language to which the currently running interpreter conforms. For CPython this is the same as `sys.version_info`.
+
+
+No support for complex expressions
+^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
+
+Type checkers are only required to support the above patterns, and are not required to evaluate complex expressions involving these variables.
+For example, the pattern ``sys.platform == "linux"`` is supported but other syntax variants such as ``platform == "linux"`` and ``"win" not in sys.platform`` are not supported.
```
Josverl: @erictraut , @AlexWaygood 

Alex proposed elsewhere to change `not supported` to `not mandated`

as this section also uses that, I wanted to check if there is a need to change the wording here as well.

Also on the light of [Jelle's functionality scan](https://discuss.python.org/t/proposal-to-improve-support-for-other-python-platforms-in-the-typing-specification/91877/7)

docs/spec/directives.rst
AlexWaygood: I think "not mandated" is clearer that type checkers are _permitted_ to support something additionally if they wish to, but that they are not _required_ to do so. So in general I prefer "not mandated".

The one exception for me might be comparisons against length-3-or-greater tuples for `sys.version_info` or `sys.implementation.version`. I think it's usually actually incorrect for a type checker to support those unless the type checker actually either infers the micro version of the Python install the user has locally or offers a configuration knob for the micro version to be specified. To my knowledge, no type checker currently does. So for comparisons between `sys.version_info` or `sys.implementation.version` and tuples of >=3 length, I think the stronger language of "not supported" (or maybe even stronger than that) is probably still appropriate

docs/spec/directives.rst
AlexWaygood: I fully agree with Eric here FWIW. Adding support for this will add significant complexity to type checkers, and this is the first time I've seen it requested. If we want to ask type checkers to add support for `in`/`not in` comparisons with `sys.platform` and `sys.implementation.name`, I think it should be a wholly separate proposal to the proposal that asks type checkers to add initial support for comparisons against `sys.implementation.version` and `sys.implementation.name` in the same way that they already do for `sys.version_info` and `sys.platform`.

Josverl:
Dear TC, 
Is there a decision on this PR yet, or have I missed a step in the process?

If so kindly let me know how to proceed.

docs/spec/directives.rst
carljm:
```suggestion
Support for checks against named attributes of `sys.version_info` is not required.
```

docs/spec/directives.rst
```diff
@@ -145,23 +145,141 @@ left undefined by the typing spec at this time.
...
+
+Type checkers should support the following comparison patterns:
+    * ``sys.platform == <string literal>``
+    * ``sys.platform != <string literal>``
+    * ``sys.platform in <tuple of string literals>``
+    * ``sys.platform not in <tuple of string literals>``
+
+    Common values: ``"linux"``, ``"darwin"``, ``"win32"``, ``"emscripten"``, ``"wasi"``
+
+The membership checks ``in``  and ``not in``, only support simple containment testing to a tuple of literal strings.
```
JelleZijlstra:
```suggestion
The membership checks ``in`` and ``not in`` only support simple containment testing with a tuple of literal strings.
```

JelleZijlstra (commented):
I'm generally supportive but I think this change needs some more editing, and I'd like support in at least some type checkers (at least an open PR) before we start mandating this pattern.

docs/spec/directives.rst
Josverl: Reworded to 
> Comparisons checks are only supported against the first two elements of the version tuple. 
> Use of named attributes is not mandated.

Josverl:
Dear typing council,
I have been waiting patiently , but there seems to be no progress on this.
As far as I can see I have actioned - or responded - to all feedback, 
Reviewers have not responded to the updates they requested. Eric has stepped down for the TC since - so that is understandable.

However now it appears that this PR is in a Catch-22 situation;  
Waiting for One or more typecheckers to implement this - but for typecheckers to implement this it should be in the spec.

Lacking response I plan close this in a few weeks

docs/spec/directives.rst
```diff
@@ -154,23 +154,141 @@ left undefined by the typing spec at this time.
...
+       # Python 3.11 and lower
+
+sys.platform checks
+^^^^^^^^^^^^^^^^^^^
+
+Type checkers should support the following comparison patterns:
+    * ``sys.platform == <string literal>``
+    * ``sys.platform != <string literal>``
+    * ``sys.platform in <tuple of string literals>``
+    * ``sys.platform not in <tuple of string literals>``
```
jorenham: Ruff ([PLR6201](https://docs.astral.sh/ruff/rules/literal-membership/)) will report an error for this tuple membership check, and I tend to agree with ruff that a set literal would be more idiomatic here.

docs/spec/directives.rst
```diff
@@ -154,23 +154,141 @@ left undefined by the typing spec at this time.
...
-Don't expect a checker to understand obfuscations like
-``"".join(reversed(sys.platform)) == "xunil"``.
+sys.version_info checks
+^^^^^^^^^^^^^^^^^^^^^^^^
+
+Type checkers should support the following comparison patterns:
+    * ``sys.version_info >= <2-tuple>``
+    * ``sys.version_info < <2-tuple>``
+
+Comparisons checks are only supported against the first two elements of the version tuple.
```
jorenham: As I mentioned a while back in https://discuss.python.org/t/proposal-to-improve-support-for-other-python-platforms-in-the-typing-specification/91877/13, there are genuine use cases for also comparing against the patch version, e.g. in typeshed: https://github.com/python/typeshed/blob/4f84ac178fb23541475f9038840c5b149834374f/stdlib/heapq.pyi#L9-L11

So I propose we loosen this restriction to also allow for 3-tuples.

docs/spec/directives.rst
```diff
@@ -154,23 +154,141 @@ left undefined by the typing spec at this time.
...
+
+.. code-block:: python
+   :caption: Example `sys.version_info`
+   :emphasize-lines: 2
+
+   import sys
+   if sys.version_info >= (3, 12):
+       # Python 3.12+
+   else:
+       # Python 3.11 and lower
```
jorenham: It might be good to include `elif` in this example, so that it's clear that `elif` is also expected to be supported, and not just `if` and `else`. For example as:

```suggestion
   elif sys.version_info >= (3, 11):
       # Python 3.11.*
   else:
       # Python 3.10 and lower
```

docs/spec/directives.rst
jorenham: In the wild I've seen things like `sys.platform.startswith("freebsd")` a couple of times, because in this case there's also a version number in the platform string (e.g. `"freebsd8"`). So how about we also allow `sys.platform.startswith(<string literal>)`?

docs/spec/directives.rst
```diff
@@ -154,23 +154,141 @@ left undefined by the typing spec at this time.
...
+    if platform == "linux":
+        # Linux specific code
+    if "win" not in sys.platform:
+        # Non-Windows specific code
+
+
+Configuration
+^^^^^^^^^^^^^
+
+Type checkers should provide configuration or CLI options to specify target sys.version, sys.platform, sys.implementation.name and sys.implementation.version.
```
jorenham:
```suggestion
Type checkers should provide configuration or CLI options to specify target ``sys.version``, ``sys.platform``, ``sys.implementation.name`` and ``sys.implementation.version``. 
```

davidhalter:
> Waiting for One or more typecheckers to implement this - but for typecheckers to implement this it should be in the spec.

This is not the case at all. It is very typical for type checkers to implement features even if they are not in the spec. It is very helpful to have feedback from type checker authors to be able to understand potential problems with spec changes.

I think this has worked quite well in the past.and if a feature is useful enough there is usually a type checker that will implement it.

docs/spec/directives.rst
carljm: As also discussed in that thread, I think mandating support for 3-tuple comparisons opens up a can of worms regarding what "support" for 3-tuple comparisons even means. We cannot require type checkers to always precisely support such comparisons, because type checkers may not have access to precise micro Python versions: type checker configurations (and `pyproject.toml` etc) typically do not specify micro versions, and I don't think we want to require users to always start requiring them. (Whereas I think in practice all type checkers already do require users to always provide major/minor version in some way, explicitly or implicitly.) So what is the expected behavior if micro Python version is not available to the type checker?

For example, ty currently "supports" `if sys.version_info >= (3, 14, 1)` (when Python version is configured to 3.14) in the "sound" way, by treating it as "unknown truthiness" and considering that either path may be taken (since actual Python version could be 3.14.0 or 3.14.1+). Would that behavior be considered to meet the bar required of type checkers here, or not? In some sense that behavior is a "regression" from writing `if sys.version_info >= (3, 14)`, where ty would always infer a definite truthiness.

("Get the micro version from the runtime Python environment" is not a reliable fallback. It requires running the Python executable, which is avoided by at least ty since it can ~double the overall runtime of checking a small project using a fast type checker. And type checking -- e.g. for a small project without dependencies -- doesn't even necessarily require a Python runtime environment to begin with.)

carljm:
```suggestion
Comparison checks are only supported against the first two elements of the version tuple. 
```

docs/spec/directives.rst
```diff
@@ -154,23 +154,141 @@ left undefined by the typing spec at this time.
...
-``"".join(reversed(sys.platform)) == "xunil"``.
+sys.version_info checks
+^^^^^^^^^^^^^^^^^^^^^^^^
+
+Type checkers should support the following comparison patterns:
+    * ``sys.version_info >= <2-tuple>``
+    * ``sys.version_info < <2-tuple>``
+
+Comparisons checks are only supported against the first two elements of the version tuple. 
+Use of named attributes is not mandated.
```
carljm:
```suggestion
Type checkers are not expected to support comparisons with named attributes of `sys.version_info`.
```

docs/spec/directives.rst
carljm: ty already supports `sys.platform.startswith`; I don't have any objection there.

Supporting set literals will be a little tricky in ty, but should be doable, and I'm not opposed to requiring support for it. I don't think the performance motivation of PLR6201 typically applies much to `sys.version_info` comparisons, but it is awkward if this rule is generally being applied in a codebase and has to be specifically ignored for `sys.version_info` checks.

docs/spec/directives.rst
```diff
@@ -154,23 +154,141 @@ left undefined by the typing spec at this time.
 Version and platform checking
 -----------------------------
 
-Type checkers are expected to understand simple version and platform
-checks, e.g.::
+Type checkers should support narrowing based on:
```
carljm: I don't think "narrowing" is the right term here.
```suggestion
Type checkers should understand code paths as definitely reachable or not reachable due to comparison tests against these symbols:
```

docs/spec/directives.rst
```diff
@@ -154,23 +154,141 @@ left undefined by the typing spec at this time.
...
+
+.. code-block:: python
+   :caption: Example `sys.implementation.name`
+   :emphasize-lines: 2,4
+
+   import sys
+   if sys.implementation.name == "cpython":
+       # CPython-specific stub
+   if sys.implementation.name == "micropython":
+       # MicroPython-specific stub
```
carljm, line 242: This raises similar questions as the three-tuple `sys.version_info` feature request above. Type checkers may commonly not have any source of information as to the Python implementation the checked code is expected to run under. What should be their behavior in this case?

I think for `sys.implementation.name` a reasonable choice would be "assume `cpython` unless known otherwise". Is that just a choice that is up to the individual type checker, or should it be standardized? It seems like if libraries begin to use these comparisons under the assumption type checkers will understand it, it can become a compatibility concern how type checkers behave here.

docs/spec/directives.rst
```diff
@@ -154,23 +154,141 @@ left undefined by the typing spec at this time.
...
+
+sys.implementation.version checks
+^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
+
+Type checkers should support the following comparison patterns:
+    * ``sys.implementation.version >= <2-tuple>``
+    * ``sys.implementation.version < <2-tuple>``
+
+Comparisons checks are only supported against the first two elements of the implementation version tuple.
+Use of named attributes is not mandated.
```
carljm:
```suggestion
Type checkers are not required to support comparisons against named attributes of `sys.implementation.version`.
```

docs/spec/directives.rst
```diff
@@ -154,23 +154,141 @@ left undefined by the typing spec at this time.
...
+   :emphasize-lines: 2,4
+
+   import sys
+   if sys.implementation.name == "cpython":
+       # CPython-specific stub
+   if sys.implementation.name == "micropython":
+       # MicroPython-specific stub
+
+
+sys.implementation.version checks
```
carljm, line 245: Same as above, I think we should probably be explicit here that when a type checker has no "implementation" information, it should assume "CPython, and implementation version matches `sys.version_info`".

It's less clear to me what should happen if a type-checker is told that the implementation is not CPython, but is not given any specific version information. I guess this could be an error? Otherwise I'm not sure how type-checkers should guess at the implementation version.

docs/spec/directives.rst
```diff
@@ -154,23 +154,141 @@ left undefined by the typing spec at this time.
...
+^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
+
+Type checkers should support the following comparison patterns:
+    * ``sys.implementation.version >= <2-tuple>``
+    * ``sys.implementation.version < <2-tuple>``
+
+Comparisons checks are only supported against the first two elements of the implementation version tuple.
```
carljm:
```suggestion
Comparison checks are only supported against the first two elements of the implementation version tuple.
```

docs/spec/directives.rst
```diff
@@ -154,23 +154,141 @@ left undefined by the typing spec at this time.
...
+   import sys
+   if sys.implementation.name == "pypy" and sys.implementation.version >= (7, 3):
+       # PyPy version 7.3 and above
+   if sys.implementation.name == "micropython" and sys.implementation.version >= (1, 24):
+       # MicroPython version 1.24 and above
+
+.. note::
+
+    ``sys.implementation.version`` is a tuple, in the same format as sys.version_info. However it represents the version of the Python implementation rather than the version of the Python language. 
+    This has a distinct meaning from the specific version of the Python language to which the currently running interpreter conforms. For CPython this is the same as `sys.version_info`.
```
carljm: This seems like it belongs more as introductory text in this section, rather than as a note.

docs/spec/directives.rst
```diff
@@ -154,23 +154,141 @@ left undefined by the typing spec at this time.
...
+
+
+No support for complex expressions
+^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
+
+Type checkers are only required to support the above patterns, and are not required to evaluate complex expressions involving these variables.
```
carljm: If the expectation is that `and` and `or` boolean expressions are supported, that should be explicitly discussed here. Rather than saying "are not required to evaluate complex expressions" (which is vague), we should simply specify the forms that must be supported, and that no other form is required.

docs/spec/directives.rst
```diff
@@ -154,23 +154,141 @@ left undefined by the typing spec at this time.
...
+For example, the pattern ``sys.platform == "linux"`` is supported but other syntax variants such as ``platform == "linux"`` and ``"win" not in sys.platform`` are not mandated.
+
+Therefore checkers are **not required** to understand obfuscations such as:
```
carljm, line 275: The "reversed" example is an obfuscation, but `"win" not in sys.platform` is not an obfuscation, and `from sys import platform; platform == "linux"` is definitely not an obfuscation.

Let's avoid value judgments and just give a clear presentation of the forms that are supported, and some example forms that type checkers are not required to support.

docs/spec/directives.rst
```diff
@@ -154,23 +154,141 @@ left undefined by the typing spec at this time.
...
+
+    ``sys.implementation.version`` is a tuple, in the same format as sys.version_info. However it represents the version of the Python implementation rather than the version of the Python language. 
+    This has a distinct meaning from the specific version of the Python language to which the currently running interpreter conforms. For CPython this is the same as `sys.version_info`.
+
+
+No support for complex expressions
+^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
+
+Type checkers are only required to support the above patterns, and are not required to evaluate complex expressions involving these variables.
+For example, the pattern ``sys.platform == "linux"`` is supported but other syntax variants such as ``platform == "linux"`` and ``"win" not in sys.platform`` are not mandated.
```
carljm: Not sure why we need to give exactly the same examples both in prose and then below in a code snippet.

carljm (commented):
Thanks for your patience, @Josverl. I'm not sure we've fully reached consensus yet on what should be supported here, but I also generally support making this clearly specified. I also think this still needs some editing to give a clear and unambiguous presentation of the requirements.

docs/spec/directives.rst
jorenham: My instinct in case of `sys.version_info >= (x, y, z)` and `sys.version_info < (x, y, z)` when statically only `x` and `y` are known is to use the latest available `z`, because that's what I expect to be most common in practice. I'm sure there will be edge-cases that this could be problematic for, but on the macro I this this'll solve more problems that it'll cause.

But having just written this, I now see how the complexity of this solution might indeed not be worth it for a problem as niche as rare as this one.

So how about we don't *require* type-checkers to support 3-tuples, but explicitly state that type-checkers *may* choose to also support 3-tuples?

docs/spec/directives.rst
carljm: I guess an alternative behavior to specify here would be that Python type checkers should not assume any particular implemention -- that is, should not treat these checks in any special way -- _unless_ a particular known implementation has been configured.

Josverl:
Thanks all for the renewed attention and comments.
 - I'll try to clarify where I can do 
 - and keep a list of topics that need a decision one way or another

docs/spec/directives.rst
carljm: > use the latest available `z`

Type checkers don't really have a way to know the latest available `z` at any point in time, either, so I think this rule would have to be something more like "assume `z` is `∞`", or something like that. Which will lead to weird behavior in some edge cases (imagine a new feature is introduced in 3.15 beta and someone introduces an `if sys.version_info <= (3, 14)` in typeshed when it is introduced, and immediately type checkers all start assuming every user is running 3.15 beta or later). But you may be right that in practice it does the right thing most of the time.

> So how about we don't _require_ type-checkers to support 3-tuples, but explicitly state that type-checkers _may_ choose to also support 3-tuples?

I think in general the assumption of this entire PR should be that type checkers can always choose to support more than what is required here, and the wording in all sections should be such as to leave that possibility open. If this is the route we take, I don't know that we need to specially emphasize it for 3-tuples, but we can.

docs/spec/directives.rst
Josverl: change to set and added `sys.platform.startswith(<string literal>)`

carljm: To be clear, I think that both tuple and set literals should be specified as supported -- it would be surprising IMO if sets work and tuples don't.

This is perhaps an argument against supporting sets -- it opens a bit of a slippery slope: why not lists? etc

docs/spec/directives.rst
Josverl: fair point. there should be no guessing involved, And defaulting to CPyton would be the most logical thing to do.

As an example a device that happens to be connected:
```
MicroPython v1.28.0 on 2026-04-06; Wio Terminal D51R with SAMD51P19A
Type "help()" for more information.
>>> import sys
>>> sys.version
'3.4.0; MicroPython v1.28.0 on 2026-04-06'
>>> sys.implementation
(name='micropython', version=(1, 28, 0, ''), _machine='Wio Terminal D51R with SAMD51P19A', _mpy=7942, _build='SEEED_WIO_TERMINAL')
```
If I want to typecheck an app for this device and firmware I would need to supply the typechecker with:

- sys.implementation.name="micropython"
- sys.implementation.version=(1,28)
- sys.version = (3,9) * 

using the relevant configuration options for that checker

* MicroPython uses a subset of features from 3.5-3.11, generally 3.9/3.10 make a good base fit.
BUt that is just one implementation.

If not provided explicit information through: typechecker config, environment or switches, or detected python runtime sys.implementation.version should fall-back to sys.version.

docs/spec/directives.rst
Josverl: In my mind it was clear that the default is the current behaviour: CPython.
I have now made that explicit.

I do not think it should be the reposinsibility of the type checker to determine *what* the runtime runtime environment is.
But it must be able to accept that as a (set of) config parameter(s) , same as if I typecheck a app for 3.10 windows syntax from a 3.14 venv on Linux.

docs/spec/directives.rst
Josverl: - booleans operators are already on lines 165~167.
- reworded to be more definitive

docs/spec/directives.rst
jorenham: > I think in general the assumption of this entire PR should be that type checkers can always choose to support more than what is required here, and the wording in all sections should be such as to leave that possibility open. If this is the route we take, I don't know that we need to specially emphasize it for 3-tuples, but we can.

In theory, yes. But there have been many time where type-checker maintainers weren't willing to implement a certain feature, purely because it wasn't in the typing spec. So that's why I suggested adding this as explicitly optional.

docs/spec/directives.rst
Josverl: Added a table in the config section to add clarity.

docs/spec/directives.rst
Josverl: my original proposal was tuples, as that is one of the core building blocks of Python.
I don't think that for this scale sets have a real benefit over tuples.

I'd like it to be straight and narrow though , 
requiring tuple + set = is still ok , though I notice that ruff mentions **'This rule is unstable and in [preview](https://docs.astral.sh/ruff/preview/).'**
 adding lists would start to 🛝 
 
 Pausing edits until there is consensus

Josverl:
Again thanks for the feedback and suggestions to improve the spec.

I processed the comments and suggestion - and commented or marked them as resolved to keep track.
Please feel free to Unresolve or add additional comments as needed.
I have extended the `Configuration` section to help clarify how a type checker could be configured including default values or approaches for defaults.

Open topics.

**1. 2-Tuple or 3-Tuple version comparisons.**
  [see comments](https://github.com/python/typing/pull/2173#discussion_r3745092032)
  - requirement is for 2-Tuple checks 
  - type checkers are allowed to also process 3-Tuple checks
  
**2. Should `in` / `not in` membership checks use `tuple`, `set` or `list`.**
   intent was tuples, was updated to set , now waiting for consensus

    * ``sys.platform in <tuple/set/list of string literals>``
    * ``sys.platform not in <tuple/set/list of string literals>``
    * ``sys.implementation.name in <tuple/set/listof string literals>``
    * ``sys.implementation.name not in <tuple/set/list of string literals>``
    
**3. Scope**
  [as raised](https://github.com/python/typing/pull/2173#discussion_r2838124588) by @AlexWaygood 
  Is the scope of this change too large (it has grown since that comment) and should there be a implementation timeline /   staged implementation.
  I have refrained from answering this - as I am not the one that can do the implementation.

  
  I think the commits should be squashed when merging - I've kept them separate for now.

carljm:
We could eliminate open-question (2) and at least partially address (3) by just considering containment checks as out of scope to specify for now. Considering they don't add any new _capability_, just allow writing more concise code, they aren't really necessary. We could leave this out of the spec / conformance suite for now and up to individual type checkers. Similarly maybe for `startswith`. Leaving it out doesn't _prevent_ type checkers from supporting it. And including it in the spec / conformance suite doesn't guarantee type-checker support, either, so it's unclear how soon some of these difficult-to-support patterns will be usable in practice anyway, even if they are included here.

jorenham:
> We could eliminate open-question (2) and at least partially address (3) by just considering containment checks as out of scope to specify for now. Considering they don't add any new _capability_, just allow writing more concise code, they aren't really necessary.

+1

> Similarly maybe for `startswith`.

Unlike containment checks, `startswith` can't always be expressed using equality checks. For example `sys.platform` on FreeBSD on Python<3.14 includes the version number ([docs](https://docs.python.org/3/library/sys.html#sys.platform)) , so it can be `"freebsd7"`, `"freebsd8"`, ....,  `"freebsd15"`. This list will grow each two (?) years too. So if you want to check if you're on FreeBSD, you'll have to spell out each previous version + a bunch of future versions. So at least until Python 3.13 is EOL, I think that we should keep `startswith`.

Josverl:
> by just considering containment checks as out of scope to specify for now

That would be disappointing, especially as one of the first responses in May'25 in DPO included :
  >  Supporting both in and not in operators is straightforward.
  
But reality is that there different opinions based on different considerations, and 

`.startwith()` was added based on review comments. ~~More than happy to remove that to simplify matters.~~

Josverl:
Not closed in anger - just a typo

Josverl:
> I think that we should keep `startswith`.

Is there any type checker that support this currently?

AlexWaygood:
> > I think that we should keep `startswith`.
> 
> Is there any type checker that support this currently?

ty

carljm:
At least for ty, `startswith` support is trivial (already there!), don't know how difficult for other type checkers. Point taken that (unlike containment checks) it really does introduce a new (and sometimes needed capability).

Regarding containment checks, for ty tuples are trivial (also already supported!) but mutable types are not. ty handles this quite differently from other type checkers: we use full type inference, not special-cased AST-matching. This makes some thing a lot easier, and some things harder. For mutable container types, we would normally not infer e.g. `["foo", "bar"]` as `list[Literal["foo", "bar"]]` but as `list[str]` (same for `set`), so we normally would lose the literal precision. It should be doable, just not straightforward and I'm not sure how quickly we'd prioritize it. (But once we did support it, we would support it for all containment checks of a literal type against a literal list/set of literal types, not just for `sys.platform` and friends specifically.)

I suspect that for type checkers that do AST matching, containment checks for tuple/list/set should all be equally straightforward.

jorenham:
> Regarding containment checks, for ty tuples are trivial (also already supported!) but mutable types are not. ty handles this quite differently from other type checkers: we use full type inference, not special-cased AST-matching.

Would `frozenset` help?

carljm:
> Would `frozenset` help?

Not much. It would require that we implement https://github.com/astral-sh/ty/issues/2280, which probably mostly requires the same machinery. (The problem here is that `frozenset` doesn't have a literal syntax, so first you have to build a literal container of some other kind, and then pass it to the `frozenset` constructor, so we still end up promoting literals in that process.)

To be clear, I don't want you to weight this ty implementation issue too much -- I'd rather first hear from other typing council members what they think a good scope is here.

Josverl:
Gentle reminder for additional feedback.

docs/spec/directives.rst
```diff
@@ -154,23 +154,157 @@ left undefined by the typing spec at this time.
...
-``"".join(reversed(sys.platform)) == "xunil"``.
+sys.version_info checks
+^^^^^^^^^^^^^^^^^^^^^^^^
+
+Type checkers should support the following comparison patterns:
+    * ``sys.version_info >= <2-tuple>``
+    * ``sys.version_info < <2-tuple>``
+
+Comparison checks are only supported against the first two elements of the version tuple.
+It should be noted that type checkers may choose to also support the 3-tuple ``sys.version_info >= <3-tuple>``.
```
JelleZijlstra, line 180:
```suggestion
Type checkers may choose to also support the 3-tuple ``sys.version_info >= <3-tuple>``.
```

docs/spec/directives.rst
```diff
@@ -154,23 +154,157 @@ left undefined by the typing spec at this time.
...
+The membership checks ``in`` and ``not in`` only support simple containment testing with a set of literal strings.
+
+.. code-block:: python
+   :caption: Example `sys.platform`
+   :emphasize-lines: 2,4
+
+   import sys
+   if sys.platform == 'win32':
+       # Windows specific definitions
+   if sys.platform in ("linux", "darwin"):
```
JelleZijlstra, line 216: That's a tuple, not a set, and the current text says only sets are supported. My instinct would be to support both, but I wouldn't mind narrowing it down if that's the consensus.

docs/spec/directives.rst
```diff
@@ -154,23 +154,157 @@ left undefined by the typing spec at this time.
...
+
+
+sys.implementation.version checks
+^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
+
+``sys.implementation.version`` is a tuple, in the same format as sys.version_info. However it represents the version of the Python implementation 
+rather than the version of the Python language. This has a distinct meaning from the specific version of the Python language to which the currently 
+running interpreter conforms. For CPython this is the same as `sys.version_info`.
+
+Type checkers should support the following comparison patterns:
```
JelleZijlstra, line 252: This is only meaningful if you've first checked for `sys.implementation.name`, right? Should we say something about this?

docs/spec/directives.rst
```diff
@@ -154,23 +154,157 @@ left undefined by the typing spec at this time.
...
+    if platform == "linux":
+        # Typecheckers will not be required to understand this import alias for sys.platform
+    if "win" not in sys.platform:
+        # Typecheckers will not be required to understand this reversed membership check
+
+
+Configuration
+^^^^^^^^^^^^^
+
+Type checkers must provide configuration or CLI options to specify target ``sys.version``, ``sys.platform``, ``sys.implementation.name`` and ``sys.implementation.version``.
```
JelleZijlstra, line 295: Why "must"? In pycroscope I would implement this simply by using the Python that pycroscope itself is running on.

docs/spec/directives.rst
Josverl: Indeed it is no longer consistent due to other changes.
I prefer to revert to only tuple, as that seems to be best supportable by all, with only ruff issuing a warning currently.

docs/spec/directives.rst
Josverl: Does Pyroscope run on all implementations of Python? 
If so, that would satisfy that, though not in a manner I anticipated.

docs/spec/directives.rst
JelleZijlstra: Is defaulting to sys.version_info really the right call here? On MicroPython that's apparently meaningless, since the version is 1.x instead of 3.x.

docs/spec/directives.rst
jorenham: I recently learned that a set in this case will br optimized using the peephole optimizer, so the performance benefits are real.

Josverl: But ty cannot support it, so the disadvantages are even more real.
If both of the Astral tools could work with a common type, that would be helpful.

davidhalter: > I recently learned that a set in this case will be optimized using the peephole optimizer, so the performance benefits are real.

I don't not what the peephole optimizer optimizes here, but unless I see actual numbers I would not be surprised if small tuples are as fast.

> But ty cannot support it, so the disadvantages are even more real.

Please do not argue like that. There are good arguments above why we should or should not support sets. It is possible for Ty to support it and Carl even argued that it likely should be supported (and then argued against it for another reason). Even if Ty doesn't support it, it still might be helpful to have it in the spec, because other people use other type checkers.

docs/spec/directives.rst
JelleZijlstra: I haven't tried it, but if it doesn't (and somebody asks for the support), I would fix this by getting it to run under that implementation.

JelleZijlstra:
Here's a summary of what type checkers already support (courtesy of Codex):

Proposed behavior | mypy | Pyright | ty | Pyrefly
-- | -- | -- | -- | --
sys.version_info >= (M, m) and < (M, m) | Yes | Yes | Yes | Yes
sys.platform == literal and != literal | Yes | Yes | Yes | Yes
sys.platform.startswith(literal) | Yes | No | Yes | Yes
sys.platform in tuple | No | No | Yes | Yes
sys.platform not in tuple | No | No | Yes | Yes
sys.implementation.name == / != | No | No | No | No
sys.implementation.name in / not in | No | No | No | No
sys.implementation.version >= / < | No | No | No | No
Combining recognized checks with not, and, and or | Yes | Yes | Yes | Yes
Configure target Python version | Yes | Yes | Yes | Yes
Configure target sys.platform | Yes | Partial | Yes | Yes
Configure implementation name or version | No | No | No | No

So nobody supports any `sys.implementation` checks. Support for slightly more advanced `sys.platform` checks is a bit varied. I had Codex look through past GitHub issues and it wasn't able to find much demand for `in`/`not in` checks, which makes me think that part is maybe not worth the trouble. On the other hand, ty and Pyrefly already support those (in Pyrefly's case, apparently as a result of some of the discussion around this PR), and it doesn't seem like an onerous feature to add.

docs/spec/directives.rst
Josverl: I can not think of anything other than a cross implementation lookup table, that would come with its own maintenance and distribution chalenges, to solve this for all or even most Python implementations.
The proposed default is intened to make things simpler to understand for humans on CPython,
on other platforms I think is is acceptable to require this to be specified in a <configurastion>.toml or .json     

And as I mentioned before - it is quite similar to type checking for Windows+ Python 3.10 from 3.14 venv on Linux

docs/spec/directives.rst
Josverl: Are you referring to:
> For CPython this is the same as `sys.version_info`.


If so , I have clarified this in the prose.

docs/spec/directives.rst
Josverl: Fair point, even for typechecker that run outside the target environment, it might be feasable to probe the target environment for the specified ttributes. So I do agree that the spec should not disallow that.
I have reworded to:

> Type checkers must be able to retrieve the information from the python implementation's runtime environment, or provide configuration  ...

docs/spec/directives.rst
Josverl: I have changed this section to:
I hope that is acceptable to all , if not i would really appreceate clear recomendation on how to improve this. 
```rst
Type checkers must support the following comparison patterns:
   <snip> 
    * ``sys.platform in <tuple of string literals>``
    * ``sys.platform not in <tuple of string literals>``
Type checkers may also support the following comparison patterns:
    * ``sys.platform in <set of string literals>``
    * ``sys.platform not in <set of string literals>``
```

docs/spec/directives.rst
```diff
@@ -154,23 +154,160 @@ left undefined by the typing spec at this time.
...
+
+================================  ==========================  ==============  ===========================================================================
+  Symbol                            Suggested Format            Example         Suggested Default
+================================  ==========================  ==============  ===========================================================================
+``sys.version``                   string ``"major.minor"``    ``"3.11"``      The version of the Python interpreter used to run the type checker.
+``sys.platform``                  lowercase string            ``"linux"``     The platform of the Python interpreter used to run the type checker.
+``sys.implementation.name``       lowercase string            ``"cpython"``   ``"cpython"`` unless configured otherwise.
+``sys.implementation.version``    string ``"major.minor"``    ``"3.14"``      The value used for ``sys.version`` unless configured otherwise.
```
carljm: Could we restrict this fallback to CPython and specify what happens when another implementation is configured without an implementation version? I think this still leaves the question from the earlier discussion unresolved.

For example, if the checker is configured for PyPy and Python 3.11, borrowing the language version gives an implementation version of `(3, 11)`. The example above then becomes false:

```python
if sys.implementation.name == "pypy" and sys.implementation.version >= (7, 3):
    ...
```

That would discard the branch even if the target is PyPy 7.3. For non-CPython implementations, could we require an explicit version, retrieve it from the target interpreter, or leave the comparison unknown when neither is available?

docs/spec/directives.rst
```diff
@@ -154,23 +154,160 @@ left undefined by the typing spec at this time.
...
+       ...
+
+
+sys.implementation.name checks
+^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
+
+Type checkers should support comparison patterns:
+    * ``sys.implementation.name == <string literal>``
+    * ``sys.implementation.name != <string literal>``
+    * ``sys.implementation.name in <set of string literals>``
```
carljm, line 230: This should also be updated for consistency with the sets vs tuples discussion above.

docs/spec/directives.rst
```diff
@@ -154,23 +154,160 @@ left undefined by the typing spec at this time.
...
+   if sys.implementation.name == "pypy" and sys.implementation.version >= (7, 3):
+       # PyPy version 7.3 and above
+   if sys.implementation.name == "micropython" and sys.implementation.version >= (1, 24):
+       # MicroPython version 1.24 and above
+
+
+No support for complex expressions
+^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
+
+Type checkers are required to support the above patterns, and are not required to evaluate other comparisons or other syntax variants.
```
carljm: Could we qualify "the above patterns" to preserve the explicitly optional cases? As written, this also requires three-element version comparisons and platform set membership, which the earlier sections say type checkers may choose to support. I think it should be unambiguous that a checker can omit those cases and still conform.

docs/spec/directives.rst
```diff
@@ -154,23 +154,160 @@ left undefined by the typing spec at this time.
 Version and platform checking
 -----------------------------
 
-Type checkers are expected to understand simple version and platform
-checks, e.g.::
+Type checkers should understand code paths as definitely reachable or not reachable due to comparison tests against these symbols:
```
carljm, line 157: It seems like we haven't really discussed the conformance suite in this PR yet, but I think we should add conformance tests for whatever new requirements we decide on in this PR.

`conformance/tests/directives_version_platform.py` currently covers basic version and platform comparisons, but none of the implementation checks, membership checks, `startswith`, or boolean combinations added here.

I would expect coverage of both reachable and unreachable branches, including combinations using `not`, `and`, and `or`. It would also be useful to distinguish required tuple membership from optional set membership and three-element version comparisons, so the tests pin down the intended minimum support.

docs/spec/directives.rst
```diff
@@ -154,23 +154,160 @@ left undefined by the typing spec at this time.
...
+
+Configuration
+^^^^^^^^^^^^^
+
+Type checkers must be able to retrieve the information from the python implementation's runtime environment, or provide configuration or CLI options to specify target ``sys.version``, ``sys.platform``, ``sys.implementation.name`` and ``sys.implementation.version``.
+
+================================  ==========================  ==============  ===========================================================================
+  Symbol                            Suggested Format            Example         Suggested Default
+================================  ==========================  ==============  ===========================================================================
+``sys.version``                   string ``"major.minor"``    ``"3.11"``      The version of the Python interpreter used to run the type checker.
```
carljm: Nit: should this be `sys.version_info`, here and in the configuration prose and implementation-version default? That's the symbol the supported checks use; `sys.version` is the descriptive interpreter version string, including build information. Keeping `"major.minor"` as the suggested configuration encoding seems fine.

docs/spec/directives.rst
```diff
@@ -154,23 +154,160 @@ left undefined by the typing spec at this time.
...
-      # Posix specific definitions
+The comparison patterns for these variables are described in more detail in the following paragraphs.
 
-Don't expect a checker to understand obfuscations like
-``"".join(reversed(sys.platform)) == "xunil"``.
+sys.version_info checks
+^^^^^^^^^^^^^^^^^^^^^^^^
+
+Type checkers should support the following comparison patterns:
+    * ``sys.version_info >= <2-tuple>``
```
carljm, line 176: This leaves the meaning of `<2-tuple>` ambiguous in some critical ways. Does a variable which is earlier bound to a 2-tuple count? What about `(major, get_minor())`?

I think the intent here is that it must be an actual tuple literal containing two integer literals. We should make that explicit.

docs/spec/directives.rst
```diff
@@ -154,23 +154,160 @@ left undefined by the typing spec at this time.
...
+   if sys.implementation.name == "micropython" and sys.implementation.version >= (1, 24):
+       # MicroPython version 1.24 and above
+
+
+No support for complex expressions
+^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
+
+Type checkers are required to support the above patterns, and are not required to evaluate other comparisons or other syntax variants.
+
+Therefore checkers are **not required** to understand obfuscations such as:
```
carljm, line 278: My previous comment here was marked resolved but was not addressed. Not all of the patterns below are "obfuscations" and we should not describe them pejoratively.

docs/spec/directives.rst
carljm: After considering it more, I think we should just leave sets out entirely. I wouldn't even mention them as optional: there's no point, since type checkers are always free to support anything they want beyond the spec.

I don't believe there is any actual performance benefit at this scale, and it's not necessary to add complexity to the spec or to type checkers when comparing to a tuple is perfectly usable and functional. The ruff rule is unstable and I think if this spec change is merged, the ruff rule should just be updated to specifically exclude these typing-supported comparisons, so there is no conflict. I don't think the ruff rule should motivate adding sets to the spec.

docs/spec/directives.rst
```diff
@@ -154,23 +154,160 @@ left undefined by the typing spec at this time.
...
+    * ``sys.platform.startswith(<string literal>)``
+    * ``sys.platform in <tuple of string literals>``
+    * ``sys.platform not in <tuple of string literals>``
+Type checkers may also support the following comparison patterns:
+    * ``sys.platform in <set of string literals>``
+    * ``sys.platform not in <set of string literals>``
+
+    Common values: ``"linux"``, ``"darwin"``, ``"win32"``, ``"emscripten"``, ``"wasi"``
+
+The membership checks ``in`` and ``not in`` only support simple containment testing with a set of literal strings.
```
carljm, line 210: This should be updated for consistency with the decision about sets vs tuples.

docs/spec/directives.rst
```diff
@@ -154,23 +154,160 @@ left undefined by the typing spec at this time.
...
+   if sys.implementation.name == "cpython":
+       # CPython-specific stub
+   if sys.implementation.name == "micropython":
+       # MicroPython-specific stub
+
+
+sys.implementation.version checks
+^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
+
+``sys.implementation.version`` is a tuple, in the same format as sys.version_info. However it represents the version of the Python implementation
```
carljm, line 251:
```suggestion
``sys.implementation.version`` is a tuple, in the same format as ``sys.version_info``. However, it represents the version of the Python implementation 
```

docs/spec/directives.rst
```diff
@@ -154,23 +154,160 @@ left undefined by the typing spec at this time.
...
+   if sys.implementation.name == "micropython":
+       # MicroPython-specific stub
+
+
+sys.implementation.version checks
+^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
+
+``sys.implementation.version`` is a tuple, in the same format as sys.version_info. However it represents the version of the Python implementation 
+rather than the version of the Python language. This has a distinct meaning from the specific version of the Python language to which the currently 
+running interpreter conforms. For CPython (``sys.implementation.name == "cpython"``) this is the same as `sys.version_info`.
```
carljm, line 253:
```suggestion
running interpreter conforms. For CPython (``sys.implementation.name == "cpython"``) this is the same as ``sys.version_info``.
```

carljm, line 253: This should be word-wrapped to around 80 columns -- it doesn't look like we've specified that anywhere, but that's the existing de facto convention in almost all spec files.

docs/spec/directives.rst
```diff
@@ -154,23 +154,160 @@ left undefined by the typing spec at this time.
...
+    if platform == "linux":
+        # Typecheckers will not be required to understand this import alias for sys.platform
+    if "win" not in sys.platform:
+        # Typecheckers will not be required to understand this reversed membership check
+
+
+Configuration
+^^^^^^^^^^^^^
+
+Type checkers must be able to retrieve the information from the python implementation's runtime environment, or provide configuration or CLI options to specify target ``sys.version``, ``sys.platform``, ``sys.implementation.name`` and ``sys.implementation.version``.
```
carljm, line 298: Should be `sys.version_info`, not `sys.version`.

Also word-wrap to 80 columns (throughout).

docs/spec/directives.rst
carljm, line 303:
```suggestion
``sys.version_info``                   string ``"major.minor"``    ``"3.11"``      The version of the Python interpreter used to run the type checker.
```

docs/spec/directives.rst
carljm, line 306: I think the fallback to `sys.version_info` must apply only when the implementation is `cpython`. If a non-CPython implementation is specified, then it should be an error not to also specify an implementation version (unless the type checker can derive it from the runtime environment.)

It's simply wrong to fall back to `sys.version_info` for `sys.implementation.version` if a non-CPython implementation has been specified.
```suggestion
``sys.implementation.version``    string ``"major.minor"``    ``"3.14"``      The value used for ``sys.version_info`` unless configured otherwise.
```

docs/spec/directives.rst
carljm: Commented on the table below, but I agree with @JelleZijlstra that "fallback to `sys.version_info`" is only acceptable if the implementation is CPython. If a non-CPython implementation is specified, and no implementation version is specified or inferable from the runtime environment, that must be a configuration error. (I don't love specifying type checker configuration errors, but we cannot specify that type checkers must fall back to a known-wrong `sys.implementation.version` in this scenario.

docs/spec/directives.rst
```diff
@@ -154,23 +154,160 @@ left undefined by the typing spec at this time.
...
+   :emphasize-lines: 2,4
+
+   import sys
+   if sys.implementation.name == "pypy" and sys.implementation.version >= (7, 3):
+       # PyPy version 7.3 and above
+   if sys.implementation.name == "micropython" and sys.implementation.version >= (1, 24):
+       # MicroPython version 1.24 and above
+
+
+No support for complex expressions
```
carljm, line 273: Let's be more careful with the wording here. "Complex" is not clearly defined, so it's not clear what this heading is saying. Not all unsupported forms can be reasonably called "complex". And "no support" also implies forbidding type checkers from supporting something, which we are not ever doing.

```suggestion
Type checkers are not required to support any forms not explicitly listed as required
```

docs/spec/directives.rst
```diff
@@ -154,23 +154,160 @@ left undefined by the typing spec at this time.
...
+^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
+
+Type checkers are required to support the above patterns, and are not required to evaluate other comparisons or other syntax variants.
+
+Therefore checkers are **not required** to understand obfuscations such as:
+
+.. code-block:: python
+    :caption: Examples of unsupported or overly complex version/platform checks
```
carljm, line 281:
```suggestion
    :caption: Examples of checks that type checkers are not required to support
```

docs/spec/directives.rst
Josverl: So can I consider the tuple/set discussion decided in favor of (literal) tuples ony?

davidhalter: @carljm You wrote:

> I think we should just leave sets out entirely

What is your reasoning about that? Why not support sets/lists as well? To me that would have been the most natural way. I understand that type checkers may support more constructs, but I think it's generally preferable for type checkers to have the same behavior.

My intuition is that normal users would not see a difference between sets / tuples / lists and just use the one they feel most comfortable.

carljm: @davidhalter This same logic could be used to expand the required support in various other ways, too. It's not very intuitive to support `import sys; sys.platform` but not `from sys import platform` or `from sys import platform as plat`, for example. IMO if we are looking to improve intuitive consistency, that's a more obvious change than adding containment support for more types. 

Ultimately we are talking about special casing sufficient patterns to allow expressing what needs to be expressed; we have to draw the line somewhere, so why require support for extra things that don't add expressive utility? If sets and lists, why not dictionary keys? Why not variables that have been assigned a list or set or tuple? All of these are also intuitive extensions from a user perspective.

davidhalter: > If sets and lists, why not dictionary keys?

Because nobody in their right mind would use `sys.platform in {...}.keys()`, while a normal user would probably like to use `sys.platform in [...]` or `sys.platform in {...}`.

I just don't really understand why we should omit this. Do you have reservations because of inference of sets/lists where you would do something like `sys.platform in CONSTANT`?

Jelle implemented it for Mypy already and it looks pretty straightforward: https://github.com/python/mypy/pull/21913

> It's not very intuitive to support import sys; sys.platform but not from sys import platform or from sys import platform as plat, for example. [...] that's a more obvious change than adding containment support for more types.

I'm open to that as well. I probably disagree that it's more obvious, but I do agree that it's unintuitive.

carljm: > Because nobody in their right mind

Not `{...}.keys()`, perhaps, but `SUPPORTED_PLATFORMS = {'linux': ...details..., ...}` then `if sys.platform in SUPPORTED_PLATFORMS`, certainly. It's rarely correct to assert that some feature is something nobody in their right mind would use.

docs/spec/directives.rst
Josverl: > My previous comment here was marked resolved but was not addressed.

Aplogies for that.

docs/spec/directives.rst
Josverl: > but I think we should add conformance tests for whatever new requirements we decide on in this PR.

Makes sense to me, but TBH I personally would need to study how to define these, And happy to do that either in this PR or as a follow-up , but perhaps agree on the 'rules' first

docs/spec/directives.rst
carljm: > Do you have reservations because of inference of sets/lists where you would do something like `sys.platform in CONSTANT`?

Yes, in ty we already support `if sys.platform in CONSTANT` if `CONSTANT` is defined above as a tuple, because tuple types inherently preserve full membership information. It's not feasible to do this accurately for mutable container types, so the level of support in ty will inherently be inconsistent between tuples and list/sets -- the latter will only work if present as an inline literal.

But ultimately I don't feel strongly about this; https://github.com/astral-sh/ruff/pull/28676 will make lists and sets work in ty. In general all membership tests against inline set/list literals can now potentially infer known `Literal[True]` or `Literal[False]` results, with no special-casing of `sys.platform` checks. If the consensus here is to mandate support for lists and sets, I'm OK going ahead with that.

davidhalter: I think if you support `sys.platform in CONSTANT`, your reasoning of only supporting tuples makes much more sense. If a type checker chooses to not support that it feels inconsistent to not support lists/sets.

I still lean towards supporting lists/sets, but I'm totally fine if people disagree and we go the tuple-only way. Maybe @jorenham, @JelleZijlstra or @rchen152 can quickly weigh in here.

JelleZijlstra: I'd prefer also supporting lists and sets. That way, users will be more likely to have code work the way they expect regardless of arbitrary choices, and that's generally a good thing. The cost is that type checkers will need to implement some more complicated support, but that cost seems low enough here that it's worth choosing the user-friendlier version.

jorenham: Sets also have my preference, because for containment checks sets are the idiomatic choice, for which there's also a ruff rule: https://docs.astral.sh/ruff/rules/literal-membership/

docs/spec/directives.rst
Josverl: I understand your point, but in not quite sure what notation works best for this , 

Is it suffient to say ``sys.version_info >= <2-tuple of int literals>``

or would this be clearer : 
- ``sys.version_info >= <tuple[int, int]>`` 
- ``sys.version_info >= <tuple[Literal[int], Literal[int]]>``

docs/spec/directives.rst
Josverl: Do I understand correctly that there is consensus to support **tuple** , **set** and **list** of literals  ?

docs/spec/directives.rst
carljm: I would probably just append ", where ``<2-tuple>`` is a literal tuple of two literal integers"

docs/spec/directives.rst
carljm: All typing council members have now commented in this thread in favor of (or fine with) going ahead with tuple/list/set support, except for @rchen152. Given [pyrefly already supports all three](https://pyrefly.org/sandbox/?project=v2.pZTRDoIgGIVfhekFtjVZtz1F99EFS2dsDpxgoa137-cXtzDjplvm-cY8H-fXZMYvzwXtsCc4K7tWWKgBKyh41ko1OJ7t4YKV6B8SrriY5KMHrj5GqnDY4phiPjeZr8D00RVzROaUYp43mZfA9NEVc9p9d5yT03xAwu-CdQM5mgE2FEovucphPmtys7YzR8aWuO4bVitW6athUYIRr7O4C4nLHOOIRicNYP-Ri85u0ZRa1JsFXyRteQM), I would not expect @rchen152 to object. So I think it's safe to consider this consensus now.

docs/spec/directives.rst
Josverl: Corrected to version_info 

In order to blanace width with readability, I have wrapped the table at 110 columns, which is still slightly narrower than the HTML render.
Wrapping to 80 would leave less than 1 word per line for the 4th column

docs/spec/directives.rst
Josverl: I have tried to improve.
Please let me know if this is better

docs/spec/directives.rst
Josverl: as above : I have tried to improve.
Please let me know if this is better

docs/spec/directives.rst
Josverl: I think there might be a misunderstanding;
IIRC I never mentioned a cross platform fallback to CPython, and I can't find the word fallback in the prose either
nonetheless I have updated the table entry for sys.implementation.version to : 

On CPython default to the value used for sys.version. On other implementations, the value must be provided according to the type checker’s configuration options.

I hope that sufficiently clarifies this , if not;  I would welcome a text suggestion to improve.

docs/spec/directives.rst
Josverl: I have reworded the sub-section heading and prose and example comments based on the above feedback.

docs/spec/directives.rst
Josverl: I have reworded the sub-section heading and prose and example comments based on the above feedback.

docs/spec/directives.rst
Josverl: That was never my intent and I do not think I ever mentioned that in the spec.
Either way I have now tried to explicitly stated that for non-`cpython` :  `sys.implementation.version` should be explicitly provided to the type-checker and that is should raise an error or warning if it is used and no value is provided.

docs/spec/directives.rst
Josverl: Thanks, corrected.

Issue #2174, opened 2026-02-12
[Conformance][RFC] Adjust `NoReturn` conformance test to model the spec more closely


migeed-z:
The conformance test for [specialtypes_never.py (lines 28-33)](https://github.com/python/typing/blob/main/conformance/tests/specialtypes_never.py#L28-L33) seems stricter than what the spec requires:

```
def func2(x: int) -> int:
    if x > 0:
        return x
    stop()
    ...  # E?: This statement can be marked unreachable
    return "whatever works"  # No type error
```

The [spec's example](https://typing.python.org/en/latest/spec/special-types.html#noreturn) for a similar scenario says that  the error "might be not reported by some checkers" but the conformance test uses `# No type error`, requiring all checkers to suppress the error.

There's also an inconsistency: lines 32 and 33 are equally unreachable after `stop()`, but the test allows an optional error on one (# E?) while forbidding any error on the other. If this is about reachability, it seems that both should be treated the same?

The `NoReturn` recognition part is already tested by `func2` itself.  If a checker didn't recognize `stop()` as `NoReturn`, it would flag the function for not returning `int` on all paths. It seems that the lines after `stop()` are testing dead-code error suppression policy, which the spec leaves to the implementation?

Suggested change:  Would it make sense to remove lines 32-33 to get:

```
def func2(x: int) -> int:
    if x > 0:
        return x
    stop()
```

This still tests that checkers recognize `NoReturn` calls, without mandating a specific error suppression policy. Would appreciate any thoughts on whether this reading is correct.

carljm:
I agree with this change. I don't think the spec should mandate error suppression policy in dead code. It could be reasonable and useful to error on `return "whatever works"` as "clearly wrong regardless of reachability".

IMO it makes more sense to just test that there is no error for the implicit `None` return, as you propose.

JelleZijlstra:
PR was merged

PR #2175, opened 2026-02-13
Do not mandate a specific dead code policy in `NoReturn` conformance test

migeed-z:
### Summary:
Changed conformance test to prevent mandating a specific policy for dead code, since the spec does not require one. 
The full discussion is here https://github.com/python/typing/issues/2174.

### Test plan:
I pinned the existing versions of all existing type checkers and followed the instructions otherwise, so we do not get unrelated changes in the PR.

### DIFF
--- a/conformance/tests/specialtypes_never.py
+++ b/conformance/tests/specialtypes_never.py
@@ -24,13 +24,11 @@ def func1(x: int) -> NoReturn:  # E: implicitly returns None
 # > The checkers will also recognize that the code after calls to such functions
 # > is unreachable and will behave accordingly.
 
-
+# No error on implicit None return after stop()
 def func2(x: int) -> int:
     if x > 0:
         return x
     stop()
-    ...  # E?: This statement can be marked unreachable
-    return "whatever works"  # No type error
 
 
 # The spec previously said that NoReturn is only valid in a function return type,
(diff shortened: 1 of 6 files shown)


conformance/tests/specialtypes_never.py
```diff
@@ -21,16 +21,10 @@ def func1(x: int) -> NoReturn:  # E: implicitly returns None
         sys.exit(1)
 
 
-# > The checkers will also recognize that the code after calls to such functions
-# > is unreachable and will behave accordingly.
```
carljm, line 25: I think this comment is still useful and relevant.

conformance/tests/specialtypes_never.py
```diff
@@ -21,16 +21,10 @@ def func1(x: int) -> NoReturn:  # E: implicitly returns None
...
-
-
 def func2(x: int) -> int:
```
carljm, line 28: I think the fact that the purpose of this test is "no error due to implicit None return" is not immediately obvious, and should be highlighted, either with an inline comment or a comment immediately above the snippet.

migeed-z: Added a comment above the snippet

conformance/tests/specialtypes_never.py
```diff
@@ -24,13 +24,11 @@ def func1(x: int) -> NoReturn:  # E: implicitly returns None
 # > The checkers will also recognize that the code after calls to such functions
 # > is unreachable and will behave accordingly.
 
-
+# No error on implicit None return after stop();
```
carljm:
```suggestion
# No error on implicit None return after stop()
```

JelleZijlstra (approved):

carljm:
@erictraut does this change look OK to you?

erictraut:
Yes, looks good to me!

Issue #2176, opened 2026-02-13
Incorrect example of Protocol type check in documentation


e1turin:
On the page "Protocols — typing documentation" in section ["`type[]` and class objects vs protocols"](https://typing.python.org/en/latest/spec/protocol.html#type-and-class-objects-vs-protocols)  is given wrong example: instead of class instances are given class references. Even more, error and success cases are mismatched:
```python
from typing import Any, Protocol

class ProtoA(Protocol):
    def meth(self, x: int) -> int: ...
class ProtoB(Protocol):
    def meth(self, obj: Any, x: int) -> int: ...

class C:
    def meth(self, x: int) -> int: ...

a: ProtoA = C  # Type check error, signatures don't match!
b: ProtoB = C  # OK
```
Should be this:
```python
a: ProtoA = C()  # OK
b: ProtoB = C()  # Error
```

e1turin:
I suggest such a change: https://github.com/python/typing/pull/2177

JelleZijlstra:
No, this is correct. Class objects can match protocols.

carljm:
Arguably this example is incorrect as written, but for the subtler reason that the `obj` parameter in `ProtoB.meth` should be positional-only, otherwise its name should matter for assignability.

e1turin:
Well, now I understood what was going on. But are you really saying that this is correct behavior?:

```py
a: ProtoA = C  # Type check error, signatures don't match!
```

carljm:
> But are you really saying that this is correct behavior?:
> 
> a: ProtoA = C  # Type check error, signatures don't match!

Yes. `C.meth` has the signature `(self: C, x: int) -> int`. `ProtoA.meth` has the signature `(x: int) -> int`. The former signature is not assignable to the latter.

The protocol is interpreted as an instance type, so the method is bound; the `self` argument "disappears". But `C.meth` accessed off the class is an unbound method, so it still has its `self` argument. The idea of the example is that that `self` ends up mapping to the `obj: Any` argument in `ProtoB.meth`.

e1turin:
Oh, I see now. Thank you!

PR #2177, opened 2026-02-13
Fix wrong example in Protocol documentation (#2176)

e1turin:
Provide fix for documentation to make example correct. And apply common code style to make it look better.


JelleZijlstra:
The example is correct. This was discussed several times on the peps repo before we moved to the typing spec.

PR #2178, opened 2026-02-16
Conformance suite: Minor tweaks to several assertions for better compatibility with ty

AlexWaygood:
This PR makes several tweaks to the conformance suite to resolve issues where ty would appear to be failing assertions the conformance suite makes. In reality, all of the locations edited in this PR appear to be causing ty failures due to the fact that the suite was making overly specific assertions that made undue assumptions.

These are all minor changes that do not change the intent of the tests, so I've bundled them together.

### DIFF
--- a/conformance/tests/generics_scoping.py
+++ b/conformance/tests/generics_scoping.py
@@ -12,12 +12,12 @@ def fun_2(x: T) -> T:  # and here could be different
     return x
 
 # One of these two should pass; either is acceptable:
-assert_type(fun_1(1), int)  # E[fun1-int]
-assert_type(fun_1(1), Literal[1])  # E[fun1-int]
+assert_type(fun_1(1), int)  # E[fun1]
+assert_type(fun_1(1), Literal[1])  # E[fun1]
 
 # One of these two should pass; either is acceptable:
-assert_type(fun_2("a"), str)  # E[fun1-str]
-assert_type(fun_2("a"), Literal["a"])  # E[fun1-str]
+assert_type(fun_2("a"), str)  # E[fun2]
+assert_type(fun_2("a"), Literal["a"])  # E[fun2]
 
 # > A type variable used in a method of a generic class that coincides
 # > with one of the variables that parameterize this class is always bound
--- a/conformance/results/zuban/generics_scoping.toml
+++ b/conformance/results/zuban/generics_scoping.toml
@@ -1,8 +1,8 @@
 conformance_automated = "Fail"
 conformant = "Partial"
 errors_diff = """
-Lines 15, 16: Expected error (tag 'fun1-int')
-Lines 19, 20: Expected error (tag 'fun1-str')
+Lines 15, 16: Expected error (tag 'fun1')
+Lines 19, 20: Expected error (tag 'fun2')
 Lines 49, 50: Expected error (tag 'method-str')
 Lines 53, 54: Expected error (tag 'method-bytes')
 """


conformance/tests/annotations_generators.py
```diff
@@ -190,4 +190,4 @@ async def generator30() -> AsyncIterator[int]:
     yield
 
 
-assert_type(generator30, Callable[[], AsyncIterator[int]])
+assert_type(generator30(), AsyncIterator[int])
```
AlexWaygood: ty infers a "function-literal" type for `generator30`, not a `Callable` type. A function-literal type is more precise than a `Callable` type: it retains, for example, the information that `generator30` is an instance of `types.FunctionType`, which is not true for all arbitrary `Callable` types.

I don't believe the spec states anywhere that type checkers _must_ infer a `Callable` type for function objects (only that functions must be considered to inhabit `Callable` types), so I think this change is appropriate.

conformance/tests/generics_defaults.py
```diff
@@ -41,8 +41,12 @@ class OneDefault(Generic[T, DefaultBoolT]): ...
 
 class AllTheDefaults(Generic[T1, T2, DefaultStrT, DefaultIntT, DefaultBoolT]): ...
 
+def needs_allthedefaults_or_subclass(x: type[AllTheDefaults[Any, Any, str, int, bool]]) -> None: ...
+
+# do not use `assert_type` here: some type checkers infer a more precise type
+# than `type[]` for class objects
+needs_allthedefaults_or_subclass(AllTheDefaults)  # OK
 
-assert_type(AllTheDefaults, type[AllTheDefaults[Any, Any, str, int, bool]])
```
AlexWaygood: Similarly here, the previous assertion was failing because ty infers a "class-literal" type for a class object, rather than a `type[]` type. Unlike a `type[]` type, a class-literal type is a singleton type that excludes subclasses of the class in question.

conformance/tests/generics_syntax_scoping.py
```diff
@@ -102,23 +102,24 @@ def method3[T](self, x: T):  # E
 T = int(0)
 
 
-class Outer2[T]:
```
AlexWaygood, line 105: the assertions in this test ran into issues because ty inferred `T` as having type `Literal[""]` where `T = str("")`.

ty also rejected the `assert_type(T, complex)` call lower down after the `T = 3j` assignment, because we view `complex` in a type expression as immediately expanding to the union `int | float | complex`, and the second argument to `assert_type` takes a type expression -- but `complex` (the type inferred for `T`) is not equivalent to `int | float | complex`.

conformance/tests/generics_defaults.py
```diff
@@ -41,8 +41,12 @@ class OneDefault(Generic[T, DefaultBoolT]): ...
 
 class AllTheDefaults(Generic[T1, T2, DefaultStrT, DefaultIntT, DefaultBoolT]): ...
 
+def needs_allthedefaults_or_subclass(x: type[AllTheDefaults[Any, Any, str, int, bool]]) -> None: ...
+
+# do not use `assert_type` here: some type checkers infer a more precise type
+# than `type[]` for class objects
+needs_allthedefaults_or_subclass(AllTheDefaults)  # OK
```
JelleZijlstra: This is a bit unfortunate because there are plausible wrong behaviors that would pass this test (for example, defaulting everything to Any).

conformance/tests/generics_defaults.py
carljm: Why doesn't this case emit `assert-type-unspellable-type`?

AlexWaygood: because the error code you're referring to is `assert-type-unspellable-subtype`, not `assert-type-unspellable-type`, and while `<class 'AllTheDefaults'>` is _assignable_ to `type[AllTheDefaults[Any, Any, str, int, bool]]`, I don't think it's a _subtype_ of `type[AllTheDefaults[Any, Any, str, int, bool]]`...

conformance/tests/generics_defaults.py
AlexWaygood: yeah. I see that this is accidentally "fixing" an assertion failure on mypy, I think for that reason.

conformance/tests/generics_defaults.py
carljm: ugh, yeah...

AlexWaygood:
I'll revert the changes to the defaults-related tests for now, since they're clearly a bit more problematic than the others, but this PR was intended to bundle together several relatively uncontroversial changes

conformance/tests/annotations_generators.py
```diff
@@ -190,4 +190,6 @@ async def generator30() -> AsyncIterator[int]:
     yield
 
 
-assert_type(generator30, Callable[[], AsyncIterator[int]])
+async def uses_generator30() -> None:
+    async for x in generator30():
+        assert_type(x, int)
```
AlexWaygood, line 195: ty infers a "function-literal" type for `generator30`, not a `Callable` type. A function-literal type is more precise than a `Callable` type: it retains, for example, the information that `generator30` is an instance of `types.FunctionType`, which is not true for all arbitrary `Callable` types.

I don't believe the spec states anywhere that type checkers _must_ infer a `Callable` type for function objects (only that functions must be considered to inhabit `Callable` types), so I think this change is appropriate.

I tried doing just `assert_type(generator30(), AsyncIterator[int])`, but then zuban (reasonably) complained that the result of `generator30()` was never awaited, which wasn't relevant to what the test was trying to assert.

conformance/tests/generics_scoping.py
```diff
@@ -11,8 +11,13 @@ def fun_1(x: T) -> T:  # T here
 def fun_2(x: T) -> T:  # and here could be different
     return x
 
-assert_type(fun_1(1), int)
-assert_type(fun_2('a'), str)
+# One of these two should pass; either is acceptable:
+assert_type(fun_1(1), int)  # E[fun1-int]
+assert_type(fun_1(1), Literal[1])  # E[fun1-int]
```
carljm: How does the conformance checker handle it if there's an error on both of these lines? It's the intent of the test that that should be considered non-conforming.

AlexWaygood: I believe a tagged error like this is still non-optional: an error must be emitted on at least one of these lines. This seems to be reflected by the fact that zuban is no longer considered conformant on this PR branch, if you look at the changes to the reported results -- it apparently doesn't emit an error on either of these lines. (That seems like a bug in zuban to me -- `int` and `Literal[1]` are not equivalent types, so `fun_1(1)` cannot be equivalent to both `int` _and_ `Literal[1]`.)

AlexWaygood: yes, as per https://github.com/python/typing/tree/main/conformance#test-case-syntax:

> `# E[tag]`, where `tag` is an arbitrary string: must appear multiple times in a file with the same tag. Exactly one line with this tag must raise an error.

JelleZijlstra (approved):

conformance/tests/generics_scoping.py
```diff
@@ -11,8 +11,13 @@ def fun_1(x: T) -> T:  # T here
...
+
+# One of these two should pass; either is acceptable:
+assert_type(fun_1("a"), str)  # E[fun1-str]
+assert_type(fun_1("a"), Literal["a"])  # E[fun1-str]
```
rchen152: This should use `fun2` like the original assertion it replaces.

```suggestion
assert_type(fun_2("a"), str)  # E[fun1-str]
assert_type(fun_2("a"), Literal["a"])  # E[fun1-str]
```

conformance/results/results.html
```diff
@@ -307,7 +307,7 @@ <h3>Python Type System Conformance Test Results</h3>
 <tr><th class="column col1">&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;generics_scoping</th>
 <th class="column col2 conformant">Pass</th>
 <th class="column col2 conformant">Pass</th>
-<th class="column col2 conformant">Pass</th>
+<th class="column col2 not-conformant">Unknown</th>
```
rchen152: I think this needs a `conformant = ...` in the corresponding toml file to indicate whether the result is partial or unsupported.

rchen152 (approved):

conformance/tests/generics_scoping.py
AlexWaygood: oops, great catch -- thanks

conformance/tests/generics_scoping.py
rchen152: cc @davidhalter - FYI this conformance change turned up what looks like a bug in zuban.

conformance/results/results.html
AlexWaygood: thanks, applied

conformance/tests/generics_scoping.py
davidhalter: @rchen152 Thanks, it looks like something about the nesting causes issues. Will investigate.

davidhalter: I quickly fixed this (on master branch). Zuban inferred essentially an intersection type of `int & Literal[1]`, which is relevant for some type inference details. Now it's still inferred this way, but the assert_type will only assert the type of `Literal[1]` and not `int`.

I don't think this was necessarily a bug in Zuban, but it's probably easier to work with the conformance tests if I implement it this way.

erictraut: `int & Literal[1]` is equivalent to `Literal[1]`, and the spec says that `assert_type` verifies type equivalency.

PR #2184, opened 2026-02-20
Fix incorrect type name in TypedDict extra_items assignment test

azimsh3r:
The test references Movie, but the assignment target is MovieBase2. This appears to be a leftover from a previous test variant.

### DIFF
--- a/conformance/tests/typeddicts_extra_items.py
+++ b/conformance/tests/typeddicts_extra_items.py
@@ -219,7 +219,7 @@ class MovieWithYear2(TypedDict, extra_items=int | None):
     year: int | None
 
 details3: MovieWithYear2 = {"name": "Kill Bill Vol. 1", "year": 2003}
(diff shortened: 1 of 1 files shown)


conformance/tests/typeddicts_extra_items.py
```diff
@@ -219,7 +219,7 @@ class MovieWithYear2(TypedDict, extra_items=int | None):
     year: int | None
 
 details3: MovieWithYear2 = {"name": "Kill Bill Vol. 1", "year": 2003}
-movie3: MovieBase2 = details3  # E:'year' is not required in 'Movie', but it is required in 'MovieWithYear2'
+movie3: MovieBase2 = details3  # E:'year' is not required in 'MovieBase2', but it is required in 'MovieWithYear2'
```
JelleZijlstra:
```suggestion
movie3: MovieBase2 = details3  # E: 'year' is not required in 'MovieBase2', but it is required in 'MovieWithYear2'
```
while we're here

JelleZijlstra (approved):

carljm (approved):

PR #2188, opened 2026-02-21
conformance: allow emitting an error on missing return with Any annotation

JelleZijlstra:
Pycroscope produces an error on the existing open() function in this file
because it is annotated as returning Any, but doesn't have a return statement.
There's some existing discussion of this case in https://github.com/python/mypy/issues/10297
showing that it is intentional for mypy to not error here, but I don't think
we need to prohibit type checkers from emitting an error in this case
if they choose.


AlexWaygood (approved):
I agree that whether or not a type checker emits a lint for this pattern should have no bearing on whether or not this test passes, since this test is trying to test something wholly unrelated to that question

PR #2194, opened 2026-02-25
Allow type checkers to infer more precise types for tuples

davidhalter:
This is consisent with the latest changes where `f(1)` could either be an int or a `Literal[1]`

```
    def f[T](x: T) -> T: ...
    assert_type(f(1), int)
    # or
    assert_type(f(1), Literal[1])
```

The same is possible with tuples:

```
    def f[*Ts](*x: *Ts) -> tuple[*Ts]: ...
    assert_type(f(1), tuple[int])
    # or
    assert_type(f(1), tuple[Literal[1]])
```

Mypy fails the test `generics_typevartuple_callable.py` now, because it doesn't error on any of the assert_type below:

```
def func3(*args: * tuple[int, *Ts, T]) -> tuple[T, *Ts]:
    raise NotImplementedError

assert_type(func3(1, "", 3j, 3.4), tuple[float, str, complex])  # E[fun3]
assert_type(func3(1, "", 3j, 3.4), tuple[float, Literal[''], complex])  # E[fun3]
```

This is also how Zuban used to work, but I think it's better to be precise here and it should probably fail one of the two (like Eric suggested in https://github.com/python/typing/pull/2178#discussion_r2843516765).

Ran the automated tests again. A few changes:

- Zuban's passes a test again (due to assert_type changes) 
- Picked up a pyrefly upgrade that fixes a test

### DIFF
--- a/conformance/tests/generics_typevartuple_args.py
+++ b/conformance/tests/generics_typevartuple_args.py
@@ -17,8 +17,11 @@ def args_to_tuple(*args: *Ts) -> tuple[*Ts]:
     raise NotImplementedError
 
 
-assert_type(args_to_tuple(1, "a"), tuple[int, str])  # E[args_to_tuple]
-assert_type(args_to_tuple(1, "a"), tuple[Literal[1], Literal["a"]])  # E[args_to_tuple]
+def takes_int_and_str(tup: tuple[int, str]) -> None:
+    raise NotImplementedError
+
+
+takes_int_and_str(args_to_tuple(1, "a"))
 
 
 class Env:
@@ -30,8 +33,7 @@ def exec_le(path: str, *args: * tuple[*Ts, Env], env: Env | None = None) -> tupl
 
 
 assert_type(exec_le("", Env()), tuple[()])  # OK
-assert_type(exec_le("", 0, "", Env()), tuple[int, str])  # E[exec_le]
-assert_type(exec_le("", 0, "", Env()), tuple[Literal[0], Literal['']])  # E[exec_le]
+takes_int_and_str(exec_le("", 0, "", Env()))
 exec_le("", 0, "")  # E
 exec_le("", 0, "", env=Env())  # E
 
--- a/conformance/tests/generics_typevartuple_args.py
+++ b/conformance/tests/generics_typevartuple_args.py
@@ -7,7 +7,7 @@ Tests for the use of TypeVarTuple with "*args" parameter.
 # > If *args is annotated as a type variable tuple, the types of the individual
 # > arguments become the types in the type variable tuple.
 
-from typing import TypeVarTuple, assert_type, Literal
+from typing import TypeVarTuple, assert_type
 
 
 Ts = TypeVarTuple("Ts")
@@ -15,13 +15,6 @@ Ts = TypeVarTuple("Ts")
 
 def args_to_tuple(*args: *Ts) -> tuple[*Ts]:
     raise NotImplementedError
-
-
-def takes_int_and_str(tup: tuple[int, str]) -> None:
-    raise NotImplementedError
-
-
-takes_int_and_str(args_to_tuple(1, "a"))
 
 
 class Env:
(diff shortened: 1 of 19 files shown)


conformance/tests/generics_typevartuple_args.py
```diff
@@ -17,7 +17,8 @@ def args_to_tuple(*args: *Ts) -> tuple[*Ts]:
     raise NotImplementedError
 
 
-assert_type(args_to_tuple(1, "a"), tuple[int, str])
+assert_type(args_to_tuple(1, "a"), tuple[int, str])  # E[args_to_tuple]
```
JelleZijlstra: To avoid the two options, could we put this in a function that takes args of type `int` and `str`? Then `tuple[int, str]` would be unambiguously right.

davidhalter: Thanks Jelle, that's a really good idea, it also feels more readable this way for people that are not used to the special `# E[...]` comments.

I pushed this. This makes Mypy pass `generics_typevartuple_callable.py` again and doesn't really change anything other than Zuban passing the tests with its latest release.

AlexWaygood: Hmm, I weakly prefer the version prior to https://github.com/python/typing/pull/2194/commits/c1ffe96722d0bba2a0b1ccec216d6f2f68c317d0. The new version could pass even if the type checker completely fails to solve the TypeVarTuple and just falls back to  `*tuple[Any, ...]` for its inference of `*Ts`. The version with the `# E[]` tagged errors was precise

davidhalter: Didn't think about that. That's obviously a problem. I'm happy to revert...

JelleZijlstra: I meant something like this:

```
def func(x: int, y: str):
    assert_type(args_to_tuple(x, y), tuple[int, str])
```

I agree the current version is worse.

AlexWaygood: @JelleZijlstra's version there works for me, yes!

davidhalter: @AlexWaygood @JelleZijlstra Thank you for the good feedback. Kind of stupid of me to not think of this in the first place. I think the change is now completely uncontroversial.

conformance/tests/generics_typevartuple_callable.py
```diff
@@ -46,4 +46,5 @@ def func3(*args: * tuple[int, *Ts, T]) -> tuple[T, *Ts]:
     raise NotImplementedError
 
 
-assert_type(func3(1, "", 3j, 3.4), tuple[float, str, complex])
+def has_int_and_str(x: int, y: str):
+    assert_type(func3(x, y, 3j, 3.4), tuple[float, str, complex])
```
AlexWaygood: even this will probably not pass ty when we (soon) add support for `TypeVarTuple`, because we infer a float literal as "just `float`" whereas we understand the symbol `float` in a type expression (here the second argument to `assert_type`) as always meaning `int | float` due to the `float` special case. "Just `float`" and `int | float` are not equivalent types, so we would not pass this assertion.

The same also goes for "just complex" vs `int | float | complex`.

So we could future-proof this even more:


```suggestion
def has_int_and_str(a: int, b: str, c: float, d: complex):
    assert_type(func3(a, b, c, d), tuple[float, str, complex])
```

davidhalter: I have committed your suggestion. I don't mind this change at all. Promotion in Zuban has the same int | float type, but it's just not shown to the user, which might confuse users more or less depending on the situation.

AlexWaygood (approved):
this makes sense to me, thanks!

JelleZijlstra (approved):

AlexWaygood:
I don't think this is controversial, but I'll wait for sign-off from another typing-council member before landing

carljm (approved):

PR #2200, opened 2026-02-28
Add version locking and CI for the conformance suite

JelleZijlstra:
The current system where we always pick the latest version of every type checker causes some changes to randomly pick up unrelated upgrades to type checkers, which makes it more difficult to make focused changes.

We were also not checking in CI that the conformance results were correct. This change introduces a GitHub action to do that, which exposed a bug due to a recent change in the typevartuple test.

### DIFF
--- a/conformance/tests/generics_typevartuple_callable.py
+++ b/conformance/tests/generics_typevartuple_callable.py
@@ -42,9 +42,9 @@ assert_type(func2(callback1), tuple[str, int, complex])
 assert_type(func2(callback2), tuple[str])
 
 
-def func3(*args: * tuple[int, *Ts, T]) -> tuple[T, *Ts]:
+def func3(*args: *tuple[int, *Ts, T]) -> tuple[T, *Ts]:
     raise NotImplementedError
 
 
 def has_int_and_str(a: int, b: str, c: float, d: complex):
-    assert_type(func3(a, b, c, d), tuple[float, str, complex])
+    assert_type(func3(a, b, c, d), tuple[complex, str, float])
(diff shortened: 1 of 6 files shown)


conformance/results/pyright/qualifiers_final_decorator.toml
```diff
@@ -1,6 +1,5 @@
 conformant = "Pass"
 output = """
-qualifiers_final_decorator.py:8:6 - warning: Import "_qualifiers_final_decorator" could not be resolved from source (reportMissingModuleSource)
```
JelleZijlstra, line 3: Not sure why this changed

conformance/tests/generics_typevartuple_callable.py
```diff
@@ -42,9 +42,9 @@ def callback2(a: int, d: str) -> tuple[str]:
...
 
 
-def func3(*args: * tuple[int, *Ts, T]) -> tuple[T, *Ts]:
+def func3(*args: *tuple[int, *Ts, T]) -> tuple[T, *Ts]:
     raise NotImplementedError
 
 
 def has_int_and_str(a: int, b: str, c: float, d: complex):
-    assert_type(func3(a, b, c, d), tuple[float, str, complex])
+    assert_type(func3(a, b, c, d), tuple[complex, str, float])
```
JelleZijlstra, line 50: This was wrong. We call func3 with arguments (int, str, float, complex), so we get Ts = (str, float) and T = complex. func3 returns `tuple[T, *Ts]`, so (complex, str, float).

davidhalter: Yeah, sorry for that. I probably pulled in a change from Alex without rerunning tests.

AlexWaygood: My bad really, I think I gave you the buggy suggestion!

PR #2201, opened 2026-02-28
Extend conformance tests for dataclass hashability

AlexWaygood:
ty currently fails all the assertions in `dataclasses_hash.py`. Insofar as these tests themselves go, this is fair: ty [has some unfortunate behaviour around how it treats `typing.Hashable`](https://github.com/astral-sh/ty/issues/1162) right now, which we'd like to fix. Our behaviour regarding hashability is suboptimal and something we'd like to improve; it's definitely the case that _some_ tests in this file should fail for ty.

However, I also don't think a 100% failure rate for ty on this file is an accurate reflection of the state of ty's dataclasses implementation. ty _does_ synthesize a callable `__hash__` method for hashable dataclasses, and it _does_ synthesize a `__hash__ = None` attribute for unhashable dataclasses. The reason why it fails 100% of the assertions in this file currently isn't because of a gap in its dataclasses implementation; it's because we simplify unions too eagerly and this causes issues for our `Protocol` implementation.

This PR therefore extends the existing tests with some added assertions that don't depend on these unrelated features so heavily. The existing tests that accurately expose a shortcoming in ty's behaviour are retained.

### DIFF
--- a/conformance/results/mypy/dataclasses_hash.toml
+++ b/conformance/results/mypy/dataclasses_hash.toml
@@ -6,19 +6,14 @@ Does not report when dataclass is not compatible with Hashable protocol.
 """
 output = """
 dataclasses_hash.py:14: error: Expression is of type "Callable[[object], int]", not "None"  [assert-type]
-dataclasses_hash.py:29: error: Expression is of type "Callable[[object], int]", not "None"  [assert-type]
-dataclasses_hash.py:40: error: Expression is of type "Callable[[object], int]", not "None"  [assert-type]
-dataclasses_hash.py:55: error: Expression is of type "Callable[[object], int]", not "None"  [assert-type]
-dataclasses_hash.py:69: error: Expression is of type "Callable[[object], int]", not "None"  [assert-type]
-dataclasses_hash.py:85: error: Expression is of type "Callable[[DC6], int]", not "None"  [assert-type]
-dataclasses_hash.py:102: error: Expression is of type "Callable[[object], int]", not "None"  [assert-type]
+dataclasses_hash.py:36: error: Expression is of type "Callable[[object], int]", not "None"  [assert-type]
 """
 conformance_automated = "Fail"
 errors_diff = """
 Line 17: Expected 1 errors
 Line 18: Expected 1 errors
-Line 43: Expected 1 errors
-Line 44: Expected 1 errors
+Line 39: Expected 1 errors
+Line 40: Expected 1 errors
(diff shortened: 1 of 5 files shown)


conformance/results/mypy/dataclasses_hash.toml
```diff
@@ -1,11 +1,24 @@
-conformant = "Partial"
+conformant = "Unsupported"
```
AlexWaygood, line 1: It's not clear to me that mypy has implemented support for any of the functionality tested in `dataclasses_hash.py`, so I downgraded its score from "Partial" to "Unsupported". I'm very happy to revert this change if anybody feels this is unfair or if I've missed something!

conformance/tests/generics_typevartuple_callable.py
```diff
@@ -47,4 +47,4 @@ def func3(*args: * tuple[int, *Ts, T]) -> tuple[T, *Ts]:
 
 
 def has_int_and_str(a: int, b: str, c: float, d: complex):
-    assert_type(func3(a, b, c, d), tuple[float, str, complex])
+    assert_type(func3(a, b, c, d), tuple[complex, str, float])
```
AlexWaygood, line 50: this is the same issue that was identified in https://github.com/python/typing/pull/2200#discussion_r2867149665

conformance/tests/dataclasses_hash.py
```diff
@@ -20,15 +23,24 @@ class DC2:
     a: int
 
 
-v2: Hashable = DC2(0)
+# Because `DC2` is frozen, type checkers should synthesize
+# a callable `__hash__` method for it, and therefore should
+# emit a diagnostic here:
+assert_type(DC2.__hash__, None)  # E
```
JelleZijlstra: An assert_type() that is meant to fail feels a little weird to me, maybe do something like `x: Callable[..., Any] = DC2.__hash__`? That is, we assert it's a callable, but don't worry about the exact shape of callable.

AlexWaygood: Oh, I thought we already had some like this, but I guess they're all tagged errors.

I'm happy to go with something like what you propose -- the only disadvantage is that if the type checker infers `DC2.__hash__` as being `Any`, the assertion will still pass? But I guess that's pretty unlikely here, and I guess all my current assertion does is assert that the type checker shouldn't infer `None`, which isn't much better, if at all

AlexWaygood: I pushed this change

JelleZijlstra (approved):

PR #2202, opened 2026-02-28
Conformance suite: rewrite `directives_version_platform.py` to not assume that type checkers will ignore all errors in unreachable code

AlexWaygood:
The typing spec states:

> Type checkers are expected to understand simple version and platform checks

ty does! However, it currently fails 100% of the assertions in `conformance/tests/directives_version_platform.py`, because it attempts to provide some type checking of code even if that code is in a region deemed to be unreachable. All the assertions in this file currently assume that a good way to test whether a type checker understands `sys.version_info` or `sys.platform` checks is to assert that they will emit no errors for obviously incorrect code in unreachable branches. But as far as I know, this is not described anywhere as a requirement for spec-conformant type checkers.

This PR rewrites the tests in this file so that they still assert that type checkers properly understand simple version and platform checks, but so that they no longer assume that type checkers will silence all errors in regions of code that they deem to be unreachable. With these changes, ty passes 100% of the assertions in this file. The conformance score of all other type checkers on this file is unchanged.

### DIFF
--- a/conformance/tests/directives_version_platform.py
+++ b/conformance/tests/directives_version_platform.py
@@ -66,7 +66,7 @@ def test(a: int, b: str):
     this_raises = val10  # E?: `val10` is undefined, but support for `os.name` is optional
     does_not_raise = val11  # E? should not error if `os.name` control flow is supported, but might be flagged as possibly undefined otherwise
 
-    if os.name != "bogus_platform":
+    if os.name != "bogus_os":
         val12 = ""
     else:
         val13 = ""


conformance/tests/directives_version_platform.py
```diff
@@ -6,40 +6,70 @@
...
+
+    if os.name == "bogus_os":
+        val10 = ""
+    else:
+        val11 = ""
+
+    this_raises = val10  # E?: `val10` is undefined, but support for `os.name` is optional
+    does_not_raise = val11  # E? should not error if `os.name` control flow is supported, but might be flagged as possibly undefined otherwise
+
+    if os.name != "bogus_platform":
```
carljm: looks like this predates the PR, but seems like it should be this instead (doesn't change the effect of the test, just naming clarity):
```suggestion
    if os.name != "bogus_os":
```

PR #2203, opened 2026-03-02
spec: do not mandate inference details around TypeVarTuple

JelleZijlstra:
I don't think this case should error; a type checker can solve Ts
to something like (object,) or (int | str,). This is not all that
different from the first call to func4(), where the two calls with (0,)
and (1,) lead to Ts being solved to a common supertype.


conformance/tests/generics_typevartuple_args.py
```diff
@@ -72,8 +72,7 @@ def func4(*args: tuple[*Ts]):
 
 
 func4((0,), (1,))  # OK
-func4((0,), (1, 2))  # E
-func4((0,), ("1",))  # E
```
carljm, line 76: To me it seems most consistent with how typevar solving works everywhere else in the type system to actually specify that this should not be an error. It looks like that would match the behavior of mypy and zuban, but not of pyrefly or pyright.

If there's opposition to that change, I would still probably favor leaving this in, with E? and a comment -- that seems more useful to future discussion of this case.

(As far as I can see from a quick scan, this behavior was not specified in the PEP.)

carljm (approved):

conformance/tests/generics_typevartuple_args.py
erictraut: This comes from [this section](https://typing.python.org/en/latest/spec/generics.html#type-variable-tuple-equality) of the spec, which was copied from the PEP. This section was added to the PEP late in the process — after I had reviewed the PEP and implemented it in pyright. I was very unhappy when I learned about the addition because, as you point out, it introduces an inconsistency with the way that TypeVar solving works in other cases. 

Line 76 of the conformance test is arguably different than the special case spelled out in the spec, which says "If the same TypeVarTuple instance is used in _multiple places_". `Ts` appears only once in the signature of `func4`, but because it is defining the type of a variadic parameter, `Ts` appears multiple times from the perspective of a constraint solver. Constraints are supplied for each argument that maps to that variadic parameter.

Consider the following example. The spec clearly dictates that the call to func3 is an error. I think the call to func4 should likewise be an error.

```python
def func3[*Ts](a: tuple[*Ts], b: tuple[*Ts], /): ...

func3((1,), ("a",)) # Clearly an error according to the spec

def func4[*Ts](*args: tuple[*Ts]): ...

func4((1,), ("a",)) # For consistency, should be an error too
```

So, unless we amend the spec, I think the conformance test should remain as is (with the possible addition of a better comment to explain the above).

I'm in favor of modifying the spec and removing this entire section.

carljm: Ah, thanks @erictraut -- I missed that section when scanning the PEP text. I agree that that section of the PEP/spec text is applicable here, and also that it's unfortunately inconsistent with the usual behavior of Python typevar solving. I'm not sure any of the PEP authors are still active in Python typing, or we could ask what motivated that late addition to the PEP text.

There are use cases for stricter typevar solving that doesn't ever synthesize wider types, but those use cases exist for regular typevars too, not just for variadic ones. So it seems better to go for consistency, and introduce a new orthogonal type system feature for stricter solving. (One possible shape for that feature is a `NoInfer` qualifier -- though it's not clear how that could be used in the `*args` case.)

conformance/tests/generics_typevartuple_basic.py
```diff
@@ -54,7 +54,7 @@ def __init__(self, shape: tuple[Shape]):  # E: not unpacked
         self._shape: tuple[*Shape] = shape
 
     def get_shape(self) -> tuple[Shape]:  # E: not unpacked
-        return self._shape
+        raise NotImplementedError
```
JelleZijlstra, line 57: This was causing mypy to emit an error because it resolves the invalid `tuple[Shape]` to `tuple[Any]`, and then considers `self._shape` incompatible with a single-element tuple. I don't think the conformance suite should enforce the behavior of type checkers with erroneous annotations.

conformance/tests/generics_typevartuple_args.py
JelleZijlstra: Thanks for the background! I have now amended this PR to propose changing the spec to remove this rule. I will post on Discuss before opening a vote for the Typing Council.

conformance/tests/generics_typevartuple_basic.py
davidhalter, line 99: @JelleZijlstra Is this an error because TypeVarTuples are invariant?

I played with this in Zuban for a bit and got this:

```
Revealed type is "tests.generics_typevartuple_basic.Array[tests.generics_typevartuple_basic.Height | tests.generics_typevartuple_basic.Width]"
```

I it likely that this will also change or not?

JelleZijlstra: Yes, I think this is analogous to a case like this with plain TypeVars:
```python
def f[T](x: list[T], y: list[T]): ...

def g(a: list[int], b: list[str]):
    f(a, b)
```
Because of invariance, there is no solution and the call should error.

> it likely that this will also change or not?

Do you mean whether we will relax the rule that TypeVarTuple is always invariant? I am not planning to propose that, but I saw some interest on Discourse recently in doing that, and it's a logical next step if we accept ParamSpec variance.

PR #2205, opened 2026-03-05
Add conformance results invariant validator

JelleZijlstra:
The conformance results have some implicit invariants in them, but it's easy to miss them since there's no validation.

### DIFF
--- a/conformance/src/validate_results.py
+++ b/conformance/src/validate_results.py
@@ -91,7 +91,7 @@ def _validate_result(file: Path, results_dir: Path, info: dict[str, Any]) -> lis
             f"conformance_automated={automated!r}"
         )
 
-    if not conformant_is_pass:
+    if conformant == "Partial":
         notes = info.get("notes", "")
         if not isinstance(notes, str) or not notes.strip():
             issues.append(


conformance/src/validate_results.py
```diff
@@ -0,0 +1,105 @@
...
+            return issues
+    elif isinstance(conformant, str):
+        if conformant not in ("Pass", "Partial", "Unsupported"):
+            issues.append(f"{rel_path}: invalid conformance status {conformant!r}")
+        conformant_is_pass = conformant == "Pass"
+    else:
+        issues.append(f"{rel_path}: conformant must be a string when present")
+        return issues
+
+    if conformant_is_pass != automated_is_pass:
```
carljm, line 88: This is the only one I wonder about. Conformance is human-scored for a reason. In cases where a conformance test accidentally tests something unspecified and unrelated to the topic of the test, is it reasonable for a human scorer to mark `conformance` as `"Pass"` even if `conformance_automated` is `"Fail"`? (Ideally this would be temporary, pending a PR to improve the conformance suite.) Or do we want to prohibit this, in order to better motivate improving the conformance suite? It seems misleading to mark a test file as "Partial" due to an automated-scoring mismatch not actually related to the topic of the test.

carljm (approved):
Nice, thank you!

conformance/src/validate_results.py
JelleZijlstra: I think in that case, we should fix the test so the mismatch goes away.

PR #2207, opened 2026-03-06
don't state that `type[Callable]` is invalid

KotlinIsland:
see: https://discuss.python.org/t/callable-isnt-a-special-form/106407

there is no reason to disallow `type[Callable]`, i can understand the motivation for disallowing `type[Literal[1]]` (although i disagree with it), i can't see any reason why `type[Callable]` isn't allowed when `type[tuple[int]]` is

this is true regardless of whether or not `Callable` is a special form (it's not, see https://github.com/python/typing/pull/2214)

motivation:
- https://github.com/astral-sh/ruff/pull/23753

### DIFF
--- a/conformance/results/mypy/generics_paramspec_basic.toml
+++ b/conformance/results/mypy/generics_paramspec_basic.toml
@@ -1,4 +1,4 @@
-conformant = "Pass"
+conformant = "Partial"
 output = """
 generics_paramspec_basic.py:10: error: String argument 1 "NotIt" to ParamSpec(...) does not match variable name "WrongName"  [misc]
 generics_paramspec_basic.py:15: error: Invalid location for ParamSpec "P"  [valid-type]
--- a/conformance/results/mypy/specialtypes_type.toml
+++ b/conformance/results/mypy/specialtypes_type.toml
@@ -17,6 +17,9 @@ specialtypes_type.py:143: error: "<typing special form>" has no attribute "unkno
 specialtypes_type.py:144: error: "<typing special form>" has no attribute "unknown"  [attr-defined]
 specialtypes_type.py:145: error: "type[type]" has no attribute "unknown"  [attr-defined]
 specialtypes_type.py:146: error: "<typing special form>" has no attribute "unknown"  [attr-defined]
+specialtypes_type.py:179: error: Variable "typing.TypedDict" is not valid as a type  [valid-type]
+specialtypes_type.py:179: note: See https://mypy.readthedocs.io/en/stable/common_issues.html#variables-vs-type-aliases
+specialtypes_type.py:180: error: type[...] can't contain "Literal[...]"  [valid-type]
 """
 conformance_automated = "Fail"
 errors_diff = """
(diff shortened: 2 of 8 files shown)


KotlinIsland:
> Hm, Callable is definitely considered a special form. 

why would it be considered a special form? i don't see it as any different to a class like `Iterable`

> What are you hoping to use `type[Callable]` for?

i have run into problems where type checkers have given incorrect errors because `Callable` was considered a special form and not a normal class, see the discussion here: https://discuss.python.org/t/callable-isnt-a-special-form/106407 for some examples

JelleZijlstra:
> why would it be considered a special form?

It is listed specifically in https://typing.python.org/en/latest/spec/annotations.html#type-and-annotation-expressions . It's special because it can be parameterized differently from other classes, though perhaps ParamSpec means its specialness can be reduced.

davidhalter:
I think it's also important to mention that Callable is very much not a normal class in the standard library: https://github.com/python/cpython/blob/b28e5f58ebf8d9badcc893cd189cb1575d4e3614/Lib/typing.py#L2830

and in typeshed: https://github.com/python/typeshed/blob/07ffb67b924d5f532f8b8b72a4902f58faae6aca/stdlib/typing.pyi#L236

AlexWaygood:
> I think it's also important to mention that Callable is very much not a normal class in the standard library: [python/cpython@`b28e5f5`/Lib/typing.py#L2830](https://github.com/python/cpython/blob/b28e5f58ebf8d9badcc893cd189cb1575d4e3614/Lib/typing.py#L2830)

I don't think that this point, specifically, is a very strong argument against this change. While it's true that `typing.Callable` is not a class at runtime, this is also true for `typing.Iterable`, `typing.Collection`, `typing.Container`, etc... _many_ classes that typeshed pretends are defined in the typing module are _actually_ defined as classes in the `collections.abc` module at runtime, with only aliases existing in the standard library's `typing.py`.

I find the fact that `Callable` has unique semantics in type expressions and cannot be expressed as a `Protocol` to be much stronger reasons for continuing to describe it as a special form in the spec. That doesn't necessarily mean that it can't be considered a "class-like" special form that is an instance of `type`.

If we want to consider `type[Callable]` a valid type annotation, we should state clearly what that means and what the semantics should be.

KotlinIsland:
> It is listed specifically in ...

@JelleZijlstra  this is true, i will split this change to address these two separate concerns:
1. `type[Callable]` is valid regardless of if `Callable` is a special type or not
2. `Callable` is not a special type
 
> It's special because it can be parameterized differently from other classes, though perhaps ParamSpec means its specialness can be reduced

if we use a parameter spec, when what specialization remains? to me it seems non-special at that point

> I find the fact that `Callable` has unique semantics in type expressions and cannot be expressed as a `Protocol` to be much stronger reasons for continuing to describe it as a special form in the spec.

@AlexWaygood what are the unique semantics? and why can it not be expressed as a `Protocol`? all the other classes that follow this pattern (`Iterable` etc) follow this pattern of being expressed as a `Protocol`

PR #2208, opened 2026-03-06
Add conformance results for ty

AlexWaygood:


conformance/tests/.pyre/pyre.stderr
```diff
@@ -0,0 +1,104 @@
+2025-08-20 16:31:38,192 [PID 9073] INFO No binary specified, looking for `pyre.bin` in PATH
```
JelleZijlstra: remove?

conformance/src/type_checker.py
```diff
@@ -188,6 +189,66 @@ def parse_errors(self, output: Sequence[str]) -> dict[int, list[str]]:
...
+    def name(self) -> str:
+        return "ty"
+
+    def install(self) -> bool:
+        try:
+            self.get_version()
+            return True
+        except (CalledProcessError, FileNotFoundError):
+            print(
+                "Unable to run pyrefly. Install conformance dependencies with "
```
JelleZijlstra: hmmmm

AlexWaygood: Lol

conformance/tests/.pyre/pyre.stderr
AlexWaygood: Oops, thanks

conformance/src/type_checker.py
AlexWaygood:
```suggestion
                "Unable to run ty. Install conformance dependencies with "
```

conformance/results/ty/constructors_callable.toml
```diff
@@ -0,0 +1,44 @@
+conformance_automated = "Fail"
+conformant = "Unsupported"
```
carljm: This should be a `Partial` with notes, not an `Unsupported`. (I can draft the notes if you want.) We definitely support treating a constructor as a callable, and we pass a number of the tests in this file -- looks like there are several cases where our handling currently differs from the spec.

conformance/results/ty/dataclasses_match_args.toml
```diff
@@ -0,0 +1,11 @@
+conformance_automated = "Fail"
+conformant = "Pass"
```
carljm: I know I said we could do this if it made sense, but @JelleZijlstra 's [new invariant-checker](https://github.com/python/typing/pull/2205) won't like this. We could either put up a PR to make this assertion less pedantic (is this really required by the spec?), or I intend to put up a PR that will remove this `Unknown |` on the ty side shortly, so we could also just mark this `Partial` until then.

conformance/results/ty/generics_variance.toml
```diff
@@ -0,0 +1,19 @@
+conformance_automated = "Fail"
+conformant = "Unsupported"
```
carljm, line 2: This "Unsupported" doesn't feel right for the name `generics_variance`, because we support a lot of variance checking, but I guess what this file is really testing is all about enforcing not using the wrong explicit-variance legacy TypeVar in the wrong place, and we don't yet do that...

conformance/results/ty/literals_interactions.toml
```diff
@@ -0,0 +1,17 @@
+conformance_automated = "Fail"
+conformant = "Partial"
+notes = """
+Deliberately does not allow `str` to be narrowed to literal string types through equality or containment checks due to the possibility of `str` subclasses that could have unexpected equality semantics.
```
carljm, line 4: I think we should make a PR to remove these assertions from the conformance suite. Unsound narrowing behavior is not required by the spec and shouldn't be asserted by the conformance suite.

conformance/results/ty/specialtypes_type.toml
```diff
@@ -0,0 +1,37 @@
+conformance_automated = "Fail"
+conformant = "Partial"
+notes = """
+Allows arbitrary attributes to be accessed on `TA` where `TA = typing.Type[typing.Any]` or `TA = type[typing.Any]`.
```
carljm, line 4: This behavior is definitely correct; I don't think the conformance suite should require something different. (Just making a note of places where we should update the conformance suite.)

conformance/results/ty/dataclasses_match_args.toml
JelleZijlstra: Another option is to use `ignore_errors` to ignore certain errors.

conformance/results/ty/specialtypes_type.toml
carljm: Oh I mis-read this; it's about accessing attributes on a type alias object, not on an inhabitant of `type[Any]`. Yeah this is a ty bug.

conformance/results/ty/constructors_callable.toml
AlexWaygood: I think you're more up to speed on exactly what we do and don't support currently in this area — would appreciate it if you could write up the notes for this one, thanks!

carljm: Done.

conformance/results/ty/dataclasses_match_args.toml
AlexWaygood: The test here is asserting that if you have this:

```py
@dataclass(match_args=True)
class DC5:
    __match_args__ = ()
    x: int
```

then the type checker is meant to respect the "manual" `__match_args__` assignment over the synthesized one that would normally be created from `@dataclass(match_args=True)`. I.e., `__match_args__` will _not_ be `("x",)`, it will be `()`. That feels like a reasonable thing for the test to assert here; it's important that type checkers should get that right, even if it's not explicitly mandated by the spec.

I guess I'll use the `ignore_errors` field for now, because it feels pretty silly to mark ty as only "partially conformant" on `dataclasses_match_args.py` just because of this: our result is clearly in keeping with the spirit of what the test is trying to assert.

conformance/results/ty/generics_variance.toml
AlexWaygood: `dataclass_transform_converter.py` should also not really be marked as "Unsupported", but I asked David to write me some notes for that one and he opened https://github.com/astral-sh/ruff/pull/23088 instead 😆

conformance/results/ty/literals_interactions.toml
AlexWaygood: Agreed. But I'll do it as a followup.

conformance/results/ty/generics_variance.toml
carljm: I tried the same with "constructors callable" but there were some subtleties in the fix that deserved more careful attention :)

conformance/results/ty/literals_interactions.toml
AlexWaygood: https://github.com/python/typing/pull/2209

conformance/results/ty/specialtypes_type.toml
AlexWaygood: I believe the reason for this is that we fallback to looking up attribute access on the typeshed stub for `types.GenericAlias`. Typeshed includes a fictitious `types.GenericAlias.__getattr__` method on the grounds that type checkers will anyway have to heavily special-case attribute access on generic alias objects.

PR #2210, opened 2026-03-07
`directives_type_checking.py`: do not assert that type checkers must ignore all type errors in `if not TYPE_CHECKING` blocks

AlexWaygood:
This test currently asserts that type checkers should not emit errors for clearly incorrect code in `if not TYPE_CHECKING` blocks. Maybe that's reasonable behaviour, and maybe it's not, but it's never been specified in any PEP and it's not currently mandated by the spec. If we want to enforce that type checkers have to have this behaviour, then I think it should go via the normal process for modifications to the spec. In the meantime, I don't think it should be something asserted by the conformance suite.

This PR changes `directives_type_checking.py` so that it still asserts that type checkers should understand the `TYPE_CHECKING` constant as always-true, but so that it no longer asserts that type checkers should silence all diagnostics in these code regions, which goes far beyond what the spec mandates.

### DIFF
--- a/conformance/results/mypy/directives_type_checking.toml
+++ b/conformance/results/mypy/directives_type_checking.toml
@@ -1,6 +1,6 @@
 conformant = "Pass"
 output = """
-directives_type_checking.py:22: error: Incompatible types in assignment (expression has type "str", variable has type "int")  [assignment]
+directives_type_checking.py:14: error: List item 0 has incompatible type "str"; expected "int"  [list-item]
 """
 conformance_automated = "Pass"
 errors_diff = """
--- a/conformance/results/pyrefly/directives_type_checking.toml
+++ b/conformance/results/pyrefly/directives_type_checking.toml
@@ -3,5 +3,5 @@ conformance_automated = "Pass"
 errors_diff = """
 """
 output = """
-ERROR directives_type_checking.py:22:14-16: `Literal['']` is not assignable to `int` [bad-assignment]
+ERROR directives_type_checking.py:14:20-27: `list[str]` is not assignable to `list[int]` [bad-assignment]
 """
--- a/conformance/results/pyright/directives_type_checking.toml
+++ b/conformance/results/pyright/directives_type_checking.toml
@@ -1,7 +1,7 @@
 conformant = "Pass"
 output = """
-directives_type_checking.py:22:14 - error: Type "Literal['']" is not assignable to declared type "int"
-  "Literal['']" is not assignable to "int" (reportAssignmentType)
+directives_type_checking.py:14:21 - error: Type "list[str]" is not assignable to declared type "list[int]"
+  "Literal['foo']" is not assignable to "int" (reportAssignmentType)
 """
 conformance_automated = "Pass"
 errors_diff = """
--- a/conformance/results/ty/directives_type_checking.toml
+++ b/conformance/results/ty/directives_type_checking.toml
@@ -2,5 +2,5 @@ conformance_automated = "Pass"
 errors_diff = """
 """
 output = """
-directives_type_checking.py:22:14: error[invalid-assignment] Object of type `Literal[""]` is not assignable to `int`
+directives_type_checking.py:14:20: error[invalid-assignment] Object of type `list[int | str]` is not assignable to `list[int]`
 """
--- a/conformance/results/zuban/directives_type_checking.toml
+++ b/conformance/results/zuban/directives_type_checking.toml
@@ -2,5 +2,5 @@ conformance_automated = "Pass"
 errors_diff = """
 """
 output = """
-directives_type_checking.py:22: error: Incompatible types in assignment (expression has type "str", variable has type "int")  [assignment]
+directives_type_checking.py:14: error: List item 0 has incompatible type "str"; expected "int"  [list-item]
 """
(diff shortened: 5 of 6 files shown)


JelleZijlstra:
It's true that the spec doesn't say anything here, but I do feel that ty is in the wrong for emitting errors within `if not TYPE_CHECKING`. In my view, the point of doing `if not TYPE_CHECKING` is to do crazy runtime things that the type checker may not understand.

AlexWaygood:
> In my view, the point of doing `if not TYPE_CHECKING` is to do crazy runtime things that the type checker may not understand.

I agree that that's one of the things it's often used for. And for this very reason, we deliberately suppress many errors in unreachable regions that we would normally emit in regions we infer as being reachable. We just don't suppress literally *all* errors in these regions — but that's what the test here currently mandates.

rchen152:
Hmm. I agree with Jelle that I don't think type checkers should emit errors in `if not TYPE_CHECKING` blocks. It's true that the spec doesn't say anything about it, so I wouldn't object too strenuously to this being taken out of the conformance tests, but it seems like it might be better to change the spec.

AlexWaygood:
I'm open to a spec change here if somebody wants to propose that. I agree that there's an argument for doing so. Until somebody makes such a proposal, however, I still don't personally think that ty should be marked down on its conformance score due to behaviour that is not mentioned in the spec or any accepted PEP

davidhalter:
> I'm open to a spec change here if somebody wants to propose that. I agree that there's an argument for doing so. Until somebody makes such a proposal, however, I still don't personally think that ty should be marked down on its conformance score due to behaviour that is not mentioned in the spec or any accepted PEP

Can I make the same argument for the `type: ignore[...]` issue that affects Zuban and Mypy (#2153)? It is currently also not specified. I'm open to having the spec being the single source of truth, but in that case I would like to be consistent. I'm a bit torn on this issue, because it feels like the spec is kind of incomplete.

It feels like it's very hard to get spec changes merged and I therefore doubt that these kind of things are going to be added, since it takes a lot of effort to get even tiny changes through. I don't think this is necessarily a bad thing, for the record.

AlexWaygood:
We recently changed our behaviour in ty to suppress all diagnostics in unreachable code (except for some very special-cased ones such as `reveal_type` diagnostics). So this PR is no longer necessary for ty's conformance score.

I still personally feel inclined to land it regardless, on the grounds that the conformance suite shouldn't be asserting things not mentioned by the spec; this limits the room for type checkers to experiment with behaviour and hurts innovation. But I don't feel too strongly anymore either way, especially if folks are interested in changing the spec on this point to more clearly state that all type-checking errors should be ignored in `if not TYPE_CHECKING` blocks.

AlexWaygood:
(I also pushed some updates so that, rather than deleting the old tests outright, it just makes them optional assertions. That way we still document whether the tested type checkers suppress diagnostics in `if not TYPE_CHECKING` blocks -- as they currently all do -- without mandating that they must.)

carljm:
I agree with @JelleZijlstra and @rchen152 that type checkers should not emit type diagnostics in `if not TYPE_CHECKING` blocks, and that it would make sense to just specify this. I posted on Discuss, since that's part of the process for a spec change: https://discuss.python.org/t/clarifying-that-type-checkers-should-not-emit-diagnostics-in-if-not-type-checking/106902

I would favor waiting the required week for any discussion there, then updating this PR to remove the `# E?` and update the spec to require silencing diagnostics in `if not TYPE_CHECKING`.

@davidhalter regarding https://github.com/python/typing/pull/2153, I certainly support that change in some form, just haven't taken a close yet since the PR is still in draft.

davidhalter:
> I still personally feel inclined to land it regardless, on the grounds that the conformance suite shouldn't be asserting things not mentioned by the spec; this limits the room for type checkers to experiment with behaviour and hurts innovation.

I feel like in general there are just so many things that are not part of the spec, since it is very incomplete. A lot of things are just "implicit" knowledge on how Python work and how Pyright/Mypy have worked in a similar way over the past few years. In my opinion this is fine and not a problem. I would want to avoid stripping the conformance tests of all of the assumptions and rather try to make it easier to add the missing "implicit" parts to the spec. But I don't have strong feelings here.

carljm:
> rather try to make it easier to add the missing "implicit" parts to the spec

In what way should it be easier? It's really not that hard: make a post on Discuss, wait a week, submit a PR. The level of difficulty is really determined entirely by how controversial the clarification is, which seems like roughly how it should be.

> I would want to avoid stripping the conformance tests of all of the assumptions

When there are things codified in the conformance suite but not mentioned in the spec (which implies "not mentioned in a PEP", which implies a lower level of community scrutiny), _and_ there is a type checker which has a reasoned perspective on how it should work that differs from the conformance suite assertion, innovation should not be constrained simply by "well, everyone else has always done it this way, and someone landed a conformance suite test at some point asserting that". That said, I think that innovation (at least, innovation that wants to claim conformance!) _should_ be constrained by whether the community / typing council feel it's actually a good idea! So type checkers _should_ continue to push back on unspecified assumptions encoded in the conformance suite, and there are (at least) two possible outcomes to that: we can say "oh yeah, it does seem reasonable to try different approaches here" and adjust/remove the assertion, or we can say "no, actually, this seems like a point where consistency has value, and we should clarify the spec." I think this particular PR falls into the latter case, but I don't agree that we should adopt an attitude that whatever the conformance suite happens to assert is de facto specified.

PR #2215, opened 2026-03-09
parameter specification/type variable tuple variance

KotlinIsland:
discussion: https://discuss.python.org/t/parameter-specifications-should-have-variance/106452

- [x] support in basedpyright https://github.com/DetachHead/basedpyright/pull/1771
- [x] support in pycharm: https://github.com/JetBrains/intellij-community/commit/7196a70ee4e147f4a02fee80b1575ea03b9e3210
- [x] cpython tvt variance: https://github.com/python/cpython/pull/148212
- [x] typeshed: https://github.com/python/typeshed/pull/15670 / https://github.com/python/typeshed/pull/15725
- [x] typing_extensions: https://github.com/python/typing_extensions/pull/741
- [x] typing-council: https://github.com/python/typing-council/issues/59


conformance/tests/generics_paramspec_semantics.py
```diff
@@ -142,3 +142,38 @@ def three(**kwargs: int) -> int:
...
+in_int_old: ContravariantParamSpecOld[int] = in_obj_old  # OK
+in_obj_old = in_int_old  # E
+
+OutP = ParamSpec("OutP", covariant=True)
+
+class CovariantParamSpecOld(Generic[OutP]):
+    def in_f(self, *args: OutP.args, **kwargs: OutP.kwargs): ...  # E
+    def out_f(self) -> Callable[OutP, None]: ...  # OK
+
+out_int_old: ContravariantParamSpecOld[int] = ContravariantParamSpecOld()
```
JelleZijlstra: This should presumably be using CovariantParamSpecOld, not ContravariantParamSpecOld.

conformance/tests/generics_paramspec_semantics.py
```diff
@@ -142,3 +142,38 @@ def three(**kwargs: int) -> int:
...
+
+class ContravariantParamSpec[**InP]:
+    def f(self, *args: InP.args, **kwargs: InP.kwargs): ...
+
+in_obj: ContravariantParamSpec[object] = ContravariantParamSpec()
+in_int: ContravariantParamSpec[int] = in_obj  # OK
+in_obj = in_int  # E
+
+
+class CovariantParamSpec[**OutP]:
```
JelleZijlstra: This class should not actually be covariant. Callable's argument is a contravariant position, so the variance flips.

You can see that in current type checkers with a simpler example that just uses a TypeVar:

```python
from typing import Callable

class X[T]:
    def meth(self) -> Callable[[T], None]:
        raise NotImplementedError

def take_xint(x: X[int]): ...
def take_xobj(x: X[object]): ...

def f(xobj: X[object], xint: X[int]):
    take_xint(xobj)  # ok
    take_xobj(xint)  # error
```
`X` here is contravariant.

A correct covariant class would look like this:

```python
from typing import Callable

class UseCallback[T]:
    def meth(self, callback: Callable[[T], None]):
        raise NotImplementedError

def take_xint(x: UseCallback[int]): ...
def take_xobj(x: UseCallback[object]): ...

def f(xobj: UseCallback[object], xint: UseCallback[int]):
    take_xint(xobj)  # error
    take_xobj(xint)  # ok
```

JelleZijlstra (changes requested):
I like adding this if we can get it specified nicely, but this PR is not ready; the proposed test is incorrect.

Also, https://typing.python.org/en/latest/spec/generics.html#paramspec-variables still says variance on ParamSpec is unsupported; this should be updated.

I'd also like to see an implementation in at least one type checker, even if only as a draft PR, so we can be confident this is something that can be feasibly implemented.

davidhalter:
I personally would also like to see tests added for param spec variance inference, since that would probably need to be supported as well.

JelleZijlstra:
The test cases do have variance inference, though as I noted some of the cases are wrong. But it would probably be useful to have a few more cases, and I'd recommend putting the tests for paramspec variance in their own file so we can track type checker support more precisely.

KotlinIsland:
> I'd also like to see an implementation in at least one type checker

I have wip support in PyCharm, but I could also add it to basedpyright 

it was very straightforward to implement

davidhalter:
> The test cases do have variance inference, though as I noted some of the cases are wrong. But it would probably be useful to have a few more cases, and I'd recommend putting the tests for paramspec variance in their own file so we can track type checker support more precisely.

Oh right, sorry I didn't see them, because I thought they would be in a different file. I'm also very much +1 on putting those tests into a different file (for example `generics_paramspec_variance.py`).

KotlinIsland:
support has landed in pycharm and cpython

JelleZijlstra:
https://typing.python.org/en/latest/spec/generics.html#paramspec-variables stills says that we don't support variance in ParamSpec, that should be fixed in this PR.

JelleZijlstra:
Also can you open an issue on python/typing-council asking for a formal pronouncement?

KotlinIsland:
https://github.com/python/typing-council/issues/59

JelleZijlstra:
Noticed some more issues:

- `generics_typevartuple_basic.py` still has a line `Ts1 = TypeVarTuple("Ts1", covariant=True)  # E`. But of course that's legal now. I'd delete the line and leave all variance testing to the new file generics_typevartuple_variance.py.
- The description of the variance inference algorithm says "In the ``upper`` specialized class, specialize the target type parameter with an ``object`` instance". But that doesn't make sense for ParamSpec and TypeVarTuple. I think TypeVarTuple should instead be specialized with `*tuple[object, ...]`. For `ParamSpec` perhaps `Callable[..., object]` is right but not 100% sure.
- The way the tests are written produces some incidental errors, like from ty ```generics_paramspec_variance.py:28:23: error[empty-body] Function always implicitly returns `None`, which is not assignable to return type `(...) -> Unknown` ``` and mypy ```generics_typevartuple_variance.py:18: error: Missing return statement  [empty-body]```. Should add something like `raise NotImplementedError`.

KotlinIsland:
> `Callable[..., object]` is right but not 100% sure.

not `Callable`, and not `...`, as this is the gradual form. but `(*object, **object)`, which is a non-denotable psuedo representation of the widest value for a parameter specification

JelleZijlstra:
I don't think that's right. The form we want here should be a supertype of every other possible value, and something like a one-parameter callable is not a subtype of `(*: object, **: object)`.

AlexWaygood:
In ty we have a type that we spell as `Top[Callable[..., object]]` in ty's type display (a fully static type describing the infinite union of all possible `Callable` types). It's not representable using standard Python notation currently.

JelleZijlstra:
We only need assignability not subtyping here, so I think `...` is enough.

conformance/tests/generics_typevartuple_basic.py
```diff
@@ -60,9 +60,8 @@ def method1(*args: Shape) -> None:  # E: not unpacked
         ...
 
 
-# > TypeVarTuple does not yet support specification of variance, bounds, constraints.
+# > TypeVarTuple does not yet support specification of bounds, constraints.
 
-Ts1 = TypeVarTuple("Ts1", covariant=True)  # E
```
carljm, line 65: This is is used by `Array3` below, so line 107 is now failing for the wrong reason (name error) in all type checkers.

It might be better for the `Array3` test to define its own TypeVarTuples right above that test, instead of relying on these TypeVarTuples doing double duty.

docs/spec/generics.rst
```diff
@@ -678,8 +678,8 @@ Prior to 3.12, the ``ParamSpec`` constructor can be used.
    P = ParamSpec("WrongName") # Rejected because P =/= WrongName
 
 The runtime should accept ``bound``\ s and ``covariant`` and ``contravariant``
-arguments in the declaration just as ``typing.TypeVar`` does, but for now we
-will defer the standardization of the semantics of those options to a later PEP.
+arguments in the declaration just as ``typing.TypeVar`` does, but for ``bound`` we
+will defer the standardization of the semantics of this option to a later PEP.
```
carljm:
```suggestion
arguments in the declaration just as ``typing.TypeVar`` does.

We defer the standardization of the semantics of the ``bound`` option to a later PEP.
```

docs/spec/generics.rst
```diff
@@ -2712,25 +2711,27 @@ The algorithm for computing the variance of a type parameter is as follows.
 
 For each type parameter in a generic class:
 
-1. If the type parameter is variadic (``TypeVarTuple``) or a parameter
-specification (``ParamSpec``), it is always considered invariant. No further
-inference is needed.
+1. If the type parameter comes from a traditional
+``TypeVar``/``TypeVarTuple``/``ParamSpec`` declaration and is not specified
+as ``infer_variance`` (see below), its variance is specified by the
```
carljm:
```suggestion
``TypeVar``/``TypeVarTuple``/``ParamSpec`` declaration and is not constructed
with ``infer_variance=True`` (see below), its variance is specified by the
```

docs/spec/generics.rst
```diff
@@ -2712,25 +2711,27 @@ The algorithm for computing the variance of a type parameter is as follows.
...
 constraints of the type parameter). In the ``upper`` specialized class,
-specialize the target type parameter with an ``object`` instance. This
-specialization ignores the type parameter's upper bound or constraints. In the
-``lower`` specialized class, specialize the target type parameter with itself
-(i.e. the corresponding type argument is the type parameter itself).
+specialize the target type parameter with:
+
+ an ``object`` instance for a type variable.
+ a ``*tuple[object, ...]`` value for a type variable tuple.
+ a ``...`` value for a parameter specification.
```
carljm: I don't think using `...` here is sound. Because it is both assignable-to and assignable-from any paramspec, it means that a case like the `ContravariantParamSpec` test will wrongly infer the paramspec as covariant (because `lower` is assignable to `upper`) before it even checks the contravariant direction.

I don't think there is any way to make the variance-inference algorithm described here correct and support inference of ParamSpec variance, without introducing the new concept of a "top signature".

(I'm also not sure how much value there is in including this particular variance-inference algorithm in the spec. I think it describes pyright's algorithm. I know that ty and pyrefly do not use this algorithm. I don't know about mypy or zuban.)

conformance/tests/generics_paramspec_variance.py
```diff
@@ -0,0 +1,50 @@
+"""
+Tests variance of ParamSpec.
```
carljm, line 2: I think we should also have tests for invariant paramspecs (would have to be used more than once in the class)

conformance/tests/generics_typevartuple_variance.py
```diff
@@ -0,0 +1,53 @@
+"""
+Tests variance of TypeVarTuple.
```
carljm, line 2: Here too, I think we should have tests for invariant TypeVarTuple.

carljm (changes requested):
Thanks! Definitely support this change.

carljm:
> We only need assignability not subtyping here, so I think `...` is enough.

But we also need assignability to fail in the "wrong" direction, and this workaround fails that requirement.

docs/spec/generics.rst
KotlinIsland: > (I'm also not sure how much value there is in including this particular variance-inference algorithm in the spec. I think it describes pyright's algorithm. I know that ty and pyrefly do not use this algorithm. I don't know about mypy or zuban.)

i agree, but i think this point is out of the scope of this change

JelleZijlstra: We need to describe the algorithm somehow, don't we? Other type checkers use a different algorithm in practice, but to be conformant they need to follow an algorithm that has the same results as that described in the spec.

carljm: Depending on the situation I think it can be fine (or preferable) for the spec to describe the required results without specifying a particular algorithm to reach them. In this case I'm not sure that would be any simpler (and in any case I don't think we should remove this algorithm in this PR.)

I think for this PR we can fairly simply describe the nature of the top signature (the union of all possible callable signatures) without having to go too deep into it.

KotlinIsland:
i've pushed changes, it's good to review again

conformance/tests/generics_typevartuple_variance.py
```diff
@@ -0,0 +1,65 @@
...
+
+out_int: CovariantTypeVarTuple[int] = CovariantTypeVarTuple[object]()  # E
+out_obj: CovariantTypeVarTuple[object] = CovariantTypeVarTuple[int]()  # OK
+out_multiple: CovariantTypeVarTuple[float, float] = CovariantTypeVarTuple[
+    int,  # OK
+    object,  # E
+]()
+
+
+InTs = TypeVarTuple("InTs", contravariant=True)
```
JelleZijlstra: This is a problem because we run the tests on 3.12, and on 3.12 typing.TypeVarTuple doesn't have a `contravariant` argument, so type checkers would be in their right to complain about this line. We should use `typing_extensions.TypeVarTuple` instead, though even that is a bit problematic because we haven't released a version of typing-extensions with the change yet either. (Hopefully soon.)

JelleZijlstra (approved):
Agree with this idea, we might have to tweak the exact PR a bit more.

conformance/tests/generics_typevartuple_variance.py
KotlinIsland: i added tests for invariant tvt

docs/spec/generics.rst
KotlinIsland: i removed the `...`

docs/spec/generics.rst
```diff
@@ -2712,25 +2712,28 @@ The algorithm for computing the variance of a type parameter is as follows.
...
 constraints of the type parameter). In the ``upper`` specialized class,
-specialize the target type parameter with an ``object`` instance. This
-specialization ignores the type parameter's upper bound or constraints. In the
-``lower`` specialized class, specialize the target type parameter with itself
-(i.e. the corresponding type argument is the type parameter itself).
+specialize the target type parameter with:
+
+ an ``object`` instance for a type variable.
+ a ``*tuple[object, ...]`` value for a type variable tuple.
+ a for a parameter specification, a 'top signature' value, i.e. a type
```
davidhalter: I assume `a for a` is not wanted? The whole paragraph feels a bit weird and the formatting might be a bit off.

conformance/tests/generics_typevartuple_basic.py
```diff
@@ -104,5 +103,5 @@ def func3(x: Array[Height], y: Array[Width], z: Array[Height, Width]):
 # > Only a single type variable tuple may appear in a type parameter list.
 
 
-class Array3(Generic[*Ts1, *Ts2]):  # E
+class Array3[*Ts1, *Ts2]:  # E
```
davidhalter: Personally I would prefer to the both the new and the old syntax.

davidhalter (commented):
I like the general approach. I think it would also be useful to have at least one type checker with a proper implementation of this new feature.

KotlinIsland:
> proper implementation of this new feature.

see the mr description, it's implemented in pycharm and basedpyright

docs/spec/generics.rst
KotlinIsland: addressed points 1 and 3. do you have a suggestion on how to improve the wording?

conformance/tests/generics_typevartuple_basic.py
KotlinIsland: done

docs/spec/generics.rst
davidhalter: I would probably go with a list here:

```suggestion
-  an ``object`` instance for a type variable
- a ``*tuple[object, ...]`` value for a type variable tuple
- a "top signature" value for a parameter specification, i.e. a type that represents the super type of every possible signature.
```

conformance/tests/generics_typevartuple_variance.py
```diff
@@ -0,0 +1,79 @@
...
+class CovariantTypeVarTuple[*OutTs]:
+    def f(self) -> tuple[*OutTs]:
+        raise NotImplementedError
+
+
+out_int: CovariantTypeVarTuple[int] = CovariantTypeVarTuple[object]()  # E
+out_obj: CovariantTypeVarTuple[object] = CovariantTypeVarTuple[int]()  # OK
+out_multiple: CovariantTypeVarTuple[float, float] = CovariantTypeVarTuple[
+    int,  # OK
+    object,  # E
```
rchen152: I don't think we should require that the error be reported on this exact line. Generally, the conformance tests allow an error to be reported on any reasonable line. Furthermore, I tested this snippet:
```

T1 = TypeVar("T1", covariant=True)
T2 = TypeVar("T2", covariant=True)

class C(Generic[T1, T2]): pass

x: C[float, float] = C[
    int,
    object,
]()
```
in mypy, pyright, pyrefly, and ty, and they all reported the error either on the first line of `x`'s definition or on the entire definition, which presumably the current conformance tests accept, so it would be odd to have a different requirement for TypeVarTuple than for TypeVar.

docs/spec/generics.rst
KotlinIsland: done

conformance/tests/generics_typevartuple_variance.py
KotlinIsland: okay, addressed with a matrix of four assignments

conformance/results/ty/generics_paramspec_variance.toml
```diff
@@ -0,0 +1,14 @@
+conformant = "Unsupported"
```
carljm: I think Partial is more fair here. ty does already support ParamSpec variance; what we don't do yet is error when a legacy ParamSpec (or any TypeVar) is used in a position that violates its explicit variance.

```suggestion
conformant = "Partial"
notes = """
Does not error when a ParamSpec is used in a position that violates its explicit variance.
"""
```

rchen152 (approved):

carljm (approved):

davidhalter (approved):

jorenham (approved):
The lack of co/contra variance for `TypeVarTuple`s has caused problems for me on several occasions in the past. So thanks for patching this hole in the spec!

The changes to the variance algorithm also look like the way to go. It's too bad we don't have a general notion of the "top signature" like we do for the general top type (`object`) and the top product type (`tuple[object, ...]`). But I expect that the way you informally describe it here will be understood just fine by anyone that needs to know.

KotlinIsland:
@carljm good to review now

conformance/results/basilisk/generics_paramspec_variance.toml
```diff
@@ -0,0 +1,26 @@
+conformant = "Unsupported"
```
carljm: Basilisk is no longer in the conformance suite, so all the Basilisk result files should be removed.

conformance/results/pycroscope/generics_typevartuple_basic.toml
```diff
@@ -1,15 +1,16 @@
-conformance_automated = "Pass"
+conformant = "Unsupported"
```
carljm: If there's just one error diff, then this should likely be Partial, not Unsupported, and there should be a "notes" field added describing the nature of the gap.

conformance/tests/generics_typevartuple_variance.py
```diff
@@ -0,0 +1,79 @@
...
+
+from typing import Generic
+from typing_extensions import TypeVarTuple
+
+class InvariantTypeVarTuple[*InOutTs]:
+    a: tuple[*InOutTs]
+
+in_out_obj: InvariantTypeVarTuple[object] = InvariantTypeVarTuple[int]()  # E
+in_out_int: InvariantTypeVarTuple[int] = InvariantTypeVarTuple[object]()  # E
+in_out_int = InvariantTypeVarTuple[int]()  # E
```
carljm: Why is this expected to be an error? It seems like we are assigning an instance of `InvariantTypeVarTuple[int]` to a variable annotated as `InvariantTypeVarTuple[int]` -- that should be a valid assignment under invariance.

conformance/tests/generics_typevartuple_variance.py
```diff
@@ -0,0 +1,79 @@
...
+    def f(self) -> tuple[*OutTs]:
+        raise NotImplementedError
+
+
+out_int: CovariantTypeVarTuple[int] = CovariantTypeVarTuple[object]()  # E
+out_obj: CovariantTypeVarTuple[object] = CovariantTypeVarTuple[int]()  # OK
+out_multiple1: CovariantTypeVarTuple[float, float] = CovariantTypeVarTuple[int,  int]()  # OK
+out_multiple2: CovariantTypeVarTuple[float, float] = CovariantTypeVarTuple[int,  object]()  # E
+out_multiple3: CovariantTypeVarTuple[float, float] = CovariantTypeVarTuple[object,  int]()  # E
+out_multiple4: CovariantTypeVarTuple[float, float] = CovariantTypeVarTuple[object,  object]()  # E
```
carljm: I wouldn't recommend to use `float` in these tests, as it entangles this test unnecessarily with the whole "complex/float special case" situation. If we were to ever remove that special case, this test would need rewriting. It's easily avoidable by using `int/bool` instead, or just defining some custom classes.

conformance/tests/generics_typevartuple_variance.py
```diff
@@ -0,0 +1,79 @@
...
+    def in_f(self, *args: *InTs) -> None:  # OK
+        raise NotImplementedError
+
+    def out_f(self) -> tuple[*InTs]:  # OK
+        raise NotImplementedError
+
+
+obj_old: InvariantTypeVarTupleOld[object] = InvariantTypeVarTupleOld[int]()  # E
+int_old: InvariantTypeVarTupleOld[int] = InvariantTypeVarTupleOld[object]()  # E
+
```
carljm, line 71: Show here a successful assignment with matching types?

conformance/tests/generics_typevartuple_variance.py
```diff
@@ -0,0 +1,79 @@
...
+out_multiple4: CovariantTypeVarTuple[float, float] = CovariantTypeVarTuple[object,  object]()  # E
+
+
+Ts = TypeVarTuple("Ts")
+
+class InvariantTypeVarTupleOld(Generic[*Ts]):
+    def in_f(self, *args: *InTs) -> None:  # OK
+        raise NotImplementedError
+
+    def out_f(self) -> tuple[*InTs]:  # OK
```
carljm: These two method definitions use `InTs` (defined below) where they should use `Ts` (defined above)

conformance/results/pycroscope/generics_typevartuple_variance.toml
```diff
@@ -0,0 +1,34 @@
...
+Line 33: Unexpected errors ['./generics_typevartuple_variance.py:33:0: Incompatible assignment: expected ./generics_typevartuple_variance.py.CovariantTypeVarTuple[object], got ./generics_typevartuple_variance.py.CovariantTypeVarTuple[int] [incompatible_assignment]']
+Line 34: Unexpected errors ['./generics_typevartuple_variance.py:34:0: Incompatible assignment: expected ./generics_typevartuple_variance.py.CovariantTypeVarTuple[float | int, float | int], got ./generics_typevartuple_variance.py.CovariantTypeVarTuple[int, int] [incompatible_assignment]']
+Line 54: Unexpected errors ["./generics_typevartuple_variance.py:54:7: Error calling (name: str, /, *, default: TypeForm[object] = <no argument given>, bound: TypeForm[object] = <no argument given>, covariant: bool = Literal[False], contravariant: bool = Literal[False], infer_variance: bool = Literal[False]) -> typing_extensions.TypeVarTuple (with impl): TypeVarTuple.__new__() got an unexpected keyword argument 'contravariant' [incompatible_call]"]
```
carljm: These errors show that the conformance suite is running with an older version of ty-extensions (4.15.0 is shown in `uv.lock` still in this PR) which doesn't support the new TypeVarTuple parameters. We need to set a minimum ty-extensions version of 4.16.0.

conformance/results/ty/generics_typevartuple_basic.toml
```diff
@@ -1,20 +1,21 @@
 conformance_automated = "Pass"
+conformant = "Partial"
```
carljm: This should not be marked "Partial" when automated conformance is "Pass" -- this is causing CI to fail.

docs/spec/generics.rst
```diff
@@ -2712,25 +2712,28 @@ The algorithm for computing the variance of a type parameter is as follows.
...
 
-1. If the type parameter is variadic (``TypeVarTuple``) or a parameter
-specification (``ParamSpec``), it is always considered invariant. No further
-inference is needed.
-
-2. If the type parameter comes from a traditional ``TypeVar`` declaration and
-is not specified as ``infer_variance`` (see below), its variance is specified
-by the ``TypeVar`` constructor call. No further inference is needed.
+1. If the type parameter comes from a traditional
+``TypeVar``/``TypeVarTuple``/``ParamSpec`` declaration and is not constructed
```
carljm, line 2718: What happens to existing callback protocols that use a legacy `ParamSpec` without explicit variance? We already have this accepted conformance case in `protocols_variance.py`:

```python
P = ParamSpec("P")
R = TypeVar("R", covariant=True)

class Callback(Protocol[P, R]):  # OK
    def __call__(self, *args: P.args, **kwargs: P.kwargs) -> R: ...
```

Under this rule, `P` is invariant by default, but it is used contravariantly here, and the protocol spec says this should be an error (declared variance must match usage). Mypy currently exempts `ParamSpec` and `TypeVarTuple` from that check specifically because their variance was previously unspecified.

I think we should update that `protocols_variance.py` conformance test to declare `P` as contravariant.

I'm not sure if we need to do something about the backwards-incompatibility. I guess we can leave it up to type-checkers how to handle this.

conformance/tests/generics_typevartuple_variance.py
```diff
@@ -0,0 +1,79 @@
...
+in_out_obj: InvariantTypeVarTuple[object] = InvariantTypeVarTuple[int]()  # E
+in_out_int: InvariantTypeVarTuple[int] = InvariantTypeVarTuple[object]()  # E
+in_out_int = InvariantTypeVarTuple[int]()  # E
+
+
+class ContravariantTypeVarTuple[*InTs]:
+    def f(self, t: tuple[*InTs]):
+        raise NotImplementedError
+
+in_obj: ContravariantTypeVarTuple[object, object] = ContravariantTypeVarTuple[int]()  # E
```
carljm: This assignment changes the tuple arity as well as the element type, so it fails regardless of whether the `TypeVarTuple` is invariant, covariant, or contravariant. Could we use `ContravariantTypeVarTuple[object] = ContravariantTypeVarTuple[int]()` for the negative case and test different arities separately? Otherwise we never check the forbidden direction with the same arity.

conformance/results/ty/generics_typevartuple_variance.toml
```diff
@@ -0,0 +1,24 @@
+conformant = "Unsupported"
```
carljm: I don't think `Unsupported` matches these results. Once the incorrect expected error on line 16 and the undefined `InTs` references are fixed, ty supports the variance inference and assignment cases; its remaining gap is not reporting invalid uses of an explicitly variant `TypeVarTuple`. Could we mark this `Partial` with an explanatory note, as we already do for ty's `ParamSpec` variance results?

docs/spec/generics.rst
```diff
@@ -2712,25 +2712,28 @@ The algorithm for computing the variance of a type parameter is as follows.
...
+with ``infer_variance=True`` (see below), its variance is specified by the
```
carljm, line 2719: This now specifies `infer_variance=True` for traditional `ParamSpec` and `TypeVarTuple`, but the "Auto Variance For TypeVar" section below still only describes `TypeVar`, and neither new conformance file tests this case. Could we update that section and cover inferred legacy variance for both constructors, including rejection of `covariant=True, contravariant=True` and explicit variance combined with `infer_variance=True`?

conformance/tests/generics_paramspec_variance.py
```diff
@@ -0,0 +1,58 @@
...
+
+class CovariantParamSpec[**OutP]:
+    def f(self, fn: Callable[OutP, None]) -> None:
+        raise NotImplementedError
+
+
+out_int: CovariantParamSpec[int] = CovariantParamSpec[object]()  # E
+out_obj: CovariantParamSpec[object] = CovariantParamSpec[int]()  # OK
+
+InP = ParamSpec("InP", contravariant=True)
```
carljm: Could we also cover a traditional `P = ParamSpec("P")` with no variance arguments? The file tests inferred inline invariance and explicitly covariant/contravariant legacy variables, but not the legacy default-invariant case. I would expect rejected assignments in both directions plus a successful same-specialization assignment.

conformance/tests/generics_paramspec_variance.py
```diff
@@ -0,0 +1,58 @@
...
+
+in_out_obj: InvariantParamSpec[object] = InvariantParamSpec[int]()  # E
+in_out_int: InvariantParamSpec[int] = InvariantParamSpec[object]()  # E
+
+
+class ContravariantParamSpec[**InP]:
+    def f(self, *args: InP.args, **kwargs: InP.kwargs): ...
+
+in_obj: ContravariantParamSpec[object] = ContravariantParamSpec[int]()  # E
+in_int: ContravariantParamSpec[int] = ContravariantParamSpec[object]()  # OK
```
carljm, line 22: All of the `ParamSpec` specializations here have one required positional parameter. Could we add cases involving keyword-only or positional-only parameters, parameter names, defaults, and differing callable arities, perhaps also `Concatenate`? Otherwise an implementation that compares only a tuple of parameter types could pass without implementing actual signature assignability.

docs/spec/generics.rst
```diff
@@ -2712,25 +2712,28 @@ The algorithm for computing the variance of a type parameter is as follows.
...
+constructor call. No further inference is needed.
 
-3. Create two specialized versions of the class. We'll refer to these as
+2. Create two specialized versions of the class. We'll refer to these as
```
carljm, line 2722: What should the "dummy type instance" be when one of the other parameters is a `TypeVarTuple` or `ParamSpec`? An ordinary dummy class isn't a valid specialization for either kind. Could we spell out kind-compatible substitutions and cover a mixed example such as:

```python
class Mixed[T, *Ts, **P]:
    def f(self, x: T, /, *args: P.args, **kwargs: P.kwargs) -> tuple[*Ts]: ...
```

I would expect `T` and `P` to be contravariant and `Ts` to be covariant, independently.

conformance/tests/generics_typevartuple_variance.py
```diff
@@ -0,0 +1,79 @@
...
+in_int: ContravariantTypeVarTuple[int] = ContravariantTypeVarTuple[object]()  # OK
+
+
+class CovariantTypeVarTuple[*OutTs]:
+    def f(self) -> tuple[*OutTs]:
+        raise NotImplementedError
+
+
+out_int: CovariantTypeVarTuple[int] = CovariantTypeVarTuple[object]()  # E
+out_obj: CovariantTypeVarTuple[object] = CovariantTypeVarTuple[int]()  # OK
```
carljm, line 43: Could we also exercise empty and arbitrary-length packs, such as `C[()]` and `C[*tuple[int, ...]]`, including fixed-length versus unbounded specializations? The inference algorithm explicitly uses `*tuple[object, ...]`, but the current cases cover only fixed packs of one or two elements.

carljm (changes requested):
The direction here remains correct IMO, but I think there are some updates needed.

conformance/tests/generics_typevartuple_variance.py
KotlinIsland: agreed, i think it was just an oversight

conformance/results/pycroscope/generics_typevartuple_variance.toml
KotlinIsland: correction: typing-extensions, not ty-extensions

docs/spec/generics.rst
KotlinIsland: good point, i agree with you. in my opinion this notion of variance matching on protocols is more like a lint rule, so i see no great need to enforce the behaviour here

conformance/tests/generics_typevartuple_variance.py
KotlinIsland: changed to:
```py
in_obj: ContravariantTypeVarTuple[object, object] = ContravariantTypeVarTuple[object, int]()  # E
```

conformance/tests/generics_paramspec_variance.py
KotlinIsland: without the ability to denote any of these forms (apart from arity), i'm having trouble find a way to test them

KotlinIsland: okay, i think i managed to get some of these cases using an invariant box

conformance/tests/generics_paramspec_variance.py
```diff
@@ -0,0 +1,128 @@
...
+
+    def usage(self) -> Callable[P, None]:
+        """infer contravariance"""
+        raise NotImplementedError
+
+
+box = Box(InitP(f))
+
+kw_p = InitP(kw)
+box.t = kw_p  # OK
```
carljm: I think the expected results in both signature-shape blocks are backwards. `InitP` produces a `Callable[P, None]`, so assigning a keyword-only or positional-only callback where `InitP(f)` is expected is unsafe: callers can use a call form the replacement callback does not accept. The defaulted callback is safe in this direction.

For `OutitP`, which consumes a `Callable[P, None]`, the keyword-only and positional-only cases should be accepted and the defaulted case rejected. Once the annotations are fixed, the parameter-name expectations need to flip too. Could we correct the expectations on lines 58, 60, 62, 64, 79, 81, 83, and 85? The ty results already expose these mismatches.

conformance/tests/generics_paramspec_variance.py
```diff
@@ -0,0 +1,128 @@
...
+
+
+box = Box(OutitP(f))
+
+kw_p = OutitP(kw)
+box.t = kw_p  # E
+pos_p = OutitP(pos)
+box.t = pos_p  # E
+names_p = OutitP(f)
+_: OutitP[int]() = names_p  # OK
```
carljm: Another call expression in a type expression, and assignment to `_` which is not a good idea since it's likely to lead to reuse of the same name, causing duplicate-declaration errors for some type checkers.

conformance/tests/generics_paramspec_variance.py
```diff
@@ -0,0 +1,128 @@
...
+default_p = OutitP(default)
+box.t = default_p  # OK
+arity_p = OutitP(arity)
+box.t = arity_p  # E
+
+
+# old style
+P = ParamSpec("InP")  # OK
```
carljm: `P = ParamSpec("InP")` violates the matching-name rule, so all six checkers reject the declaration that is supposed to exercise default invariance. `InvariantParamSpecOld.f` also uses the unrelated `InP` rather than its own `P`, and `InvP2`/`InvP3` use mismatched constructor names, allowing those negative cases to pass for the wrong reason.

docs/spec/generics.rst
carljm, line 2722: What should the dummy specialization be when one of the other parameters is a `TypeVarTuple` or `ParamSpec`? A concrete anonymous class is a valid dummy only for an ordinary `TypeVar`; it cannot stand in for a variadic pack or parameter signature. That means the algorithm as written here still cannot construct the specializations needed for the newly added `Mixed[T, *Ts, **P]` conformance example (which looks great). Could we spell out kind-compatible dummy substitutions for all three parameter kinds here?

conformance/tests/generics_paramspec_variance.py
```diff
@@ -0,0 +1,128 @@
...
+
+
+class OutitP[**P]:  # covariant
+    def __init__(self, fn: Callable[P, None]): ...
+
+    def usage(self, fn: Callable[P, None]):
+        """infer covariance"""
+
+
+box = Box(OutitP(f))
```
carljm: `box` was already inferred as `Box[InitP[...]]` above, and `kw_p`, `pos_p`, `names_p`, `default_p`, and `arity_p` likewise already have `InitP` types. Reusing those names for `OutitP` produces six unrelated incompatible-assignment errors in the mypy results. We should give the second block distinct names or isolate the two scenarios in separate scopes.

docs/spec/generics.rst
carljm, line 2719: This now specifies `infer_variance=True` for traditional `ParamSpec` and `TypeVarTuple`, but neither conformance file contains a successful declaration or assignments demonstrating inferred covariance and contravariance. Every existing use is marked `# E`, so a checker that rejects `infer_variance` as an unknown option can still appear to pass.

conformance/uv.lock
```diff
@@ -280,28 +280,28 @@ requires-dist = [
 
 [[package]]
 name = "typing-extensions"
-version = "4.15.0"
+version = "4.16.0"
```
carljm, line 285: I think updating the transitive lockfile entry to `typing-extensions` 4.16.0 does not actually enforce the minimum version these new `TypeVarTuple` tests require. `conformance/pyproject.toml` does not declare `typing-extensions` at all, so an unlocked or refreshed environment can still install an older unsupported version. Could we add `typing-extensions>=4.16.0` as a direct conformance dependency?

conformance/tests/generics_typevartuple_variance.py
```diff
@@ -0,0 +1,91 @@
...
+out_multiple1: CovariantTypeVarTuple[int, int] = CovariantTypeVarTuple[bool,  bool]()  # OK
+out_multiple2: CovariantTypeVarTuple[int, int] = CovariantTypeVarTuple[bool,  object]()  # E
+out_multiple3: CovariantTypeVarTuple[int, int] = CovariantTypeVarTuple[object,  bool]()  # E
+out_multiple4: CovariantTypeVarTuple[int, int] = CovariantTypeVarTuple[object,  object]()  # E
+out_variadic_int: CovariantTypeVarTuple[*tuple[int, ...]] = CovariantTypeVarTuple[*tuple[object, ...]]()  # E
```
carljm, line 48: These additions compare fixed packs with fixed packs, unbounded packs with unbounded packs, and empty packs with themselves, but they still never compare a fixed-length specialization with an unbounded specialization. Could we cover both directions between something like `C[int]` and `C[*tuple[int, ...]]` for invariant, covariant, and contravariant classes? That would directly exercise the unbounded-pack behavior discussed in the inference algorithm.

conformance/results/ty/generics_paramspec_variance.toml
```diff
@@ -0,0 +1,40 @@
...
+Line 64: Expected 1 errors
+Line 79: Expected 1 errors
+Line 81: Expected 1 errors
+Line 110: Expected 1 errors
+Line 121: Expected 1 errors
+Line 58: Unexpected errors ['generics_paramspec_variance.py:58:1: error[invalid-assignment] Object of type `InitP[(*, a: int)]` is not assignable to attribute `t` of type `InitP[(a: int)]`']
+Line 60: Unexpected errors ['generics_paramspec_variance.py:60:1: error[invalid-assignment] Object of type `InitP[(a: int, /)]` is not assignable to attribute `t` of type `InitP[(a: int)]`']
+Line 83: Unexpected errors ['generics_paramspec_variance.py:83:4: error[invalid-type-form] Function calls are not allowed in type expressions', 'generics_paramspec_variance.py:83:4: error[missing-argument] No argument provided for required parameter `fn` of `OutitP.__init__`']
+Line 85: Unexpected errors ['generics_paramspec_variance.py:85:1: error[invalid-assignment] Object of type `OutitP[(a: int = 1)]` is not assignable to attribute `t` of type `OutitP[(a: int)]`']
+Line 91: Unexpected errors ['generics_paramspec_variance.py:91:15: error[mismatched-type-name] The name passed to `ParamSpec` must match the variable it is assigned to: Expected "P", got "InP"']
```
carljm: These errors do not look like "Does not error when a ParamSpec is used in a position that violates its explicit variance." So either the "notes" field above is incomplete and there's another gap in ty's implementation, or some of these tests are incorrect.

conformance/results/ty/generics_typevartuple_variance.toml
```diff
@@ -0,0 +1,29 @@
+conformant = "Partial"
+conformance_automated = "Fail"
+notes = """
+Does not error when a ParamSpec is used in a position that violates its explicit variance.
```
carljm:
```suggestion
Does not error when a TypeVarTuple is used in a position that violates its explicit variance.
```

conformance/tests/generics_paramspec_variance.py
```diff
@@ -0,0 +1,128 @@
...
+
+
+box = Box(InitP(f))
+
+kw_p = InitP(kw)
+box.t = kw_p  # OK
+pos_p = InitP(pos)
+box.t = pos_p  # OK
+names_p = InitP(f)
+_: InitP[int]() = names_p  # E
```
carljm: Call expressions are not permitted in type expressions; all type checkers error on that, but I'm pretty sure it's not what this line is supposed to be testing.

conformance/tests/generics_typevartuple_variance.py
```diff
@@ -0,0 +1,91 @@
...
+out_variadic_object: CovariantTypeVarTuple[*tuple[object, ...]] = CovariantTypeVarTuple[*tuple[int, ...]]()  # OK
+out_empty: CovariantTypeVarTuple[()] = CovariantTypeVarTuple[()]()  # OK
+
+Ts = TypeVarTuple("Ts")  # OK
+InvTs1 = TypeVarTuple("InvTs1", covariant=True, contravariant=True)  # E
+InvTs2 = TypeVarTuple("InvTs2", covariant=True, infer_variance=True)  # E
+InvTs3 = TypeVarTuple("InvTs3", covariant=True, infer_variance=True)  # E
```
carljm: This seems like it duplicates the previous line?

docs/spec/generics.rst
```diff
@@ -2779,8 +2782,9 @@ Since ``lower`` is assignable to ``upper``, ``T3`` is covariant.
 Auto Variance For TypeVar
```
carljm: We should update this header - either to mention all three, or remove the specific mention of `TypeVar`.

docs/spec/generics.rst
carljm: This seems perhaps out of place in the typing spec? Unless it is trying to say that type checkers should also model the presence of this `__infer_variance__` attribute and its value? If we keep it, we should update it to mention when this attribute was added to `TypeVarTuple` and `ParamSpec` at runtime. But I'd be inclined to just remove it.

carljm (changes requested):
Thanks! Some things still need attention.

carljm (changes requested):
CI says conformance results need updating; can you do that and then re-request review?

carljm (approved):
Thank you! I pushed a few additional small changes (in separate commits so you can review each change if you like) -- this now looks good to me.

The most notable change is that I found three more places where the conformance suite used an implicitly-invariant ParamSpec in a location where it should be contravariant; thus a compliant checker should error on those uses, per the current spec / conformance suite. Fixing these involved a choice between unpalatable options:

1. The most obvious fix is to make those ParamSpecs explicitly `contravariant=True`, but this errors on four type checkers (mypy, pyright, zuban, pyrefly) which don't yet allow ParamSpecs to use that argument, making those type checkers fail on conformance tests that aren't really related to ParamSpec variance.
2. Another option is to define those ParamSpecs using PEP 695 syntax, which automatically opts into inferred variance, and allows those tests to stay focused on their core purpose. Unfortunately this seems to expose a crash in pycroscope.

I've gone with option (2) for now. @JelleZijlstra let me know if you are OK with landing this for now, would like to propose another alternative, or would like time to fix the pycroscope crash before we land this.

JelleZijlstra:
Let's land it, I think I need to do a round of fixing bugs in pycroscope :)

PR #2221, opened 2026-03-16
Allow errors for unsound Self definitions on non-final classes

davidhalter:
This comes after https://github.com/python/typing/pull/2148 which wants to make this a mandatory error.

But for now it should be totally fine to allow type checkers to error here, since it can clearly be shown to be unsound, see also:

https://discuss.python.org/t/disallow-access-to-class-variables-with-self/105434

I'm happy to revisit the original pull request, but for now I hope this should suffice. I will rerun the automated tests if and once no changes are required.

### DIFF
--- a/conformance/results/results.html
+++ b/conformance/results/results.html
@@ -345,7 +345,7 @@
 <tr><th class="column col1">&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;generics_self_advanced</th>
 <th class="column col2 partially-conformant"><div class="hover-text">Partial<span class="tooltip-text" id="bottom"><p>Does not infer the type of an unannotated `self` parameter to be type `Self`.</p><p>Does not retain `Self` when calling method that returns `Self`.</p><p>Does not infer the type of an unannotated `cls` parameter to be type `type[Self]`.</p><p>Does not retain `Self` when accessing attribute through `type[Self]`.</p></span></div></th>
 <th class="column col2 conformant">Pass</th>
-<th class="column col2 partially-conformant"><div class="hover-text">Partial<span class="tooltip-text" id="bottom"><p>Doesn't allow accessing `Self` in a classmethod</p></span></div></th>
+<th class="column col2 conformant">Pass</th>
 <th class="column col2 conformant"><div class="hover-text">Pass*<span class="tooltip-text" id="bottom"><p>Treats attributes not initialized on the class as instance-only</p></span></div></th>
 <th class="column col2 conformant">Pass</th>
 </tr>
--- a/conformance/results/zuban/generics_self_advanced.toml
+++ b/conformance/results/zuban/generics_self_advanced.toml
@@ -1,11 +1,5 @@
-conformant = "Partial"
-notes = """
-Doesn't allow accessing `Self` in a classmethod
-"""
-conformance_automated = "Fail"
+conformance_automated = "Pass"
 errors_diff = """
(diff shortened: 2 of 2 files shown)


davidhalter:
@JelleZijlstra Would you mind reviewing this? IMO it's a very uncontroversial change. 

The problem is (quoting from [discuss](https://discuss.python.org/t/disallow-access-to-class-variables-with-self/105434)):

```
from typing import Self, assert_type

class ParentB:
    a: list[Self]
    @classmethod
    def method4(cls) -> None:
        cls.a = [cls()]

class ChildB(ParentB):
    @classmethod
    def method3(cls) -> None:
        assert_type(cls, type[Self])
        assert_type(cls.a, list[Self]) # This can be a different type, but is not in conformance tests
        assert_type(cls.a[0], Self)  # Same
        print(cls.a)

ParentB.method4()
ChildB.method3()  # Prints [<__main__.ParentB object>]
```

There is currently no part of the spec that describes this.

conformance/tests/generics_self_advanced.py
```diff
@@ -40,6 +40,9 @@ def method2(self) -> None:
     @classmethod
     def method3(cls) -> None:
         assert_type(cls, type[Self])
-        assert_type(cls.a, list[Self])
-        assert_type(cls.a[0], Self)
+        # Allow type checkers to error here, since Self definitions on
+        # non-final classes are unsound.
+        a = cls.a  # E?
```
carljm, line 48: It's really unclear to me where a type checker aiming for soundness here should emit an error. It would be defensible IMO to emit it on `a: list[Self]` above, or to emit it on `class ChildB(ParentB)`, or to emit it on `cls.a` as here. But if `cls.a` is an error here, `self.a` in the method above should be just as much an error.

carljm: I don't necessarily object to just marking optional errors here in the places where Zuban emits them -- if other type checkers want to emit an error elsewhere, they can make a case for that in a separate PR.

I do think it's confusing to mark `cls.a` as an optional error but not `self.a`, because it suggests that one is more unsound than the other, which is not the case.

davidhalter: Should I just mark the other as an optional error as well? I don't particularly care. I still want to think about how the spec could handle these cases (both instance and class variables), but that is a separate issue.

Do you want me to change the comments? I'm not sure what exactly you would like me to change.

carljm: Agree this needs more of a re-evaluation. For now I think my preference would be to handle `self.a` and `cls.a` the same, allowing either to optionally be an error.

The comment is fine as written, but I guess should be duplicated in the method above for the `self` case also? Or perhaps just included once further above, in a way that applies to both.

davidhalter: I think that makes sense. I changed the PR accordingly.

conformance/results/zuban/typeforms_typeform.toml
```diff
@@ -1,15 +1,9 @@
-conformant = "Partial"
-notes = """
-TypeForm[Any] is not considered equivalent to itself.
-"""
-conformance_automated = "Fail"
+conformance_automated = "Pass"
```
JelleZijlstra: I'd prefer to keep this in a separate PR now that we allow fixing type checker versions, I can prepare a PR now to update it though.

davidhalter: Makes sense, I'll try to revert it.

davidhalter: Zuban is now at the same version as before and the results are much more understandable

PR #2240, opened 2026-04-06
Add conformance test for nested Optional behavior

BHUVANSH855:
This PR adds a conformance test for nested Optional types.

Specifically, it verifies that `Optional[Optional[T]]` behaves as `Optional[T]`, 
and ensures that such values are not treated as non-optional types without proper handling.

The test includes:
- Validation of type collapsing behavior for nested Optional types
- A negative case where `Optional[Optional[int]]` is incorrectly used as `int`

This helps improve coverage of edge cases related to Optional and None handling 
in the typing specification.

### DIFF
--- a/conformance/tests/specialtypes_optional_nested.py
+++ b/conformance/tests/specialtypes_optional_nested.py
@@ -1,0 +1,16 @@
+"""
+Tests for behavior of nested Optional types collapsing.
+
+Optional[Optional[T]] should behave as Optional[T].
+"""
+
+from typing import Optional, assert_type
+
+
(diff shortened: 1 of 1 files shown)


JelleZijlstra:
Why does this need a new conformance test? The behavior here is obvious and flows from general rules.

BHUVANSH855:
> Why does this need a new conformance test? The behavior here is obvious and flows from general rules.

Thanks for the feedback!

I understand that this behavior is already covered by general rules, so this test may be redundant.

Closing this PR and will look for a more suitable case to contribute.

PR #2242, opened 2026-04-07
Update generics reference to use PEP 695 type parameter syntax

BHUVANSH855:
This PR updates the `reference/generics.rst` documentation to reflect the modern type parameter syntax introduced in PEP 695.

Changes include:

* Updating initial examples (e.g., `Stack`, `Box`) to use `class Foo[T]` syntax
* Adding a note explaining the transition from `TypeVar`/`Generic` to PEP 695
* Retaining mention of older syntax for backward compatibility

This improves consistency with the spec documentation and reduces confusion for users of Python 3.12+.

Fixes #2227


docs/reference/generics.rst
```diff
@@ -96,12 +110,12 @@ can be used as a base class for another class (generic or non-generic). For exam
     data: StrDict[int, int]  # error: "StrDict" expects no type arguments, but 2 given
     data2: StrDict  # OK
 
-   # This is a user-defined generic class
-   class Receiver(Generic[T]):
-       def accept(self, value: T) -> None: ...
+    # This is a user-defined generic class
```
JelleZijlstra, line 107: This still uses the old syntax. Please make sure all examples (other than the one mentioning the old syntax explicitly) use the new syntax.

BHUVANSH855: Thanks for the review @JelleZijlstra, I’ve updated the remaining example to use PEP 695 syntax so that all examples are now consistent, with the old syntax kept only in the explicitly marked compatibility section.

docs/reference/generics.rst
```diff
@@ -6,6 +6,13 @@ code. These types are interesting in that they are parametrised by other types!
 A ``list[str]`` isn't just a list, it's a list of strings. Types with type
 parameters like this are called *generic types*.
 
+.. note::
+
+   Python 3.12 introduced :pep:`695`, which allows defining generics using
+   type parameter syntax (e.g., ``class Foo[T]:`` and ``def func[T](x: T) -> T:``).
+   The older ``TypeVar`` and ``Generic``-based syntax remains supported
+   for compatibility with earlier Python versions.
```
srittau: I wouldn't mention the PEP. This is mostly historic information and not really relevant to users. Also, the wording suggests that the user is already familiar with both syntaxes. I would rather move this note to the bottom of the section (before the old syntax example) and reword it to says something along the line that the current syntax was only introduced with Python 3.12, and here's the legacy syntax for older Python versions.

BHUVANSH855: Thanks for the feedback @srittau, I’ve removed the PEP reference, moved the note next to the legacy syntax example, and reworded it to clarify that the current syntax was introduced in Python 3.12 and the older syntax is for earlier versions.

srittau:
Thanks so far, but I also think the whole "Defining subclasses of generic classes" needs to be rewritten, as with Python 3.12 syntax, `Generic` isn't used anymore.

BHUVANSH855:
> Thanks so far, but I also think the whole "Defining subclasses of generic classes" needs to be rewritten, as with Python 3.12 syntax, `Generic` isn't used anymore.

@srittau I made a few final refinements:

* Updated wording of the note for clarity
* Fully aligned subclassing examples with PEP 695 syntax
* Removed outdated discussion of `Generic`

Everything should now consistently reflect the Python 3.12+ style.

Kindly review it again.

BHUVANSH855:
Hi @srittau @JelleZijlstra,

Just checking in regarding the latest updates to this PR.
Kindly let me know if any further changes are required.

Thank you!

docs/reference/generics.rst
```diff
@@ -96,12 +106,12 @@ can be used as a base class for another class (generic or non-generic). For exam
...
-   class Receiver(Generic[T]):
-       def accept(self, value: T) -> None: ...
+    # This is a user-defined generic class
+    class Receiver[T]:
+        def accept(self, value: T) -> None: ...
 
-   # This is a generic subclass of Receiver
-   class AdvancedReceiver(Receiver[T]): ...
+    # This is a generic subclass of Receiver
+    class AdvancedReceiver(Receiver[T]): ...
```
JelleZijlstra: This one should also have `[T]`, because it's independently generic. Please check for other similar issues throughout the file.

BHUVANSH855: hi @JelleZijlstra,

Thanks for the feedback! I’ve updated `AdvancedReceiver` to include its own type parameter and checked for similar cases throughout the file to ensure consistency.

BHUVANSH855:
Hi @srittau @JelleZijlstra,

Just checking in regarding the latest updates to this PR.
Kindly let me know if any further changes are required.

Thank you!

BHUVANSH855:
Hi @srittau @JelleZijlstra,

Just checking in regarding the latest updates to this PR.
Kindly let me know if any further changes are required.

Thank you!

docs/reference/generics.rst
JelleZijlstra:
```suggestion
   def first[T](seq: Sequence[T]) -> T:
```

JelleZijlstra:
```suggestion
   def first[T](seq: Sequence[T]) -> T:
```

JelleZijlstra:
```suggestion
   def last[T](seq: Sequence[T]) -> T:
```
(Also remove the `T = TypeVar` above.)

JelleZijlstra:
```suggestion
       def set_scale[T: Shape](self: T, scale: float) -> T:
```

JelleZijlstra (changes requested):
There are lots of remaining places in the document using the old syntax. Look for all places that still use `T = TypeVar("T")`-style syntax and replace them.

BHUVANSH855:
Thanks for the review! @JelleZijlstra 

I’ve now updated the generics reference to consistently use the PEP 695 type parameter syntax in all applicable examples, including generic functions, methods, class methods, and bounded type parameters.

I kept `TypeVar` usage only in sections where it is still required (such as variance, constraints, decorators, and ParamSpec), so the document stays accurate for those advanced cases.

Please let me know if anything else should be refined.

BHUVANSH855:
Hi @JelleZijlstra,

Just checking in regarding the latest updates to this PR.
Kindly let me know if any further changes are required.

Thank you!

JelleZijlstra:
The document still has many old-style ParamSpec and TypeVar declarations further down. It also says "By default, all user-defined generics are invariant." which is no longer correct.

BHUVANSH855:
> The document still has many old-style ParamSpec and TypeVar declarations further down. It also says "By default, all user-defined generics are invariant." which is no longer correct.

Thanks for the review @JelleZijlstra,

I’ve updated the invariance statement and added the missing imports in the variance examples. I also ensured formatting consistency in the code blocks.

Please let me know if anything else needs adjustment.

BHUVANSH855:
> The document still has many old-style ParamSpec and TypeVar declarations further down. It also says "By default, all user-defined generics are invariant." which is no longer correct.

Thanks for the review @JelleZijlstra,

I’ve updated the invariance statement and added the missing imports in the variance examples. I also ensured formatting consistency in the code blocks.

Please let me know if anything else needs adjustment.

JelleZijlstra:
Please stop pinging on this PR. I'm a bit tired of it because of the number of times I've had to ask you to complete the change. I'll look at it when I find the time and energy; in the meantime please do not send any more pings.

PR #2247, opened 2026-04-07
Clarify callable assignability to require consistent parameter mapping

### LINKED ISSUE #2224: Spec/Conformance: subtyping callables with non-constant parameter mapping.
Consider the example below, which errors with all tested type checkers (mypy/pyright/ty/pyrefly):

```python
from typing import Protocol

class Interval: ...

class Make(Protocol):
    def __call__(self, /, lower: float, upper: float) -> Interval: ...

def make_impl(
    string_or_lower: str | float | None = None,
    /,
    lower: float | None = None,
    upper: float | None = None,
) -> Interval: ...

def test() -> None:
    _f: Make = make_impl  # ❌️ type checkers error here.
```

I believe from a pure type theory POV, this assignment should be legal, because all legal arguments to `Make` are also legal arguments to `make_impl`. The spec phrases it in the same spirit:

> A callable type B is [assignable](https://typing.python.org/en/latest/spec/glossary.html#term-assignable) to a callable type A if the return type of B is assignable to the return type of A and the input signature of B accepts all possible combinations of arguments that the input signature of A accepts. All of the specific assignability rules described below derive from this general rule.

And I couldn't find anything else in https://typing.python.org/en/latest/spec/callables.html#assignability-rules-for-callables that would disallow this assignment.

It seems the type-checkers try to match the `KEYWORD_OR_POSITIONAL` parameters by name, which is incorrect. `Make` has 3 legal call signatures:

1. `Make(float, float)`
2. `Make(float, upper=float)`
3. `Make(lower=float, upper=float)`

and all these 3 call signatures are supported by `make_impl`, but the parameter mapping is not constant:

1. `make_impl(float, float)`    ->  `{lower:string_or_lower, upper:lower}`
2. `Make(float, upper=float)` -> `{lower:string_or_lower, upper:upper}`
3. `Make(lower=float, upper=float)` -> `{lower:lower, upper:upper}`

So either the spec should demand a constant parameter mapping, or this example should be added to the conformance tests.

BHUVANSH855:
Fixes #2224

The current typing spec states that a callable type B is assignable to a callable type A if B accepts all possible combinations of arguments that A accepts. However, it does not explicitly define whether the mapping between parameters must remain consistent across different call patterns.

In practice, all major type checkers (mypy, pyright, etc.) reject cases where parameter mapping is non-constant, even if all valid calls are technically accepted.

This PR clarifies the spec by stating that callable assignability requires a consistent parameter mapping. If arguments map to different parameters depending on how they are passed (positional vs keyword), the callable is not considered assignable.

This aligns the spec with existing type checker behavior and resolves the ambiguity highlighted in the issue.


carljm (changes requested):
I don't think this is the direction we should go in the spec. It seems to me that the requirement for a constant mapping is a quality-of-implementation issue. What the spec already describes is the intended and desired behavior; implementations just aren't fully implementing it. The spec should not prohibit a type checker from attempting to do better.

BHUVANSH855:
Thanks for the feedback @carljm,

That makes sense — I understand that the spec describes the intended behavior, even if current type checkers don't fully support it.

I've updated the PR to remove the spec change and instead added a conformance test capturing this scenario, so the gap between the spec and current implementations is made explicit.

Kindly review it again 
@JelleZijlstra @srittau

JelleZijlstra:
This doesn't properly update the conformance results tracking. I'd rather not make any changes to the repo here right now.

PR #2248, opened 2026-04-07
Clarify positional argument handling in TypedDict constructor

BHUVANSH855:
This PR updates the TypedDict constructor specification to clarify handling of positional arguments.

The current spec states that only keyword arguments are allowed, but this does not fully reflect runtime behavior and differences across type checkers.

This change:
- Allows type checkers to accept a single positional argument of TypedDict type
- Explicitly discourages other positional forms (e.g., dicts, sequences)
- Adds conformance tests to define expected behavior

This aims to improve consistency while avoiding unsafe or ambiguous cases.


BHUVANSH855:
Hi @srittau @JelleZijlstra,

Just checking in regarding the latest updates to this PR.
Kindly let me know if any further changes are required.

Thank you!

carljm:
I am not convinced that we should discourage type checkers from accepting additional forms.

BHUVANSH855:
> I am not convinced that we should discourage type checkers from accepting additional forms.

Thanks for the feedback @carljm,

I’ve updated the wording to remove the restriction on additional positional forms and instead clarify that only the specified behavior is required, while type checkers may choose to support additional forms.

Please let me know if this looks good now.
Kindly review it again.

docs/spec/typeddict.rst
```diff
@@ -266,7 +266,15 @@ The TypedDict constructor
...
 TypedDict types are callable at runtime and can be used as a constructor
 to create values that conform to the TypedDict type. The constructor
-takes only keyword arguments, corresponding to the items of the TypedDict.
+primarily takes keyword arguments, corresponding to the items of the TypedDict.
+
+Additionally, type checkers may allow a call with a single positional argument
+whose type is a TypedDict. In this case, the resulting value is initialized from
+that argument.
+
+Type checkers may choose to accept additional forms of positional arguments,
```
carljm, line 275:
```suggestion
Type checkers may choose to accept additional forms of positional arguments, and/or mixed positional/keyword calls,
```

carljm (commented):
Thanks. I'm generally fine with this leave-it-up-to-type-checkers approach; it's good to add something to the spec text to make it more explicit, even if we don't strictly require a particular behavior. Would like to know what @JelleZijlstra and others think.

conformance/tests/typeddict_constructor_positional.py
```diff
@@ -0,0 +1,18 @@
...
+
+# keyword arguments (OK)
+TD1(a=1)
+
+# positional TypedDict (OK)
+td2: TD2 = {"a": 1, "b": "x"}
+TD1(td2)
+
+# invalid positional arguments
+TD1({"a": 1})  # E
```
JelleZijlstra: Why do we require an error? These both seem like valid calls.

JelleZijlstra (commented):
I'm not sure this change is an improvement. If we're going to expand the spec here, I'd rather we clearly list out some patterns that we want to require. It's going to require some work though to sort out what is reasonable to require.

docs/spec/typeddict.rst
BHUVANSH855: Thanks for the feedback! @carljm 

I initially added wording to clarify that type checkers may accept additional positional forms, but based on the discussion, I’ve simplified the spec to avoid implying specific behavior beyond what is clearly supported.

For now, the spec only describes the keyword-based constructor and the case of a single TypedDict positional argument, without constraining or explicitly encouraging other forms.

Happy to refine this further if we reach consensus on specific patterns that should be required.

conformance/tests/typeddict_constructor_positional.py
BHUVANSH855: Good point — I’ve updated the conformance tests to remove the requirement for errors in those cases, since such calls can be valid and behavior may vary across type checkers.

Regarding the spec change, I’ve also simplified the wording to avoid over-specifying behavior. I agree that if we expand this area further, it would be better to clearly define specific patterns to require.

Happy to help explore and document those patterns if needed.
@JelleZijlstra

BHUVANSH855:
Hi @carljm @JelleZijlstra,

Just checking in regarding the latest updates to this PR.
Kindly let me know if any further changes are required.

Thank you!

BHUVANSH855:
Hi @carljm @JelleZijlstra,

Just checking in regarding the latest updates to this PR.
Kindly let me know if any further changes are required.

Thank you!

JelleZijlstra:
My previous comment stands.

carljm:
@BHUVANSH855 This is not an "easy bug fix" to the spec where the ideal behavior is clear and we just need to settle on the right wording and land the PR. The useful next steps toward landing this would not involve updates to this PR at all; they would involve building consensus among type checker authors and the typing community about what behavior makes sense to standardize for TypedDict constructors. This could involve posting on discuss.python.org with a proposal to invite feedback (this is a requirement for non-trivial typing spec changes anyway), it could involve research into what patterns are actually found in the ecosystem at what frequency, it could involve surveying the actual behavior of current type checkers.

Closing this PR for now to more clearly indicate that it's not ready to land, and it won't become ready to land with simple tweaks. We can revisit it if/when we have a clear consensus plan.

PR #2250, opened 2026-04-08
Improve Step 5 of overload call evaluation.

rchen152:
Makes two modifications to step 5 of overload call evaluation:
1. When materializing arguments to check whether we can eliminate overloads, skip arguments that have the same parameter type in all overloads.
2. When we still have multiple overloads after the materialization filter, try to find a return type that all materializations of all other return types are assignable to.

Updates the conformance tests as well. All type checkers already do (1). Mypy and pyrefly pass the conformance test for (2). I believe mypy has a heuristic that approximates the new rule; pyrefly implements it exactly.


carljm:
I think it's preferable to keep PRs updating type-checker versions separate from PRs implementing conformance suite and spec changes.

docs/spec/overload.rst
```diff
@@ -397,6 +419,24 @@ Example 4::
...
+      def f(self) -> T:
+          return self.x
+
+  @overload
+  def example5(x: A[None]) -> A[None]: ...
+  @overload
+  def example5(x: A[Any]) -> A[Any]: ...
+
+  def test(x: Any):
+      # Steps 5 eliminates the first overload because there exists a
```
carljm:
```suggestion
      # Step 5 eliminates the first overload because there exists a
```

docs/spec/overload.rst
```diff
@@ -268,39 +268,61 @@ If so, eliminate overloads that do not have a variadic parameter.
...
-the corresponding parameter type for each of the remaining overloads. If so,
-eliminate all of the subsequent remaining overloads.
+For each of the remaining overloads, determine whether all arguments satisfy at
+least one of the following conditions:
 
-Consider the following example::
+- All possible :term:`materializations <materialize>` of the argument's type are
+  assignable to the corresponding parameter type, or
+- The parameter types corresponding to this argument in all of the remaining overloads
+  are :term:`equivalent`.
```
JelleZijlstra: I think this is correct but there's some subtlety in the wording that we should cover in the test cases:

- "The parameter types corresponding to this argument". That may be a very different parameter in each of the overloads; e.g. maybe one takes `*args` and the other has explicit arguments. So my reading is that to check this, you have to create for each overload a mapping between argument and parameter type, and consult that. What is the parameter type corresponding to an unpacked argument, though?
- "the remaining overloads". So we only check this overload and all following overloads, not ones above it?

conformance/tests/overloads_evaluation.py
```diff
@@ -345,3 +348,51 @@ def check_example7(v1: list[Any], v2: Any) -> None:
 
     ret3 = example7(v1, v2)
     assert_type(ret3, Any)
+
+
+@overload
+def example8(x: str, y: Literal['o1']) -> str: ...  # E? some type checkers report unsafe overlap between example8's overloads
```
carljm: This unnecessary complexity would be eliminated (without reducing the strength of the test) by changing `str` here to `bool` (and changing the rest of the test accordingly).

conformance/tests/overloads_evaluation.py
```diff
@@ -345,3 +348,51 @@ def check_example7(v1: list[Any], v2: Any) -> None:
...
+@overload
+def example9(x: A[Any]) -> A[Any]: ...
+
+
+def example9(x: A[Any]) -> A[Any]:
+    return x
+
+
+def check_example9(x: Any):
+    # Steps 5 eliminates the first overload because there exists a
```
carljm:
```suggestion
    # Step 5 eliminates the first overload because there exists a
```

rchen152:
> I think it's preferable to keep PRs updating type-checker versions separate from PRs implementing conformance suite and spec changes.

Point taken, sorry! I was being a bit sloppy because I didn't want to open two PRs XD

Version update here: https://github.com/python/typing/pull/2254

conformance/tests/overloads_evaluation.py
rchen152: Good idea, thanks!

docs/spec/overload.rst
rchen152: Great points, thanks!
* I added a clarification about unpacked arguments and some more tests.
* Ah, I meant all of the candidate overloads that are remaining at the beginning of Step 5, not just the overloads following the one that we're checking. I replaced "remaining overloads" with "candidate overloads" - hopefully that's clearer.

rchen152:
Thanks for the feedback! I believe I've addressed all the comments.

davidhalter:
Thanks Rebecca for looking into this. I think we really need changes here. Please give me a few days to respond before you call for a vote. I have a few more general thoughts.

docs/spec/overload.rst
```diff
@@ -268,39 +268,64 @@ If so, eliminate overloads that do not have a variadic parameter.
...
+      example2(x, 'o1')
 
 This rule eliminates overloads that will never be chosen even if the
 caller eliminates types that include ``Any``.
 
-If the call involves more than one argument, all possible materializations of
-every argument type must be assignable to its corresponding parameter type.
-If this condition exists, all subsequent remaining overloads should be eliminated.
+If the call involves more than one argument, every argument must satisfy one of
+the above conditions.
```
davidhalter: You wrote above:

> determine whether all arguments satisfy at least one of the following conditions

I feel like this part is redundant and it confuses me more than it helps.

docs/spec/overload.rst
```diff
@@ -268,39 +268,64 @@ If so, eliminate overloads that do not have a variadic parameter.
...
 
 Once this filtering process is applied for all arguments, examine the return
 types of the remaining overloads. If these return types include type variables,
-they should be replaced with their solved types. If the resulting return types
-for all remaining overloads are :term:`equivalent`, proceed to step 6.
+they should be replaced with their solved types. Eliminate every overload for
+which there exists a :term:`materialization <materialize>` of another
+overload's return type that is not assignable to this overload's return type.
```
davidhalter: This is clever. I like this much better than Mypy's type erasure (where `list[int]` and `list[str]` become `list[Any]`), because this generalizes better. I think the sentence is a bit hard to understand - at least for me - but I doubt I would be able to come up with something more understandable.

conformance/tests/overloads_evaluation.py
```diff
@@ -345,3 +348,119 @@ def check_example7(v1: list[Any], v2: Any) -> None:
...
+
+def example11(x: str, y: int, z: str) -> bool | int:
+    return True
+
+
+def check_example11(x: Any):
+    # `*x` maps to `(y: int, z: str)` in both overloads, so the second overload
+    # can be eliminated.
+    ret = example11('o1', *x)
+    assert_type(ret, bool)
```
davidhalter: I would personally like to add

```
    ret2 = example11('o1', x, x)
    assert_type(ret2, bool)
```

It's both interesting and useful information for type checker devs. E.g. Mypy passes that while they fail the other one.

davidhalter (commented):
Thanks a lot for this PR. I feel like this is taking overloads in a much better direction. Projects like scipy-stubs will benefit a lot from this.

I think we can (and probably should) still add `AnyOf` in the future, but the improved return types already fix most of the issues I have had with the current spec.

Sorry for taking so much time for such a small review, but it took me many hours to understand Mypy's code around overloads good enough to make sure we don't omit something.

docs/spec/overload.rst
rchen152: Agreed that this phrasing is clunky, but I spent a while wordsmithing it, and this was the best I could come up with :/

carljm: I think this can be phrased more simply in terms of "top materialization" (the union of all possible materializations of a gradual type), but that's not a term that has been formally defined in the spec yet, so that would just introduce a different problem :)

rchen152:
> Sorry for taking so much time for such a small review, but it took me many hours to understand Mypy's code around overloads good enough to make sure we don't omit something.

No worries! Part of the process is waiting at least a week for feedback anyways. Thanks for checking mypy's code.

docs/spec/overload.rst
JelleZijlstra: I think an equivalent formulation is: define a type AllReturn that is the top materialization of the union of the return types of all remaining overloads. Remove all overloads for which the top materialization of the return type is not equivalent to AllReturn.

(Not suggesting we use this wording, but for me at least it helps me understand what's being suggested.)

docs/spec/overload.rst
```diff
@@ -268,39 +268,61 @@ If so, eliminate overloads that do not have a variadic parameter.
...
 
 This rule eliminates overloads that will never be chosen even if the
 caller eliminates types that include ``Any``.
 
-If the call involves more than one argument, all possible materializations of
-every argument type must be assignable to its corresponding parameter type.
-If this condition exists, all subsequent remaining overloads should be eliminated.
-
 Once this filtering process is applied for all arguments, examine the return
 types of the remaining overloads. If these return types include type variables,
```
JelleZijlstra: This is not new to this change but I think replacing all type variables is wrong, since there may be free type variables from an outer scope (if an overload contains references to a class-scoped type variable, and it is being called from another method of that class).

JelleZijlstra:
I think this makes sense to me but because it is so subtle, it's probably good if someone tries to implement it in at least one other type checker. I'll try to do that for pycroscope, but it may be a little while (the current overload implementation in pycroscope doesn't follow the spec's step structure though the results match the current spec).

carljm:
I'm also working on implementation in ty.

carljm:
Unfortunately, working on implementation in ty pretty quickly led me to doubt whether the currently proposed rule is adequate, and the conformance tests here specify desirable behavior. See https://discuss.python.org/t/spec-change-ambiguous-arguments-in-overload-call-evaluation/106891/5 for details.

conformance/tests/overloads_evaluation.py
```diff
@@ -345,3 +348,121 @@ def check_example7(v1: list[Any], v2: Any) -> None:
...
+def example9(x: str | bytes, y: str) -> bool | int:
+    return True
+
+
+def check_example9(x: Any):
+    # All three overloads are candidates. The parameter types corresponding to
+    # argument `x` are `str` and `bytes`, which are not equivalent, so none of
+    # the overloads can be eliminated. We pick the most general return type.
+    ret1 = example9(x, 'o1')
+    assert_type(ret1, int)
```
carljm, line 395: I implemented in ty the additional rule that a candidate overload return type is only considered "most general" if it is also assignable to all matched overloads. The only expectation in these conformance tests that changes as a result is this one; it reverts back to `Unknown/Any`.

If you are happy with the `int` result in this case for pyrefly, I can discuss with the ty team, but I think we might prefer if the conformance suite gave leeway for a type checker to give a more gradual result here. (We might prefer a type that relies on intersections.)

PR #2262, opened 2026-04-16
Add PEP 800 (disjoint bases)

JelleZijlstra:
Accepted here: https://discuss.python.org/t/pep-800-solid-bases-in-the-type-system/99910/41

### DIFF
--- a/conformance/results/ty/directives_disjoint_base.toml
+++ b/conformance/results/ty/directives_disjoint_base.toml
@@ -1,0 +1,17 @@
+conformance_automated = "Fail"
+conformant = "Partial"
+notes = """
+Does not reject @disjoint_base on TypedDict or Protocol definitions.
+"""
+errors_diff = """
+Line 118: Expected 1 errors
+Line 123: Expected 1 errors
+"""
+output = """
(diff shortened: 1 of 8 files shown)


carljm (approved):

conformance/results/ty/directives_disjoint_base.toml
```diff
@@ -0,0 +1,17 @@
...
+conformant = "Partial"
+notes = """
+Does not reject @disjoint_base on TypedDict or Protocol definitions.
+"""
+errors_diff = """
+Line 118: Expected 1 errors
+Line 123: Expected 1 errors
+"""
+output = """
+directives_disjoint_base.py:69:7: error[instance-layout-conflict] Class will raise `TypeError` at runtime due to incompatible bases: Bases `Left` and `Right` cannot be combined in multiple inheritance
```
JelleZijlstra, line 11: hm, this error is actually inaccurate; there's no runtime error if the class is just marked with `@disjoint_base` at runtime.

carljm: I think the error is correct for the intended use of disjoint_base, though. I'm not sure it would be preferable to adjust the error message to account for misuse on classes that aren't actually disjoint bases.

davidhalter (approved):

PR #2264, opened 2026-04-18
conformance: remove incorrect error from `Literal` test

KotlinIsland:
```python
from typing import Literal, Union


a: dict[(int, str)]  # valid
b: Union[(int, str)]  # valid
c: Literal[(1, 2)]  # invalid???
```


carljm:
This is specifically mandated in the spec, with rationale, so it shouldn't be changed in the conformance suite unless the spec text is updated.

The rationale for forbidding this is that it's unclear if the user is trying to express `Literal[1, 2]` or `tuple[Literal[1], Literal[2]]`. I think this rationale makes sense, and there's nothing significant to be gained by supporting this syntax.

PR #2277, opened 2026-05-01
Add sentinels (PEP 661) to the spec

JelleZijlstra:
PEP-661 was accepted and implemented for Python 3.15. The main typing innovation is that sentinel objects are usable directly in type annotations.

In this proposed text I chose to drop the requirement that the name of the sentinel variable must match the name of the sentinel itself. When I converted sentinels to the new class in CPython itself (https://github.com/python/cpython/pull/149084/changes#diff-ca80d44aeeadd5e4966195b99aff106abb3e7b0b5356b7499125ee021bd84777), I encountered some that had a pre-existing repr with special characters in it, and it seemed best to leave it alone (`UNNAMED_SECTION = sentinel("<UNNAMED_SECTION>")`). And in the PEP discussion, there were questions about sentinels defined in classes: to make those picklable, you'd have to prefix the sentinel name with the class name (https://discuss.python.org/t/pep-661-sentinel-values/9126/345). I feel there isn't a strong typing reason for *requiring* that the two names must match. It may be a good code quality rule to require that the two names match, but type checkers should be able to support sentinels in annotations fine even if they don't match.

No type checker supports the PEP fully yet, though pyright apparently has experimental support. I validated the conformance tests in this PR by using pycroscope, which does support it but currently enforces the same-name rule.


docs/spec/special-types.rst
```diff
@@ -53,6 +53,46 @@ are highly dynamic.
...
+      return x
+
+  f(OTHER)  # Error, OTHER is not an int or MISSING
+  f(MISSING)  # OK, MISSING is a valid argument
+
+Sentinels may be created using the ``sentinel()`` built-in in Python 3.15
+and higher. ``typing_extensions`` provides a backport of this function. For
+historical reasons the object was first introduced under the name
+``typing_extensions.Sentinel``, and later ``typing_extensions.sentinel`` was
+added as an alias; type checkers should support both.
```
cdce8p, line 78: Not sure about `typing_extensions.Sentinel`. If most type checkers already supported it fine, but as it's not, it might be easier in the long term to just drop the requirement here.

carljm: It looks like at the moment `typing_extensions.sentinel` doesn't exist yet in typeshed. As long as `typing_extensions.sentinel` and `typing_extensions.Sentinel` are the same kind of object in typeshed (e.g. both classes), or one is literally an alias of the other, I expect it asks very little of type-checkers to support both.

That said, if there really aren't any uses of `typing_extensions.Sentinel` yet in the wild, I agree it seems like it would reduce the potential for confusion if we didn't support it.

conformance/tests/specialtypes_sentinels.py
```diff
@@ -0,0 +1,40 @@
+from unittest.mock import sentinel
```
davidhalter: This import does not seem to be used.

conformance/tests/specialtypes_sentinels.py
```diff
@@ -0,0 +1,40 @@
...
+
+from typing_extensions import Sentinel, assert_type
```
davidhalter, line 1: The spec states that:

> For historical reasons the object was first introduced under the name ``typing_extensions.Sentinel``, and later ``typing_extensions.sentinel``

Should we test both or is this good enough? I'm not sure how strict conformance tests usually mirror the spec in these cases.

davidhalter (commented):
Thanks for this. I like that sentinels are added to the language.

Would it be worth adding a test with multiple sentinels (i.e. `int | MISSING1 | MISSING2`) to the conformance tests? I'm not sure this is necessary, just putting it out there.

conformance/tests/specialtypes_sentinels.py
JelleZijlstra: I didn't because there is no released version of typing-extensions yet that provides the `sentinel` name, so type checkers would be in their right to reject that name at the moment.

docs/spec/special-types.rst
JelleZijlstra: It's already being used by Pydantic: https://github.com/pydantic/pydantic/blob/597fa692eaaa90132f683e88ca994f28e185463b/pydantic-core/python/pydantic_core/__init__.py#L148

cdce8p: > It's already being used by Pydantic: https://github.com/pydantic/pydantic/blob/597fa692eaaa90132f683e88ca994f28e185463b/pydantic-core/python/pydantic_core/__init__.py#L148

Is that really an issue though if almost no type checker supports it? To be clear I'm not talking about removing it from `typing_extensions`. That would break already released pydantic versions. I'd just suggest only enforcing that type checkers support the lower case variants. For pydantic that would essentially mean to switch to `sentinel`, which I'd guess they'll do anyway as soon as it's available, and update their type checker once those support it. Until then they will just continue to see the error messages they already see now.

JelleZijlstra: I'd prefer to support both since it doesn't seem like it would add a significant burden to type checkers.

cdce8p: > I'd prefer to support both since it doesn't seem like it would add a significant burden to type checkers.

Sure, once support for PEP 661 is implemented, supporting `typing_extensions.Sentinel` will be trivial. That reduces the incentive for projects to migrate to `sentinel` though, thus forcing `typing-extensions` to keep it around for longer.

Mypy still has code for `mypy_extensions.TypedDict` which I'd love to clean up, though I suspect it might still be a long while before that will happen. Better to not enforce supporting something outdated to begin with. _If type checkers decide they still want to support it, so be it, but I'd recommend against dictating it in the spec._

davidhalter (approved):
I quickly implemented this in Zuban. I haven't found any issues and latest Zuban implements sentinels. (No need to update here, we can do that later after its merged).

docs/spec/special-types.rst
Viicos: On the Pydantic side, we [explicitly document](https://pydantic.dev/docs/validation/latest/concepts/experimental/#missing-sentinel) the feature using sentinels as experimental, meaning (at least for us) users should be prepared for things to break, so our end we are fine _not_ supporting `Sentinel`.

carljm (approved):
This looks fine to me. Typechecker support should be updated; at least ty and zuban support this now (haven't checked others).

conformance/tests/specialtypes_sentinels.py
```diff
@@ -0,0 +1,40 @@
...
+
+# > Sentinel objects may be used in type annotations if they are defined using
+# > a simple assignment of the form ``NAME = sentinel('NAME')`` in the
+# > global scope or in a class body that is not within a function.
+
+MISSING = Sentinel("<MISSING>")  # name is not required to match the variable name
```
carljm, line 7: If it's still generally considered good style where possible to have the names match (and if it isn't, why is this case different from e.g. `TypeVar` etc?), type checkers may want to provide this as a style warning/lint. Should we allow that here with an `# E?`

See discussion in https://github.com/astral-sh/ty/issues/4091

JelleZijlstra: I'd prefer to do this by turning off the error code for ty in our checks. That way, we enforce that the type checker has a mode that allows this; a type checker that always errors on mismatched names is not compliant.

In general I'd rather avoid "E?" as it makes the spec weaker in its checks.

carljm: Hmm, we were planning to make this the same error code that enforces name matching on `TypeVar` and all other similar forms. I think if we disabled that error code for the conformance suite, it might cause us to fail some other checks? So I guess that gets back to my question above: is there a reason the conformance suite expectation for sentinels should be the reverse of what it is for those other forms with a similar name pattern?

JelleZijlstra: I'd be willing to relax the check for TypeVar etc. too but that's a little out of scope for this PR.

I do think the case for having a mismatched name is stronger for sentinels, which have more runtime uses than other objects where we might want a name match.

carljm: Yeah I see the rationale for it being different from TypeVar etc. We can just make it a different error code in ty. Feels somewhat odd to have two different error codes for what looks like an identical pattern, but if it's expected that some users will want to enforce same-name for TypeVar et al but not for sentinel, then it's worth it.

PR #2278, opened 2026-05-02
Add pycroscope to the conformance suite

JelleZijlstra:


conformance/pyproject.toml
```diff
@@ -4,6 +4,7 @@ version = "0.1.0"
 requires-python = "==3.12.*"
 dependencies = [
     "mypy",
+    "pycroscope==0.4.0",
```
AlexWaygood: Pinning it here as well as in uv.lock probably means that `uv lock --upgrade-package=pycroscope` won't work.

Speaking of, you probably want to add pycroscope here too (though I generally just run `uv lock --upgrade-package=ty` manually when updating ty's results): https://github.com/python/typing/blob/6483940ffaf4ab6b515513fbda8b81661372e2f3/conformance/scripts/bump_type_checkers.py#L10

AlexWaygood (approved):

conformance/uv.lock
```diff
@@ -170,6 +206,7 @@ dependencies = [
 [package.metadata]
 requires-dist = [
     { name = "mypy" },
+    { name = "pycroscope", specifier = "==0.4.0" },
```
AlexWaygood: Maybe it doesn't matter, but this looks stale now that you've removed the specifier from pyproject.toml

JelleZijlstra: I didn't want to figure out the uv invocation to update it, and bump_type_checkers.py updated some other stuff, so I made #2279 first to bump the others.

JelleZijlstra: I suppose I could have edited the file manually

AlexWaygood (approved):

PR #2279, opened 2026-05-02
Bump type checkers

JelleZijlstra:
Pyrefly, ty, and zuban all pass one more test now.


conformance/results/ty/generics_defaults.toml
AlexWaygood, line 1: Not totally sure (I'm on my phone), but we might be able to remove the line in the `notes` field here about not fully supporting defaults for `ParamSpec`?

AlexWaygood (approved):

conformance/results/ty/generics_defaults.toml
JelleZijlstra: The error ```generics_defaults.py:200:17: error[invalid-type-form] The first argument to `Callable` must be either a list of types, ParamSpec, Concatenate, or `...` ``` is still something with ParamSpecs. That line is just `y: Callable[P, None]` where P is a ParamSpec with a default.

AlexWaygood: IIRC I _think_ the root cause of that one is that the class is generic over a TypeVarTuple as well as a ParamSpec, and we still don't support TypeVarTuple at all, so that confuses us. (I don't feel strongly about this being updated here now though, it doesn't matter that much.)

JelleZijlstra: Ah true, I didn't notice that.

AlexWaygood (approved):
You gotta regen the html file now, but LGTM other than that!

PR #2282, opened 2026-05-10
Update ty results for v0.0.35

charliermarsh:


conformance/results/ty/overloads_consistency.toml
AlexWaygood, line 1: We pass all tests here, so I think "Pass" is accurate. However, I might add a `notes` field that points out that we currently only do the check for non-generic functions. (You'll need to regen the HTML file afterwards -- our conformance will be displayed as "Pass*" in the rendered HTML, and the `notes` message will display as a tooltip when folks hover over the cell.)

AlexWaygood (approved):

conformance/results/ty/overloads_consistency.toml
JelleZijlstra: I suspect ty isn't alone in that, type checkers are OK with https://github.com/JelleZijlstra/unsoundness/blob/main/examples/overload/generic_overlap.py#L6-L11 after all (we were just talking about that case on pydis).

JelleZijlstra (approved):

conformance/results/ty/overloads_consistency.toml
AlexWaygood: Right... ty is the only type checker not to complain about this, for example, though:

```py
from typing import overload

@overload
def mixed_generic[T](x: T) -> T: ...
@overload
def mixed_generic(x: int) -> int: ...
def mixed_generic(x: str) -> object:
    return x
```

Multiplay gist: 10c760796b5da6d1614ec8341c0fc3bc

PR #2287, opened 2026-05-16
added variance inference test with method scoped type variable

### LINKED ISSUE #2281: [Conformance/Spec] Type Checker Divergence in Variance Inference
**Update:** Since v0.0.36 `ty` now also considers `Foo` covariant in `T` in the example below

Currently, all type checkers pass the [`generics_variance_inference` portion of the conformance test suite](https://htmlpreview.github.io/?https://github.com/python/typing/blob/main/conformance/results/results.html). However, there is some divergence with generic methods that have bound self-type: 
[[mypy]](https://mypy-play.net/?gist=2b75369999f5d7278f1ff3ab56edd03c), [[pyright]](https://pyright-play.net/?code=MYGwhgzhAEBiD28DaAVAugLgFDV9AJgKYBm0A5oQC4AUEhIxAlNALQB80KG0AdHzniKkw%2BfEgDKaWvWLcEySQBpo8SgAtCAJ24gAlhEoS0ygPTN2cREe58eWLEOiVCBgPrB4ANzCbdYAHbAhNQ%2BZHJWuv6UaOYc8kjwAEYAVoTA0dh40JpUAK6a-tChuADE0AC2AJ4ADpXKAF65iQHKlJXcgDLkgPB-yrW%2BZGqUvZU5xCDt0ICg5F1AA), [[ty]](https://play.ty.dev/f6f5b8e9-9c4e-4ef4-a559-53662c76f5f1), [[pyrefly]](https://pyrefly.org/sandbox/?project=N4IgZglgNgpgziAXKOBDAdgEwEYHsAeAdAA4CeSIAxlKnHAAQBiuuA2gCoC6iAOuvQPqYYYegHMYAFwAUcGFDABKegFoAfPXaJ6hXX0FCR9VJkysAyp1nyw2niGZtL9gDT1ckgBYwATtscWnG4A9Mrq9PYBzkg6euh8wqKS8JIA%2BpS4AG6oPhAYlDDSOWL%2BLKwQ6JKcYRoBuNgAVjCUVbz8gj5SAK4%2B-MUCAMT0ALakZG4AXl3YGG6SpNqAMuSA8H9uZLlinpJrpJ1gUAv0gKDky3wgLiDrIgeEkrjDUBRDAAq716T0aFh4%2BPQZ6JAxD1UJIILh0IQ%2BENzDAYPQtpJiHBEMFgld9qRCLgfGJgjB0MFMLhKHBgv9AcDQeDgvQwNjjNloKhsLA-uCKT4QWD%2BLhiFT0HBIegyF5wSpMr44Nz6ABeCIgADMhAAjAAWex8Vi%2BHzYuCcPhddAQYbEbHJTAqTAQTotCAS2X0ADkEDE6GxMEdfDdkhUnQAjl1rTALQBrGCkFSoSgFOgOx0Adxy6E96BAAF8LlHQRLGNAYBQvjgCCRyGmgA)

```python
class Foo[T]:
    def get(self) -> T: ...
    def add[S](self: Foo[S], other: list[S], /) -> Foo[S]: ...

def test_covariance(arg: Foo[int]) -> Foo[object]:
    return arg  # mypy, zuban, ty: ❌️, pyright, pyrefly: ✅️
```

Following the [variance inference specification](https://typing.python.org/en/latest/spec/generics.html#variance-inference), I believe `Foo` should be covariant in `T`, which is what `pyright` and `pyrefly` say, because `Foo[Dummy].add` is assignable to `Foo[object].add`. Intuitively, since `S` is method and not class scoped, and `T` does not appear in the signature of `Foo.add`, it has no bearing on variance.

However, `mypy`, `ty` and `zuban` think `Foo` is invariant in `T`.  In https://github.com/python/mypy/pull/19466#discussion_r2432203931,  @ilevkivskyi argued that for an instance, `def add[S](self: Foo[S], other: list[S], /) -> Foo[S]: ...` is indistinguishable from `def add(self, other: list[T], /) -> Foo[T]: ...` and therefore `Foo` should be invariant in `T`.

This argumentation reveals another divergence between type checkers when it comes to subclassing generics: [[mypy]](https://mypy-play.net/?gist=0869df5e6345f426cba8dfe419ddbc5f), [[ty]](https://play.ty.dev/0e291fd1-d71f-4cd2-ae2f-5c1f8b58278a), [[pyrefly]](https://pyrefly.org/sandbox/?project=N4IgZglgNgpgziAXKOBDAdgEwEYHsAeAdAA4CeSIAxlKnHAAQBiuuA2gCoC6iAOuvQPqYYYegHMYAFwAUcGFDABKegFoAfPXaJ6hXX0FCR9VJkysAyp1nyw2niGZtL9gDT1ckgBYwATtqgQcJIWnG4A9Mrq9PaOIfbauoR8fNS0DOYArtjSsRDokpyKvPyCAMT0ALakZG4AXlkYbpKkbmQ%2BIlCk2oCg5IDwf62kPhBinpLagDLkvfqCwqImmNYKbh7efvQBQax5BZEa9pnY8Tq6IC4gbR2khJK4FVAU5QAKg5f0aFh4%2BPSUuOiQYhkfKhJBBfkl0OVzDAYPRRpJiHBEGEwhcwJ1CLgfGIwjB0GFMLhKHAwj8-sNAcDQXj6GBMcYAG6oaCobCwb6-f4UkG-dzEbnoODgsheX4qem%2BOBU%2BgAXmiIAAzIQAIwAFnsfFYvh8mLgnD4GXQEAqxExkhgmBUmAg7UoIPFMvoAHJhuhMTBHXxXZIVO0AI4Za3mlQAaxgpBUqEolHgDFljoA7qgfOgPegQABfM6Ru0wRjQGAUd44AgkcjpoA), [[pyright]](https://pyright-play.net/?code=MYGwhgzhAEBiD28DaAVAugLgFDV9AJgKYBm0A5oQC4AUEhIxAlNALQB80KG0AdHzniKkw%2BfEgDKaWvWLcEySQBpo8SgAtCAJ24gAlhEoS0ygPTN2cREe58eWLKEgxxAVwBG1eUl0A7SmkZsPGgAYmgAWwBPAAdI5QAvdzAfZUo46FjNEhBI7kBQckB4P%2BVM3TI1Sm5AGXICgVwhaBF8aQZlVQ1taD0Dbz8A1g5XNxs%2BIA)

```python
class Foo[T]:
    def get(self) -> T: ...
    def add[S](self: Foo[S], other: list[S], /) -> Foo[S]: ...

class Sub(Foo[int]):
    # mypy, zuban, ty, pyrefly: ✅️, pyright: ❌️
    def add(self, other: list[int]) -> Sub: ...
```

(I believe `pyright` is correct to error here, as for example a function that expects a `Foo[int]` could in principle try to call `type(arg).add(Foo[str](), ["a", "b", "c"])`)

Notably, if we remove the `self: Foo[S]` annotation, then the inference changes:
[[mypy]](https://mypy-play.net/?gist=942e7831ef27d9365ab249a2f48cc1b0), [[pyright]](https://pyright-play.net/?code=MYGwhgzhAEBiD28DaAVAugLgFDV9AJgKYBm0A5oQC4AUEhIxAlNALQB80KG0AdHzniKkw%2BfEgDKaWvWIAaaPEoALQgCduIAJYRKEtPID0zdnER7ufHlixDolQjoD6weADcwqzWAB2wQtQ8ybgRkTW9KNGMOEKR4ACMAK0JgCOw8aFUqAFdVb2hA3ABiaABbAE8ABzL5SmroKs8yJUp5BpIQOoAvLLifbkBQckB4P%2BtQSBhxHuoYsIjGNLxi2oHB0sq6hs0mluhu3u9WssziDu5AGXJh9NsRfGkGeUUVdWgtHSQZyNYOCbiLPiA), [[pyrefly]](https://pyrefly.org/sandbox/?project=N4IgZglgNgpgziAXKOBDAdgEwEYHsAeAdAA4CeSIAxlKnHAAQBiuuA2gCoC6iAOuvQPqYYYegHMYAFwAUcGFDABKegFoAfPXaJ6hXX0FCR9VJkysAyp1nywAGnq5JACxgAnbVAhxJFzvYD0yur0PCDMbJah2rqEfHzCopLwkgD6lLgAbqiuEBiUMNLZYtrhrBDokpxBGqW42ABWMJSVvPyCrlIArq78RQIAxPQAtqRk9pKk9mQ5Yk6SU6QdYFCT9ABendgY2oCg5IDwf3Ho1LQM5pvSpeWViq0GgxO7e8Ojq9MQs-PrmxgLSyvagBlyA5tAQJYymawKeyOFzueiebxlCpVVQaUJnbBRHS6EC2EDTEQrQiSXBDKAUQYABUWhNI9DQWDw%2BHo6XQkDE3VQkgguHQsXQg3MMBg9DmkmIcEQ-n8BOWpEIuFcYn8MHQ-kwuEocH8rPZnO5vP89DAiuMWWgqGwsBZvL1ri5PP4uGIBvQcH5ZGcvJUGTccEd9AAvCEQABmQgARgALKE%2BKw3K5FXBOHxOugIENiIqkpgVJgIB1mhBfUH6AByd7oRUwMt8KuSFQdACOnQLMFzAGsYKQVKhKPk6KWywB3bLoWvoEAAXzxfe5vsY0BgFAZOAIJHIU6AA), [[ty]](https://play.ty.dev/36714b64-0db5-486f-9a30-30a9ff645b58)

```python
class Foo[T]:
    def get(self) -> T: ...
    def add[S](self, other: list[S], /) -> Foo[S]: ...

def test_covariance(arg: Foo[int]) -> Foo[object]:
    return arg  # mypy, ty, pyright, pyrefly, zuban: ✅️

class Sub(Foo[int]):
    # ty: ✅️ mypy, pyright, zuban, pyrefly: ❌️
    def add(self, other: list[int]) -> Sub: ...
```

## What's needed

Type checker divergence in variance inference is quite insidious; I believe the spec should be clarified in this regard and examples like the ones in this issue should be added to the conformance suite.

I believe that `pyright` is the only type checker with correct behavior here at the moment, but maybe @erictraut can comment on these examples, and as there is some disagreement, at least from the `mypy` maintainers, I believe this needs some discussion.

## References / Other Discussions

- https://github.com/python/mypy/issues/19439
- https://github.com/python/mypy/issues/20644
- https://github.com/python/mypy/issues/18334
- https://github.com/python/mypy/pull/19466

randolf-scholz:
Fixes #2281

This adds a conformance test checking that method scoped type-variables do not influence variance inference, in accordance with the spec.

### DIFF
--- a/conformance/tests/generics_variance_inference.py
+++ b/conformance/tests/generics_variance_inference.py
@@ -192,3 +192,14 @@ class ShouldBeContravariant2[T](Parent_Contravariant[T]):
 
 c1: ShouldBeContravariant2[int] = ShouldBeContravariant2[float]()  # OK
 c2: ShouldBeContravariant2[float] = ShouldBeContravariant2[int]()  # E
+
+class ShouldBeCovariant7[T]:
+    def get(self) -> T:
+        raise NotImplementedError
+
+    def add[S](self: "ShouldBeCovariant7[S]", other: list[S]) -> "ShouldBeCovariant7[S]":
+        raise NotImplementedError
+
+
+d1: ShouldBeCovariant7[object] = ShouldBeCovariant7[int]()  # OK
+d2: ShouldBeCovariant7[int] = ShouldBeCovariant7[object]()  # E
--- a/conformance/results/mypy/generics_variance_inference.toml
+++ b/conformance/results/mypy/generics_variance_inference.toml
@@ -1,4 +1,9 @@
-conformant = "Pass"
+conformant = "Partial"
+notes = """
+polymorphic functions with bound self type influence variance inference,
+which is not part of the variance inference algorithm as described in the spec.
+See gh-2281 for details.
+"""
 output = """
 generics_variance_inference.py:24: error: Incompatible types in assignment (expression has type "ClassA[float, int, int]", variable has type "ClassA[int, int, int]")  [assignment]
 generics_variance_inference.py:25: error: Incompatible types in assignment (expression has type "ClassA[float, int, int]", variable has type "ClassA[float, float, int]")  [assignment]
@@ -23,7 +28,10 @@ generics_variance_inference.py:169: error: Incompatible types in assignment (exp
 generics_variance_inference.py:170: error: Incompatible types in assignment (expression has type "ShouldBeInvariant6[int]", variable has type "ShouldBeInvariant6[float]")  [assignment]
 generics_variance_inference.py:181: error: Incompatible types in assignment (expression has type "ShouldBeCovariant6[float]", variable has type "ShouldBeCovariant6[int]")  [assignment]
 generics_variance_inference.py:194: error: Incompatible types in assignment (expression has type "ShouldBeContravariant2[int]", variable has type "ShouldBeContravariant2[float]")  [assignment]
+generics_variance_inference.py:204: error: Incompatible types in assignment (expression has type "ShouldBeCovariant7[int]", variable has type "ShouldBeCovariant7[object]")  [assignment]
+generics_variance_inference.py:205: error: Incompatible types in assignment (expression has type "ShouldBeCovariant7[object]", variable has type "ShouldBeCovariant7[int]")  [assignment]
 """
-conformance_automated = "Pass"
+conformance_automated = "Fail"
 errors_diff = """
+Line 204: Unexpected errors ['generics_variance_inference.py:204: error: Incompatible types in assignment (expression has type "ShouldBeCovariant7[int]", variable has type "ShouldBeCovariant7[object]")  [assignment]']
 """
--- a/conformance/results/pycroscope/generics_variance_inference.toml
+++ b/conformance/results/pycroscope/generics_variance_inference.toml
@@ -25,4 +25,5 @@ output = """
 ./generics_variance_inference.py:170:0: Incompatible assignment: expected generics_variance_inference.ShouldBeInvariant6[float | int], got generics_variance_inference.ShouldBeInvariant6[int] [incompatible_assignment]
 ./generics_variance_inference.py:181:0: Incompatible assignment: expected generics_variance_inference.ShouldBeCovariant6[int], got generics_variance_inference.ShouldBeCovariant6[float | int] [incompatible_assignment]
 ./generics_variance_inference.py:194:0: Incompatible assignment: expected generics_variance_inference.ShouldBeContravariant2[float | int], got generics_variance_inference.ShouldBeContravariant2[int] [incompatible_assignment]
+./generics_variance_inference.py:205:0: Incompatible assignment: expected generics_variance_inference.ShouldBeCovariant7[int], got generics_variance_inference.ShouldBeCovariant7[object] [incompatible_assignment]
 """
--- a/conformance/results/pyrefly/generics_variance_inference.toml
+++ b/conformance/results/pyrefly/generics_variance_inference.toml
@@ -26,4 +26,5 @@ ERROR generics_variance_inference.py:169:31-58: `ShouldBeInvariant6[float]` is n
 ERROR generics_variance_inference.py:170:33-58: `ShouldBeInvariant6[int]` is not assignable to `ShouldBeInvariant6[float]` [bad-assignment]
 ERROR generics_variance_inference.py:181:31-58: `ShouldBeCovariant6[float]` is not assignable to `ShouldBeCovariant6[int]` [bad-assignment]
 ERROR generics_variance_inference.py:194:37-66: `ShouldBeContravariant2[int]` is not assignable to `ShouldBeContravariant2[float]` [bad-assignment]
+ERROR generics_variance_inference.py:205:31-59: `ShouldBeCovariant7[object]` is not assignable to `ShouldBeCovariant7[int]` [bad-assignment]
 """
--- a/conformance/results/pyright/generics_variance_inference.toml
+++ b/conformance/results/pyright/generics_variance_inference.toml
@@ -79,6 +79,10 @@ generics_variance_inference.py:194:37 - error: Type "ShouldBeContravariant2[int]
   "ShouldBeContravariant2[int]" is not assignable to "ShouldBeContravariant2[float]"
     Type parameter "T@ShouldBeContravariant2" is contravariant, but "int" is not a supertype of "float"
       "float" is not assignable to "int" (reportAssignmentType)
+generics_variance_inference.py:205:31 - error: Type "ShouldBeCovariant7[object]" is not assignable to declared type "ShouldBeCovariant7[int]"
+  "ShouldBeCovariant7[object]" is not assignable to "ShouldBeCovariant7[int]"
+    Type parameter "T@ShouldBeCovariant7" is covariant, but "object" is not a subtype of "int"
+      "object" is not assignable to "int" (reportAssignmentType)
 """
 conformance_automated = "Pass"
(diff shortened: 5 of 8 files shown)


randolf-scholz:
Actually, let's delay until ty 0.0.36 release (https://github.com/astral-sh/ty/issues/1459)

davidhalter:
I understand why you want to clarify this and I'm generally supportive.

Would this be considered a spec change? I'm not quite sure if this should be documented and therefore might need a discuss thread. In any case I would like to at least hear the opinions of the Mypy devs here.

conformance/results/basilisk/generics_variance_inference.toml
```diff
@@ -1,6 +1,7 @@
 conformant = "Pass"
```
carljm: This should be updated to Partial if this is merged.

conformance/tests/generics_variance_inference.py
```diff
@@ -192,3 +192,14 @@ class ShouldBeContravariant2[T](Parent_Contravariant[T]):
 
 c1: ShouldBeContravariant2[int] = ShouldBeContravariant2[float]()  # OK
 c2: ShouldBeContravariant2[float] = ShouldBeContravariant2[int]()  # E
+
+class ShouldBeCovariant7[T]:
+    def get(self) -> T:
+        raise NotImplementedError
+
+    def add[S](self: "ShouldBeCovariant7[S]", other: list[S]) -> "ShouldBeCovariant7[S]":
```
carljm, line 200: I'm not sure it's correct for this method not to affect variance. (Assuming the class were named `C` instead for brevity) `C[object]` has method `add` taking `list[object]`, where `C[int]` has method `add` taking `list[int]`. That means neither `C[object]` nor `C[int]` can safely be substituted for the other, and `C` should be invariant in `T`.

So it seems to me that if the algorithm in the spec says this class should be covariant, that might be a bug in the spec text.

randolf-scholz: I believe this argument in its current form relies on circular reasoning, because whether or not `C[int].add` allows `list[object]` argument depends on whether or not `C` is considered covariant in `S` in the first place.

Let's remember that `x.add(items)` is basically syntactic sugar for `type(x).add(x, items)`. Then, given `x: C[int]` and `items: list[objects]` our constraints are:

1. `C[int] <: C[S]` from assigning `x` to `self`
2. `list[object] <: list[S]`  from assigning `items` to `other`

The first constraint is equivalent to `S=int` if we **presume** that `C` is invariant in `S`, and equivalent to `int <: S` if `C` is presumed covariant in `S`. In the latter case `S=object` is a solution.

Let's make the example more concrete:

```python
class C[T]:
    def __init__(self, vals: list[T]) -> None:
        self._vals: Sequence[T] = vals

   def get(self, idx: int) -> T:
       return self._vals[idx]
       
    def add[S](self: C[S], items: list[S]) -> C[S]:
        return C([*self._vals, *items])
```

Do you have a concrete example were treating `C` covariantly leads to a runtime error? Then I'd gladly change this PR to assert invariance.

carljm: Yes, I believe you are right! I will approve the PR if you fix my other comment (which will also fix the CI).

carljm (approved):
I think this is correct, both in principle and per the current language in the spec. So I think we should merge this. If someone would like to advocate a change to the language in the spec, that would be a separate thing.

randolf-scholz:
Rebased after branches divergenced due to removal of basilisk.

PR #2289, opened 2026-05-18
Clarify generic type parameter ordering

### LINKED ISSUE #2059: Spec: More precision on type parameter order
The spec currently isn't very precise on the order in which type parameters appear in a class. (Given a class C that is generic over type parameters T1 and T2, does `C[int, str]` mean that T1 is bound to int or that T2 is bound to int?)

I think the rule should be:

- If PEP 695 syntax is used, the order is the order in which the type parameters appear in the type parameter list.
- If the class has `Generic[...]` as a syntactic base, the order is the order of the type arguments to `Generic`.
- If the class has `Protocol[...]` as a syntactic base, the order is the order of the type arguments to `Protocol`. This applies only if Protocol is subscripted, not if the class inherits from bare `Protocol`.
- Otherwise, the order is the order in which the type parameters appear syntactically in the base class list.

As far as I know, this is how all type checkers currently behave, but we just found out it's not how the runtime behaves when Protocol is involved: https://github.com/python/cpython/issues/137191#issuecomment-3132748382.

In these examples, all classes are generic over two parameters T1 and T2, and in all cases the type parameter order is T1, T2:

```python
from typing import Generic, Protocol, TypeVar

class C1[T1, T2]: ...  # order is T1, T2

T1 = TypeVar("T1")
T2 = TypeVar("T2")

class C2(C1[T2, T1], Generic[T1, T2]): ...  # order is T1, T2

class GenericProto[T1, T2](Protocol):  ...  # order is T1, T2

class Proto2(GenericProto[T2, T1], Protocol[T1, T2]): ...  # order is T1, T2

class Proto3(GenericProto[T1, T2], Protocol): ...  # order is T1, T2

class C3(C2[T1, T2]): ...  # order is T1, T2
```

We should change the spec to make this rule explicit. The only current rule appears to be in https://typing.python.org/en/latest/spec/generics.html#arbitrary-generic-types-as-base-classes "Type variables are applied to the defined class in the order in which they first appear in any generic base classes".

BHUVANSH855:
## Summary

Clarify how type parameter ordering is determined for generic classes.

This expands the existing wording in the generics spec to explicitly
cover:

* `Generic[...]` base classes
* parameterized `Protocol[...]` base classes
* Python 3.12 generic class syntax

The updated wording also clarifies that a bare `Protocol` base class
does not affect type parameter ordering.

Closes #2059

### DIFF
--- a/docs/spec/generics.rst
+++ b/docs/spec/generics.rst
@@ -505,6 +505,51 @@ That ``Child`` definition is equivalent to::
   class Child(Parent1[T1, T3], Parent2[T2, T3], Generic[T1, T3, T2]):
       ...
 
+When ``Generic[...]`` is present in the base class list, the order of
+type parameters is determined by the order of type arguments to
+``Generic``::
+
+  from typing import Generic, TypeVar
+
+  T1 = TypeVar('T1')
+  T2 = TypeVar('T2')
+
+  class Base(Generic[T1, T2]):
+      ...
+
+  class Child(Base[T2, T1], Generic[T1, T2]):
+      ...
+
+In this example, ``Child`` has type parameter order ``T1``, ``T2``.
+
+Similarly, when a parameterized ``Protocol[...]`` base class is present,
+the order of type parameters is determined by the order of type
+arguments to ``Protocol``. A bare ``Protocol`` base class does not
+affect type parameter ordering::
+
+  from typing import Protocol, TypeVar
+
+  T1 = TypeVar('T1')
+  T2 = TypeVar('T2')
+
+  class ProtoBase(Protocol[T1, T2]):
+      ...
+
+  class ProtoChild(ProtoBase[T2, T1], Protocol[T1, T2]):
+      ...
+
+  class BareProtoChild(ProtoBase[T1, T2], Protocol):
+      ...
+
+In both examples above, the type parameter order is ``T1``, ``T2``.
+
+For classes using the Python 3.12 generic class syntax, the type
+parameter order is determined by the order of parameters in the type
+parameter list::
+
+  class NewStyle[T1, T2]:
+      ...
+
 A type checker should report an error when the type variable order is
 inconsistent::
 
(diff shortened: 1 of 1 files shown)


JelleZijlstra:
@BHUVANSH855 please do not keep merging main into your branch unless there is an actual merge conflict.

docs/spec/generics.rst
```diff
@@ -505,6 +505,51 @@ That ``Child`` definition is equivalent to::
...
+  class ProtoBase(Protocol[T1, T2]):
+      ...
+
+  class ProtoChild(ProtoBase[T2, T1], Protocol[T1, T2]):
+      ...
+
+  class BareProtoChild(ProtoBase[T1, T2], Protocol):
+      ...
+
+In both examples above, the type parameter order is ``T1``, ``T2``.
```
JelleZijlstra: There are three examples.

docs/spec/generics.rst
```diff
@@ -505,6 +505,51 @@ That ``Child`` definition is equivalent to::
...
+
+For classes using the Python 3.12 generic class syntax, the type
```
JelleZijlstra: This should be first, before the legacy Generic[] syntax.

JelleZijlstra (changes requested):
I think we should add a section header "Type parameter ordering", starting at line 487 or so. Currently this is in a section "Arbitrary generic types as base classes", which isn't really apposite.

JelleZijlstra:
Also, is this already covered sufficiently in the conformance suite? What are the relevant tests that we have?

BHUVANSH855:
> Also, is this already covered sufficiently in the conformance suite? What are the relevant tests that we have?

Addressed the requested changes:

* added a dedicated “Type parameter ordering” section header
* reordered the section so the Python 3.12 generic class syntax appears first
* updated “both examples” to “all of the examples above”

There is already conformance coverage for `Protocol[...]` ordering in `conformance/tests/protocols_generic.py`. In particular, the `Proto1(Iterable[T_co], Protocol[S, T_co])` example together with the specialization checks verifies that type parameter ordering follows the `Protocol[...]` subscript order.

This PR expands the spec wording to make the broader ordering rules explicit, including `Generic[...]`, parameterized `Protocol[...]`, bare `Protocol`, and PEP 695 syntax.

PR #2293, opened 2026-05-21
Bump type checker versions in conformance tests.

rchen152:
As far as I can tell, nothing significant has changed conformance-wise, but several type checkers have new versions, so we might as well update.

Updated mypy v1.20.2 -> v2.1.0
Updated pyrefly v0.63.1 -> v1.0.0
Updated ty v0.0.35 -> v0.0.38
Updated zuban v0.7.1 -> v0.7.2


conformance/results/pyrefly/constructors_callable.toml
JelleZijlstra, line 3: you might be able to remove a few of these lines

JelleZijlstra (approved):

PR #2296, opened 2026-05-28
Remove questionable dataclass descriptor conformance assertions

### LINKED ISSUE #2259: Conformance suite: Questionable tests in dataclasses_descriptors.py

```python
@dataclass
class DC2:
    x: Desc2[int]
    y: Desc2[str]
    z: Desc2[str] = Desc2()


assert_type(DC2.x, list[int])
assert_type(DC2.y, list[str])
assert_type(DC2.z, list[str])
```

```python

assert_type(dc2.x, int)
assert_type(dc2.y, str)
assert_type(dc2.z, str)
```

BHUVANSH855:
## Summary

Remove questionable `Desc2` conformance assertions from `dataclasses_descriptors.py`.

The existing assertions do not align with actual runtime behavior for non-data descriptors in dataclasses:

* `DC2.x` and `DC2.y` raise `AttributeError` at runtime because no descriptor instance exists in the class dictionary for those fields.
* `dc2.x` and `dc2.y` evaluate to the stored `Desc2` instances because non-data descriptors are shadowed by instance attributes.
* The `z` case is also subtle due to descriptor access during dataclass default extraction.

Since this behavior is currently under-specified and differs across type checkers and runtime semantics, this PR removes the questionable assertions and replaces them with explanatory comments.

Fixes #2259


davidhalter:
Closing, because there is also #2299, which seems more promising and tries to fix the exact same issue.

PR #2297, opened 2026-05-29
Conformance: test @override checks on __init__ and __new__

### LINKED ISSUE #2222: Conformance: test that `@override` checks are honored for `__init__` / `__new__`
When explicitly marking a method with `@override`, type checkers ought to test assignability (https://typing.python.org/en/latest/spec/class-compat.html#override).

Hence, even methods like `__init__` and `__new__`, which are normally not checked against their parent classes, should be checked when explicilty decorated with `@override`.

```python
from typing import override

class Foo:
    def __init__(self, x: int) -> None: ...

class Bar(Foo):
    @override
    def __init__(self, y: int) -> None: ...  # should error
```

This is useful in cases when we really need constructors to be compatible.
I suggest adding a version of the example above to the conformance test suite.

ashishpatel26:
Closes #2222.

The `@override` section of the spec states that a method decorated with
`@override` "should [be treated] as a type error unless that method is
overriding a method or attribute in some ancestor class, and the type of the
overriding method is assignable to the type of the overridden method." There is
no exemption for `__init__`/`__new__`, so although constructors are normally
not checked against their parents, an explicit `@override` should still trigger
the assignability check.

This adds cases to `tests/classes_override.py`:

- `ChildC1`: `@override` `__init__`/`__new__` that *are* assignable to the
  parent — no error.
- `ChildC2`: `@override` `__init__`/`__new__` with an incompatible parameter
  type — error (`# E[init]`, `# E[new]`).
- `ChildC3`: incompatible `__init__`/`__new__` *without* `@override` — allowed,
  to document the normal constructor exemption.

### Scoring

Ran the conformance tool against the locked checker versions:

- **pyrefly** flags the incompatible constructor overrides → conformant.
- **mypy, pyright, pycroscope, ty, zuban** do not currently apply the
  assignability check to `__init__`/`__new__` even with `@override` → scored
  `Partial` with a note.

I marked the incompatible cases as mandatory errors (`# E`) since the spec text
is unconditional. If maintainers feel constructor override-checking should be
left optional, these could be relaxed to `# E?` and the five `Partial` scores
returned to `Pass` — happy to adjust.


JelleZijlstra (approved):

jorenham (commented):
In typeshed the `object.__init__` and `object.__new__` are defined as ([src](https://github.com/python/typeshed/blob/79859b31dbd949a4799371976d6b142e6a568f4d/stdlib/builtins.pyi#L112-L113))

```py
class object:
    # --snip--
    def __init__(self) -> None: ...
    def __new__(cls) -> Self: ...
```

Every class without base implicitly inherits from `object`. So does that mean that e.g.

```py
class A:
    @override
    def __init__(self, x: int) -> None: ...
```

should now also be rejected? This technically indeed overrides `object.__init__`, but I'm not sure if reporting an error is actually useful in this case.

JelleZijlstra:
I think it should, yes. It is not a particularly useful error, but the user did put that override decorator there, so the type checker should honor it.

jorenham (approved):

rchen152:
Thanks @ashishpatel26!

PR #2299, opened 2026-06-01
Conformance: align dataclass descriptor tests with runtime

### LINKED ISSUE #2259: Conformance suite: Questionable tests in dataclasses_descriptors.py

ashishpatel26:
Closes #2259.

`dataclasses_descriptors.py` keeps coverage for the non-data descriptor (`Desc2`, only `__get__`) used as a dataclass field, but aligns the expected behavior with runtime descriptor lookup:

- `DC2.x` / `DC2.y` class access is optional-error because those fields have no descriptor object in the class namespace.
- `DC2.z` class access is `list[str]`, because class access invokes `Desc2.__get__(None, owner)`.
- `dc2.x` / `dc2.y` / `dc2.z` are the stored `Desc2[...]` objects, because non-data descriptors are shadowed by the instance `__dict__` after dataclass `__init__` assigns the fields.

The result TOMLs and summary HTML have been regenerated. With the runtime-correct asserts, all six checkers are scored `Partial` for `dataclasses_descriptors`.


davidhalter:
What - in your opinion - would be the proper solution for these cases? I generally want to avoid removing tests and rather expand the docs or correct the tests. There are a lot of cases where the spec is lacking and I want to avoid removing them all.

In my opinion it is very valuable to standardize descriptors on Dataclasses. I'm not particularly invested in how they are standardized, but I would like to have that to be one specific way.

JelleZijlstra:
As I wrote in the issue, pyrefly's behavior seemed the most correct to me. I'm not sure this behavior needs to be specified in the typing spec in much detail since what I view as the correct behavior primarily reflects the runtime.

davidhalter:
> pyrefly's behavior seemed the most correct to me. I'm not sure this behavior needs to be specified in the typing spec in much detail since what I view as the correct behavior primarily reflects the runtime.

I agree that this does not necessarily need to be specified. However I would still prefer to make the pyrefly behavior part of the tests than to drop the test.

ashishpatel26:
Thanks both. I reworked this to keep the non-data descriptor coverage rather than deleting it, and aligned the `DC2` case with runtime descriptor lookup:

- `DC2.x` / `DC2.y` class access remains optional-error because those fields have no class attribute at runtime.
- `DC2.z` class access is asserted as `list[str]`, since class access invokes `Desc2.__get__(None, owner)`.
- `dc2.x` / `dc2.y` / `dc2.z` are asserted as the stored `Desc2[...]` objects, because a non-data descriptor is shadowed by the instance `__dict__` after dataclass `__init__` assigns the fields.

I regenerated the result files and summary HTML for this test. With the runtime-correct asserts, all six checker result TOMLs are now scored `Partial` for `dataclasses_descriptors`.

conformance/tests/dataclasses_descriptors.py
```diff
@@ -55,14 +64,22 @@ def __get__(self, instance: object | None, owner: Any) -> list[T] | T:
...
+    z: Desc2[str] = Desc2()  # E?: a non-data descriptor default may be rejected
 
 
-assert_type(DC2.x, list[int])
-assert_type(DC2.y, list[str])
+# ``x`` and ``y`` are not present in the class namespace, so accessing them on
+# the class is an ``AttributeError`` at runtime; a type checker may report an
+# error here. ``z`` is present in the class namespace, so class access runs
+# ``Desc2.__get__(None)`` and yields ``list[str]``.
+assert_type(DC2.x, Desc2[int])  # E?
```
JelleZijlstra, line 74: An assert_type with an `E?` doesn't really test anything. This line will throw AttributeError at runtime so I think the only reasonable thing to test for is that `DC2.x` should emit an error. I don't feel strongly though that we need to require type checkers to do that; an alternative is to simply omit these lines.

JelleZijlstra:
A bit sad that even the type checkers that attempt hardest to support this faithfully (pyrefly and pycroscope) still fail these tests. Perhaps we should make some things a little more lenient.

conformance/tests/dataclasses_descriptors.py
davidhalter: Is there an argument why we shouldn't require that? I don't have a strong opinion here, but it feels like in dataclasses these fields should never be available. (The same goes for `DC2.x = 1`, which should probably also be an error.

And this probably means that we should test for `DC2.x # E` without an assert_type. Zuban currently also doesn't add an error there, which is wrong.

davidhalter (commented):
> A bit sad that even the type checkers that attempt hardest to support this faithfully (pyrefly and pycroscope) still fail these tests. Perhaps we should make some things a little more lenient.

Aren't these just small oversights? Has anyone tried to pass these new tests? It feels like these new changes are not that hard to implement, but I haven't tried yet.

conformance/tests/dataclasses_descriptors.py
carljm: I think we need to be careful about over-specifying the behavior of dataclasses in type checkers based on the particular runtime behavior of stdlib dataclasses. There's a lot of stdlib dataclass behavior that is effectively accidental. And everything we specify for dataclasses in the typing spec implicitly applies to type checking of all libraries using `dataclass_transform`. Is a dataclass-transform library which explicitly assigns a value or a descriptor to the class object in its class transform for each field non-conformant with the dataclass-transform spec?

I do not think type checkers should be required to error on `DC2.x` here.

carljm: No type checker currently errors on this:

```py
class C:
    x: int
    
reveal_type(C.x)
```

Given that dataclass behavior encompasses arbitrary third-party dataclass transforms (and arbitrary behavior people add to their own dataclasses), why should we specify a greater level of confidence about the absence of this attribute for dataclasses than we do for regular classes?

conformance/tests/dataclasses_descriptors.py
```diff
@@ -55,14 +64,22 @@ def __get__(self, instance: object | None, owner: Any) -> list[T] | T:
...
+# ``Desc2`` is a non-data descriptor, the instance ``__dict__`` shadows it and
+# ``__get__`` never runs, so each attribute keeps the assigned ``Desc2`` object
+# (even ``z``, which is also present in the class namespace).
 dc2 = DC2(Desc2(), Desc2(), Desc2())
-assert_type(dc2.x, int)
-assert_type(dc2.y, str)
-assert_type(dc2.z, str)
+assert_type(dc2.x, Desc2[int])
+assert_type(dc2.y, Desc2[str])
+assert_type(dc2.z, Desc2[str])
```
carljm, line 85: This opens up some new questions. At runtime, if this field were `init=False`, or the dataclass provided its own custom `__init__` (that is, if in one way or another the dataclass does not actually assign a value for `z` in each instance dict), the descriptor on the class could run and this might be `str` instead of `Desc2[str]`. Should type checkers attempt to model this also?

conformance/tests/dataclasses_descriptors.py
JelleZijlstra: Carl puts it better than I would have, but I'm also concerned about overspecifying things that aren't important for compatibility. I think the tests as they were before this PR are wrong because type checkers that faithfully implement the runtime behavior are penalized, but we could also just leave this behavior out of the conformance suite.

davidhalter: > Given that dataclass behavior encompasses arbitrary third-party dataclass transforms (and arbitrary behavior people add to their own dataclasses), 

I think that dataclass transforms are a good example why we do not want to generalize it to that. At least in Zuban I would probably only emit the error in a normal dataclass. 

> why should we specify a greater level of confidence about the absence of this attribute for dataclasses than we do for regular classes?

I think from a practical reason it almost never makes sense to assign to a dataclass, since they have an auto-generated `__init__` that will fill the `__dict__` field. Regular classes are different here, users tend to do almost anything with them. I still lean toward adding an error here, but I realize that especially dataclass_transform makes this a bit of a weird edge case. (Also this error could only be emitted if there's no customized `__init__` and `init=True`). I now understand that we don't want to force that on type checkers.

rchen152: Fwiw I think the part of this test file with `Desc1` and `DC1` is worth keeping, because it reflects valid code that type checkers ought to handle consistently and correctly.

The `Desc2` and `DC2` tests I'm ambivalent on. I think there's a reasonable argument for keeping the class definitions themselves, as they show a clear logic error that checkers could usefully detect, but I don't feel strongly about it. Everything from line 68 onwards seems to be essentially documenting messy runtime behavior, and my vote would be delete it entirely.

PR #2300, opened 2026-06-04
[conformance] Add examples that require bidi inference of list literals after specialization inference

dcreager:
The conformance suite currently contains the following example:

https://github.com/python/typing/blob/94514b12d6d6ee61fd0204699950265ceb401a8e/conformance/tests/constructors_callable.py#L178-L186

The important point is that `r8` is a generic callable. It has the following signature:

```
[T](x: list[T], y: list[T]) -> Class8[T]
```

Previously, the last line of this example would _require_ that if you pass in list literals of different types, it should be a type inference error. However, the union of the two lists' respective element types is a valid solution — in this case, `T = int | str`.

I've updated these examples so that it's not an error if they succeed. We don't require the more difficult inference result. (This example holds for a simpler generic function, so most of the new tests are in `generics_basic.py`.)

I've also added additional tests for examples where there is genuinely no valid specialization: when one of the arguments is not a list at all, and where the typevar is bounded and the argument is a list that cannot possibly satisfy the upper bound.

### DIFF
--- a/conformance/results/mypy/constructors_callable.toml
+++ b/conformance/results/mypy/constructors_callable.toml
@@ -15,6 +15,10 @@ Line 107: Unexpected errors ['constructors_callable.py:107: error: Expression is
 Line 118: Unexpected errors ['constructors_callable.py:118: error: Incompatible return type for "__new__" (returns "Class6Proxy", but must return a subtype of "Class6")  [misc]']
 Line 128: Unexpected errors ['constructors_callable.py:128: error: Expression is of type "Class6", not "Class6Proxy"  [assert-type]', 'constructors_callable.py:128: error: Too few arguments  [call-arg]']
 Line 145: Unexpected errors ['constructors_callable.py:145: error: Expression is of type "Class6Any", not "Any"  [assert-type]', 'constructors_callable.py:145: error: Too few arguments  [call-arg]']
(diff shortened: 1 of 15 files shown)


JelleZijlstra:
I agree with changing this, but the proposal to require type checkers to infer the union also doesn't feel right. This test isn't about testing precise inference behaviors, it's about testing class constructors. Type inference behavior in cases like this test isn't specified precisely, so we shouldn't force a specific behavior in the conformance suite.

The same behavior is observable in a simpler function:
```python
def f[T](x: list[T], y: list[T]):
    pass

f([1], ["x"])
```

is currently rejected by all type checkers in multiplay.

dcreager:
> I agree with changing this, but the proposal to require type checkers to infer the union also doesn't feel right. This test isn't about testing precise inference behaviors, it's about testing class constructors. Type inference behavior in cases like this test isn't specified precisely, so we shouldn't force a specific behavior in the conformance suite.

I moved most of the tests over to `generics_basic`, and made the main example from the PR body `E?` so that we do not require the specific inference result.

dcreager:
(Do you typically squash-merge? If not do you prefer that I merge the commits on this branch to make a single clean atomic commit?)

carljm:
We do usually squash merge, I don't think the branch commit history matters at all.

carljm (approved):

carljm:
Going ahead and merging this -- seem uncontroversial to allow this inference, but not require it.

PR #2301, opened 2026-06-05
conformance: make open TypedDict extra kwargs optional

charliermarsh:
## Summary

#2272 made errors optional when an extra keyword is passed to `**kwargs: Unpack[OpenTypedDict]`, modifying the expectation in `callables_kwargs.py`. But `typeddicts_extra_items.py` contains the same case:

```py
class MovieNoExtra(TypedDict):
    name: str

def unpack_no_extra(**kwargs: Unpack[MovieNoExtra]) ->
None: ...

unpack_no_extra(name="No Country for Old Men",
year=2007)  # E?
```

This PR marks the duplicate case optional, for consistency with #2272.


JelleZijlstra (approved):
Agree this seems exactly analogous to the change in #2272.

Issue #1779, opened 2024-06-17
[spec] better clarify the difference between a runtime type object and a static type


carljm:
See https://github.com/python/typing/pull/1743#discussion_r1641580302

carljm:
I think there is overlap between this and clarifying the use of the term "concrete type": https://github.com/python/typing/issues/1781

Might make sense to address them both in a single PR.

PR #2302, opened 2026-06-08
docs: add concrete type definition and update protocol references

### LINKED ISSUE #1781: [spec] define or replace the term "concrete type"
This term is used in a few places in the spec, but never defined. I think it is intended to mean "a type specified by the name of a Python class." We should either give a clear definition for the term, or replace its usages.

BHUVANSH855:
## Summary

This PR introduces a formal definition for **concrete types** in the typing specification and updates protocol-related references to point to the new definition.

## Changes

- Added a new **Concrete types** section to `docs/spec/concepts.rst`.
- Defined a concrete type as a type specified by the name of a Python class (e.g. `int`, `str`, `list`, or a user-defined class).
- Updated references in `docs/spec/protocol.rst` to use the new `:ref:\`concrete-type\`` reference.
- Clarified protocol assignability rules by linking the term "concrete type" to its definition.

## Related Issues

Closes #1781

Related to #1779

### DIFF
--- a/docs/spec/concepts.rst
+++ b/docs/spec/concepts.rst
@@ -169,6 +169,17 @@ is a subtype of ``str``, because ``MyStr`` represents a subset of the values
 represented by ``str``. Such types can be called "nominal types" and this is
 "nominal subtyping."
 
+.. _`concrete-type`:
+
+Concrete types
+--------------
+
+A **concrete type** is a type specified by the name of a Python class, for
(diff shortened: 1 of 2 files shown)


BHUVANSH855:
Hi @JelleZijlstra @srittau,

Any feedback !

davidhalter:
@BHUVANSH855 Please don't tag people here. Both of them are notified by your PRs anyway and I'm sure they have seen them. If they think the PR is worth it they will give you feedback.

PR #2305, opened 2026-06-11
Switch from requirements files to pyproject.toml

srittau:


srittau:
This works: https://typing--2305.org.readthedocs.build/en/2305/, also tested the build locally (`make -C docs venv && make -C docs html`). uv is used only by readthedocs internally as far as I understand, not required locally by developers.

pyproject.toml
```diff
@@ -0,0 +1,21 @@
+[project]
+name = "typing"
+version = "1"  # dummy
+requires-python = ">= 3.12"  # for building the documentation etc.
+
+[dependency-groups]
+test = [
```
AlexWaygood, line 7: if you call this group `dev` rather than `test`, then uv will automatically install it when you're running tests etc. locally using `uv run` commands (it's treated as a "default dependency group" for local development)

srittau: That's why I was explicitly not naming it `dev`, so that it doesn't get installed by default when building docs in this doc-centric repository. But I'll defer to you what the best option is.

AlexWaygood: Oh, fair enough! I don't feel strongly

AlexWaygood (approved):

PR #2309, opened 2026-06-22
spec: expand Never/NoReturn with precise subtyping rules

### LINKED ISSUE #1458: NoReturn and Never are underspecified in the spec
Never was added without a PEP as a way to directly indicate that something is uninhabited in 3.11

This brings up something from <https://github.com/CarliJoy/intersection_examples/issues/5#issuecomment-1699490563> where we now have a way to express what appears to be an LSP violation that type checks (incorrectly) properly via structural subtyping by mishandling Never as a subtype of other types. Never is not a subtype of all other types. As the bottom type, it is uninhabited and does not follow subtyping rules.

```py
from typing import Never

class A:
    def foo(self, x: int) -> None: ...
    
class B(A):
    foo: Never   # __getattr__(B, "foo") -> raise NotImplementedError
    
x: A = B()  # type checks 
reveal_type(B.foo)
```

https://mypy-play.net/?mypy=latest&python=3.11&flags=strict&gist=a4279b36d82c1a28d7b17be9a4bbcbdf

I believe under LSP, B is no longer a safe drop-in replacement for A. It's important to note that ongoing work to formalize the type system, including Never does not treat Never as a subtype of all other types, see the ongoing work here: http://bit.ly/python-subtyping

ashishpatel26:
## Summary

Resolves #1458.

The `Never` section in the spec was four lines with no subtyping rules. Type checkers (mypy, pyright, pyrefly, zuban) all converge on a consistent set of behaviours, but none were written down, making the spec unexecutable for implementers and leaving the community guessing.

This PR codifies what the implementations already agree on:

- `Never` is a subtype of every fully static type (bottom type property)
- No type other than `Never` (and `Any`) is a subtype of `Never`
- `Never | T` is equivalent to `T` for any `T`
- In covariant type parameter positions, `Container[Never]` is assignable to `Container[T]` for any `T`; in invariant positions, the normal invariance rules apply
- `type[Never]` is a valid annotation representing an uninhabitable class
- Clarify that `NoReturn` may appear in non-return-type positions (the old restriction was removed)

## Changes

| File | Change |
|------|--------|
| `docs/spec/special-types.rst` | Expand `Never` from 4 lines to a full subtyping specification |
| `conformance/tests/specialtypes_never.py` | Add 6 new test cases covering union collapse, `type[Never]`, and invariance |
| `conformance/results/*/specialtypes_never.toml` | Updated results for mypy, pyright, pyrefly, zuban (all Pass); ty marked Partial (misses two of the new error cases) |

## Conformance results snapshot

| Checker | Result | Notes |
|---------|--------|-------|
| mypy | Pass | All 5 expected errors flagged |
| pyright | Pass | All 5 expected errors flagged |
| pyrefly | Pass | All 5 expected errors flagged |
| zuban | Pass | All 5 expected errors flagged |
| ty | Partial | Misses `object → Never` and `list[Never] → list[int]` errors |

## Out of scope

The LSP/mutable-attribute debate from #1458 (whether `class B(A): foo: Never` should be an error) is left for a separate issue. That requires formalising mutable attribute override rules — a larger change that should not block landing the uncontroversial subtyping rules documented here.


docs/spec/special-types.rst
```diff
@@ -95,14 +95,77 @@ is unreachable and will behave accordingly::
...
+static type.  This means that a value of type ``Never`` is :term:`assignable`
+to a variable of any type ``T``::
+
+  from typing import Never
+
+  def f(x: Never) -> None:
+      v1: int = x   # OK — Never is a subtype of int
+      v2: str = x   # OK — Never is a subtype of str
+
+No type other than ``Never`` itself (and :ref:`Any <any>`) is a subtype of
```
JelleZijlstra: This is wrong, Any is not a subtype of Never

docs/spec/special-types.rst
```diff
@@ -95,14 +95,77 @@ is unreachable and will behave accordingly::
...
+  class Box(Generic[T]):
+      pass
+
+  def f() -> Box[int]:
+      return Box[Never]()  # Error — Box is invariant in T
+
+``type[Never]``
+^^^^^^^^^^^^^^^^^
+
+``type[Never]`` represents the class object for a type that has no instances.
```
JelleZijlstra: I don't know what this is supposed to mean.

JelleZijlstra (changes requested):
I don't think this verbiage is necessary. This isn't the kind of issue that can be fixed by adding some text without much understanding.

docs/spec/special-types.rst
ashishpatel26: You're right. `Any` is *assignable* to `Never` (as this file states: "Any is assignable to every type"), but that's the consistency/assignability relation, not the formal subtype relation — the spec treats these as distinct concepts. Removed `(and Any)`.

docs/spec/special-types.rst
ashishpatel26: Good point, the phrasing was unclear. Replaced with: "`type[Never]` is a subtype of `type[T]` for every type `T`, just as `Never` is a subtype of every type `T`. In practice a variable receives this type only in provably unreachable code — for example, when narrowing a `type[int] | type[str]` through all possible branches."

docs/spec/special-types.rst
```diff
@@ -95,14 +95,77 @@ is unreachable and will behave accordingly::
...
+
+  def f(x: Never) -> None:
+      v1: int = x   # OK — Never is a subtype of int
+      v2: str = x   # OK — Never is a subtype of str
+
+No type other than ``Never`` itself (and :ref:`Any <any>`) is a subtype of
+``Never``.  In particular, ``object`` is *not* a subtype of ``Never``::
+
+  def g(x: object) -> Never:
+      return x  # Error — object is not assignable to Never
```
grievejia: This example is technically correct, but `object` feels like an odd witness here. Since `object` is the "top" type, readers already expect it not to be a subtype of almost anything else, so it does not really illustrate what is special about `Never`.

docs/spec/special-types.rst
```diff
@@ -95,14 +95,77 @@ is unreachable and will behave accordingly::
...
+
+Because ``Never`` is a subtype of every type ``T``, the union ``Never | T`` is
```
grievejia: We'd want to be careful with the wording here. `Never` is a subtype of every *fully static type* `T`. The subtype relation is not defined for gradual types such as `Any` or types containing `Any`.

docs/spec/special-types.rst
grievejia: I think this claim is also too strong if read syntactically. `Never` can be a supertype of other uninhabitable types, such as `tuple[Never, int]` or, in systems with intersections, `A & B` where A and B are disjoint. These are semantically equivalent to `Never`, but spelling that out here probably adds more complexity than this section needs.

Suggested simplification: something along the line of "For ordinary inhabited types such as `int` or `str`, values are not assignable to Never".

docs/spec/special-types.rst
```diff
@@ -95,14 +95,77 @@ is unreachable and will behave accordingly::
...
+
+``Never`` may appear as a type argument.  When used with a covariant
+type parameter, ``Container[Never]`` is a subtype of ``Container[T]`` for
+every type ``T``, because ``Never`` is a subtype of every type.  This is
+useful to type an empty container whose element type is not yet known::
+
+  from typing import Never
+
+  def empty_list() -> list[Never]:
+      return []
```
grievejia, line 149: I’m not sure the `list[Never]` example is the best way to illustrate the intended rule. If the point is that an empty list can be used where the element type is not yet known, that seems clearer with a contextual or generic target type, e.g. returning `list[T]` from [].

`list[Never]` is also a bit subtle: it is inhabited by the empty list, but because list is invariant, it is not assignable to `list[int]`. So it doesn’t really demonstrate the useful “bottom type flows into any element type” intuition; it mostly demonstrates invariance.

Maybe use a covariant container like `Sequence[Never]` for the example?

docs/spec/special-types.rst
ashishpatel26: Good point. Switched to `int` as the witness and rephrased to avoid the too-strong 'no type other than Never' claim (uninhabitable types like `tuple[Never, int]` are also subtypes of Never). New text: 'For ordinary inhabited types such as `int` or `str`, no value is assignable to `Never`.' with `def g(x: int) -> Never: return x  # Error`.

docs/spec/special-types.rst
ashishpatel26: Good catch. Added 'fully static' qualifier throughout — now reads 'every fully static type T' in the union equivalence, the covariant type argument section, and the `type[Never]` section.

docs/spec/special-types.rst
ashishpatel26: Agreed — addressed together with the object-witness comment (r3454364249). The new text avoids the too-strong universal claim: 'For ordinary inhabited types such as `int` or `str`, no value is assignable to `Never`.' This sidesteps the uninhabitable-type edge cases you mention (`tuple[Never, int]`, intersections) without needing to enumerate them.

docs/spec/special-types.rst
ashishpatel26: Correct — `list` is invariant so `list[Never]` is not assignable to `list[int]`, which directly contradicts the covariant-container intuition the example was trying to convey. Switched to `Sequence[Never]` (`Sequence` is covariant) and renamed the function to `empty_sequence`.

PR #2312, opened 2026-06-22
typeddict: clarify that .get() with an unknown literal key is not an error

### LINKED ISSUE #2054: Discrepancy between spec and tests on indexing TypedDicts with unknown literal keys
*Splitting issue from PR https://github.com/python/typing/pull/1978*

Spec says that indexing with unknown literal keys should be reported as an error:

> **The use of a key that is not known to exist should be reported as an error, even if this wouldn’t necessarily generate a runtime type error.** These are often mistakes, and these may insert values with an invalid type if structural assignability hides the types of certain items. For example, d[‘x’] = 1 should generate a type check error if ‘x’ is not a valid key for d (which is assumed to be a TypedDict type).

But it's not reflected in conformance tests (https://github.com/python/typing/blob/2d88da2a407b556f5b52b4c6ad1bd659873bb9ac/conformance/tests/typeddicts_operations.py) and there's a bit of discrepancy between type checkers how they handle some cases (most of them reject all cases that may cause runtime error / allow cases that do not cause runtime error ).

Would be great if we can either adapt conformance tests to the spec or amend the spec to remove the ambiguity.

Related discussion - https://discuss.python.org/t/typeddict-operations-with-unknown-literal-keys/89653 (also has information on state of this issue in different type checkers).

ashishpatel26:
## Summary

Fixes #2054.

The spec's `.get()` exception bullet only specified the return type for the variable-key case ("cannot be determined statically"), leaving two cases undefined:

- `d.get("known_key")` — a literal key that exists in the TypedDict
- `d.get("unknown_key")` — a literal key **not** defined in the TypedDict

This PR expands the bullet to cover all three cases explicitly:

1. **Known literal key** — return type is `T | None` for non-required keys; type checkers may return `T` or `T | None` for required.
2. **Unknown literal key** — NOT an error; return type is the union of all value types combined with `None`.
3. **Non-literal key** — unchanged: union of all value types combined with `None`.

### Conformance test changes

The test had `movie.get("other")  # E?` — the `# E?` marker is a wildcard that treats errors as optional (neither requiring nor forbidding one). Now that the spec is explicit, this becomes a plain call with no error marker.

`pycroscope` is the only checker that currently flags `movie.get("other")` as an error; its TOML is updated to `Partial`/`Fail`. All other checker TOMLs are updated for the +2-line shift caused by replacing the "not clear" comments with spec-quoting comments.

## Files changed

- `docs/spec/typeddict.rst` — expanded `.get()` bullet with three explicit sub-cases
- `conformance/tests/typeddicts_operations.py` — removed `# E?`, updated comments
- `conformance/results/*/typeddicts_operations.toml` — updated line numbers; pycroscope updated to Partial/Fail

## Test plan

- [ ] Conformance suite run locally confirms line numbers in all TOMLs are correct
- [ ] Only pycroscope has a `conformance_automated = "Fail"` result for this test (as expected)
- [ ] `validate_results.py` passes (all Fail results have explicit `conformant` field)

### DIFF
--- a/docs/spec/typeddict.rst
+++ b/docs/spec/typeddict.rst
@@ -757,9 +757,19 @@ This section discusses some specific operations in more detail.
   should be allowed for TypedDict objects, for an arbitrary expression
   ``e`` with type ``str``.  The motivation is that these are safe and
   can be useful for introspecting TypedDict objects.  The static type
-  of ``d.get(e)`` should be the union of all possible item types in ``d``
-  if the string value of ``e`` cannot be determined statically.
-  (This simplifies to ``object`` if ``d`` is :term:`open`.)
+  of ``d.get(e)`` depends on what is known about the value of ``e``:
+
+  - If ``e`` is a string literal equal to a defined key of ``d``, the
+    return type is the declared type of that key combined with ``None``
+    for non-required keys.  For required keys, type checkers may return
+    either the declared type ``T`` or ``T | None``.
+  - If ``e`` is a string literal that is **not** a defined key of ``d``,
+    no error should be reported.  The return type is the union of all
+    possible value types in ``d`` combined with ``None``.
+  - If the string value of ``e`` cannot be determined statically, the
+    return type is the union of all possible value types in ``d``
+    combined with ``None``.  (This simplifies to ``object | None`` if
+    ``d`` is :term:`open`.)
 
 * ``clear()`` is not safe on :term:`open` TypedDicts since it could remove required items, some of which
   may not be directly visible because of :term:`structural`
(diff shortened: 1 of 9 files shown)


docs/spec/typeddict.rst
```diff
@@ -757,9 +757,19 @@ This section discusses some specific operations in more detail.
...
   can be useful for introspecting TypedDict objects.  The static type
-  of ``d.get(e)`` should be the union of all possible item types in ``d``
-  if the string value of ``e`` cannot be determined statically.
-  (This simplifies to ``object`` if ``d`` is :term:`open`.)
+  of ``d.get(e)`` depends on what is known about the value of ``e``:
+
+  - If ``e`` is a string literal equal to a defined key of ``d``, the
+    return type is the declared type of that key combined with ``None``
+    for non-required keys.  For required keys, type checkers may return
+    either the declared type ``T`` or ``T | None``.
```
JelleZijlstra, line 765: Why leave this as an option? If the key is required, then it must be present so `.get()` can't return None.

docs/spec/typeddict.rst
```diff
@@ -757,9 +757,19 @@ This section discusses some specific operations in more detail.
...
+  - If ``e`` is a string literal that is **not** a defined key of ``d``,
+    no error should be reported.  The return type is the union of all
+    possible value types in ``d`` combined with ``None``.
+  - If the string value of ``e`` cannot be determined statically, the
+    return type is the union of all possible value types in ``d``
+    combined with ``None``.  (This simplifies to ``object | None`` if
```
JelleZijlstra, line 771: Which is just `object`.

PR #2319, opened 2026-07-10
Conformance: remove non-data-descriptor tests in dataclasses_descriptors.py

### LINKED ISSUE #2259: Conformance suite: Questionable tests in dataclasses_descriptors.py

ashishpatel26:
## Summary
- The `Desc2`/`DC2` test in `dataclasses_descriptors.py` asserted type-checker behavior for non-data descriptors (objects with `__get__` but no `__set__`) used as dataclass fields.
- Those assertions don't match runtime behavior: fields typed with a non-data descriptor are stored as plain instance attributes on init, so the descriptor's `__get__` never actually fires on instance access.
- Per the issue, pyrefly gets this right (and thus fails the test), while mypy/pyright pass the test but implement the wrong behavior.
- Removed the incorrect assertions; kept the `Desc1` (data descriptor) case, which is correctly specified and confirmed against runtime.
- Regenerated conformance results for `dataclasses_descriptors.toml` across all checkers and `results.html`; reconciled `conformant` flags that referenced the now-removed behavior.

Fixes #2259

## Test plan
- [x] `uv run python src/main.py` — reran full conformance suite, confirmed only `dataclasses_descriptors` results changed for the intended checkers
- [x] `uv run python src/validate_results.py` — 987 files validated, 0 invariant violations
- [x] Diff scoped to `conformance/tests/dataclasses_descriptors.py`, the 4 affected `results/*/dataclasses_descriptors.toml`, and `results.html`

### DIFF
--- a/conformance/results/mypy/dataclasses_descriptors.toml
+++ b/conformance/results/mypy/dataclasses_descriptors.toml
@@ -1,18 +1,6 @@
-conformant = "Partial"
-notes = """
-Assumes descriptor behavior only when field is assigned in class body.
-Does not correctly evaluate type of descriptor access.
-"""
+conformant = "Pass"
 output = """
-dataclasses_descriptors.py:61: error: Cannot access instance-only attribute "x" on class object  [misc]
-dataclasses_descriptors.py:62: error: Cannot access instance-only attribute "y" on class object  [misc]
-dataclasses_descriptors.py:66: error: Expression is of type "Desc2[int]", not "int"  [assert-type]
-dataclasses_descriptors.py:67: error: Expression is of type "Desc2[str]", not "str"  [assert-type]
 """
-conformance_automated = "Fail"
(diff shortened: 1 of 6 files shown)


davidhalter:
I'm closing this, since you already opened a pull request for the same issue: https://github.com/python/typing/pull/2299.

Please next time check at least the issue you're trying to fix for your *own* pull requests and other pull requests. Also please check for others who have tried to solve the issue. This is just wasting maintainer time.

PR #2320, opened 2026-07-11
Fix `pre-commit` hook `end-of-file-fixer` interfering with CI.

randolf-scholz:
Encountered this in https://github.com/python/typing/actions/runs/29147006108/job/86530507754:  `uv run --python 3.12 --frozen python src/main.py` produces `results.html` without trailing newline, which then gets added by  `end-of-file-fixer`, which then causes `Assert conformance results are up to date` CI to fail.

1. ~~excluded `conformance/results/results.html` from `end-of-file-fixer`~~
Set `keep_trailing_newline=True` in `jinja2.Environment` in `reporting.py`
2. ran `pre-commit run -a` which applied `end-of-file-fixer` and `trailing-whitespace` to some files.

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -5,7 +5,6 @@ repos:
       - id: trailing-whitespace
         exclude: conformance/results/.*/.*\.toml
       - id: end-of-file-fixer
-        exclude: conformance/results/results\.html
       - id: check-yaml
       - id: check-toml
       - id: check-merge-conflict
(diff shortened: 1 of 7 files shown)


srittau:
I think we should rather fix `base.html` to include a final end-of-line.

randolf-scholz:
@srittau `base.html` already had a trailing newline, but it was possible to fix by setting `keep_trailing_newline=True` in `jinja2.Environment`.

PR #2322, opened 2026-07-22
Recommend Never for return types as well

srittau:
Per discussion in python/cpython#132941

### DIFF
--- a/docs/guides/modernizing.rst
+++ b/docs/guides/modernizing.rst
@@ -249,18 +249,14 @@ return types. For example, replace::
     from typing import NoReturn
 
     def f(x: int, y: NoReturn) -> None: ...
+    def g() -> NoReturn: ...
 
 with::
 
     from typing import Never  # or typing_extensions.Never
 
     def f(x: int, y: Never) -> None: ...
-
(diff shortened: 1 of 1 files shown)


JelleZijlstra (approved):

davidhalter (approved):

ngnpope:
This change updated [this section](https://typing.python.org/en/latest/guides/modernizing.html#typing-noreturn), but there are other places that should probably also be reworded:

https://typing.python.org/en/latest/spec/special-types.html#never has the following:

> The `Never` type is equivalent to `NoReturn`, which is discussed above. The `NoReturn` type is conventionally used in return annotations of functions, and `Never` is typically used in other locations, but the two types are completely interchangeable.

https://typing.python.org/en/latest/spec/special-types.html#noreturn should probably also say that `Never` is preferred.

https://typing.python.org/en/latest/spec/constructors.html has some references to `NoReturn` that could be changed to `Never`?

On the https://typing.python.org/en/latest/spec/special-types.html page, perhaps `Never` should come above `NoReturn` and the examples for `NoReturn` should be moved to `Never`, with `NoReturn` being a simple statement saying that `Never` is preferred?

PR #2324, opened 2026-07-30
Conformance: broaden enum-expansion tests

### LINKED ISSUE #2323: enums_expansion.py test3 misses cases of the union-of-all-member = enum equivalency
The Enum Literal Expansion section states that [a complete union of all literal members is equivalent to the enum type](https://typing.python.org/en/latest/spec/enums.html#enum-literal-expansion). `test3` in `[conformance/tests/enums_expansion.py](https://github.com/python/typing/blob/main/conformance/tests/enums_expansion.py)` is meant to cover this, but doesn't catch some cases.

Both `pyright-1.1.411` and `zuban-0.9.0` reject the following.
```python
class Answer(enum.Enum):
    Yes = 1
    No = 2

a: Answer = ...
x: Literal[Answer.Yes, Answer.No] = a   # both report an error here
```

pyright: `Type "Answer" is not assignable to declared type "Literal[...]"`; zuban: `Incompatible types in assignment`.

Suggested test 

```python
def test4(a: Answer) -> None:
    x: Literal[Answer.Yes, Answer.No] = a  # E?
    assert_type(a, Literal[Answer.Yes, Answer.No])
```

This came up from the discussion at https://discuss.python.org/t/treat-literal-of-all-elements-of-enum-equal-to-the-enum-in-type-annotations/108317

tapetersen:
The spec's Enum Literal Expansion section states that a type checker
should treat a complete union of all literal members as equivalent to
the enum type. Add test to cover cases failing in some checkers.

Also refreshes all three quoted spec passages, which had drifted from
the section as rewritten in https://github.com/python/typing/commit/a6eeccf4870f95357c7cc37332e350d147966b35 ("new type system concepts section",

Fixes https://github.com/python/typing/issues/2323.

### DIFF
--- a/conformance/results/basilisk/enums_expansion.toml
+++ b/conformance/results/basilisk/enums_expansion.toml
@@ -1,8 +1,13 @@
-conformant = "Pass"
-conformance_automated = "Pass"
+conformant = "Partial"
+notes = """
+Does not accept an enum-typed value where a complete union of all literal members is declared, although it treats the two as equivalent for `assert_type`.
+"""
+conformance_automated = "Fail"
 errors_diff = """
+Line 85: Unexpected errors ['enums_expansion.py:85:5: error: Type mismatch: `x` is annotated `Literal[Answer.Yes, Answer.No]` (answer.yes | answer.no) but assigned answer [assignment_compatibility]']
 """
 output = """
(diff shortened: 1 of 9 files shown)


JelleZijlstra (approved):

carljm (approved):
Needs rebasing, but LGTM

tapetersen:
@carljm Rebased on main and thanks.

davidhalter:
I fixed this in Zuban as well, merging now, updating the tests later.

PR #2326, opened 2026-07-31
Add section about metaclass constructors

Viicos:


docs/spec/constructors.rst
```diff
@@ -488,3 +490,289 @@ callable.
...
+
+Keyword arguments in a direct metaclass call (such as the last two calls in the
+example above) require no special handling: they are validated as part of
+evaluating the call using the standard constructor call rules.
+
+Regardless of the evaluated return type of the implied metaclass call, a
+:keyword:`class` statement defines a class, and type checkers should evaluate the
+type of the bound name accordingly (``type[MyClass1]`` in the example above).
+The implied metaclass call is evaluated only for the purpose of validating
+its arguments.
```
Viicos, line 674: This doesn't represent the runtime behavior:

```python
class Meta(type):
    def __new__(cls, *args, **kwargs):
        return 1

class Class(metaclass=Meta): ...

Class
#> 1
```

But all type checkers currently model this. It is pretty unusual for something like this to happen, and type checkers may rely on the assumption that a `class` statements always results in a type object.

docs/spec/constructors.rst
```diff
@@ -488,3 +490,289 @@ callable.
...
+    class MyClass2(Base, flag=""):  # Type error: wrong type for "flag"
+        pass
+
+    class MyClass3(Base, other=1):  # Type error: Base.__init_subclass__() got an unexpected keyword argument 'other'
+        pass
+
+    class MyClass4(other=1):  # Type error: MyClass4.__init_subclass__() takes no keyword arguments
+        pass
+
+A metaclass :meth:`!__init__` method has no effect on this rule: when the
```
Viicos, line 741: Perhaps type checkers could also enforce consistency between `__init_subclass__()` and `__new__()`/`__init__()` in some way?

Viicos:
Apologies, I keep facing a Github bug opening PRs on the upstream repo. Meant to open this on my fork.

PR #2327, opened 2026-08-01
Update typeddicts_class_syntax.py test to use a custom metaclass.

rchen152:
The TypedDict chapter of the spec says (emphasis mine):
> It is invalid to specify a base class other than TypedDict, Generic, or another TypedDict type in a class-based TypedDict definition. **It is also invalid to specify a custom metaclass.**

It looks like most type checkers interpret this to mean that the appearance of `metaclass=...` is illegal in a TypedDict definition.

Pyrefly interprets this to mean that a non-`type` metaclass is illegal, so a redundant `metaclass=type` is fine.

IMO either interpretation is defensible, but the conformance test currently uses `metaclass=type`, which only accommodates the first. This PR changes the test to use a non-`type` metaclass, which is unambiguously an error.

This doesn't actually change pyrefly's conformance results because pyrefly has a bug that causes it to emit a different, nonsensical error on the `metaclass=type` line, which was mistakenly taken as pyrefly passing the test. I'm in the process of fixing that bug, which is how I discovered that pyrefly behaves slightly differently from the other checkers here.


davidhalter (approved):
This seems to map with runtime behavior:

```
>>> from typing import *
>>> class BadTypedDict2(TypedDict, metaclass=type): pass
... 
>>> class CustomMeta(type):
...     pass
... 
>>> class BadTypedDict2(TypedDict, metaclass=CustomMeta): pass
... 
TypeError: metaclass conflict: the metaclass of a derived class must be a (non-strict) subclass of the metaclasses of all its bases
```

JelleZijlstra (approved):

PR #2316, opened 2026-07-05
Add Basilisk to conformance results

MelbourneDeveloper:
Adds Basilisk to the typing conformance report.

Basilisk is published as `basilisk-python` on PyPI and installs a `basilisk` console script, matching the conformance suite's locked `uv` dependency model.

Local verification run from this branch:

```bash
cd conformance
uv sync --python 3.12 --frozen
uv run --python 3.12 --frozen python src/main.py
uv run --python 3.12 --frozen python src/validate_results.py
```

The run validated 987 conformance result files with no invariant violations and left `conformance/results` clean. Basilisk reports 141/141 manual Pass, automated Pass, and zero `errors_diff` entries with `basilisk-python==0.27.0`.

PR #2330, opened 2026-08-05
Remove Basilisk from conformance results

MelbourneDeveloper:
Reverts #2316. I'm the author of Basilisk, and I'm asking for it to be taken out of the conformance results. Our published score doesn't measure what the table says it measures, so it needs to be removed.

## Why

Many of Basilisk's rules match against raw source text and hard-coded typing symbol names instead of resolved symbols on the AST. The result is that our score reflected the exact spelling of the test files rather than conformance to the specification.

A [mutation harness](https://gist.github.com/sharkdp/3f3266fd9c67d22137e2b6c015c5f206) makes this unambiguous. It renames imported typing symbols and adjusts whitespace. It should not move any checker's score. Ours moves. I reproduced this myself and 113 of 141 tests are affected. We will investigate exactly how much work is necessary to fix this.

## How it happened

Despite our best efforts to enforce accuracy, we relied too heavily on AI to validate the work we did, and we focused too heavily on conformance results instead of thoroughly testing real world scenarios. It looks as though we didn't pick up code that clearly should have only been just a placeholder and that code proliferated. This should have been predictable and I take full responsibility for letting this happen. 

I am sorry for breaking trust, especially when the whole aim of Basilisk as a tool is to build trust. We have an absolute commitment to building the trust of the Python community so requesting this removal is a necessary step until we can work through the issues. 

## Resubmission

I'm asking that you please leave the door open to make another attempt in future. We still have full confidence that Basilisk is a worthy type checker.

Before we submit again, we'll hold ourselves to the test that proved the issues. If a rename or a whitespace change moves our score, the rule isn't implemented and the number doesn't count. We'll publish that measurement alongside any resubmission, and if the honest score is lower, we'll publish the lower score.

We will do a full audit to remove any code that pattern-matches source text. We will track it publicly in our repo. We will take as long as it takes to get this right and not submit again until we've got proof that we're not making the same mistakes again.

Sincere apologies to the maintainers here. You accepted our submission in good faith and our carelessness led to misleading results being posted here.


AlexWaygood (approved):
Thank you! I really appreciate the honesty and self-reflection here.

> I'm asking that you please leave the door open to make another attempt in future. We still have full confidence that Basilisk is a worthy type checker.

I can't speak for the other maintainers of this repo, but I certainly would have no problem with this, assuming the underlying issues here are addressed

DetachHead:
honestly i don't understand why a 100% vibecoded type checker was even treated as a serious project in the first place. and since even this apology looks AI generated i don't see why the door should be left open if there doesn't even appear to be any remorse from a human at all. is the developer even aware that this PR was made?

carljm:
@DetachHead (my personal opinion, not speaking for the typing council)

a) The PR to add Basilisk was accepted in the first place because this situation had never happened before, and thus we didn't have any existing guidelines around what qualifies a tool to be added to the conformance suite, beyond "it can run the conformance suite and someone submitted a PR to add it". This situation has made it clear that we need those guidelines and a bit more due diligence before we accept such pull requests in future.

b) IMO the whole "door open" thing is a bit meaningless / a given. I guess we _could_ choose to ban this author from ever submitting something to the conformance suite again, but IMO that would be an over-reaction. We are going to be more careful in future about adding any new type checker, and that certainly includes (probably to an even greater degree) any future re-submission of Basilisk.

yangdanny97:
> the whole aim of Basilisk as a tool is to build trust

Without participating in the typing community and understanding the problems people are trying to solve, you tell AI to loop until the conformance tests are passing then go around bragging and talking smack about other tools on social media.

Is that what you call trust-building? 🤡

PR #2335, opened 2026-08-09
Add conformance tests to disallow ClassVar/Final in NamedTuple and TypedDict

yangdanny97:
The spec for `Final` says:

> Final cannot be used as a qualifier for a [TypedDict](https://typing.python.org/en/latest/spec/typeddict.html#typeddict) item or a [NamedTuple](https://typing.python.org/en/latest/spec/namedtuples.html#namedtuple) field. Such usage also generates an error at runtime.

https://typing.python.org/en/latest/spec/qualifiers.html#semantics-and-examples

The spec for `ClassVar` says

> ClassVar cannot be used as a qualifier for a [TypedDict](https://typing.python.org/en/latest/spec/typeddict.html#typeddict) item or a [NamedTuple](https://typing.python.org/en/latest/spec/namedtuples.html#namedtuple) field. Such usage also generates an error at runtime.

https://typing.python.org/en/latest/spec/class-compat.html#classvar

I believe neither rule is currently exercised in the conformance tests. This PR adds tests exercising those rules and regenerates the conformance results.

This should be an easy one to review, the diff is just big because the new imports change the line numbers of a bunch of diagnostics


yangdanny97:
Every type checker implements the rule already, except for Pycroscope which seems to allow ClassVar in NamedTuple

jorenham (approved):
Makes sense; thanks!

JelleZijlstra (approved):

rchen152 (approved):

PR #2338, opened 2026-08-11
Update unpack TypedDict kwargs forwarding spec + add conformance tests

yangdanny97:
Previously, this section of the spec was not exercised at all in the conformance tests, and also was not implemented by any type checker.

This PR updates the spec to account for closed & extra_items TypedDicts, and adds conformance tests.

There are 3 asserted errors:
- open TypedDict kwargs being unpacked into function w/o kwargs (no type checker implements this)
- extra_items TypedDict kwargs being unpacked into function w/o kwargs (only ty implements this)
- extra_items TypedDict kwargs being unpacked into function w/ incompatibly-typed kwargs (ty, pyright, and zuban implement this)

This would supersede https://github.com/python/typing/pull/1960, which proposed that we delete the section of the spec entirely. 
@rchen152's comment in https://discuss.python.org/t/typing-spec-inconsistency-for-unpacking-typed-dict-kwargs/79640 favors deletion.


carljm:
This issue is nuanced, because many unpacking behaviors in Python are unsafe. If you unpack a `list[int]` into a function that takes `(int, int)` (but not `*args: int`), every type checker allows that, even though the list could obviously contain fewer or more than two elements. The collective decision seems to be that since length is not part of what the `list` type encodes, we trust the user rather than erroring.

In ty we made the decision not to error on the first case here, because we saw large mypy-primer fallout if we enforced this rule, and we felt it was analogous to the case of `list[int]` -- an open `TypedDict` _may_ contain extra elements (just like a list _may_ contain any number of elements), but the type doesn't explicitly say that it does, so we can be forgiving here.

We decided to be stricter with `extra_items`, in part because there is no widespread usage yet, so usage will adapt to what type checkers permit, and in part because in this case the user has explicitly declared their intent for the `TypedDict` to contain extra items, so it feels reasonable to account for that more strictly. This does mean that implicitly open typed dicts are not equivalent to `extra_items=ReadOnly[object]`, which maybe in principle they would be -- but I think this is currently true in all type checkers.

I think ty's compromise is reasonable, but I'm not totally convinced it's the best option, open to other resolutions here. But enforcing this rule on all implicitly open TypedDicts will have a significant impact on existing real-world code.

yangdanny97:
For Pyrefly my current thought is that this would be off-by-default and enabled in strict mode.

carljm:
That seems reasonable. I'm not sure if behavior that we wouldn't turn on by default in our own type checkers should be encoded as a conformance suite requirement, though?

It seems to me that a strict mode which requires safety in unpacking an open TypedDict should probably also require safety in unpacking a list/Sequence?

yangdanny97:
So do you think that we should just merge #1960 and delete this section from the typing spec entirely?

Or weaken the first assertion to "may" and "E?"

carljm:
I would probably land this with the first assertion weakened (and discussion of that in the spec also), but that's not a strong preference, it just matches what we decided to do in ty, so naturally it already makes sense to me :) Very open to other opinions here.

yangdanny97:
Does it make sense to differentiate between implicitly open and explicitly open (`closed=False`), by making errors mandatory for the latter?

carljm:
> Does it make sense to differentiate between implicitly open and explicitly open (`closed=False`), by making errors mandatory for the latter?

It might. Currently in ty we only apply the extra strictness with `extra_items=`, not with `closed=False`. I think the appeal of this is that keeps `closed` itself boolean rather than tri-valued (that is, `closed=False` is the default and providing it explicitly makes no difference), but `closed=False` alone is not the same thing as explicitly providing an `extra_items` type -- the latter more explicitly says "I expect there to be extra items in this dict of a particular type". But all of this is a judgment call, there's no clear right or wrong answer. Or rather, the "right" answer in the abstract is probably the universal strictness you initially proposed, it's just a question of whether that's workable in the ecosystem.

JelleZijlstra:
The current TypedDict spec says that `closed=False` means the TypedDict is open, i.e. consistent with the default. I think we should stick with the TypedDict spec chapter's clear division of TypedDicts into states, and allow unsound behavior only if the TypedDict is open (as defined by the spec; the spec change in this PR should link to the glossary definition for the term).

In the abstract I'd prefer to treat open TypedDicts similar to ones with `extra_items=ReadOnly[object]`, but I agree that's probably a bridge too far right now.

docs/spec/callables.rst
```diff
@@ -413,15 +413,13 @@ it needs completely ignoring any additional values.
 The calls to ``bar`` and ``spam`` will fail because an unexpected keyword
 argument will be passed to the ``takes_name`` function.
 
-Therefore, ``kwargs`` hinted with an unpacked ``TypedDict`` can only be passed
-to another function if the function to which unpacked kwargs are being passed
-to has ``**kwargs`` in its signature as well, because then additional keywords
-would not cause errors at runtime during function invocation. Otherwise, the
-type checker should generate an error.
+Therefore, it is only safe to pass ``kwargs`` hinted with an unpacked, non-:term:`closed` ``TypedDict``
+to another function if that function has ``**kwargs`` in its signature as well. Type checkers should
```
carljm: Should we also explicitly require possible extra-item values to be assignable to the destination's `**kwargs` annotation? Having `**kwargs` alone isn't sufficient:

```python
class Open(TypedDict):
    name: str

def target(name: str, **kwargs: int) -> None: ...

def forward(**kwargs: Unpack[Open]) -> None:
    target(**kwargs)  # Hidden extra items have type object, not int.
```

Both ty and pyright reject this, while zuban currently accepts it. I think the optional-error compromise should apply to unexpected keyword names, not to values that are incompatible with an explicitly annotated parameter. Could we spell out that distinction and cover forwarding an open TypedDict into typed `**kwargs`?

conformance/results/pyright/callables_kwargs.toml
```diff
@@ -1,36 +1,43 @@
-conformant = "Pass"
+conformant = "Partial"
+notes = """
+Does not reject passing unpacked kwargs typed with a non-closed TypedDict to a callable that has no `**kwargs`.
```
carljm: Should we remove this note here and from the other checker results? The updated spec explicitly allows an open TypedDict to be forwarded without an error, and the corresponding test is marked `E?`, so accepting that case isn't a conformance failure. (The separate note about explicitly declared `extra_items` still makes sense.)

docs/spec/callables.rst
```diff
@@ -413,15 +413,13 @@ it needs completely ignoring any additional values.
...
+error if the ``TypedDict`` has :term:`extra items`, and may error if the ``TypedDict`` is :term:`open`.
 
 In cases similar to the ``bar`` function above the problem could be worked
-around by explicitly dereferencing desired fields and using them as arguments
-to perform the function call::
+around by marking ``Animal`` with ``closed=True``, or by explicitly dereferencing desired
```
carljm, line 433: I think the same closed-TypedDict distinction also affects the earlier callable-assignment rule in this chapter, which currently says a callable without `**kwargs` can never be assigned to a destination with `**kwargs: Unpack[TypedDict]`. For example:

```python
class Closed(TypedDict, closed=True):
    name: str

class AcceptsClosed(Protocol):
    def __call__(self, **kwargs: Unpack[Closed]) -> None: ...

def takes_name(*, name: str) -> None: ...

callback: AcceptsClosed = takes_name
```

Both ty and pyright accept this assignment, and the hidden-key rationale doesn't apply because `Closed` cannot contain extra keys. Would it make sense to update that earlier rule and add closed-assignment coverage while we're clarifying this? `extra_items=Never` should behave the same way.

conformance/tests/callables_kwargs.py
```diff
@@ -132,3 +133,54 @@ def func7(*, v1: int, v3: str, v2: str = "") -> None:
...
+def takes_name_kwargs(name: str, **kwargs) -> None:
+    ...
+
+
+# > It is only safe to pass ``kwargs`` hinted with an unpacked, non-closed ``TypedDict``
+# > to another function if that function has ``**kwargs`` in its signature as well. Type checkers
+# > should error if the ``TypedDict`` has ``extra_items``, and may error if the ``TypedDict`` is open.
+
+def func8(**kwargs: Unpack[TD3]) -> None:
+    takes_name(**kwargs)  # E?: a subtype may contain unknown keys
```
carljm: Should we also cover directly unpacking ordinary TypedDict-annotated parameters or locals, not just a function's own `**kwargs`? The existing spec already demonstrates this with `baz(animal: Animal)`, and the same rules should apply regardless of where the TypedDict came from:

```python
def forward_direct(value: TD5) -> None:
    takes_name(**value)  # E: extra items may be present
    takes_name_str_kwargs(**value)  # E: int is incompatible with str
```

It would be useful to include the corresponding open and closed cases too, so a checker can't pass by special-casing `**kwargs` provenance.

conformance/tests/callables_kwargs.py
```diff
@@ -132,3 +133,54 @@ def func7(*, v1: int, v3: str, v2: str = "") -> None:
...
+
+
+# Specification: https://typing.readthedocs.io/en/latest/spec/callables.html#passing-kwargs-inside-a-function-to-another-function
+
+
+class TD3(TypedDict):
+    name: str
+
+
+class TD4(TypedDict, closed=True):
```
carljm: Could we keep an explicitly open `TypedDict(..., closed=False)` case alongside the implicitly open one, and also add `extra_items=Never` alongside `closed=True`? The discussion specifically settled that `closed=False` should behave like the default, and the TypedDict spec defines `extra_items=Never` as equivalent to `closed=True`. Without both spellings, a checker can incorrectly distinguish them and still pass these tests.

conformance/results/mypy/callables_kwargs.toml
```diff
@@ -1,29 +1,39 @@
 conformant = "Partial"
 notes = """
 Allows callable without kwargs to be assigned to callable with unpacked kwargs.
+Does not support the `closed` and `extra_items` TypedDict class arguments.
+Does not reject passing unpacked kwargs typed with a non-closed TypedDict to a callable that has no `**kwargs`.
```
carljm: Remove this note? Accepting an open TypedDict here is explicitly permitted by the updated spec, and the corresponding test is marked `E?`. The separate `extra_items` note is the actual conformance issue.

conformance/results/pycroscope/callables_kwargs.toml
```diff
@@ -1,19 +1,27 @@
-conformance_automated = "Pass"
+conformant = "Partial"
+notes = """
+Does not reject passing unpacked kwargs typed with a non-closed TypedDict to a callable that has no `**kwargs`.
```
carljm: Remove

conformance/results/pyrefly/callables_kwargs.toml
```diff
@@ -1,23 +1,30 @@
-conformant = "Pass"
-conformance_automated = "Pass"
+conformant = "Partial"
+notes = """
+Does not reject passing unpacked kwargs typed with a non-closed TypedDict to a callable that has no `**kwargs`.
```
carljm: Remove

conformance/results/zuban/callables_kwargs.toml
```diff
@@ -1,25 +1,32 @@
-conformance_automated = "Pass"
+conformant = "Partial"
+notes = """
+Does not reject passing unpacked kwargs typed with a non-closed TypedDict to a callable that has no `**kwargs`.
```
carljm: Remove

docs/spec/callables.rst
```diff
@@ -413,15 +413,13 @@ it needs completely ignoring any additional values.
 The calls to ``bar`` and ``spam`` will fail because an unexpected keyword
 argument will be passed to the ``takes_name`` function.
 
-Therefore, ``kwargs`` hinted with an unpacked ``TypedDict`` can only be passed
-to another function if the function to which unpacked kwargs are being passed
-to has ``**kwargs`` in its signature as well, because then additional keywords
-would not cause errors at runtime during function invocation. Otherwise, the
-type checker should generate an error.
+Therefore, it is only safe to pass ``kwargs`` hinted with an unpacked, non-:term:`closed` ``TypedDict``
```
carljm: This rule shouldn't be specific to forwarding `kwargs` -- but I guess that would imply a larger rewrite of this whole section. (Which I think we should do, but maybe not in this PR.) The existing text is way too focused on specific cases rather than outlining the general principles.

carljm (commented):
Thank you!

docs/spec/callables.rst
yangdanny97: I can reword this part a little bit in this PR, but let's save a bigger rewrite for later

docs/spec/callables.rst
yangdanny97: Yep, I can add that and add some tests

docs/spec/callables.rst
```diff
@@ -304,9 +304,9 @@ only::
 
 The reverse situation where the destination callable contains
 ``**kwargs: Unpack[TypedDict]`` and the source callable doesn't contain
-``**kwargs`` should be disallowed. This is because, we cannot be sure that
-additional keyword arguments are not being passed in when an instance of a
-subclass had been assigned to a variable with a base class type and then
+``**kwargs`` should be disallowed, if the TypedDict is :term:`closed`. This is because
```
carljm: I think this condition is backwards: a closed `TypedDict` cannot contain hidden keys, so assigning a callable without `**kwargs` is safe in precisely that case. The new tests also accept both `closed=True` and `extra_items=Never`, while rejecting the open case.

Should this say "if the TypedDict is not closed" instead? The copied specification quotation in the test would need the same fix.

conformance/tests/callables_kwargs.py
```diff
@@ -132,3 +152,94 @@ def func7(*, v1: int, v3: str, v2: str = "") -> None:
...
+    takes_name_str_kwargs(**value)
+    takes_name(**kwargs)
+    takes_name_kwargs(**kwargs)
+    takes_name_str_kwargs(**kwargs)
+
+
+def kwargs_extra_items(value: TDExtraItems, **kwargs: Unpack[TDExtraItems]) -> None:
+    takes_name(**value)  # E: extra items may be present
+    takes_name_str_kwargs(**value)  # E: extra items type is not compatible
+    takes_name_int_kwargs(**value)
```
carljm, line 266: Should we also cover the case where an extra item binds an explicitly declared optional parameter before it reaches `**kwargs`?

```python
class Extras(TypedDict, extra_items=int):
    name: str

def target(name: str, *, label: str = "", **kwargs: int) -> None: ...

def forward(value: Extras) -> None:
    target(**value)  # Extras could supply label=1.
```

Matching `extra_items=int` just against `**kwargs: int` is not sufficient because `label` expects `str`.

docs/spec/callables.rst
```diff
@@ -413,15 +413,14 @@ it needs completely ignoring any additional values.
...
 
-Therefore, ``kwargs`` hinted with an unpacked ``TypedDict`` can only be passed
-to another function if the function to which unpacked kwargs are being passed
-to has ``**kwargs`` in its signature as well, because then additional keywords
-would not cause errors at runtime during function invocation. Otherwise, the
-type checker should generate an error.
+Therefore, it is only safe to unpack a non-:term:`closed` TypedDict in a function call
+if that function has ``**kwargs`` in its signature, and any :term:`extra items` are assignable to the type of ``**kwargs``.
+If the function being called has ``**kwargs``, checkers should error if the TypedDict's extra items are not assignable to the type of ``**kwargs``.
+If the function being called does not have ``**kwargs``, checkers may error if the TypedDict is :term:`open`.
```
carljm: Should we explicitly state what happens when a `TypedDict` has declared, non-`Never` `extra_items` and the destination does not have `**kwargs`?

This paragraph currently requires an error for incompatible annotated `**kwargs` and permits an error for an open `TypedDict` without destination `**kwargs`, but it never says the declared-extra-items/no-`**kwargs` combination must error. The new conformance tests do require that error and downgrade several checkers for omitting it, so I think we need the corresponding normative requirement here.

docs/spec/callables.rst
```diff
@@ -413,15 +413,14 @@ it needs completely ignoring any additional values.
...
 argument will be passed to the ``takes_name`` function.
 
-Therefore, ``kwargs`` hinted with an unpacked ``TypedDict`` can only be passed
-to another function if the function to which unpacked kwargs are being passed
-to has ``**kwargs`` in its signature as well, because then additional keywords
-would not cause errors at runtime during function invocation. Otherwise, the
-type checker should generate an error.
+Therefore, it is only safe to unpack a non-:term:`closed` TypedDict in a function call
+if that function has ``**kwargs`` in its signature, and any :term:`extra items` are assignable to the type of ``**kwargs``.
+If the function being called has ``**kwargs``, checkers should error if the TypedDict's extra items are not assignable to the type of ``**kwargs``.
```
carljm: Does the reference to `extra items` here formally cover an open `TypedDict`? The glossary and TypedDict chapter define "open" and "has extra items" as distinct states, but the new tests require the possible additional values of an open `TypedDict` to be checked as `object` against annotated `**kwargs`.

Could we say explicitly that, for this rule, an open `TypedDict` is treated as having read-only additional items of type `object`, while a `TypedDict` with `extra_items=` uses its declared extra-item type? Otherwise the mandatory open-`TypedDict` errors in the tests do not follow from the defined terminology.

conformance/tests/callables_kwargs.py
```diff
@@ -132,3 +152,94 @@ def func7(*, v1: int, v3: str, v2: str = "") -> None:
...
+
+def takes_name_kwargs(name: str, **kwargs) -> None:
+    ...
+
+
+def takes_name_str_kwargs(name: str, **kwargs: str) -> None:
+    ...
+
+
+def takes_name_int_kwargs(name: str, **kwargs: int) -> None:
```
carljm, line 202: Would it be useful to also cover destinations whose `**kwargs` are themselves annotated with `Unpack[TypedDict]`?

```python
def takes_closed(**kwargs: Unpack[TDClosed]) -> None: ...

def forward(value: TDExtraItems) -> None:
    takes_closed(**value)  # Extra keys are not allowed by TDClosed.
```

Also unpacking e.g. an open `TypedDict` into a destination with `Unpack` of a `TypedDict` with `extra_items=int`?

All of the destination helpers currently use untyped or homogeneous `**kwargs`.

docs/spec/callables.rst
```diff
@@ -366,8 +366,8 @@ traditionally typed ``**kwargs`` aren't checked for keyword names.
 To summarize, function parameters should behave contravariantly and function
 return types should behave covariantly.
 
-Passing kwargs inside a function to another function
-^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
+Unpacking a TypedDict as keyword arguments
```
carljm, line 369: The new conformance test still links to `#passing-kwargs-inside-a-function-to-another-function`, but this heading will generate `#unpacking-a-typeddict-as-keyword-arguments` instead.

carljm (approved):
I think there are a couple details here in the spec (one backwards statement, one broken link) that should be addressed -- everything else here is up to you whether you want to include it in this PR. Otherwise this looks ready to me.

I'm glad we are clearing this up now, before wider adoption of closed/extra_items TypedDicts.

docs/spec/callables.rst
yangdanny97: yeah I can make that explicit

carljm (approved):
I'm happy with this.

I guess since it touches the spec and the conformance suite, we need a Discuss thread, the one-week wait, and an issue for Typing Council sign-off?

yangdanny97:
https://discuss.python.org/t/spec-change-proposal-updating-clarifying-rules-for-unpacking-typeddicts-in-function-calls/108582

https://github.com/python/typing-council/issues/64

rchen152 (approved):

davidhalter:
Sorry for taking such a long time here. I have tried to play with this quite a bit in Zuban to get a feel for it and I'm favorable to merging this PR. There have been two potential cases that I would like to add to the spec (please let me know if you disagree):

```
from typing import TypedDict
from typing_extensions import Unpack

class Person(TypedDict):
    name: str
    age: int

class Base:
    def foo(self, **kwargs: Unpack[Person]) -> None: ...
class SubGood(Base):
    def foo(self, *, name: str, age: int, extra: bool = False) -> None: ...   # Should probably error here?!
```

This is currently not an error in Zuban/Mypy (test is called `testUnpackKwargsOverrides` in Mypy), but likely should be. I want to avoid type checkers using different logic for normal callable subtyping and inheritance checking.

Another case of callable subtyping that I looked at was this one that might be worth to add:

```
from typing import Never, Unpack, TypedDict

class TDBase(TypedDict):
    x: int

class TD(TDBase, total=False):
    y: int

def a1(**kwargs: Unpack[TDBase]) -> None: .
def a2(**kwargs: Unpack[TD]) -> None: ...

if bool():
    a1 = a2  # (1) This should error
else:
    a2 = a1  # (2)
```

Before my changes to think of non-closed TypedDicts in Unpacks as essentially having `extra_items=object`, (2) had an error, now it's the other way around (1) errors. This is definitely better, it was buggy before. It feels like this isn't really part of the tests yet and I think it's central to this view of TypedDicts.

(I'm aware that this might not be an issue in other type checker, because it just widens the type. But you can also create this case with Protocols like `a1: <some protocol that mimics a1> = a2`)

----

I'm also open to the make some of the `# E?` to `# E`, but I don't have a strong opinion here, especially after Carl's discovery that the impact is not as bad.

yangdanny97:
Okay, based on the dsicussion in the forum thread, I've updated the docs and tests to require the error.

yangdanny97:
I haven't had time to look into @davidhalter's examples yet, will do so when i hvae time

docs/spec/callables.rst
```diff
@@ -413,15 +413,15 @@ it needs completely ignoring any additional values.
...
 
-Therefore, ``kwargs`` hinted with an unpacked ``TypedDict`` can only be passed
-to another function if the function to which unpacked kwargs are being passed
-to has ``**kwargs`` in its signature as well, because then additional keywords
-would not cause errors at runtime during function invocation. Otherwise, the
-type checker should generate an error.
+Therefore, it is only safe to unpack a non-:term:`closed` TypedDict in a function call
+if that function has ``**kwargs`` in its signature, and any :term:`extra items` are assignable to the type of ``**kwargs``.
+
+- If the function being called has ``**kwargs``, checkers should error if the TypedDict's :term:`extra items` are not assignable to the type of ``**kwargs``. For this rule, :term:`open` TypedDicts are treated as having extra items of type ``object``.
```
carljm: Could we spell out how this rule applies when the destination's `**kwargs` is itself annotated with `Unpack[TD]`? It's not clear what "the type of `**kwargs`" means for that comparison.

The `foo(**animal)` example accepts unknown extra items, whereas the new `takes_closed` tests reject them, even though both destinations syntactically have `**kwargs`. I think we want an open destination to accept extras of type `object`, a destination with `extra_items=T` to check them against `T`, and a closed destination to reject extra items.

Would it make sense to state that explicitly here? Otherwise the conformance requirements for unpacked destination `**kwargs` depend on an unstated interpretation.

conformance/results/pycroscope/callables_kwargs.toml
```diff
@@ -1,19 +1,48 @@
-conformance_automated = "Pass"
+conformant = "Partial"
+notes = """
+Does not check the type of an unpacked TypedDict's extra items against the target callable's `**kwargs` annotation.
```
carljm: I think need to update these notes to match the regenerated output.

The recorded diagnostics show that these checks now work: incompatible extra-item values are rejected, extra items are checked against optional named parameters, declared extras are rejected when the destination cannot accept them, and the closed/`extra_items=Never` callable assignments are accepted.

The only remaining misses in this test are the eight open-source unpacking diagnostics listed in `errors_diff`. I think we can remove the other claimed failures here.

conformance/tests/callables_kwargs.py
```diff
@@ -123,12 +143,121 @@ def func6(**kwargs: Unpack[T]) -> None:  # E: unpacked value must be a TypedDict
...
+def takes_closed(**kwargs: Unpack[TDClosed]) -> None:
+    ...
+
+
+# > It is only safe to unpack a non-:term:`closed` TypedDict in a function call if that function has
+# > ``**kwargs`` in its signature, and any :term:`extra items` are assignable to the type of ``**kwargs``.
+
+def open_implicit(value: TDOpenImplicit, **kwargs: Unpack[TDOpenImplicit]) -> None:
+    takes_name(**value)  # E: a subtype may contain unknown keys
+    takes_name_kwargs(**value)
```
carljm, line 224: Would it be useful to add a successful `**kwargs: object` case for an open source? Currently, the open-source cases only succeed with unannotated `**kwargs` and fail with `**kwargs: str`, so a checker that rejects every explicitly typed destination could still pass these additions.

For example:

```python
def takes_name_object_kwargs(name: str, **kwargs: object) -> None: ...

def forward(value: TDOpenImplicit) -> None:
    takes_name_object_kwargs(**value)  # OK
```

A destination using `Unpack` of a TypedDict with `extra_items=object` would also cover successful destination expansion. We could call it with both an open source and an `extra_items=int` source.

docs/spec/callables.rst
carljm: Small nit: could we preserve the previous section anchor with an explicit RST label before this heading?

```rst
.. _passing-kwargs-inside-a-function-to-another-function:
```

The rendered HTML now only has the new anchor, so existing links and bookmarks to the old section will no longer navigate to it.

conformance/results/zuban/callables_kwargs.toml
```diff
@@ -1,25 +1,52 @@
-conformance_automated = "Pass"
+conformant = "Partial"
+notes = """
+Does not reject unpacking an open TypedDict in a call to a callable that has no `**kwargs`.
+Does not check the type of an open TypedDict's extra items against the target callable's `**kwargs` annotation.
+Does not reject unpacking a TypedDict that sets `extra_items` in a call to a callable that has no `**kwargs`.
+Incorrectly rejects assigning a callable without `**kwargs` to a callable whose `**kwargs` is typed with a closed TypedDict or one that sets `extra_items=Never`.
```
carljm: Similar to pycroscope, I think these notes no longer match the actual results.

carljm (commented):
Looks good! A few things I noticed inline.

conformance/results/pycroscope/callables_kwargs.toml
JelleZijlstra: Yes, I implemented the earlier permissive version of this PR

conformance/tests/callables_kwargs.py
yangdanny97: Done

docs/spec/callables.rst
yangdanny97: This is what I changed it to:

Therefore, it is only safe to unpack a non-:term:`closed` TypedDict in a function call
if that function has ``**kwargs`` in its signature, and any :term:`extra items` are assignable to the type of ``**kwargs``.
For these rules, :term:`open` TypedDicts are treated as having :term:`extra items` of
type ``object``.

- If the callee has non-unpacked ``**kwargs``, checkers should error if the TypedDict's
  :term:`extra items` are not assignable to the type of ``**kwargs``.
- If the callee has unpacked ``**kwargs`` (``**kwargs: Unpack[TD]``), checkers should
  error if the TypedDict's :term:`extra items` are not assignable to ``TD``'s extra
  items, treating a :term:`closed` ``TD`` as accepting no extra items.
- If the callee does not have ``**kwargs``, checkers should error if the TypedDict is
  :term:`open`, and should error if the TypedDict declares non-``Never``
  :term:`extra items`.

Additionally, the type of :term:`extra items` must be assignable to any keyword parameters the function call does not provide.

carljm (approved):
Looks great!

PR #2343, opened 2026-08-14
Remove class-object assertion from defaults test

JelleZijlstra:
#2204 rewrote many defaults tests because type checkers are not required to infer a class expression as having a `type[C]` type, but the assertion for `slice` in `generics_defaults_referential.py` was missed. The same rationale applies here. I simply removed the line because the next line with `slice()` adequately tests the defaults.

### DIFF
--- a/conformance/results/mypy/generics_defaults_referential.toml
+++ b/conformance/results/mypy/generics_defaults_referential.toml
@@ -3,20 +3,18 @@ notes = """
 Does not correctly handle defaults referencing other `TypeVar`s.
 """
 output = """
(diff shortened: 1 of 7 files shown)


AlexWaygood (approved):
I believe the reason why I didn't change this one in https://github.com/python/typing/pull/2204 was that we emit `assert-type-unspellable-subtype` for this one in ty instead of `type-assertion-failure` (since both the asserted type and the inferred type are fully static types), and we just switch off `assert-type-unspellable-subtype` when running ty on the conformance suite

JelleZijlstra:
> we just switch off assert-type-unspellable-subtype when running ty on the conformance suite

I feel like we shouldn't be doing that, it weakens the `assert_type` tests. But I guess we can gradually burn down the cases where we need it.

PR #2344, opened 2026-08-17
Add section about metaclass constructors

Viicos:
Discussion thread: https://discuss.python.org/t/108357

This PR adds a new section about metaclass constructors, following the existing section about class constructors. Currently opened in draft, as I'm looking to first finalize the spec so that I can update conformance tests (`metaclass-constructors-checkers.md` is a temporary document to get a sense of current support in type checkers).

### DIFF
--- a/docs/spec/constructors.rst
+++ b/docs/spec/constructors.rst
@@ -3,6 +3,8 @@ Constructors
 
 Calls to constructors require special handling within type checkers.
 
+.. _`constructor-calls`:
+
 Constructor Calls
 -----------------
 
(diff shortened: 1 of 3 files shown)


docs/spec/constructors.rst
```diff
@@ -488,3 +490,289 @@ callable.
...
+
+Keyword arguments in a direct metaclass call (such as the last two calls in the
+example above) require no special handling: they are validated as part of
+evaluating the call using the standard constructor call rules.
+
+Regardless of the evaluated return type of the implied metaclass call, a
+:keyword:`class` statement defines a class, and type checkers should evaluate the
+type of the bound name accordingly (``type[MyClass1]`` in the example above).
+The implied metaclass call is evaluated only for the purpose of validating
+its arguments.
```
Viicos: This doesn't represent the runtime behavior:

```python
class Meta(type):
    def __new__(cls, *args, **kwargs):
        return 1

class Class(metaclass=Meta): ...

Class
#> 1
```

But all type checkers currently model this. It is pretty unusual for something like this to happen, and type checkers may rely on the assumption that a `class` statements always results in a type object.

Viicos: @jorenham [commented](https://github.com/Viicos/typing/pull/1#discussion_r3693709300):

Returning a non-`type` indeed seems a bit silly, but perhaps there are genuine use-cases, so that might be worth investigating. 

Either way, I think it could help to make it more clear whether it is allowed to return classes other than `cls` from `__new__`, for example:

```
In [1]: class Meta(type):
   ...:     def __new__(cls, *a, **kw):
   ...:         return int
   ...: class PhantomInt(metaclass=Meta): ...

In [2]: PhantomInt
Out[2]: int
```

Viicos: > Returning a non-`type` indeed seems a bit silly, but perhaps there are genuine use-cases, so that might be worth investigating.

As type checkers should generally mirror runtime behavior as much as possible, it would make sense, but I'd like to defer to type checker authors as it may not be possible to model this easily.

> Either way, I think it could help to make it more clear whether it is allowed to return classes other than `cls` from `__new__`, for example:

The spec covers this case by stating:

> In both cases, the metaclass call should be evaluated using the same rules described in the sections above.

In the section above. But I agree an example would be great to make it explicit (or we just incorporate this as a conformance test).

docs/spec/constructors.rst
```diff
@@ -488,3 +490,289 @@ callable.
...
+    class MyClass2(Base, flag=""):  # Type error: wrong type for "flag"
+        pass
+
+    class MyClass3(Base, other=1):  # Type error: Base.__init_subclass__() got an unexpected keyword argument 'other'
+        pass
+
+    class MyClass4(other=1):  # Type error: MyClass4.__init_subclass__() takes no keyword arguments
+        pass
+
+A metaclass :meth:`!__init__` method has no effect on this rule: when the
```
Viicos, line 747: Perhaps type checkers could also enforce consistency between `__init_subclass__()` and `__new__()`/`__init__()` in some way?

metaclass-constructors-checkers.md
```diff
@@ -0,0 +1,356 @@
...
+from typing import Any
+
+
+class Meta(type):
+    # all four checkers report an override-incompatibility error at this
+    # definition (parameter "**kwds" missing vs. type.__prepare__); that check
+    # is unrelated to validating the implied __prepare__ call below
+    @classmethod
+    def __prepare__(mcls, name: str, bases: tuple[type, ...]):  # No **kwds
+        return {}
```
Viicos, line 203: Should it be exempt from the LSP?

Viicos: @jorenham [commented](https://github.com/Viicos/typing/pull/1#discussion_r3693757178):

On the surface I guess it makes sense for at least the return type to be LSP-enforced, since those contain the class member definitions. But I wonder how much that'll matter in practice though, because I can imagine that most will just annotate the return type as `dict[str, Any]`, and not a fancy `TypedDict`.
It also feeds directly into `__new__`, which doesn't participate in the LSP, so perhaps it's more consistent also not enforce LSP for `__prepare__`.

So all things considered, I'm leaning towards grating this "constructor method status" and have it not partake in LSP.

Viicos: We could also enforce consistency between the return type of `__prepare__()` and the `namespace` argument of `__new__()`.

I'll keep this in mind as a follow up, it doesn't need to be part of this spec update.

docs/spec/constructors.rst
carljm: From the ty perspective, we consider this a bug/limitation and intend to support returning non-classes from metaclass `__new__` in the future. (It's just relatively lower priority since no other type checker supports it, either.) So I would not prefer to encode in the spec/conformance suite an expectation that type checkers must model this inconsistently with the runtime behavior. If we specify anything in this area, I'd prefer to specify runtime-consistent behavior, and allow all type-checkers to be marked non-conforming in that aspect for now.

Viicos: Changed in https://github.com/python/typing/pull/2344/commits/c1995e4012141d33f3696a6cf63e857a6496421a.

docs/spec/constructors.rst
```diff
@@ -488,3 +490,295 @@ callable.
...
+        pass
+
+    Meta("MyClass4", (), {}, key=3)  # OK
+    Meta("MyClass5", (), {}, key="")  # Type error: wrong type for "key"
+
+Keyword arguments in a direct metaclass call (such as the last two calls in the
+example above) require no special handling: they are validated as part of
+evaluating the call using the standard constructor call rules.
+
+Type checkers should honor the evaluated retunr type of the implied metaclass call,
```
ngnpope, line 670:
```suggestion
Type checkers should honor the evaluated return type of the implied metaclass call,
```

PR #2350, opened 2026-08-27
Allow non-Any gradual return types for ambiguous overloads

carljm:
The spec's requirement to infer `Any` for gradually-ambiguous overload evaluations is a regular source of discontent, as a more precise type could often still meet the graduality requirement that the inferred return type can be gradually used as if it were any of the matched-overload return types.

In fact mypy already does infer more precise types that meet this requirement. For example, mypy combines ambiguous overload returns of `list[int]` and `list[str]` into `list[Any]`, which is gradually usable as either `list[int]` or `list[str]`, and is clearly preferable to `Any` in its precision.

Adjust the spec language and the conformance suite tests to express the actual graduality requirements, instead of requiring exactly `Any`.

An inference of `Any` remains spec-compliant under this change, but more precise gradual types that still meet the graduality requirement can also be spec-compliant.

### DIFF
--- a/conformance/results/mypy/overloads_evaluation.toml
+++ b/conformance/results/mypy/overloads_evaluation.toml
@@ -3,7 +3,6 @@ notes = """
 Does not expand boolean arguments to `Literal[True]` and `Literal[False]`.
 Does not expand enum arguments to literal variants.
 Does not expand tuple arguments to possible combinations.
-Does not evaluate `Any` in some cases where overload is ambiguous.
 Evaluates `Any` in some cases where overload is not ambiguous.
 """
 conformance_automated = "Fail"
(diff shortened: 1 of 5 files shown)


carljm:
https://discuss.python.org/t/permitting-more-precise-inferred-types-from-ambiguous-overloads/108776

jorenham (approved):
I really like this; thanks!

davidhalter (approved):
Wow, if I knew it was this simple I would have done that a long time ago. I have preferred Mypy's behavior in this case and I'm happy to reintroduce it in Zuban. I originally was unhappy to remove it, because it feels much better (and I hope this improves Zuban for scipy-stubs and similar cases)...

I still feel that AnyOf might be useful, but this is great for now.

Thanks a lot!

JelleZijlstra (approved):

carljm:
This is approved by the full typing council; I'm just waiting the required seven days from the DPO post before merging it.

jorenham:
> I'm just waiting the required seven days from the DPO post before merging it.

It might be nice to automate this, e.g. using a label + a github actions cronjob, so we don't forget to merge after 7 days or merge too soon.

Issue #2351, opened 2026-08-31
Finals and Literals: Spec contradiction


srittau:
This came up in python/typeshed#16301. I assume this is mostly a clerical/historical error, but let me know if we should follow the whole spec change process.

----

The typing spec [says about `Final`](https://typing.python.org/en/latest/spec/qualifiers.html#syntax):

> Example:
> 
> `ID: Final = 1`
>
> The typechecker should apply its usual type inference mechanisms to determine the type of ID (here, likely, int).

And then a bit later:

> Type checkers should treat uses of a final name that was initialized with a literal as if it was replaced by the literal.

Both sentences are copied directly from [PEP 591](https://peps.python.org/pep-0591/), which overlapped with the Literals [PEP 586](https://peps.python.org/pep-0586/).

I assume that we should change the first sentence to "here, likely, `Literal[1]`" and change the second sentence to be a bit more concrete along these lines: 

> Type checkers should infer uses of a final name that was initialized with a literal `X` as `Literal[X]` if `X` is an allowed literal type (cf. to the literals section). Otherwise, they should infer the type of the `X`.

(And maybe extend the example after a bit.)

In fact, this is what the conformance tests already check:

https://github.com/python/typing/blob/cf943ccbea5596ef969eec6e2260097ff697b7ba/conformance/tests/qualifiers_final_annotation.py#L11-L13

(Although the tests could be extended a bit.)

carljm:
I do think a clarification to the spec text is warranted here; the current text is confusing. I think we should remove `(here, likely, int)` and instead mention the later "replacement" rule. And we should probably also clarify that the replacement rule applies to a bare `Final` annotation with valid literal initializer.

I would consider this a spec clarification, not a change in specification.

srittau:
As @carljm points out in #2352, the [Literal section of the spec](https://typing.python.org/en/latest/spec/literal.html#interactions-with-final) has a whole section about the interaction with `Final` that actually is already correct. We should probably just change the mentions in the `Final` spec part to point to that section.

PR #2352, opened 2026-08-31
Add tests for `Final`/`Literal` interaction

srittau:
This is my first time changing the conformance tests, did I do it correctly? What does "conformant is required when conformance_automated is 'Fail'" mean?

Also cf. #2351. If that issue and this PR is accepted I would like to add quotes from the (changed) spec to the tests before merging this.


carljm:
> What does "conformant is required when conformance_automated is 'Fail'" mean?

It means that if the automated conformance fails, it is your responsibility to analyze the nature of the gap for that type checker, and decide whether `conformant = 'Fail'` or `conformant = 'Partial'` is more appropriate. In the latter case you should also add a `notes` field describing the gap in conformance.

conformance/tests/qualifiers_final_annotation.py
```diff
@@ -7,10 +7,14 @@
 # Specification: https://typing.readthedocs.io/en/latest/spec/qualifiers.html#id1
 
 ID1: Final[int] = 1
+assert_type(ID1, Literal[int])
```
carljm: `Literal[int]` is not a valid type expression (and thus every type checker correctly errors on it). Same with `Literal[float]` below.

Did you mean to assert the type is just `int`? If so, I don't think that's an assertion supported by the spec. https://typing.python.org/en/latest/spec/literal.html#interactions-with-final explicitly says that in this exact scenario, an inference of either `int` or `Literal[1]` is fine.

carljm (changes requested):

conformance/tests/qualifiers_final_annotation.py
srittau: Interesting, another contradiction in the spec.

carljm:
What is your intention for this PR vs https://github.com/python/typing/pull/2353 ? I think it would be simpler and preferable to just have one combined PR that modifies both the spec text and the conformance suite, rather than two separate PRs.

srittau:
> What is your intention for this PR vs #2353 ? I think it would be simpler and preferable to just have one combined PR that modifies both the spec text and the conformance suite, rather than two separate PRs.

Sure, I can combine them if you prefer. Considering that both are more or less independent (since the are only refactorings/documentation of the status quo) I thought it would be easier to review separately.

carljm:
I don't have strong feelings; if you prefer to separate them, that's fine. I would personally find it easier to review them together; it makes it easier to validate that we have the right tests for the clarified spec text.

srittau:
Will merge when I get home.

PR #2353, opened 2026-08-31
Unify and fix`Literal`/`Final` interaction

### LINKED ISSUE #2351: Finals and Literals: Spec contradiction

srittau:
- Add an "Inference Rules" subsection to the `Final` section.
- Move the inference rules for class vars there.
- Remove contradictory paragraph about literal handling from `Final` section.
- Merge the "Interactions with Final" section from literals into the new "Inference Rules" section, add a link, and trim slightly.

I've decided to move the interaction section to the "qualifiers" chapters,
since `Literal` seems much closer linked to `Final` than vice versa.

Closes: #2351


docs/spec/qualifiers.rst
```diff
@@ -159,14 +160,6 @@ The generated ``__init__`` method of :doc:`dataclasses` qualifies for this
 requirement: a bare ``x: Final[int]`` is permitted in a dataclass body, because
 the generated ``__init__`` will initialize ``x``.
 
-Type checkers should infer a final attribute that is initialized in a class
-body as being a class variable, except in the case of :doc:`dataclasses`, where
-``x: Final[int] = 3`` creates a dataclass field and instance-level final
-attribute ``x`` with default value ``3``; ``x: ClassVar[Final[int]] = 3`` is
-necessary to create a final class variable with value ``3``. In
-non-dataclasses, combining ``ClassVar`` and ``Final`` is redundant, and type
-checkers may choose to warn or error on the redundancy.
```
srittau, line 168: This got moved to the "Inference Rules" section.

docs/spec/qualifiers.rst
```diff
@@ -193,20 +186,52 @@ with ``Final`` to prevent mutating such values::
...
 
-Type checkers should treat uses of a final name that was initialized
-with a literal as if it was replaced by the literal. For example, the
-following should be allowed::
-
-   from typing import NamedTuple, Final
-
-   X: Final = "x"
-   Y: Final = "y"
-   N = NamedTuple("N", [(X, int), (Y, int)])
```
srittau, line 204: This got removed completely as it's redundant.

docs/spec/qualifiers.rst
```diff
@@ -193,20 +186,52 @@ with ``Final`` to prevent mutating such values::
...
+^^^^^^^^^^^^^^^
+
+In the example below, we know that ``foo`` will always be equal to
+exactly ``3``. A type checker can use this information to deduce that ``foo``
+is valid to use in any context that expects a ``Literal[3]``::
+
+    def expects_three(x: Literal[3]) -> None: ...
+
+    foo: Final = 3
+    expects_three(foo)  # Type checks, since 'foo' is Final and equal to 3
```
srittau: The introductory part was reshuffled a bit from the original section to be a better fit here. The rest is identical.

srittau:
I just noticed that the introductory sentence about "normal inference" is actually important. I will change the PR when I get home.

srittau:
The spec used to say:

> The typechecker should apply its usual type inference mechanisms to determine the type of ``ID`` (here, likely, ``int``)"

before I replaced that sentence with a link to the new inference rules section. The problem is that this contradicts the explicit guidance in the section I copied over from the literals spec:

> Type checkers are not obligated to understand any other uses of Final. For
example, whether or not the following program type checks is left unspecified: [...]

My suggestion: We leave this original sentence out for now, reverting back to "left unspecified" for now. But I did plan to open a discuss thread anyway to define a few more supported cases, like `X: Final = cls()`, `Y: Final = 3.2`, and remove the ambiguity surrounding `Z: Final[int] = 3`.

docs/spec/qualifiers.rst
```diff
@@ -91,8 +91,9 @@ Syntax
 
     ID: Final = 1
 
-  The typechecker should apply its usual type inference mechanisms to
-  determine the type of ``ID`` (here, likely, ``int``). Note that unlike for
+  The typechecker should apply the inference mechanisms
```
carljm: The rules below don't attempt to cover every possible right-hand-side of an assignment to a `Final` annotated attribute; they are specific to valid literal values. So I think we still need the "typechecker should apply its usual type inference mechanisms" default fallback language here, with the literal rules below as additional constraints. I don't think we can just refer to the rules below as though they fully specify inference of all `Final` assignments.

Currently this seems to drop any requirement for type checkers to support e.g. `x: Final = C()` or `x: Final = ['a', 'b']`, because those right-hand-sides are not valid `Literal` values.

conformance/tests/qualifiers_final_annotation.py
```diff
@@ -6,10 +6,27 @@
...
+# > Type checkers are not obligated to understand any other uses of Final.
+
+ID3: Final[int] = 1
+expects_one(ID3)  # E?: May or may not be accepted by type checkers
+
+ID4: Final = 2 - 1
+expects_one(ID4)  # E?: May or may not be accepted by type checkers
+
+ID5: Final = range(1)
+assert_type(ID5, range)  # E?: May or may not be inferred by type checkers
```
carljm: All three of these tests are kind of useless as conformance tests, but the first two I don't mind, since I think they still provide useful clarification that this behavior is not specified either way. But this third one is a really strange test, since we assert that `range(1)` may or may not be inferred as `range` -- which is not even a literal type anyway. So it really feels like this test is saying nothing at all. I would just remove it.

docs/spec/qualifiers.rst
carljm: I don't think this is redundant or should be removed. The position of `X` and `Y` in this `NamedTuple` declaration is not an ordinary "call expects a literal type" position; it's syntax-sensitive type-factory metadata that is special-cased by type checkers; they must be able to understand as describing a specific field name. I don't think the rules below are sufficient to clarify that this example must work.

Also the conformance tests still quote this deleted paragraph.

docs/spec/qualifiers.rst
```diff
@@ -193,20 +186,52 @@ with ``Final`` to prevent mutating such values::
...
+The ``Final`` qualifier serves as a shorthand for declaring that a variable
+is *effectively Literal*.
+
+Type checkers are expected to
+support this shortcut. Specifically, given a variable or attribute assignment
+of the form ``var: Final = value`` where ``value`` is a valid parameter for
+``Literal[...]``, type checkers should understand that ``var`` may be used in
+any context that expects a ``Literal[value]``.
+
+Type checkers are not obligated to understand any other uses of Final. For
```
carljm: I find this sentence confusing, so if we're clarifying here, I think we should clarify it. Obviously type checkers are generally expected to understand any use of `Final` as requiring finality (i.e. not accepting later reassignments). So why would we say they are "not obligated to understand any other uses of Final"? It seems like what we really mean is more like this:
```suggestion
Type checkers are not obligated to understand any other uses of Final as implying a Literal type. For
```

carljm (commented):
I think combining these sections makes sense. Left some inline comments.

I think this is a clarification, not a substantive spec change, so I don't know that we need the full process including DPO post here. But it's a hefty enough rewording / rearrangement that I do think we should try to get at least most of the typing council to approve it.

docs/spec/qualifiers.rst
```diff
@@ -193,20 +186,52 @@ with ``Final`` to prevent mutating such values::
...
+    foo: Final = 3
+    expects_three(foo)  # Type checks, since 'foo' is Final and equal to 3
+
+The ``Final`` qualifier serves as a shorthand for declaring that a variable
+is *effectively Literal*.
+
+Type checkers are expected to
+support this shortcut. Specifically, given a variable or attribute assignment
+of the form ``var: Final = value`` where ``value`` is a valid parameter for
+``Literal[...]``, type checkers should understand that ``var`` may be used in
```
JelleZijlstra: This wording is a bit vague. Should we just say that type checkers must infer `Literal[...]`?

conformance/tests/qualifiers_final_annotation.py
```diff
@@ -6,10 +6,27 @@
...
-ID1: Final[int] = 1
+def expects_one(x: Literal[1]) -> None: ...
 
-ID2: Final = 1
-assert_type(ID2, Literal[1])
+# > The Final qualifier serves as a shorthand for declaring that a variable is effectively Literal.
+
+ID1: Final = 1
+assert_type(ID1, Literal[1])
+expects_one(ID1)
```
JelleZijlstra: This seems redundant with the `assert_type`.

docs/spec/qualifiers.rst
srittau: I think this was my main issue. As written, this reads as if any other inference is unsupported. I will just rewrite the section completely to make the rules very clear and make the whole language a bit more "standardy" rather than "PEPy".

docs/spec/qualifiers.rst
srittau: I've restored this section for now, although it's really vague and ambiguous. That makes it sound as if `Final` works like C's `#define`. Maybe the best idea would be to have a dedicated "Interaction with NamedTuple" section, since this seems to be a special case.

srittau:
I've now completely rewritten the inference rules section (except the part finals in class bodies). I think I captured the intent of the various bits and pieces previously scattered around the specs. I'm using concrete language better suited for a spec than the previous "PEP" language that was more explanatory. 

Personally, I'd prefer the "explicit type" section to just read "Type checkers should use an explicit type if one is provided.", but that would be a direct contradiction to the previous example that type checkers may accept:

```python
    bar1: Final[int] = 3
    expects_three(bar1)  # May or may not be accepted by type checkers
```

I've gone with the old vague language for now, but I think there is very much room for improvement here in the future.

I've also restored  the  `NamedTuple` section in a dedicated "Interaction with NamedTuple" section and reworded it to apply to `NamedTuple`s only for now, since the previous language was extremely vague. Please let me know if there are cases missing that should be added.

Finally, I've reworked the conformance tests to follow the reworked section. I've introduced a new way to mark result groups as requiring at least one success (as pyrefly accepts both `assert_type(X, float)` and `assert_type(X, float | int)`. I will extract that into a separate PR. The reworked conformance tests reveal a few instances of type checkers being more liberal in what type of literals they support (especially `Literal[3.14]`, but pycroscope also accepts `Literal[range(0, 3)]`. I'm not sure what the best approach here is.

docs/spec/qualifiers.rst
```diff
@@ -207,6 +198,55 @@ following should be allowed::
...
+
+    bare1: Final = 3  # infer bare1 as Final[Literal[3]]
+
+* If the value is an expression whose result is a valid parameter for
+  ``Literal[...]``, type checkers may infer that ``Literal`` or use standard
+  inference rules. For example::
+
+    bare2: Final = 1 + 2  # may infer bare2 as Final[Literal[3]] or Final[int]
+
+* In all other cases, type checkers should use standard inference rules.
```
davidhalter: This is technically not correct, because immutable containers are just fine inferring literals, e.g.

```
x: Final = (1,)
```

should IMO infer `tuple[Literal[1]]`. It's at least what Mypy and Pyright do. I haven't checked the other type checkers.

conformance/tests/qualifiers_final_annotation.py
```diff
@@ -6,10 +6,31 @@
...
+# > inference rules.
+
+bare2: Final = 1 + 2
+assert_type(bare2, Literal[3])  # E[bare2]: either Literal[3] or int
+assert_type(bare2, int)  # E[bare2]: either Literal[3] or int
+
+# > In all other cases, type checkers should use standard inference rules.
+
+bare3: Final = 3.14  # infer bare3 as Final[float] or Final[float | int]
+assert_type(bare3, float)  # E[bare3!]: either float or float | int
```
carljm: These exact-type assertions reject checkers that _are_ applying their normal inference, just as the spec requires.

We should just not use `float` as an example, because the `float/int` special case is not well specified, so we can't make very many assertions around it that are clearly supported by the spec.

docs/spec/qualifiers.rst
```diff
@@ -207,6 +198,55 @@ following should be allowed::
...
+
+Type checkers should use the following inference rules for ``Final`` with an
+explicit type:
+
+* If the right hand side is missing, type checkers should use the explicit
+  type::
+
+    typed1: Final[int]  # infer typed1 as Final[int]
+
+* Type checkers are not obligated to understand any other uses of Final. For
```
carljm: This still uses the problematic (IMO) "not obligated to understand any other uses of Final" wording.
```suggestion
* Type checkers are not obligated to infer ``Literal`` types for any other uses of Final. For
```

docs/spec/qualifiers.rst
```diff
@@ -207,6 +198,55 @@ following should be allowed::
...
+  example, whether or not the following program type checks is left
+  unspecified::
+
+    # Note: The assignment does not exactly match the form 'var: Final = value'.
+    typed2: Final[int] = 3
+    expects_three(typed2)  # May or may not be accepted by type checkers
```
carljm, line 241: The only definition of `expects_three` was deleted with the old Literal section, so this example no longer shows that the parameter is `Literal[3]` and has an unrelated undefined-name error as written. We need to add a definition of `expects_three` here.

```python
def expects_three(x: Literal[3]) -> None: ...
```

conformance/src/main.py
```diff
@@ -65,7 +65,7 @@ def run_tests(
 
 def get_expected_errors(test_case: Path) -> tuple[
     dict[int, tuple[int, int]],
-    dict[str, tuple[list[int], bool]],
+    dict[str, tuple[list[int], bool, bool]],
```
carljm: Mentioned this in the extracted PR, but the docstring below needs updating if we include this change.

docs/spec/qualifiers.rst
carljm: I agree. This should ideally not be an "example" of a broad vague requirement, it should be a comprehensive spec of what type checkers are expected to support (should also probably include field names in non-class `TypedDict` definitions, too.) But maybe that's too much to include in this PR.

docs/spec/qualifiers.rst
```diff
@@ -207,6 +198,55 @@ following should be allowed::
...
+type annotation:
+
+* If the value is a literal value which is a valid parameter for
+  ``Literal[...]``, type checkers should infer that ``Literal``. For example::
+
+    bare1: Final = 3  # infer bare1 as Final[Literal[3]]
+
+* If the value is an expression whose result is a valid parameter for
+  ``Literal[...]``, type checkers may infer that ``Literal`` or use standard
+  inference rules. For example::
```
carljm: I don't love this wording, simply because it frames "infer `Literal`" and "use standard inference rules" as opposing alternatives, when there's no reason inferring the `Literal` type couldn't also be the "standard inference" for some type checkers. (In fact I think it is for many type checkers, even in the `1 + 2` case below.)
```suggestion
  ``Literal[...]``, type checkers may infer that ``Literal`` but are not
  required to. For example::
```

docs/spec/qualifiers.rst
```diff
@@ -207,6 +198,55 @@ following should be allowed::
 item or a :ref:`NamedTuple <namedtuple>` field. Such usage also generates
 an error at runtime.
 
+.. _`final-inference`:
+
+Inference Rules
+^^^^^^^^^^^^^^^
+
+Type checkers should use the following inference rules for ``Final`` without a
+type annotation:
```
carljm: "Without a type annotation" is not clear.

```suggestion
Type checkers should use the following rules for inferring the type of ``x`` after an annotated assignment in the form ``x: Final = value``:
```

docs/spec/qualifiers.rst
```diff
@@ -207,6 +198,55 @@ following should be allowed::
...
+
+* If the value is a literal value which is a valid parameter for
+  ``Literal[...]``, type checkers should infer that ``Literal``. For example::
+
+    bare1: Final = 3  # infer bare1 as Final[Literal[3]]
```
carljm: nit: `Final` is not part of the inferred type, it's a type qualifier that comes from the annotation -- same for all comments below.
```suggestion
    bare1: Final = 3  # infer bare1 as Literal[3]
```

docs/spec/qualifiers.rst
```diff
@@ -207,6 +198,55 @@ following should be allowed::
...
+* If the value is an expression whose result is a valid parameter for
+  ``Literal[...]``, type checkers may infer that ``Literal`` or use standard
+  inference rules. For example::
+
+    bare2: Final = 1 + 2  # may infer bare2 as Final[Literal[3]] or Final[int]
+
+* In all other cases, type checkers should use standard inference rules.
+  For example::
+
+    bare3: Final = 3.14  # infer bare3 as Final[float] or Final[float | int]
```
carljm: I think we should stay away from floats here.

I don't actually think it's useful to even include examples for "type checkers should use standard inference rules", I would just leave them out entirely.

docs/spec/qualifiers.rst
carljm: For all type checkers other than mypy/zuban, `tuple[Literal[1]]` is the "standard inference rule" for `(1,)`, regardless of `Final`.

I think what this highlights is again that setting up "infer literal" vs "standard inference rules" as a dichotomy doesn't make sense. This section is specifying some cases where type checkers must infer literal types. Beyond those specified cases, this section imposes no requirements on literal inference. It's fine for mypy/zuban to infer `tuple[Literal[1]]` for `x: Final = (1,)` where they would otherwise infer `tuple[int]`; we shouldn't imply that's wrong -- but we also don't necessarily need to specify it as required. (Though we could add it as a required example, that wouldn't bother me either. It wouldn't render any type checker non-conformant.)

```suggestion
* In all other cases, type checkers may apply their preferred inference rules.
```

docs/spec/qualifiers.rst
```diff
@@ -207,6 +198,55 @@ following should be allowed::
...
+    bare4: Final = range(3)  # infer bare4 as Final[range]
+
+Type checkers should use the following inference rules for ``Final`` with an
+explicit type:
```
carljm, line 226:
```suggestion
Type checkers should use the following inference rules for ``Final[SomeType]`` annotations:
```

docs/spec/qualifiers.rst
carljm: As written, this conflicts with the requirements stated below, because `x: Final[int] = 3` is still "a final name that was initialized with a literal", but we explicitly say below that type checkers aren't required to infer `x` as `Literal[3]` in that case. So I think we at the very least need an update to the text here to fix that inconsistency.
```suggestion
Type checkers should treat uses of a name annotated with bare ``Final`` and initialized
with a literal as if it was replaced by the literal. For example, the
```

docs/spec/qualifiers.rst
```diff
@@ -207,6 +198,55 @@ following should be allowed::
...
+.. _`final-inference`:
+
+Inference Rules
+^^^^^^^^^^^^^^^
+
+Type checkers should use the following inference rules for ``Final`` without a
+type annotation:
+
+* If the value is a literal value which is a valid parameter for
+  ``Literal[...]``, type checkers should infer that ``Literal``. For example::
```
carljm: I think saying "is a literal value" here introduces some ambiguity that wasn't present in the previous wording. Are `-4`, `+1`, and `Color.RED` "literal values"? Unclear; none of them are represented in the AST as a literal value. But they are all valid `Literal` arguments.

Ideally those cases might also be covered in the tests, though arguably it's fine to leave these edge cases out of the `Final` tests if the `Literal` tests cover them, as long as the wording here is clear.

At the same time, we should take care not to imply that `Alias = Literal[1]; x: Final = Alias` invokes this rule, even though `Literal[Alias]` would be a valid type expression. The old wording was potentially subject to this mis-interpretation.

I think the clearest option here is probably just to outline the relevant cases explicitly:

```suggestion
* If ``value`` is an integer literal (optionally prefixed with unary ``+`` or ``-``),
  a string or bytes literal, ``True``, ``False``, ``None``, or an enum-member
  reference such as ``Color.RED``, type checkers should infer ``Literal[value]``.
  For example::
```

conformance/tests/qualifiers_final_annotation.py
srittau: Inferring `Literal[3.14]` is against the spec, which explicitly forbids `float` literals. That said, this is more of an issue that should be covered by the `Literal` test cases, so changing it to a list is the better option.

docs/spec/qualifiers.rst
srittau: I agree, but the syntax section uses "With no type annotation". I'm changing it to exactly the latter, but prefer the clarity of using pre-defined terms. (Even if these terms should probably be changed.)

docs/spec/qualifiers.rst
srittau: I think examples are still useful as a clarification, although two examples are not really necessary.

docs/spec/qualifiers.rst
srittau: This also uses the exact wording of the syntax section, which I think should be used.

docs/spec/qualifiers.rst
srittau: I've opened #2356 to track this.

docs/spec/qualifiers.rst
srittau: Maybe "... annotated with a `Final` where a `Literal` is inferred"? This would mean consistent handling between these cases, and we don't need to repeat the rules here directly.

docs/spec/qualifiers.rst
srittau: I don't think we should repeat the literal rules here. If we just slightly reformulate this as "If the value would be a valid parameter for `Literal[...]`, ..." it is unambiguous that `-4`, `+1`, and `Color.RED` are allowed.

carljm: I agree that would address the inclarity with `-4`, `+1`, and `Color.RED`. But as mentioned in my comment, `Literal[1]` is also a valid parameter for `Literal[...]`, and so is `Alias`, where `Alias = Literal[1]`. But `x: Final = Alias` where `Alias = Literal[1]`, and `x: Final = Literal[1]`, do not trigger this rule. So "would be a valid parameter for `Literal[...]`" seems not sufficiently precise. That's why my suggestion here does not actually just repeat the literal rules, it specifies a subset of them which apply in this case.

PR #2363, opened 2026-09-22
Adapt and clarify tests and docs around explicitly annotated receiver parameters

sharkdp:
The conformance tests currently include a test in `protocols_generic.py` which wants type checkers to accept assignability of `ConcreteHasParent` to the `HasParent` protocol:

```py
T = TypeVar("T")

class HasParent(Protocol):
    def get_parent(self: T) -> T:
        ...

class ConcreteHasParent:
    def get_parent(self) -> Self:
        return self
```

However, the `get_parent` method of the protocol seems to be more general than what `ConcreteHasParent` implements. The `T` type variable is unbounded, while the implicit `Self` annotation on the unannotated `self` received of `ConcreteHasParent.get_parent` should be bounded by `ConcreteHasParent`. The protocol should theoretically support calls like the following, through `type[HasParent]`, whereas `ConcreteHasParent` would not:

```py
def f(h: type[HasParent]):
    h.get_parent("whatever")
```

[The spec](https://typing.python.org/en/latest/spec/annotations.html#annotating-instance-and-class-methods) includes a paragraph that states:

> Note that some type checkers may apply restrictions on [the use of an explicitly annotated receiver parameter], such as requiring an appropriate upper bound for the type variable used

On the other hand, no type checker seems to do so (pyright *assumes* an implicit upper bound, but that's not supported by that statement in the spec). Given that, I think it would be best to just define an upper bound for the type variable here.

I also fixed that problem (using an unbounded type variable for as an annotation on a receiver parameter) elsewhere in the tests and the documentation, often by using `Self` instead, if it seemed like the test was about something unrelated.


conformance/results/ty/protocols_generic.toml
```diff
@@ -1,10 +1,10 @@
 conformance_automated = "Fail"
 conformant = "Partial"
 notes = """
-Only partially supports `@property` protocol members.
```
sharkdp, line 4: I am a ty maintainer. This description was outdated.

I am *not* making this change to make ty look better on the conformance tests. By disentangling the different aspects of the previous `ConcreteHasProperty*` tests, we just isolate the deficiency of ty more clearly.

conformance/results/ty/protocols_generic.toml
```diff
@@ -1,10 +1,10 @@
...
+Does not reject incompatible non-generic implementations of generic protocol methods.
```
sharkdp, line 4: Cross reference: https://github.com/astral-sh/ty/issues/2323

conformance/tests/protocols_generic.py
```diff
@@ -96,52 +98,69 @@ def get_parent(self) -> Self:
...
 
-class HasPropertyProto(Protocol):
+class HasParentProperty(Protocol):
     @property
-    def f(self: T) -> T:
-        ...
-
-    def m(self, item: T, callback: Callable[[T], str]) -> str:
+    def parent(self) -> Self:
         ...
```
sharkdp, line 105: The existing tests here were convoluted (the `ConcreteHasProperty*` assignments failed for multiple reasons). It looks to me like they wanted to tests generic properties and generic methods at the same time and failed to do so. Looking back at the history, I can see that the assertion on the `hp2: HasPropertyProto = ConcreteHasProperty2()` line has [changed from "OK" to "E"](https://github.com/python/typing/pull/1864), further illustrating that it's not really clear what was being tested there.

In the new version, I tried to stay close to the original (I can deviate further if that is desirable for educational reasons, for example), while disentangling the generic-property-related tests from the generic-method-related tests.

conformance/tests/protocols_generic.py
```diff
@@ -96,52 +98,69 @@ def get_parent(self) -> Self:
...
 
 
-class ConcreteHasProperty1:
+class ConcreteParentProperty:
     @property
-    def f(self: T) -> T:
```
sharkdp, line 110: If we really want to retain a test with a property that is generic over an *unbounded* type variable, I can add that again. Otherwise, it seems like using `Self` is sufficient here (which also makes that property generic)?

docs/guides/writing_stubs.rst
```diff
@@ -652,7 +652,7 @@ referencing a type variable.
 
 Yes::
 
-    _T = TypeVar("_T")
+    _T = TypeVar("_T", bound="Foo")
```
sharkdp: I can omit this, but it seemed sensible to add.

docs/spec/protocol.rst
```diff
@@ -365,9 +365,10 @@ The self-types in protocols follow the
 
   class Copyable(Protocol):
       def copy[C: Copyable](self: C) -> C:
+          ...
```
sharkdp, line 368: Drive-by fix.

docs/spec/protocol.rst
```diff
@@ -365,9 +365,10 @@ The self-types in protocols follow the
...
 
   class One:
-      def copy(self) -> 'One':
```
sharkdp, line 370: This seems like it violates the generic type of `Copyable.copy`, since there's no "you can use Foo instead of Self" exception for arbitrary type variables in the spec.

conformance/tests/protocols_generic.py
```diff
@@ -96,52 +107,79 @@ def get_parent(self) -> Self:
...
-        ...
+# Specification: https://typing.readthedocs.io/en/latest/spec/generics.html#use-in-protocols
+# > `Self` is valid within Protocols, similar to its use in classes:
 
-    def m(self, item: T, callback: Callable[[T], str]) -> str:
+
+class HasParentProperty(Protocol):
+    @property
+    def parent(self) -> Self:
```
carljm: Could we keep an explicitly annotated, bounded-TypeVar property variant alongside this `Self` version? After this change, there are no property getters with an explicitly annotated generic `self` left in the suite. A bounded variant would preserve coverage of generic receiver binding through `@property` while addressing the concern about the unbounded type variable.

docs/spec/protocol.rst
```diff
@@ -365,9 +365,10 @@ The self-types in protocols follow the
...
+      def copy(self) -> Self:
```
carljm: What's the rationale for this change? I think the original `-> 'One'` is already allowed, per the spec: the [generics spec](https://typing.python.org/en/latest/spec/generics.html#use-in-protocols) treats protocol `Self` as equivalent to a protocol-bounded TypeVar and allows implementations that return their concrete class. The new version with `-> Self` is also valid, but makes `One` and `Other` demonstrate closer-to-equivalent annotations instead of two distinct supported cases.

conformance/results/ty/protocols_generic.toml
carljm: I think astral-sh/ty#2924 is the better reference here; astral-sh/ty#2323 is specifically discussing unsoundness of `Self` in protocols, which the spec actually mandates should be allowed. (Though it does get a bit muddied in the comments, since in ty `Self` is simply a bounded typevar.)

carljm (approved):
Looks good to me, but I'll wait for another reviewer before merging.

docs/spec/protocol.rst
sharkdp: See my inline annotation on the negative part of the diff:

> This seems like it violates the generic type of Copyable.copy, since there's no "you can use Foo instead of Self" exception for arbitrary type variables in the spec.

You wrote:

> the generics spec treats protocol `Self` as equivalent to a protocol-bounded TypeVar and allows implementations that return their concrete class.

The actual protocol in question here,`Copyable`, does not use `Self`. It uses a (properly bounded) typevar tough, so if we combine your quote of the spec with *"`Self` is valid within Protocols [it] is treated equivalently to [bounded typevars]"*, then maybe one could conclude that returning `One` is also valid here.

It didn't seem clear from the spec to me, which is why I replaced it `One` with `Self`, but I can revert that.

sharkdp: I reverted this change.

docs/guides/writing_stubs.rst
JelleZijlstra: Shouldn't this example be using Self?

conformance/tests/protocols_generic.py
```diff
@@ -96,52 +107,79 @@ def get_parent(self) -> Self:
...
+class ConcreteParentProperty:
     @property
-    def f(self: T) -> T:
+    def parent(self) -> Self:
         return self
 
+
+class InvalidParentProperty:
+    @property
+    def parent(self) -> int:
```
JelleZijlstra: a more interesting example might be having this return `HasParentProperty`

docs/spec/protocol.rst
carljm: I agree this is really unclear in the spec. It probably deserves a separate PR that is not just a conformance suite clarification, but an actual spec clarification/change with accompanying conformance suite updates.

(Sorry I missed your comment; I've found that much easier to do accidentally now that GitHub collapses them by default.)

docs/guides/writing_stubs.rst
sharkdp: That didn't work great with the prose directly above. I rewrote the text a bit and changed the examples. Hopefully that's better now.

conformance/tests/protocols_generic.py
sharkdp: Good idea, changed.

JelleZijlstra (approved):

